PR #6, opened 2025-02-26
Return html paywall when request is from a browser

apmcdermott-cb:


example/resource.ts
```diff
@@ -7,14 +7,15 @@ const app = new Hono();
 const port = 4021;
 
 app.use(
-  "/joke",
+  '/joke',
   hono.paymentMiddleware(
-    "$0.01",
-    "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
+    '$0.01',
+    '0x209693Bc6afc0C5328bA36FaF03C514EF312287C',
```
erikreppel-cb: oh nooo we have different linting on quotes configured 😅

site/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+    `${request.nextUrl.protocol}//${request.nextUrl.host}${request.nextUrl.pathname}` as Resource;
+
+  app.use(
+    '*',
+    hono.paymentMiddleware('$0.01', facilitatorAddress as Address, {
+      description: 'Access to protected content',
+      testnet: true,
+      facilitatorUrl,
+      resource: resourceUrl,
+    })
```
erikreppel-cb: John is working on a middleware for next so we can probably use that once its in

site/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Address } from 'viem';
+import { Resource } from 'x402/types';
+
+export async function middleware(request: NextRequest) {
+  if (!request.nextUrl.pathname.startsWith('/protected')) {
+    console.log('Not a protected route, skipping middleware');
+    return NextResponse.next();
+  }
+
+  const facilitatorAddress = process.env.FACILITATOR_WALLET_ADDRESS;
```
erikreppel-cb: should be `RESOURCE_WALLET_ADDRESS`

site/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+  const facilitatorUrl =
+    process.env.NEXT_PUBLIC_FACILITATOR_URL || 'http://localhost:4020';
```
erikreppel-cb: can default to `https://x402.org/facilitator`

site/middleware.ts
erikreppel-cb: this is a really clever way of doing this!

erikreppel-cb (approved):

site/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  const res = await app.fetch(honoReq);
+
+  if (res.status === 402) {
+    return new NextResponse(await res.text(), {
+      status: 402,
+      headers: Object.fromEntries(res.headers.entries()),
+    });
+  }
```
erikreppel-cb, line 50: the only problem with this is it verifys and settles in 1 step before content is returned

example/resource.ts
apmcdermott-cb: hahah I had to turn off format on save

site/middleware.ts
apmcdermott-cb: oh truuuuue

so we'd need to wait to see if verification goes through successfully before doing the settlement and the rest

erikreppel-cb: yep. The other way to do this is just to directly call the `settle` and `verify` functions from the `server` package

erikreppel-cb: yep. The other way to do this is just to directly call the `settle` and `verify` functions from the `server` package

PR #46, opened 2025-04-11
feat: added x402-hono and hono server example

CarsonRoscoe:
Added x402-hono package and hono server example

### Manual

Tested by running the `facilitator` example along with the new `hono` server example, and then running the `axios` example client against them

### Unit
```
 ✓ src/index.test.ts (6 tests) 13ms
   ✓ configurePaymentMiddleware() > should return 402 with payment requirements when no payment header is present 10ms
   ✓ configurePaymentMiddleware() > should return HTML paywall for browser requests 0ms
   ✓ configurePaymentMiddleware() > should verify payment and proceed if valid 1ms
   ✓ configurePaymentMiddleware() > should return 402 if payment verification fails 0ms
   ✓ configurePaymentMiddleware() > should handle settlement after response 0ms
   ✓ configurePaymentMiddleware() > should handle settlement failure before response is sent 0ms
```


typescript/packages/x402-hono/src/index.ts
```diff
@@ -0,0 +1,115 @@
...
+        return c.json(
+          {
+            error: response.invalidReason,
+            paymentRequirements: toJsonSafe(paymentRequirements),
+          },
+          402,
+        );
+      }
+
+      console.log("Payment verified, proceeding");
```
John-peterson-coinbase: production ready packages should not log directly to console unless in a debug mode. Let's fast follow with allowing developers to inject a logger. We had a GH issue relating to this earlier in the week.

typescript/packages/x402-hono/vitest.config.ts
John-peterson-coinbase, line 1: Please add unit tests. It is critical that middleware behaves as expected.

typescript/packages/x402-hono/src/index.ts
```diff
@@ -0,0 +1,115 @@
...
+  Money,
+  moneySchema,
+  PaymentMiddlewareConfig,
+  PaymentRequirements,
+  Resource,
+  settleResponseHeader,
+} from "x402/types";
+import { useFacilitator } from "x402/verify";
+
+export function configurePaymentMiddleware(globalConfig: GlobalConfig) {
```
John-peterson-coinbase, line 46: Please add doc string

typescript/packages/x402-hono/src/index.ts
```diff
@@ -0,0 +1,115 @@
...
+  const { facilitatorUrl, address, network } = globalConfig;
```
John-peterson-coinbase, line 47: does network have to come from globalConfig? Should we allow specifying at the middleware config level?

CarsonRoscoe: This is a product question to me regarding how we expect users to use our app, and that we should optimize the devx for those experiences. After consideration, my thoughts are:
1. The median user will be sticking to one network, so putting it in the global config makes sense.
2. The power user may be using multiple networks, and the current spec does not support multiple networks simultaneously, they are expected to choose one at a time. Meaning, with the current spec, power users will be forced to create two endpoints (one to charge USDC on Solana, and one to charge USDC on base) if their needs require multinetworks. Putting it in the global config would not work for these users, they require it in the MiddlewareConfig

So I think the solution is to keep it in the global config for the median case, but also allow it optionally at the branch level as a `networkOverride` or `overrides: { network }` param. Middleware would then be expected to get the network by `middlewareConfig.networkOverride ?? globalConfig.network`

John-peterson-coinbase (approved):

typescript/packages/x402-hono/src/index.ts
murrlincoln: that solution makes sense, however I could also see the median user accepting as many different options as possible if we make it easy. would it be possible to eventually have it such that the server tells the user it can accept either Base or Solana, then whatever signature the user returns is parsed by the facilitator and used on the network? That might be the most seamless option for the end user

PR #58, opened 2025-04-15
feat: optional jwt signing in middleware

CarsonRoscoe:
* Added optional param to middleware to allow header injection
* Added helper function to create CDP API signed JWTs and inject them into the Auth header when using the CDP facilitator

NOTE: `@coinbase/cdp-sdk` breaks node 18 support, unless we dynamically import the `generateJwt` function. We decided to drop node 18 support all together, as Node 18's EOL is April 30th, before our official release.


typescript/packages/x402-express/src/index.ts
```diff
@@ -27,6 +27,8 @@ import {
  * @param globalConfig.facilitatorUrl - URL of the payment facilitator service
  * @param globalConfig.address - Address to receive payments
  * @param globalConfig.network - Network identifier (e.g. 'base-sepolia')
+ * @param globalConfig.apiKeyId - API key ID for the payment facilitator service
+ * @param globalConfig.apiKeySecret - API key secret for the payment facilitator service
```
John-peterson-coinbase: we should prefix these to make it clear these are for CDP

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -20,10 +36,25 @@ export function useFacilitator(url: string = "https://x402.org/facilitator") {
...
+        details: toJsonSafe(paymentRequirements),
+      },
+      {
+        headers: authOptions
+          ? {
+              Authorization: await createAuthHeader(
+                authOptions.apiKeyId,
+                authOptions.apiKeySecret,
+                url,
+                "/verify",
```
John-peterson-coinbase: the correct routes are defined here - https://github.cbhq.net/c3/cdp-api/blob/7f284eab7d6affb732a703c85203586a7f195300/openapi/openapi.yaml#L343

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -20,10 +36,25 @@ export function useFacilitator(url: string = "https://x402.org/facilitator") {
     payload: string,
     paymentRequirements: PaymentRequirements,
   ): Promise<VerifyResponse> {
-    const res = await axios.post(`${url}/verify`, {
-      payload: payload,
-      details: toJsonSafe(paymentRequirements),
-    });
+    const res = await axios.post(
+      `${url}/v2/x402/verify`,
```
John-peterson-coinbase: should we modify the url param to take `https://.../v2/x402` so that we can keep the routes here as `/verify` and `/settle` the `/v2/x402` is unique to our hosted facilitator

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -1,14 +1,30 @@
...
 
 /**
  * Creates a facilitator client for interacting with the X402 payment facilitator service
  *
  * @param url - The base URL of the facilitator service (defaults to "https://x402.org/facilitator")
+ * @param authOptions - Optional authentication options for the facilitator service
  * @returns An object containing verify and settle functions for interacting with the facilitator
  */
-export function useFacilitator(url: string = "https://x402.org/facilitator") {
+export function useFacilitator(
```
John-peterson-coinbase, line 23: should the second argument be a function to create headers that get passed along with facilitator so that we are not leaking logic specific to our facilitator service into the core method?

CarsonRoscoe: I agree generalizing this more. The challenge is making it easy enough to guide the users towards using the CDP facilitator, meaning getting their CDP api keys, while abstracting it away behind a function call.

I am content with my solution, I believe I found a healthy balance. But I encourage you to let me know if you were thinking something a little different.

typescript/packages/x402/src/shared/jwt.ts
```diff
@@ -0,0 +1,27 @@
+import { generateJwt } from "@coinbase/cdp-sdk/auth";
+import { Resource } from "../types";
+
+/**
+ * Creates an authorization header for a request to the Coinbase API.
+ *
+ * @param apiKeyId - The CDP API key ID
+ * @param apiKeySecret - The CDP API key secret
+ * @param requestHost - The host for the request (e.g. 'https://x402.org/facilitator')
```
erikreppel-cb: remove mention of CDP here, should be the way any facilitator auths

John-peterson-coinbase (approved):

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -1,14 +1,61 @@
...
+/**
+ * Creates a CDP auth header for the facilitator service
+ *
+ * @param apiKeyId - The CDP API key ID
+ * @param apiKeySecret - The CDP API key secret
+ * @returns A function that returns the auth header
+ */
+export function createCdpAuthHeaders(apiKeyId: string, apiKeySecret: string): CreateHeaders {
+  return async () => {
+    return {
```
erikreppel-cb: Generally we should avoid direct CDP references in the x402 package code, prefer createAuthHeaders

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -1,14 +1,61 @@
...
  * Creates a facilitator client for interacting with the X402 payment facilitator service
  *
- * @param url - The base URL of the facilitator service (defaults to "https://x402.org/facilitator")
+ * @param url - The base URL of the facilitator service (defaults to "https://x402.org/facilitator/v2/x402")
+ * @param createAuthHeaders - Optional function to create an auth header for the facilitator service. If using Coinbase's facilitator, use the createCdpAuthHeaders function.
  * @returns An object containing verify and settle functions for interacting with the facilitator
  */
-export function useFacilitator(url: string = "https://x402.org/facilitator") {
+export function useFacilitator(
+  url: Resource = "https://x402.org/facilitator/v2/x402",
```
erikreppel-cb: does this need to be such a long url? can we not just change the /facilitator endpoint?

erikreppel-cb (changes requested):

typescript/packages/x402/src/verify/useFacilitator.ts
John-peterson-coinbase: +1 - The current url should still be `"https://x402.org/facilitator"`

The route for the internal hosted facilitator will be `"https://api.cdp.coinbase.com/platform/v2/x402/{verify|settle}"`

CarsonRoscoe: Whether we default to `https://x402.org/facilitator` or `https://x402.org/facilitator/v2/x402` depends on what subroute we want `verify` and `settle` to look for.

i.e. https://github.com/coinbase/x402/pull/58#discussion_r2047045373
The spec for our server demands `/v2/x402/verify` and `/v2/x402/settle` as the endpoints of `https://x402.org/facilitator`. Do we expect 3rd parties to include the `/v2/x402` in the routes? if so, yes we can shorten it. If not, and we just want to enforce they have `/verify` and `/settle` routes, then we need to make our default uri longer to acomodate our extras in the routes

typescript/packages/x402/src/verify/useFacilitator.ts
CarsonRoscoe: I'm am okay with minimizing CDP references, however I do think this function is the isolated exception. It uses `@coinbase/cdp-sdk` for generating the JWT to validate against the internal hosted facilitator. It's the only place that requires explicit domain knowledge of CDP, and users will need to provide CDP keys to it.

This has been tweaked in a more recent commit. I removed all other CDP references beyond this one function so far. Please let me know if you still feel strongly about it

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -5,15 +5,55 @@ import {
...
+          "https://api.cdp.coinbase.com/platform",
+          "/v2/x402/verify",
+        ),
+      },
+      settle: {
+        Authorization: await createAuthHeader(
+          apiKeyId,
+          apiKeySecret,
+          "https://api.cdp.coinbase.com/platform",
+          "/v2/x402/settle",
```
John-peterson-coinbase: Update based on the [code example in auth package](https://github.com/coinbase/cdp-sdk/tree/main/typescript/src/auth)
```suggestion
          "api.cdp.coinbase.com",
          "/platform/v2/x402/settle",
```

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -5,15 +5,55 @@ import {
...
+ */
+export function createAuthHeaders(apiKeyId: string, apiKeySecret: string): CreateHeaders {
+  return async () => {
+    return {
+      verify: {
+        Authorization: await createAuthHeader(
+          apiKeyId,
+          apiKeySecret,
+          "https://api.cdp.coinbase.com/platform",
+          "/v2/x402/verify",
```
John-peterson-coinbase: Update based on the [code example in auth package](https://github.com/coinbase/cdp-sdk/tree/main/typescript/src/auth)
```suggestion
          "api.cdp.coinbase.com",
          "/platform/v2/x402/verify",
```

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -5,15 +5,55 @@ import {
...
  * Creates a facilitator client for interacting with the X402 payment facilitator service
  *
  * @param url - The base URL of the facilitator service (defaults to "https://x402.org/facilitator")
+ * @param createAuthHeaders - Optional function to create an auth header for the facilitator service. If using Coinbase's facilitator, use the createAuthHeaders function.
  * @returns An object containing verify and settle functions for interacting with the facilitator
  */
-export function useFacilitator(url: string = "https://x402.org/facilitator") {
+export function useFacilitator(
+  url: Resource = "https://x402.org/facilitator",
+  createAuthHeaders?: CreateHeaders,
```
John-peterson-coinbase, line 25: Once we switch over to using the internal hosted facilitator we should default both values so that we can call like

`... = useFacilitator();`

typescript/packages/x402/src/verify/useFacilitator.ts
```diff
@@ -5,15 +5,67 @@ import {
...
+  verify: Record<string, string>;
+  settle: Record<string, string>;
+}>;
+
+/**
+ * Creates a CDP auth header for the facilitator service
+ *
+ * @param apiKeyId - The CDP API key ID
+ * @param apiKeySecret - The CDP API key secret
+ * @returns A function that returns the auth headers
```
erikreppel-cb: if we're going to have CDP specific stuff, we should put it in its own export at least

CarsonRoscoe: For the sake of making movement today and not blocking the rest of the PRs, I have moved it to a separate `x402/cdp` export for now.

However, I do like your feedback yesterday about moving it into its own package entirely. I've taken note of that and want to do it in a fast-follow once the outstanding work is merged. I have a short list of fast-follows prepared

PR #61, opened 2025-04-16
feat: middleware spec update

CarsonRoscoe:
- [x] The x402 packages' verify/settle helpers have been updated to take a decoded object
- [x] The x402 middleware packages have been updated to decode the payment requirements from the user and pass them along to the facilitator
- [x] The x402 middleware packages have been updated to respond with an array of PaymentRequirements when giving a x402 error. Clients are expected to pick from the array.
- [x] The x402 axios interceptor and fetch wrapper packages have been updated to handle an array of payment requirements
- [x] Clients packages offer a selector to pick which payment requirements. If none is explicitly given, it defaults to our selector, which selects USDC by default
- [x] The default HTML paywall has been updated to expect an array of PaymentRequirements, and uses our default selector
- [x] The `extra` field should be used to request asset-related metadata required for the domain separator. If this is not set, USDC is defaulted

### Axios + Express + Facilitator with the above changes
Screenshot 2025-04-16 at 2 45 02 PM

### Fetch + Hono + Facilitator with the above changes
Screenshot 2025-04-16 at 2 45 46 PM

### TODO Tests
Once Next + Site refactor updated to include these changes, they will test the paywall addition

### DIFF
--- a/typescript/packages/x402/src/client/selectPaymentRequirements.ts
+++ b/typescript/packages/x402/src/client/selectPaymentRequirements.ts
@@ -1,0 +1,27 @@
+import { PaymentRequirements } from "../types";
+import { getUsdcAddressForChain } from "../shared/evm";
+import { getNetworkId } from "../shared/network";
+
+/**
+ * Default selector for payment requirements.
+ * Default behavior is to select the first payment requirement that has a USDC asset.
+ * If no USDC payment requirement is found, the first payment requirement is selected.
+ * 
+ * @param paymentRequirements - The payment requirements to select from.
+ * @returns The payment requirement that is the most appropriate for the user.
+ */
+export function selectPaymentRequirements(paymentRequirements: PaymentRequirements[]): PaymentRequirements {
+  const usdcPaymentRequirement = paymentRequirements.find(requirement => requirement.scheme === "exact" && requirement.asset === getUsdcAddressForChain(getNetworkId(requirement.network)));
+  if (usdcPaymentRequirement) {
+    return usdcPaymentRequirement;
+  }
+  return paymentRequirements[0];
+}
+
+/**
+ * Selector for payment requirements.
+ * 
+ * @param paymentRequirements - The payment requirements to select from.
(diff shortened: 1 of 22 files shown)


typescript/packages/x402/src/shared/json.ts
```diff
@@ -17,7 +17,7 @@ export function toJsonSafe<T extends object>(data: T): object {
    * @returns The converted value with bigints as strings
    */
   function convert(value: unknown): unknown {
-    if (value !== null && typeof value === "object") {
+    if (value !== null && typeof value === "object" && !Array.isArray(value)) {
```
CarsonRoscoe, line 20: This was needed as `toJsonSafe` was handling arrays as if they were regular objects, converting `["This"]` to `{ "0": "This" }`. See below how arrays typeof value is object, which is why it was getting caught in this first if and never falling into the separate array case

Screenshot 2025-04-16 at 2 38 23 PM

typescript/packages/x402/src/client/selectPaymentRequirements.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Default selector for payment requirements.
+ * Default behavior is to select the first payment requirement that has a USDC asset.
+ * If no USDC payment requirement is found, the first payment requirement is selected.
+ * 
+ * @param paymentRequirements - The payment requirements to select from.
+ * @returns The payment requirement that is the most appropriate for the user.
+ */
+export function selectPaymentRequirements(paymentRequirements: PaymentRequirements[]): PaymentRequirements {
+  const usdcPaymentRequirement = paymentRequirements.find(requirement => requirement.scheme === "exact" && requirement.asset === getUsdcAddressForChain(getNetworkId(requirement.network)));
```
John-peterson-coinbase, line 14: Nit: can be a follow up, but we should have the default network selector be keyed off of the connected wallet

John-peterson-coinbase (approved):

PR #75, opened 2025-04-24
feat: add payer to verify response

stat:
adding payer to verify response


CarsonRoscoe (commented):
Realizing `x402`'s `src/facilitator/facilitator.ts` file is not returning the `payer` in its response. Please add it there too

CarsonRoscoe (approved):
LGTM!

PR #165, opened 2025-05-13
feat: define 'exact' scheme for sui

bmwill:
Define the initial specification to be used for the 'exact' scheme for the Sui Blockchain.

This specification is designed to be optimally flexible for clients to be able to fully leverage PTBs (Programmable Transaction Blocks) to either provide a direct payment or perform a set of operations to swap from one asset to the asset expected by the resource server. It also provides the optionality for leveraging sponsorship/gas-less transactions.

Per the contribution guidelines this only includes a proposed specification and after its been accepted and merged we can work on providing an implementation of the spec.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

erikreppel-cb (commented):
@bmwill I think this solution is a reasonable direction

My forever concern with client signing a transaction directly from the resource server is a malicious resource server could use that as a vector to drain funds, this seems to not open that posibility because the payer is constructing the transaction, could you confirm?

Additionally, please document how to construct a transaction to transfer funds

bmwill:
> My forever concern with client signing a transaction directly from the resource server is a malicious resource server could use that as a vector to drain funds, this seems to not open that posibility because the payer is constructing the transaction, could you confirm?

Yes that is correct, the client is constructing a transaction, using the resource server's address and requested amount as inputs. Sui's data model is more closely akin to BTC's UTXO model where an Address can own a number of `Coin<T>` objects (where `T` is the particular asset/token type). For example a user could have multiple `Coin<UCSD>` objects with differing values. If a resource server wants a payment of $0.01 UCSD then the client would need to perform coin selection and pick the object(s) they want to use in their transaction and then shave off the amount requested to be sent to the destination address. Some of the `Address Balance` work described in the spec may help to simplify some of this work but would still have strong requirements that the client would be the one constructing the message/transaction to be signed.
 
> Additionally, please document how to construct a transaction to transfer funds

Transaction construction in Sui is very programatic via PTBs (Programmable Transaction Blocks). PTBs can contain up to ~1k commands or operations to perform atomically. i'm happy to outline one possible way in which a transaction can be constructed but it wouldn't be able to encompass all the possible ways to construct a transaction. Two such examples:

- Like the example above a client already has the same asset type required as payment by the resource server so they can construct a Transaction with the following commands:
  1. split $0.01 UCSD off of a `Coin<UCSD>`
  2. send the newly split coin with value $0.01 to the resource server's address
- If the client doesn't have the required asset type they could atomically swap an asset type they do have for the one that they need
  1. swap some `Coin<SUI>` (or other asset) for exactly $0.01 `Coin<UCSD>` via a liquidity pool of the client's choice
  2. send the UCSD to the resource server's address

erikreppel-cb:
@bmwill thanks for the detailed response. My last question is if there is a way for the client to construct a transaction without access to RPC? This isn't a hard requirement as I'm sure there are public Sui RPC providers that could allow for processing without creating and account, but its preferable if txs can be constructed without needing to perform reads.

I suspect the answer is no as client logic would need to determine if the client has a balance of the right `Coin<UCSD>`, but curious.

Other than that this all seems reasonable, because the differs slightly from the request flow of evm exact, if you could provide a sequence diagram of the exact proposed flow for the client, I'm happy to tentatively accept. (we'll need signoff from CB legal & security before we merge code, and Coinbase is on a recharge week this week, just to set expectations on timeline)

bmwill:
>My last question is if there is a way for the client to construct a transaction without access to RPC?

If a client is smart enough to already have stored locally the object's (their Ids, versions, and digests) then they wouldn't need to go to an RPC server. But in general yes a client would likely need to go to an RPC server to get a list of its coin objects to use in transaction construction. Some of the work that is ongoing (eg the `Address Balance` feature I mention in the spec) will improve this situation. So once that feature has been rolled out we can come back and provide an improved version of the spec.

> if you could provide a sequence diagram of the exact proposed flow for the client, I'm happy to tentatively accept.

of course. Do you have any preferred format you'd like to see for the sequence diagram? If not i'll make one up similar to the one in this repo's readme.

erikreppel-cb:
@bmwill apologies for the delay, was on vacation.

Sounds like for now the best way is via RPC. I think thats okay as a stopgap solution as long as Sui has free to use public RPC endpoints.

I'll chat with folks internally this week and run them through the design

erikreppel-cb:
another thought looking at your diagram, the diagram implies a direct hop from the client to the facilitator, is that needed during transaction construction on the client, or can that just be done when the facilitator goes to settle a transaction?

bmwill:
> another thought looking at your diagram, the diagram implies a direct hop from the client to the facilitator, is that needed during transaction construction on the client, or can that just be done when the facilitator goes to settle a transaction?

That is an optional hop if the client wants to have the facilitator sponsor the transaction and the facilitator supports doing sponsorship. Today in Sui an explicit Gas coin needs to be selected to pay for gas and so the sponsor needs to provide that to the user so that they can craft and sign the fully formed transaction.

erikreppel-cb (approved):
@bmwill have discussed with cb legal and we're comfortable with this approach contingent on a security review of the implementation, you're good to go on building!

erikreppel-cb:
@bmwill if you could sign your commits I'll merge

bmwill:
@erikreppel-cb pushed a new set of commits that i've signed. Let me know if there is anything else you need from me to get this merged.

bmwill:
@erikreppel-cb friendly ping to see if theres anything else needed to get this merged

erikreppel-cb:
@bmwill apologies, your first message skipped my inbox somehow. Running all the CI and I think we're good to go!

erikreppel-cb (approved):

PR #181, opened 2025-05-16
Add "upto"scheme specification

pawelpolak2:


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,70 @@
...
+5. Verify the spender in the `payload.authorization` is the resource server
+6. Verify the permit deadline is not expired
+7. Verify nonce is not used
+8. Verify the permit parameters are for the agreed upon ERC20 contract and chain
+9. Simulate the `permitTransferFrom` to ensure the transaction would succeed
+
+## Settlement
+
+The resource server can settle the payment by calling the `permitTransferFrom` function on the `Permit2` contract with the signature and authorization parameters from the `X-PAYMENT` header. 
+
```
erikreppel-cb, line 67: Could you expand on this section to document settlement, including an example function call in viem would be excellent

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,70 @@
...
+## Appendix
+
+The `Permit2` contract requires the user to issue an "infinite" approval for the `Permit2` contract to spend the user's tokens before it can be used, which requires an additional setup step from the user. Alternatively, we could extend this scheme to take in an "infinite" `EIP-2612` `permit` signature along with the `Permit2` signature, which would delegate the approval step to the facilitator. Unfortunately, because of a hard dependency on `msg.sender`, these two transactions cannot be batched with a regular multicall, so it would increase the first payments settlement time.
```
erikreppel-cb, line 70: So I think this is the biggest issue I have the approach for two reasons,
1. infinite approvals are imo never good, and hard to explain in a non-scary way to people new to crypto
2. having to perform a 1 time gas operation before being able to use a scheme feels like a hard UX to square, as it now requires any client using upto to have gas, which I think should be avoided.

As much as I'm a fan of permit2, I think there's a better solution to be had here, even if that means a custom contract and 2 signatures.

There's possibly a better solution here, but in my head it goes something like this.

New contract: `PermitTransfer`, has 1 function `transferPermitWithSignature`. `transferPermitWithSignature` takes in a EIP-2612 permit signature, and a second signature of a newly defined `TransferInstruction` (along with parameters of both.

- EIP-2612 signature authorizes `PermitTransfer` contract to move funds
- `TransferInstruction` directs `PermitTransfer` where funds should 
- `PermitTransfer` atomically uses the permit, then calls `transferFrom(clientAddress, resourceAddress)`

What I like about this: there's no up front operation, funds are never held in the contract for any duration of time, the client needs no gas, and there isn't now an indefinite approval to explain

pawelpolak2: Thanks for the feedback! The issue I see with using raw EIP-2612 is that it uses linear nonces which may cause the payments to fail if the client calls the resource server in parallel. I don't think there is a way to avoid this without an infinite approval. What do you think?

shafu0x: not sure why we need an infinite approval here? @pawelpolak2 @erikreppel-cb could you explain?

erikreppel-cb: agree, we should avoid an infinite approval. Permit2 typically uses an infinite approval

rsproule: feels like simplest is just: 

server calls `/settle` with the clients payment, then calls `/settle` with the surplus from its own wallet to transfer back to the client. `/settle` is just a gas sponsored transfer api 

Doing this with permit2 or an additional contract doesn't actually get you anything because you need to trust the server to "report back" with the actual price honestly, so might as well just let them do this with another 3009 sig. 

We can upstream an extension to `paymentMiddleware` where the server can just define that they want to enable surplus refunds via a flag and it will just do the double settle under the hood for them.

@shafu0x can do PR if you guys agree with this strategy

pawelpolak2: yeah i agree, i think refunds are the way to go

erikreppel-cb: @rsproule I'm not sure I understand you point. This is a same currency transfer so there isn't really a refund to give, theres no swap to another asset.

pawelpolak2: @erikreppel-cb What I'm getting is the user would sign an EIP3009 signature for the max cost of the endpoint just like in the exact scheme and then to take the actual cost of the endpoint we would take the whole upto cost and refund the excess. We can do it in a single transaction so from the perspective of a user its just like we would took less than they permitted.

pawelpolak2: Thinking about it more this would require the server to be a signer for the receiver address to issue the refunds which adds a quite a bit of setup and vulnerability.

shafu0x: @pawelpolak2 where is the vulnerability.

phdargen:
Superseded by https://github.com/coinbase/x402/pull/1074

PR #270, opened 2025-07-05
fix: Share core payment logic via PaymentMiddleware, and fix bugs in Hono/Express middlewares

ukstv:
## Description

This PR introduces a major internal and API refactor of the x402 middleware libraries across Hono, Express, and Next.js:
- Extracts core verification/settlement logic into a shared `PaymentMiddleware` class.
- Refactors existing middleware implementations to use the shared class.
- Fixes long-standing settlement timing bug in Express middleware.
- Fixes response rewriting bug in Hono middleware.
- Cleans up package dependencies to properly declare `peerDependencies`.
- Adds a higher-level `withPayment()` helper for paygating individual handlers on Next.js.

### Motivation

While building a small Next.js app with a protected GET route, I discovered a fundamental flaw: Next.js middleware cannot observe handler completion. This led to premature settlement, even if the handler failed.

To solve this in a non-middleware context (e.g., an API handler), I prototyped a wrapper that performs payment verification before executing the handler, and settlement after it completes. This became `withPayment`.

However, I ran into code duplication and tight coupling to web frameworks, which wouldn’t scale to future transports (e.g., XMTP) and currently leads to weird subtle bugs (see changes to the tests for Next middleware). So I extracted shared logic into a framework-agnostic `PaymentMiddleware` class and refactored the existing Hono and Express middlewares around it.

### New Features

**`withPayment()` API**

Ergonomic helper for wrapping individual handlers with payment verification + settlement:

```typescript
import { withPayment } from "x402-next";

export const GET = withPayment(
    async (req) => {
        return new Response("Welcome to the premium zone");
    }, 
    {
        price: "$0.01", 
        network: "base-sepolia", 
        payTo: "0xYourAddress", 
        config: {
            description: "Premium content",
        },
    }
);
```

This is especially useful in environments where full middleware support is limited (e.g. Next.js API routes).

**`PaymentMiddleware` core abstraction**

A shared, transport-agnostic class that implements:
- Payment requirement selection
- Verification (via facilitator /verify)
- Settlement (via /settle)
- transport-specific hooks to accept payments, determine if human paywall should be rendered, determine resource URL

Used internally now by Express, Hono, Next middleware, and can be reused for future transports (e.g., XMTP).

The separation of responsibilities is now clearer:
- `PaymentMiddleware` handles payment logic,
- Middleware wrappers (Express, Hono, etc.) handle actual request/response mechanics.

### Bug Fixes

**Hono**
- Before: On settlement failure, we attempted to return `c.json(...)` after the response had already been sent. That just does not work.
- After: We now properly replace `c.res` with a new `Response`, which is the correct way to override a response in Hono.

**Express**
- Before: Middleware awaited `next()`, which does not actually wait for handler completion in Express.
- After: We now monkey-patch res.end to detect when the response is available, and only then proceed to settle the payment.

**Package Dependencies**
- Before: Middleware packages incorrectly listed viem, hono, and express as hard dependencies, even when only types were used.
- After: These have been moved to peerDependencies as appropriate.

### Known Limitations
- ❗ Next.js `middleware.ts` limitation remains. Because of framework constraints, there’s still no way to wait for the page handler to finish before settling the payment. If the middleware is used directly (not via `withPayment`), payments may be settled even if the handler fails.

## Tests

Got through the usual `pnpm run test` routine. Added tests to `PaymentMiddleware`.

## Checklist

### DIFF
--- a/typescript/packages/x402-express/src/index.ts
+++ b/typescript/packages/x402-express/src/index.ts
@@ -1,25 +1,36 @@
-import { NextFunction, Request, Response } from "express";
-import { Address, getAddress } from "viem";
+import type { NextFunction, Request, Response } from "express";
+import type { Address } from "viem";
+import type { FacilitatorConfig, PaywallConfig, RoutesConfig } from "x402/types";
 import { exact } from "x402/schemes";
-import {
-  computeRoutePatterns,
-  findMatchingPaymentRequirements,
-  findMatchingRoute,
-  getPaywallHtml,
-  processPriceToAtomicAmount,
-  toJsonSafe,
-} from "x402/shared";
-import {
-  FacilitatorConfig,
-  moneySchema,
-  PaymentPayload,
-  PaymentRequirements,
-  PaywallConfig,
-  Resource,
-  RoutesConfig,
-  settleResponseHeader,
-} from "x402/types";
+import { getPaywallHtml, toJsonSafe } from "x402/shared";
+import { moneySchema, settleResponseHeader } from "x402/types";
 import { useFacilitator } from "x402/verify";
+import { PaymentMiddleware, X402Error } from "x402/middleware";
+
+type EndArgs =
+  | [cb?: () => void]
+  | [chunk: unknown, cb?: () => void]
+  | [chunk: unknown, encoding: BufferEncoding, cb?: () => void];
+
+/**
+ * Creates a deferred promise with exposed `resolve` and `reject` functions.
+ *
+ * Useful for integrating with callback-style APIs or pausing execution until
+ * an external event (like `res.end`) occurs.
+ *
+ * @returns An object containing the promise, and its associated `resolve` and `reject` functions.
+ */
+function defer<T = void>() {
+  let resolve!: (value: T | PromiseLike<T>) => void;
+  let reject!: (reason?: unknown) => void;
+
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+
+  return { promise, resolve, reject };
+}
 
 /**
  * Creates a payment middleware factory for Express
@@ -28,6 +39,7 @@ import { useFacilitator } from "x402/verify";
  * @param routes - Configuration for protected routes and their payment requirements
  * @param facilitator - Optional configuration for the payment facilitator service
  * @param paywall - Optional configuration for the default paywall
+ * @param useFacilitatorFn - Optional useFacilitator function, used in dev/testing mode
  * @returns An Express middleware handler
  *
  * @example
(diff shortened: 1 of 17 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

CarsonRoscoe (commented):
This does have a couple hidden 'breaking' interface changes with the middleware. Two that I stumbled upon while upgrading the `/examples/typesecript/express` were:

1. This PR tightens up the `resource` variable. Previously, if it was not set, it would default to a empty string. Now, it throws and error, and even if I force a empty string through it fails url validation downstream.
2. This PR tightens up the address validation

Not the worst side effects, but they do qualify as breaking changes. Will need to think about if its okay, or how we should address that

typescript/packages/x402-express/src/index.ts
```diff
@@ -73,60 +85,50 @@ export function paymentMiddleware(
...
+    payTo,
+    routes,
+    function paymentFromRequest(req) {
+      const paymentHeader = req.header("X-PAYMENT");
+      if (!paymentHeader) {
+        return undefined;
+      }
+      return exact.evm.decodePayment(paymentHeader);
+    },
+    function canRenderPaywall(req) {
```
CarsonRoscoe: I'm not a fan of the devx here regarding passing in these functions. I think the middleware can do a little more work.

In my draft, I had the middleware class have a [isWebBrowser](https://github.com/coinbase/x402/pull/272/files#diff-5e24a47550a9515125b4fad80eb71793c4dec84e3a6f7e8d7dfba8369c45ba46R451) function which took headers. This still requires that the external packages do some processing, but we are asking them to process their headers, rather than know which headers they are looking for

typescript/packages/x402/src/middleware/middleware.ts
```diff
@@ -0,0 +1,425 @@
...
+  /**
+   * Attempts to acquire and verify a payment for the request.
+   *
+   * @param request - The incoming HTTP request.
+   * @param paymentRequirements - Requirements this route expects.
+   * @returns An AcquiredPayment object on success, or undefined if paywall should be shown.
+   * @throws {X402Error} If payment is missing or invalid.
+   */
+  async acquirePayment(
+    request: TRequest,
```
CarsonRoscoe, line 413: Regarding my comment about the software doing more work, I think this is the inflection point.

Rather than taking in a TRequest, to pass into the functions, can we instead take a `headers: Record<string, string>`? From there we can process/decode the payment header ourselves, we can check if we can render the paywall ourselves, and in turn we're asking them to front a little work processing their headers, rather than defining functions to pass into the construction

typescript/packages/x402/src/middleware/middleware.ts
```diff
@@ -0,0 +1,425 @@
...
+
+  /**
+   * Attempts to acquire and verify a payment for the request.
+   *
+   * @param request - The incoming HTTP request.
+   * @param paymentRequirements - Requirements this route expects.
+   * @returns An AcquiredPayment object on success, or undefined if paywall should be shown.
+   * @throws {X402Error} If payment is missing or invalid.
+   */
+  async acquirePayment(
```
CarsonRoscoe: nit: can we rename this to `verifyPayment`? I'd like to be consistent with the verify/settle verbiage

CarsonRoscoe (commented):
One addition my #272 added while doing this refactor was allowing multi-pricing with backwards compatibility for the old way prices were declared. 

[* RouteConfig/PaymentConfig types](https://github.com/coinbase/x402/pull/272/files#diff-638c2811817919ac86b67b592580e044e318ff8b9689d719e2a9e3421c7961f1)
[* routeConfigToPaymentOptions](https://github.com/coinbase/x402/pull/272/files#diff-5e24a47550a9515125b4fad80eb71793c4dec84e3a6f7e8d7dfba8369c45ba46R156)
[* Looping over the payment options to create multiple payment requirements](https://github.com/coinbase/x402/pull/272/files#diff-5e24a47550a9515125b4fad80eb71793c4dec84e3a6f7e8d7dfba8369c45ba46R272)

While not strictly required, if you could sneak that in so this have functional parity with that PR, that'd be great :)

typescript/packages/x402/src/middleware/middleware.ts
ukstv: Yup. No problem with that at all.

ukstv:
@CarsonRoscoe  stewing over the SRP and headers got me to this point.

(1) DX is damn important, but it is a journey. We can safely bet >90% of the devs would be happy with the existing `x402-next`, `x402-hono` or `x402-express`. For them the middleware class is completely hidden.

The middleware class is for the ~1% of devs who are advanced enough to craft something special. For them, it can be more precise, more flexible, and a bit more demanding. These users often know what they're doing, or are happy to explore with the help of AI or docs.

(2) I agree, on the previous iteration, the functions passed to the middleware were repetitive in everything but reading headers.

(3) Headers vs function passing. You [said in a comment](https://github.com/coinbase/x402/pull/270#discussion_r2188806855):

> but we are asking them to process their headers, rather than know which headers they are looking for

That's a noble goal: to isolate the developer from the details of which headers are needed.

In practice, even in your PR this only works for Express, while [Next](https://github.com/coinbase/x402/pull/272/files#diff-0cc148194d248d4b89ba1943b5fc5b48d2b7d3afb767966b83127452e4e2d1c4R94-R99) and [Hono](https://github.com/coinbase/x402/pull/272/files#diff-b10fd2a8ba9bb4fcb2afd7e36a964ed8b1965a679bc0404e2f9438ef24d9d9a3R76-R80) have to know the header names.

There are subtle differences in how you read the headers between just these three. So, the current iteration still asks to provide a function, but it is now an unassuming one-liner `getHeader` passed as a parameter:
- `getHeader: (ctx, name) => ctx.req.header(name)` for Hono,
- `getHeader: (req, name) => req.headers.get(name)` for Next,
- `getHeader: (req, name) => req.header(name)` for Express.

That givies each platform its optimal access path, while removing the technicalities from both the dev and the core.
For non-HTTP transports, it's trivial to stub or adapt.

Aww, almost forgot about micro-optimisation! A function passed like this adds ~1ns to a vanilla case, whereas reconstructing a "relevant headers" object adds 50-100ns.

(3) Multi-price feature is included.

(4) Returning HTML for a paywall is still a separate function, not a class method. Putting HTML/CSS/JS in a class method makes it harder for bundlers to tree-shake. In many environments (especially edge or serverless), carrying dead code matters. This way, if you don't use the paywall HTML, it doesn't get bundled. We avoid unnecessary payload bloat by default.

(5) Had to make `routeConfigToPaymentOptions` more strict. (a) It's hard to imagine a valid payment route with no price at all. This should be considered a misconfiguration, not silently accepted. (b) Returning `[]` just defers the failure to a later point in the middleware, making it harder to trace and more confusing for the developer. Better to catch the issue early and fail loud, rather than letting it slip downstream.

(6) Re: Tightening up what is in the resource and address validation. That is a fun side effect of code deduplication. Some middlewares were more permissive, some were less. As for resource, Zod schema clearly states it is `z.string().url()`. So yes, it finally has to be a valid URL. No more silently allowing garbage or empty strings.

phdargen:
Hi @ukstv, we really appreciate the contribution! Some great bug catches in here, I think meanwhile they have all been addressed in other PRs (both in v1 and v2). In particular for next, we now have an API route wrapper that overcomes the limitations of the middleware, see https://github.com/coinbase/x402/blob/main/typescript/packages/http/next/README.md#withx402.

Please feel free to open a new PR against the v2 package if any of the issues remain

Issue #131, opened 2025-05-08
Feature Request: Add Solana support to x402


edgarpavlovsky:
To help expand payment rails for `x402`, @darkresearch would love to add Solana support to the protocol.

We're comfortable following guidelines in [CONTRIBUTING.md](https://github.com/coinbase/x402/blob/main/CONTRIBUTING.md) and expect Solana support to comply with existing (and nearly upcoming) schemes.

As a starting point, we'd love to implement `exact` scheme support. We're open to input on implementation design specifics -- the Solana ecosystem also has variety of payment infrastructure solutions that can be built on top of or from which inspiration can be taken, including:

* [Solana Pay](https://github.com/anza-xyz/solana-pay)
* [Solana Actions](https://github.com/dialectlabs/actions)

Couple considerations to kick off the conversation:

* The current `x402` implementation requires the facilitator to settle transactions -- is there room for client-side tx settlement? Given Solana speed (and relevant to other similarly performant blockchains), this could allow the facilitator to verify by tx signature on chain and support a faster process
* SPL-USDC support is likely the most interesting, but given `x402` is designed to be multi-currency we're happy to add mutlicurrency support down the line / in a separate PR.

lqk8:
Hope that X402 can support Solana—this would bring greater prosperity to the Solana ecosystem and potentially catalyze an "AI-driven economic ecosystem," such as automated resource leasing negotiations between AI Agents.

erikreppel-cb:
@edgarpavlovsky would love to collaborate!

>The current x402 implementation requires the facilitator to settle transactions -- is there room for client-side tx settlement? Given Solana speed (and relevant to other similarly performant blockchains), this could allow the facilitator to verify by tx signature on chain and support a faster process

A large reason that the facilitator settles transactions is to prevent tx reuse, without having to keep state of which transactions have been seen before, my preference would be to keep that paradigm. I don't think having the facilitator submit the tx vs the client should add much meaningful latency

>SPL-USDC support is likely the most interesting, but given x402 is designed to be multi-currency we're happy to add mutlicurrency support down the line / in a separate PR.

SPL-USDC would be a great, but we may be able to support all SPL tokens out of the box with a few tweaks (evm-exact supports any token that implements [EIP-3009](https://github.com/ethereum/EIPs/issues/3010)).

I my head exact on solana looked something like:
1. resource server responds with: `token_address`, `amount`, `fee_payer_address`, `pay_to_address`
[...]

PR #283, opened 2025-07-11
131 add solana support

notorious-d-e-v:
# Add Solana Virtual Machine (SVM) Support to x402 Protocol

## Overview

This pull request adds Solana support for X402 according to the discussion in https://github.com/coinbase/x402/issues/131 and the final [sequence diagram](https://github.com/coinbase/x402/issues/131#issuecomment-3059196377) proposed by @ngundotra.

The implementation includes an `exact` payment scheme for Solana.

### 🚀 Core SVM Implementation
- **Exact Payment Scheme for Solana**: Complete implementation of the `exact` payment scheme for Solana Virtual Machine
- **SVM Client Support**: Client-side utilities for creating and signing Solana transactions
- **SVM Facilitator**: Payment verification and settlement
- **Transaction Processing**: Full support for partially-signed transactions with fee payer sponsorship by facilitator

### 📦 Package Support
 - x402-fetch
 - x402-express
 - x402-axios (todo after feedback or in another pr)
 - x402-hono (todo after feedback or in another pr)
 - x402-next (todo after feedback or in another pr)

### 📄 Examples Support
 - clients/fetch/solana.ts
 - facilitator/express/solana.ts
 - servers/express/solana.ts
 - clients/axios (todo after feedback or in another pr)

### 🧪 Testing & Quality
- **Comprehensive Test Suite**: Unit tests for all SVM components including verification, settlement, and fee payer functionality
- **Error Handling**: Robust error handling and validation throughout the SVM implementation

### Fee Payer

In order for the facilitator to sponsor a transaction, the client needs to know the address of the facilitator's signer that will sign the transaction and pay the fee.

**Current Implementation**: Added a `/fee-payer` endpoint to the facilitator
- Made it a POST request to send `paymentRequirements` as part of the JSON body
- This allows the facilitator to return different fee payer addresses based on scheme and network in the future if necessary
- **Considerations**:
  - Could rename to `/hints` for broader use cases
  - Could simplify to GET request with query parameters for network and scheme
  - Could integrate into the `/verify` endpoint to reduce network calls
  - Benefits of `/verify` integration:
    - Future schemes may need additional hints (priority fees, blockhash, etc.)
    - Simplifies server middleware code -- only proxy the request to the facilitator verify/settle

**After Discussion and Feedback**
- Returned the fee payer as part of the `PaymentKindsResponse` from the `/supported` route on the facilitator. The facilitator can adjust the fee payer based on the network and schemes that they support.

#### Client-Side Network Calls
4 calls total while creating payment header:
- **Token Program (smart contract) Detection**: RPC call to determine if asset belongs to Token or Token-2022 program, and the decimals of the token.
- **Check if Receiver ATA Exists**: RPC call to determine if the receiver already has an ATA or whether one should be created as part of the tx instructions.
- **Simulation for gas limit**: RPC call to simulate the tx to determine the compute budget (gas limit).
- **Blockhash Retrieval**: RPC call to get blockhash for transaction expiry (validUntil) time.

#### Facilitator-Side Network Calls
3 network calls to Solana RPC during verification
- **Verify Endpoint**:
  - Lookup tables fetch for transaction decompilation during introspection
  - Associated Token Account (ATA) fetch for sender and receiver during transaction introspection
  - Transaction simulation RPC call

4 network calls to Solana RPC during settlement
- **Settle Endpoint**:
  - Calls verify again = 3 calls
  - Transaction broadcast to network

Small note here but may be a larger discussion. Is it necessary to have verify AND settle?
It seems faster and more stable for the flow to be one network call from the server to the facilitator to verify and settle.
That's what the current settle endpoint does anyway.

### Network ID Handling Considerations

Solana "network IDs" (devnet = 103, testnet = 102, mainnet = 101) may overlap with EVM network IDs.

**Open Questions**:
- Should we use [CAIP-2](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-2.md) for networks?
- CAIP-2 + CAIP-19 for assets?
- Alternative: Create our own scheme (e.g. NetworkEnum, NetworkEnum:asset) and base everything off of it

**Some Affected Components**:
- `getDefaultAsset` in `x402/src/shared/middleware.ts`
- `getUsdcAddressForChain` in `x402/src/shared/evm/usdc.ts`
- `selectPaymentRequirements` in `x402/src/client/selectPaymentRequirements.ts`
- `getNetworkId` in `x402/src/shared/network.ts`

**After Feedback & Discussion**:
- It seems like CAIP-2 + CAIP-19 will be considered for future versions
- getNetworkId will return the solana network ids for the time being and they won't clash with EVM requests since the code path only fetches the network ids after the network has been determined via the network enum.

### Transaction Introspection

**Current Support**: Only TransferChecked from token and token-2022 programs
**Open Question**: Do we want to support Transfer as well?

**After Feedback & Discussion**:
- TransferChecked is enough.
- ComputeBudgetLimit and ComputeBudgetPrice instructions have been added.
- Create ATA instructions have been added if receiver ATA does not exist

### Examples Structure

Created two different versions of extending the examples:

1. **New Directory Approach** (`examples/typescript/facilitators/svm/express`)
   - More overhead but can be more customizable
   - Allows for SVM-specific configurations and dependencies
   - Cleaner separation of concerns

2. **File Addition Approach** (`examples/typescript/clients/fetch/solana.ts`)
   - Less overhead and complexity
   - Requires importing both EVM and Solana dependencies into the same example directory
   - More compact structure

**Open Question**: which approach is preferred?

**After Feedback & Discussion**:
- we will add a `solana.ts` file and potentially `README_SVM.md` file in the existing directories instead of creating a new directory for svm.

## Testing
- [x] All existing tests pass
- [x] New SVM-specific tests added and passing
- [x] Error handling and edge case coverage

## Documentation
- Added comprehensive SVM scheme specification in `specs/schemes/exact/scheme_exact_svm.md`
- Updated examples and their READMEs

### Package Support
- [ ] x402-axios
- [ ] x402-hono  
- [ ] x402-next

### Examples (Pending Structure Decision)
- [ ] Clients
  - [ ] axios
- [ ] Servers
  - [ ] hono
  - [ ] mainnet
  - [ ] next

### Other Features
- [ ] Paywall HTML implementation
- [ ] Squads and other major programmable wallet support
- [ ] Nonce-based transaction support?

## Checklist

### Success

https://github.com/user-attachments/assets/7f44aeaa-c2f0-480c-a6c8-5e77b588a879

### Failure

https://github.com/user-attachments/assets/f865892d-d7d2-4c16-81fc-1f81242df426

---

Thank you for reading!  

It's been exciting working on this and I'm happy to adjust it as needed to fit the community's shared vision.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

typescript/packages/x402-express/src/index.ts
```diff
@@ -116,16 +125,45 @@ export function paymentMiddleware(
...
+          feePayer: "",
+        },
+      };
+      // make network call to facilitator to get the fee payer that will pay for gas
+      const feePayerResponse = await getFeePayer(tempPaymentRequirement);
+      tempPaymentRequirement.extra = {
+        ...tempPaymentRequirement.extra,
+        feePayer: feePayerResponse.feePayer,
+      };
+      paymentRequirements.push(tempPaymentRequirement);
```
ngundotra: This looks great

typescript/packages/x402/src/schemes/exact/svm/client.ts
```diff
@@ -0,0 +1,213 @@
...
+  const transferIx = await createTransferInstruction(client, paymentRequirements);
+  const rpc = getRpcClient(paymentRequirements.network);
+  const { value: latestBlockhash } = await rpc.getLatestBlockhash().send();
+  const feePayer = paymentRequirements.extra?.feePayer as Address;
+
+  return pipe(
+    createTransactionMessage({ version: 0 }),
+    tx => setTransactionMessageFeePayer(feePayer, tx),
+    tx => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, tx),
+    tx => appendTransactionMessageInstructions([transferIx], tx),
```
ngundotra: This look awesome, can you also add support for ComputeBudget + PriorityFee, and ATA creation for the destination address?

In my notes, I had an "ideal" template transaction of following form:
```
[
  ComputeBudget.setLimit, 
  ComputeBudget.setPriorityFee, 
  ATA.createTokenAccountIdempotent, 
  TokenProgram*.transferChecked({account}, to, {mint}, decimals)
]
feePayer=facilitator
signers=[facilitator, {account}]
```

In this case `facilitator` =`feePayer`.

typescript/packages/x402/src/schemes/exact/svm/client.ts
```diff
@@ -0,0 +1,213 @@
...
+ * @returns A promise that resolves to the transfer instruction
+ */
+async function createTransferInstruction(
+  client: KeyPairSigner,
+  paymentRequirements: PaymentRequirements,
+): Promise<TransferCheckedInstructionToken2022 | TransferCheckedInstructionToken> {
+  const { asset } = paymentRequirements;
+
+  const rpc = getRpcClient(paymentRequirements.network);
+  const tokenAccount = await rpc.getAccountInfo(asset as Address).send();
```
ngundotra: You should just pass `tokenAccount.decimals` to `createTransferInstructionToken` and `createTransferInstructionToken2022` to avoid duplicating RPC calls to fetch the same account.

typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts
```diff
@@ -0,0 +1,261 @@
...
+  const decodedTransaction = decodeTransactionFromPayload(svmPayload);
+  const compiledTransactionMessage = getCompiledTransactionMessageDecoder().decode(
+    decodedTransaction.messageBytes,
+  );
+  const decompiledTransactionMessage = await decompileTransactionMessageFetchingLookupTables(
+    compiledTransactionMessage,
+    rpc,
+  );
+
+  // verify that the transaction and the transfer instruction are valid
```
ngundotra: Can we verify a slightly more complicated scheme here?

```
[
  ComputeBudget.setLimit, 
  ComputeBudget.setPriorityFee, 
  ATA.createTokenAccountIdempotent, 
  TokenProgram*.transferChecked({account}, to, {mint}, decimals)
]
feePayer=facilitator
signers=[facilitator, {account}]
```

I think it's pretty important to have priority fee, compute budget, and ATA creation in the transaction as well.

typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts
```diff
@@ -0,0 +1,261 @@
...
+    throw new Error(`invalid_exact_svm_payload_transaction_receiver_ata_not_found`);
+  }
+
+  // verify that the source ATA exists
+  try {
+    await fetchToken(rpc, tokenInstruction.accounts.source.address);
+  } catch (error) {
+    console.error(error);
+    throw new Error(`invalid_exact_svm_payload_transaction_sender_ata_not_found`);
+  }
```
ngundotra: If we have to fetch both token accounts to meet existing facilitator implementations, can we batch the `fetches` into 1 network call?

typescript/packages/x402/src/shared/middleware.ts
```diff
@@ -89,6 +91,22 @@ export function findMatchingRoute(
...
+  if (network === NetworkEnum.SOLANA_MAINNET) {
+    return {
+      address: "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
+      decimals: 6,
+    };
+  }
+  // solana devnet
+  if (network === NetworkEnum.SOLANA_DEVNET) {
+    return {
+      address: "Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr",
```
ngundotra: Shouldn't this be devnet USDC: `4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU`?

typescript/packages/x402/src/types/shared/config.ts
```diff
@@ -21,9 +22,17 @@ export const config: Record<string, ChainConfig> = {
     usdcAddress: "0xcdf79194c6c285077a58da47641d4dbe51f63542",
     usdcName: "Bridged USDC",
   },
+  "solana-devnet": {
+    usdcAddress: "Gh9ZwEmdLJ8DscKNTkTqPbNwLNNBjuSzaG9Vp2KGtKJr" as SolanaAddress,
```
ngundotra: Isn't this `4zMMC9srt5Ri5X14GAgXhaHii3GnPAEERYPJgZJDncDU` ?

notorious-d-e-v:
Thank you for the review @ngundotra! Will be working these shortly.

typescript/packages/x402/src/types/shared/config.ts
notorious-d-e-v: Good catch thanks!

typescript/packages/x402/src/shared/middleware.ts
notorious-d-e-v: Good catch thanks!

typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts
notorious-d-e-v: Nice savings!

typescript/packages/x402/src/schemes/exact/svm/client.ts
notorious-d-e-v: Huge savings!

typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts
notorious-d-e-v: Thanks!

_Prioritization Fee_

How do you think we should go about this? There's the [getRecentPrioritizationFees](https://solana.com/docs/rpc/http/getrecentprioritizationfees) RPC method. 

I don't know how useful it will be, but perhaps it's a good starting point. 
Perhaps taking the max of the returned values.

_Creating Destination ATA_

There was some discussion on this before on #131 , Jon had suggested it happens out of band as a one time setup fee by the server/merchant https://github.com/coinbase/x402/issues/131#issuecomment-2878333129

I tend to agree, although in reality I think that the facilitator will end up providing this service to the server/merchant one way or another. 

The way I see it the server is a merchant and the facilitator is a value-add payment processor. 

I guess the question is, should this be part of the core code and invisible to the server/merchant? 

Or should this be out of band and required by the server to set up in a separate step.

After writing this out I'm leaning towards having it be part of the transaction, since it reduces friction for merchants onboarding to the protocol.

ngundotra: I think for now it should be part of the core code & invisible to the server/merchant. Assume the merchant doesn't ever close their ATA, and this is a 1x fee.

Prioritization fee should be small. There shouldn't be a wide range of priority fees on any of the accounts being used in this transaction. I think calling `getRecentPrioritizationFees` is fine, but should have backup default fee (likely that these accounts have not been touched in last 150 slots).

ngundotra:
@erikreppel-cb could we get a cursory review of the structure of this PR? It's 77 files, so want to get your feedback early!

CarsonRoscoe (commented):
Thank you for putting together this PR, and the foundation for helping out. This is looking promising to me.

I'm going to start going over the code, but first here are some high-level topics. 

### Fee Payer

I appreciate you calling this out, and you putting it in the extra field of payment requirements. That is appropriate. 

I am concerned about the need to add a new facilitator api endpoint to support this. This is a smell to me that future expansions will require similar custom work. I think we can come up with a more framework agnostic solution.

My proposal would be to enhance the [supported kinds response type](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/types/verify/x402Specs.ts#L106) (e.g. [x402.org/facilitator/supported](https://www.x402.org/facilitator/supported)) such that it can provide extra metadata around a specific network/scheme implementation. Essentially adding an `extra` or `metadata` field to `SupportedPaymentKindSchema` that the fee payer could go into. This achieves the same goal of allowing a different fee payer based on scheme and network if necessary.

You likely missed this since it isn't currently being used by the middleware, but we should add this API call to `useFacilitator`, and leverage it to discover the fee payer for a given network/scheme combination when checking what network/schemes the facilitator supports. 

cc: @erikreppel-cb thoughts?

###  "Is it necessary to have verify AND settle"

The short answer is yes. 

The long answer is that if we just settle before the server does work, the client will be wrongfully charged if the server throws an error during work. However, if we settle after doing the work and the payment fails, the server may have spent resources doing work it did not get paid for. Sandwiching the servers work between /verify and /settle in the middleware ensures that both parties are treated fairly. Servers don't work for free, but clients don't get charged for work that fails.

Mind you, this is just for our middleware implementations. Builders can always implement their own middleware that skips validation, or settles ahead of time, if needed. But our middleware should be robust and safe

### Network IDs

We use the term network ids over chain ids to avoid collisions between protocol families such as EVM and SVM. This is why we call it `base` and `base-sepolia` over listing the chain ids. In the evm implementation of the exact schema, we convert from network id to evm chain id once we actually need to use the chain id.

I suggest Solana does the same, and follows the same naming as cdp-sdk use for their network ids. That is, `solana`, `solana-devnet` and `solana-testnet` would be the network id names I'd suggest. 

As for CAIP-2/ CAIP-2 + CAIP-19, that is an interesting alternative. That may be the right long term approach, but it would affect the current EVM implementation. I will discuss this idea with the team and get back to you with our thoughts.

### Examples Structure

Personally, I lean towards the file addition approach. I think it maps more nicely to our packages. You see `x402-axios`, you go to `examples/typescript/clients/axios`, and you see the EVM and SVM examples for that package. I think that is easier to navigate for builders reading the repo

typescript/packages/x402/src/schemes/exact/svm/client.ts
notorious-d-e-v: The client now supports this.

Some notes

### RPC Requests
There are now 5 RPC requests happening before the client signs and sends the tx to the server.
It's a big number and ideally can be reduced.

Here are the RPC requests:
1. Get the token program that the asset belongs to -- used for creating the token transfer tx (Token or Token 2022).
2. Fetch the ATA of the receiver to see if it exists -- used to create a createATA instruction with the facilitator as the feePayer if receiver ATA doesn't exist.
3. getRecentPrioritizationFees -- used to get an the recent priority fees for the write-locked accounts in order to estimate a priority fee for the transaction
4. simulate the tx -- used to estimate the compute units spent in order to set a compute limit for the transaction (gas limit)
5. getRecentBlockhash -- used to get a recent blockhash which ultimately determines the expiry or validUntil time of a transaction

In my opinion, we can possibly remove requests 2 and 3.

#### Removing Request 2
For removing request 2, there could be two solutions.

First solution is to not create the ATA altogether and let it happen out of band between the server and facilitator. 

We already discussed this earlier and decided it would be a better UX for the merchant (server) if we keep this flow.

Second solution is to use the `CreateAssociatedTokenIdempotent` instruction. This is an instruction in the ATA program that will create the ATA if the account doesn't exist, or succeed silently if the ATA does exist.

This solution would come at the cost of higher tx fees, as the facilitator will have to pay the extra compute units required to run this transaction

Below are comparisons of our ideal txs with CreateTokenAccount vs CreateTokenAccountIdempotent

### Compute-unit comparison for handling the recipient’s Associated Token Account (ATA)

| # | RPC pre-check made? | Instruction sequence (on-chain order) | Does ATA already exist? | ATA instruction used | CU consumed | Example transaction |
|---|--------------------|----------------------------------------|-------------------------|----------------------|-------------|---------------------|
| 1 | **Yes** | `ComputeBudget.setLimit` → `ComputeBudget.setPriorityFee` → `TokenProgram.transferChecked` | ✅ | *None* (no ATA instruction) | **6 500** | [Solscan](https://solscan.io/tx/4MhLpeMC9mdxX3P8pnb7VfwNfuUuB73KN4xiUjo4Ebms2RgCxjPmVCVciYMM9pnH2cK8zR9vtfJom55D5Nd5W7jk?cluster=devnet) |
| 2 | **Yes** | `ComputeBudget.setLimit` → `ComputeBudget.setPriorityFee` → `ATA.createTokenAccount` → `TokenProgram.transferChecked` | ❌ | `createTokenAccount` (non-idempotent) | **30 044** | [Solscan](https://solscan.io/tx/3XyE4bzUacGh2wFbHH1vUBV9WNFNN7azr2SFboVWzsAn6TxodDskyQoUuxaFeDin9QQhBDriHaJs5BKuN1NoaAh9?cluster=devnet) |
| 3 | **No** | `ComputeBudget.setLimit` → `ComputeBudget.setPriorityFee` → `ATA.createTokenAccountIdempotent` → `TokenProgram.transferChecked` | ✅ | `createTokenAccountIdempotent` | **13 937** | [Solscan](https://solscan.io/tx/LnznCotzBURUdriuvPfKEVUSLf6hZHtFUSJdyeyohyPWXC6a2p5sE8FSHa4CAAvTHJ11Nxr55JsJZiYkHkhSR9L?cluster=devnet) |
| 4 | **No** | `ComputeBudget.setLimit` → `ComputeBudget.setPriorityFee` → `ATA.createTokenAccountIdempotent` → `TokenProgram.transferChecked` | ❌ | `createTokenAccountIdempotent` | **34 552** | [Solscan](https://solscan.io/tx/5xjVTCyRToGB55ECfYL499xm6Q5KKzFfyQjkM1Ynrd5n8qyJr2kxEH4trPrLpEnBTpmSLWFbctDKV8GnFX8hn8f3?cluster=devnet) |

For most transactions, there will exist a destination ATA (rows 1 and 3 above)
For few transactions, the destination ATA does not exist and needs to be created (rows 2 and 4 above).

Seeing the difference in units consumed between 1 and 3 above, my preference would be to go with number 1. 

Eat the RPC hit, and save more than half the price in transaction fees. 

Additionally, it may be possible to optimize the RPC requests more and more. For example, a scenario where all the RPC requests are done in 1 or 2 batches.  

#### Removing request 3
With regard to RPC request number 3 (`getRecentPrioritizationFees`) -- we may be able to remove this altogether.  

As @ngundotra mentioned before, it is rare that the write-locked accounts involved in these transactions will be in high demand per block. 

Probably 99% of the time `getRecentPrioritizationFees` will return 0 over the write-locked accounts for the previous 150 slots. 

Therefore we will be falling back to a default value 99% of the time.

It may be better to remove this RPC request and just use a default value, or make it so that the client can pass their preferred priority fee in.


### Compute Unit Price (Priority Fee)

As mentioned above, we will probably fall back to a default priority fee in most cases since `getRecentPrioritizationFees` will return 0 most of the time.

@ngundotra pointed out that Phantom currently sets a CU of 10 lamports for transfers.

This may be a good default value. It also may be too expensive.

When I first started writing this code I was using 1 lamport and seeing the tx fee 10x made me feel bad, lol.

I liked it better when my tx fee was $0.001.

In any case, the phantom default may be good.

And it may even be better to allow the client to pass in a configurable value. But perhaps we save that for later.

Additionally, even with a configurable value, we will have to add a check on the facilitator side to make sure the compute unit price set by the client is not outrageous. 

There is potential here for a malicious client to abuse a facilitator if they set too high of a compute unit price.


For the time being, it may be best to go with a default compute unit price of 10 lamports and make that a strict verification check on the facilitator.

notorious-d-e-v: Notes on the facilitator side

### RPC Requests

#### Verify

3 RPC requests total:
1. fetch lookupTables in order to decompile the transaction when doing transaction introspection
2. fetch Associated Token Account (ATA) of the sender and receiver (batched) while doing transaction introspection
3. simulate transaction

#### Settle

5 RPC requests total
1,2,3. from calling verify again.
4.  broadcasting the transaction to the network
5. Open a ws subscription with the RPC, waiting for notification regarding transaction confirmation after broadcasting the tx

notorious-d-e-v:
Thanks for the feedback @CarsonRoscoe!

> ### Fee Payer
> My proposal would be to enhance the [supported kinds response type](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/types/verify/x402Specs.ts#L106) (e.g. [x402.org/facilitator/supported](https://www.x402.org/facilitator/supported)) such that it can provide extra metadata around a specific network/scheme implementation. Essentially adding an `extra` or `metadata` field to `SupportedPaymentKindSchema` that the fee payer could go into. This achieves the same goal of allowing a different fee payer based on scheme and network if necessary.
> 
> You likely missed this since it isn't currently being used by the middleware, but we should add this API call to `useFacilitator`, and leverage it to discover the fee payer for a given network/scheme combination when checking what network/schemes the facilitator supports.
> 
> cc: @erikreppel-cb thoughts?

If there's already plumbing to get various metadata from the facilitator then it makes sense to use it.

I have some thoughts on [making the scheme more powerful](https://github.com/coinbase/x402/issues/291) but that is probably a larger discussion.  

Will wait for @erikreppel-cb thoughts here before making changes.

> ### "Is it necessary to have verify AND settle"
> 
> The short answer is yes.
> 
> The long answer is that if we just settle before the server does work, the client will be wrongfully charged if the server throws an error during work. However, if we settle after doing the work and the payment fails, the server may have spent resources doing work it did not get paid for. Sandwiching the servers work between /verify and /settle in the middleware ensures that both parties are treated fairly. Servers don't work for free, but clients don't get charged for work that fails.
> 
> Mind you, this is just for our middleware implementations. Builders can always implement their own middleware that skips validation, or settles ahead of time, if needed. But our middleware should be robust and safe

Sounds good! As you mentioned, a builder can customize this if they need to.  
Happy to keep it the way it is.

Overall my thought is that it is the merchant's responsibility to make sure their services are not failing.
If they're doing a bad job they should get a refund complaint from the customer since it is a clear signal that something is broken.

> ### Network IDs
> 
> We use the term network ids over chain ids to avoid collisions between protocol families such as EVM and SVM. This is why we call it `base` and `base-sepolia` over listing the chain ids. In the evm implementation of the exact schema, we convert from network id to evm chain id once we actually need to use the chain id.
> 
> I suggest Solana does the same, and follows the same naming as cdp-sdk use for their network ids. That is, `solana`, `solana-devnet` and `solana-testnet` would be the network id names I'd suggest.
> 
> As for CAIP-2/ CAIP-2 + CAIP-19, that is an interesting alternative. That may be the right long term approach, but it would affect the current EVM implementation. I will discuss this idea with the team and get back to you with our thoughts.

Sounds good!

> ### Examples Structure
> 
> Personally, I lean towards the file addition approach. I think it maps more nicely to our packages. You see `x402-axios`, you go to `examples/typescript/clients/axios`, and you see the EVM and SVM examples for that package. I think that is easier to navigate for builders reading the repo

Sounds good! It does feel cleaner.  

Maybe we can add a note for the user that the examples will have dependencies for different chains so if they are only using the EVM client, they can safely remove the SVM dependencies from package.json in their implementation.


Thank you for the feedback! I'll get started on most of those, and wait for Erik's feedback on the `SupportedPaymentKindSchema`.

typescript/packages/x402/src/schemes/exact/svm/client.ts
```diff
@@ -0,0 +1,423 @@
...
+  const sortedPrices = [...recentPrices].sort((a, b) =>
+    Number(a.prioritizationFee - b.prioritizationFee),
+  );
+  const percentileIndex = Math.floor(0.9 * (sortedPrices.length - 1));
+  const percentilePrice = sortedPrices[percentileIndex].prioritizationFee;
+  const bidPrice = Math.ceil(Number(percentilePrice));
+
+  if (bidPrice === 0) {
+    return 10_000_000; // default to 10,000,000 microlamports, same as phantom wallet
+  }
```
ngundotra: This should be set to a default of 1 microlamport to reduce cost on facilitator side. 

I think having a dynamic methodology should come later for high throughput facilitators that are processing large # of transactions. Those facilitators have an explicit interest in guaranteeing the long tail of transfers land. These apps (like Phantom) are very sensitive to even 0.1% of transfers failing, as it results in worse UX.

typescript/packages/x402/src/schemes/exact/svm/client.ts
```diff
@@ -0,0 +1,423 @@
...
+ * This function will create a transfer instruction for a token created
+ * by the token program: https://github.com/solana-program/token/
+ *
+ * @param client - The signer instance who's tokens will be debited from
+ * @param paymentRequirements - The payment requirements
+ * @param decimals - The decimals of the token
+ * @returns A promise that resolves to the transfer instruction
+ */
+async function createTransferInstructionToken(
+  client: KeyPairSigner,
```
ngundotra: These two methods can be one function, can just take a single token program as argument

typescript/packages/x402/src/schemes/exact/svm/facilitator/settle.ts
```diff
@@ -0,0 +1,231 @@
...
+    assertIsTransactionMessageWithBlockhashLifetime(decompiledTransactionMessage);
+
+    // add the blockhash lifetime to the signed transaction
+    const signedTransactionWithBlockhashLifetime = {
+      ...signedTransaction,
+      lifetimeConstraint: decompiledTransactionMessage.lifetimeConstraint,
+    };
+
+    // create the config for the transaction confirmation
+    const commitment: Commitment = "finalized";
```
ngundotra: I believe it is better to set `confirmed`, confirmations don't roll back.

typescript/packages/x402/src/schemes/exact/svm/facilitator/settle.ts
```diff
@@ -0,0 +1,231 @@
...
+      compiledTransactionMessage,
+      rpc,
+    );
+    assertIsTransactionMessageWithBlockhashLifetime(decompiledTransactionMessage);
+
+    // add the blockhash lifetime to the signed transaction
+    const signedTransactionWithBlockhashLifetime = {
+      ...signedTransaction,
+      lifetimeConstraint: decompiledTransactionMessage.lifetimeConstraint,
+    };
```
ngundotra, line 154: Is this necessary? Why wouldn't this already have a lifetimeConstraint?

notorious-d-e-v: I think this was related to typescript type hell with solana kit. 

Maybe there's another way to do it like passing it through a solana kit assert function that I don't know about.

notorious-d-e-v:
> Thanks for the feedback @CarsonRoscoe!
> 
> > ### Fee Payer
> > My proposal would be to enhance the [supported kinds response type](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/types/verify/x402Specs.ts#L106) (e.g. [x402.org/facilitator/supported](https://www.x402.org/facilitator/supported)) such that it can provide extra metadata around a specific network/scheme implementation. Essentially adding an `extra` or `metadata` field to `SupportedPaymentKindSchema` that the fee payer could go into. This achieves the same goal of allowing a different fee payer based on scheme and network if necessary.
> > You likely missed this since it isn't currently being used by the middleware, but we should add this API call to `useFacilitator`, and leverage it to discover the fee payer for a given network/scheme combination when checking what network/schemes the facilitator supports.
> > cc: @erikreppel-cb thoughts?

Hey @erikreppel-cb just want to get your thoughts / confirmation on this.

notorious-d-e-v:
@CarsonRoscoe I've went ahead and replaced the `/fee-payer` route with an `extra` field returned in the `PaymentKindsResponse` from the facilitator in the `/supported` route.

Updated the `useFacilitator` to support the `/supported` route.

I've also adjusted the examples to be files instead of directories.

The network names are now `solana` and `solana-devnet` and the network ids reflect those.

Please let me know if there's anything else.

CarsonRoscoe (commented):
Completed the initial manual testing, and I am very happy! The implementation is looking promising and well-structured. 

Screenshot 2025-08-06 at 12 11 04 PM

Below are the key feedback points and requested changes.

## Requested Changes

### 1. Example Structure Unification

#### Server Examples
- Server examples should be unified between Ethereum and Solana implementations
  - Server code is agnostic to the chain - it only deals with public addresses/keys and network definitions

#### Client Examples
- Client examples can remain separate, but I would prefer to unify them between Ethereum and Solana as well. If we can get the branching wallet logic to be very minimal, and showcase the example supporting either a evm or svm wallet via the same sdk interface, that would be ideal.
  - Client code is not agnostic to the chain - but it does use the same sdk entrypoint.

#### Facilitator Example
- Facilitator example should also be merged into one. If we can have this be one example that takes a EVM_PRIVATE_KEY and a SVM_PRIVATE_KEY, and supports base-sepolia as well as solana-devnet, that would be ideal. The learnings from here will go towards the x402.org/facilitator, such that we will have the official testnet facilitator allow builders to test on solana-devnet by default. This is why ensuring both work together cohesively is important

### 2. Signer Implementation

#### Package Level Changes
- Update `createSignerFromBase58` to support multiple private key formats in signer creation:
  - Option 1: Decoded private key
  - Option 2: Byte array input
  - Both should resolve to the same signer type

#### Example Standardization
- Currently `createSignerFromBase58` exists in:
  - Client example
  - Shared SVM folder
- Update all examples to use helpers exported from middleware
- Keep examples minimal and consistent

### 3. Documentation Updates

#### README Consolidation
- Combine separate example READMEs into core README
- Facilitator README should cover both EVM and Solana implementations
- Emphasize unified system approach:
  - Facilitators can support both EVM and SVM
  - Middleware can handle payments from either chain
  - Example: Endpoint accepting $0.10 USDC on either Base or Solana
- Avoid suggesting EVM and SVM as separate development paths. Consumers of the npm packages should be able to mix and match as needed.

### 4. Testing

#### E2E Testing
- Currently missing e2e tests for Solana, however the e2e testing package also lacks support for different protocol implementations.
- Action Item (@CarsonRoscoe ):
  - Will create PR to add Solana e2e test support
  - Will implement express/fetch test cases
- Action Item (@notorious-d-e-v ):
  - Implement e2e tests for remaining middleware and clients after support is added

## Overall Assessment

The implementation is solid and working well in practice! These requested changes are mostly around examples, testing, and minimizing duplicate code.

Thank you for all your work, this is looking fantastic, and I'm excited to get this in :)

CarsonRoscoe (commented):
Actually, am I correct in understanding we've lost `payer` in settlement response, as a result of `ExactSvmPayloadSchema` not having the equivalent `authorization.from` value to read from?

Is it possible to get the `payer` out of the `transaction` in `ExactSvmPayloadSchema`? If not, is it possible to include a payer field in `ExactSvmPayloadSchema` that the facilitator can validate is who that transaction belongs to? It would just be ideal to have the facilitator return `payer` to the resource server, as the server may choose to create a side-effect regarding it (i.e., in their own internal database, track the successful sale with that payer).

notorious-d-e-v:
Thank you for the review and feedback @CarsonRoscoe!

Question regarding the Action Items for e2e testing: do we want to get those done in this PR or in a follow-on?

Everything else is clear - I'll go ahead and get those changes in.

> Actually, am I correct in understanding we've lost `payer` in settlement response, as a result of `ExactSvmPayloadSchema` not having the equivalent `authorization.from` value to read from?
> 
> Is it possible to get the `payer` out of the `transaction` in `ExactSvmPayloadSchema`? If not, is it possible to include a payer field in `ExactSvmPayloadSchema` that the facilitator can validate is who that transaction belongs to? It would just be ideal to have the facilitator return `payer` to the resource server, as the server may choose to create a side-effect regarding it (i.e., in their own internal database, track the successful sale with that payer).

Regarding the payer in settlement response: as you mentioned, the payer can be retrieved from the transaction. 
We _do_ currently retrieve and return it to the server if verification/settlement is _successful_.

However, if verification fails before we get to checking the transfer instruction in the transaction, then `payer` is returned as an empty string. This could happen due to the following reasons during verification:

1. the `PaymentPayload` (network or scheme) does not match the `PaymentRequirement` (network or scheme), or the `PaymentPayload` network is not a supported svm network.
2. the transaction sent by the client does not match the transaction that is expected by the server/facilitator. For example, the number of instructions are different, or the required instructions are not present.
3. the receiver address (ATA) does not exist and needs to be created, but there is no instruction included in the transaction to create it.

We could add some logic to get the `payer` earlier on in the process, so that it is always available in the verification response, barring some failures like a malformed transaction, or a transfer instruction not existing inside the transaction.

We can also provide it as part of the `ExactSvmPayloadScheme`. It would have to be signed by the client's private key, and the facilitator will have to do some extra work to check that the signature resolves to the payer, and that the payer matches the sender address in the transfer instruction of the transaction.

Currently payer is marked as optional in the schema, which is why we designed it to be returned later on in the verification process. 
Should we update the docs / schema, if it is required?


>Example Standardization
>
>    Currently createSignerFromBase58 exists in:
>        Client example
>        Shared SVM folder
>    Update all examples to use helpers exported from middleware
>    Keep examples minimal and consistent

The reason I added it to the client example was so that the client doesn't have to install the x402 package in addition to the x402 fetch package.

Is it okay if the client installs both packages? 
An alternative could be to re-export the `createSignerFromBase58` function from inside the `x402-fetch`, `x402-axios`, etc client packages.

notorious-d-e-v:
Hey @CarsonRoscoe can you please check the most recent commit to see if that's what your intention was in unifying the interfaces for evm and svm?

Still have to do some work on createSigner so that it handles byte arrays, but wanted to see if the general direction so far addresses your feedback.

CarsonRoscoe (commented):
> Question regarding the Action Items for e2e testing: do we want to get those done in this PR or in a follow-on?

I suppose it can be a follow up. I'd like to get this merged in asap at this point.
As long as the E2E tests continue passing for the EVM implementation, we can merge it in and fast-follow.

> Regarding the payer in settlement response: as you mentioned, the payer can be retrieved from the transaction.
We do currently retrieve and return it to the server if verification/settlement is successful.

Noted, thank you for the clarification, I must have missed that

> The reason I added it to the client example was so that the client doesn't have to install the x402 package in addition to the x402 fetch package.

Yes, I would re-export, similar to how the x402 axios package currently re-exports the decodeXPaymentResponse function:
e.g. `export { decodeXPaymentResponse } from "x402/shared";` but for your use case

CarsonRoscoe (commented):
One concern I have is the interface for clients. 

It is valid that a server will use both SVM and EVM, and that is currently implemented, they just need to assign the right `payTo` address based on network. Easy.

It is also valid that a client will want to support both networks. For example, they may not care whether the endpoint charges them on Solana or Base, as long as its USDC. This is not currently possible.

Can we make one last change, such that the axios/fetch libraries take EITHER `Signer` OR `{ evm: evm.EvmSigner, svm: svm.SvmSigner }`, such that it is backwards compatible if you pass in one signer, but also declarative to provide up to 1 signer per implementation.

This will ripple into `selectedPaymentRequirements`, which tries to use the wallets to determine if there's a network filter. I think this Network param should be changed to `Network | Network[]`. If its a EVM-only wallet passed in, keep the current Ethereum logic for determining if a network filter is required. If it's a SVM-only wallet passed in, pass in `["solana", "solana-devnet"]` from the SvmNetworks list/mapping. If its the object structure that declared both, keep it undefined for now.

This change would then allow clients to truly support both implementations simultaneously. Whether the EVM or SVM wallet gets used when both are declared would essentially come down to the server's response and ordering of payment requirements, which I believe is appropriately agnostic. The servers' ordering essentially becomes their preferred way to receive payment, and clients are better equipped to handle both systems.

Can you get this change in @notorious-d-e-v?

Beyond that, I am aligned with all your changes. I see you have implemented `fetch`, `hono`, `next` and `express`, so I think it's just `axios` that's still missing, at which point we'll be good to merge :)

notorious-d-e-v:
Hey @CarsonRoscoe can you please take another look and provide any feedback if necessary?

notorious-d-e-v:
I'm actually thinking of making the `MultiNetworkSigner` be an object mapping `{Network: Signer, ...}` instead of `{evm: Signer, svm: Signer}` what do you think @CarsonRoscoe ? 

It will be useful especially for setting variable RPC's per network.
Example `{"base": Signer, "avalanche": Signer, "solana": Signer}`

xR0am:
This if really great work @notorious-d-e-v !

Cant wait to see this merged so I can enhance my MCP.

https://github.com/xR0am/tip-md-x402-mcp-server

CarsonRoscoe (commented):
Hey @notorious-d-e-v,

I prefer to keep the current approach as the base way we define the multi network signer.

In the current approach, there is no way for a Solana signer to be used for a Ethereum network, or vice-versa. There's an extra guard inherited from this explicit design from constraining it by families. Also, requiring it's set per-network would be a degradation in devx for circumstances where one wallet instance can satisfy many chains.

I do, however, see the benefit of allowing a custom RPC to be used for specific chains, and thus the need to assign signers per chains. However, I don't think this should replace what we have, it should be in addition to. Perhaps something like:
```ts
type MultiNetworkSigner = { evm: evm.EvmSigner; svm: svm.SvmSigner } & Partial<Record<Network, Signer>>
```
where you CAN assign a signer per network as an override, while still assigning the default for a given family?

Regardless, I'd like to do that as a fast-follow. It would be non-breaking to upgrade the interface to that if we do it right, so I am comfortable prioritizing getting this in.

Just completing a final round of testing!

CarsonRoscoe (approved):

CarsonRoscoe (approved):
Thank you so much for the contribution, fantastic work!

Merging this in! I'll fast-follow with the e2e testing (#352 should unblock), adding solana-devnet to the x402.org default facilitator, and we should potentially also follow with the network/signer changes discussed as well

ukstv:
Yahoo!!!

notorious-d-e-v:
> Thank you so much for the contribution, fantastic work!
> 
> Merging this in! I'll fast-follow with the e2e testing (#352 should unblock), adding solana-devnet to the x402.org default facilitator, and we should potentially also follow with the network/signer changes discussed as well

Woohoo! Sounds good! Big thanks to you and @ngundotra for design and feedback!

microchipgnu:
LET'S GO!

Abelhubprog:
Huge, congratulations 👏

nikshepsvn:
this is huge! will be a big unlock for the future of the agent economy

notorious-d-e-v:
> This if really great work @notorious-d-e-v !
> 
> Cant wait to see this merged so I can enhance my MCP.
> 
> https://github.com/xR0am/tip-md-x402-mcp-server

Looking forward to it!

jaycoolslm:
Awesome stuff @notorious-d-e-v

Following up on this point you raised:

> Small note here but may be a larger discussion. Is it necessary to have verify AND settle?
It seems faster and more stable for the flow to be one network call from the server to the facilitator to verify and settle.
That's what the current settle endpoint does anyway.

Has this been discussed any further? I understand your point but I'm sure this was considered in the initial design - do you know what the rationale is for the original decision?

notorious-d-e-v:
> Awesome stuff @notorious-d-e-v
> 
> Following up on this point you raised:
> 
> > Small note here but may be a larger discussion. Is it necessary to have verify AND settle?
> It seems faster and more stable for the flow to be one network call from the server to the facilitator to verify and settle.
> That's what the current settle endpoint does anyway.
> 
> Has this been discussed any further? I understand your point but I'm sure this was considered in the initial design - do you know what the rationale is for the original decision?


Yeah Carson mentions the reasoning in this comment https://github.com/coinbase/x402/pull/283#pullrequestreview-3040360843


Basically it's to try to be fair to both the buyer and seller.

PR #365, opened 2025-09-06
[WIP] Add Hedera support

jaycoolslm:
## Checklist

### DIFF
--- a/examples/typescript/clients/axios/index.ts
+++ b/examples/typescript/clients/axios/index.ts
@@ -18,13 +18,17 @@ if (!baseURL || !privateKey || !endpointPath) {
  *
  * To run this example, you need to set the following environment variables:
  * - PRIVATE_KEY: The private key of the signer
+ *   - For EVM: hex private key (0x...)
+ *   - For Solana: base58 private key
+ *   - For Hedera: ECDSA private key + account ID in format "privateKey|accountId"
  * - RESOURCE_SERVER_URL: The URL of the resource server
  * - ENDPOINT_PATH: The path of the endpoint to call on the resource server
  *
  */
 async function main(): Promise<void> {
   // const signer = await createSigner("solana-devnet", privateKey); // uncomment for solana
-  const signer = await createSigner("base-sepolia", privateKey);
+  const signer = await createSigner("hedera-testnet", privateKey); // uncomment for hedera (privateKey|accountId format)
+  // const signer = await createSigner("base-sepolia", privateKey);
 
   const api = withPaymentInterceptor(
     axios.create({
--- a/examples/typescript/clients/axios/multi-network-signer.ts
+++ b/examples/typescript/clients/axios/multi-network-signer.ts
@@ -12,19 +12,22 @@ config();
 
 const evmPrivateKey = process.env.EVM_PRIVATE_KEY as Hex;
 const svmPrivateKey = process.env.SVM_PRIVATE_KEY as string;
+const hederaPrivateKey = process.env.HEDERA_PRIVATE_KEY as string;
 const baseURL = process.env.RESOURCE_SERVER_URL as string; // e.g. https://example.com
 const endpointPath = process.env.ENDPOINT_PATH as string; // e.g. /weather
 
-if (!baseURL || !evmPrivateKey || !svmPrivateKey || !endpointPath) {
+if (!baseURL || !evmPrivateKey || !svmPrivateKey || !hederaPrivateKey || !endpointPath) {
   console.error("Missing required environment variables");
   process.exit(1);
 }
 
 /**
- * This example shows how to use the x402-axios package to make a request to a resource server that requires a payment.
+ * This example shows how to use the x402-axios package with a multi-network signer that supports EVM, SVM, and Hedera.
  *
  * To run this example, you need to set the following environment variables:
- * - PRIVATE_KEY: The private key of the signer
+ * - EVM_PRIVATE_KEY: The EVM private key (hex format, with or without 0x)
+ * - SVM_PRIVATE_KEY: The Solana private key (base58 format)
+ * - HEDERA_PRIVATE_KEY: The Hedera private key + account ID (format: "privateKey|accountId")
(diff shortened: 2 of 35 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

examples/typescript/clients/axios/index.ts
```diff
@@ -18,13 +18,17 @@ if (!baseURL || !privateKey || !endpointPath) {
  *
  * To run this example, you need to set the following environment variables:
  * - PRIVATE_KEY: The private key of the signer
+ *   - For EVM: hex private key (0x...)
+ *   - For Solana: base58 private key
+ *   - For Hedera: ECDSA private key + account ID in format "privateKey|accountId"
```
rbair23, line 23: What about ED25519 keys? I think either could work in this case.

examples/typescript/clients/axios/index.ts
```diff
@@ -18,13 +18,17 @@ if (!baseURL || !privateKey || !endpointPath) {
...
  * - RESOURCE_SERVER_URL: The URL of the resource server
  * - ENDPOINT_PATH: The path of the endpoint to call on the resource server
  *
  */
 async function main(): Promise<void> {
   // const signer = await createSigner("solana-devnet", privateKey); // uncomment for solana
-  const signer = await createSigner("base-sepolia", privateKey);
+  const signer = await createSigner("hedera-testnet", privateKey); // uncomment for hedera (privateKey|accountId format)
```
rbair23, line 30: Not sure the comment is applicable, or maybe the line should be commented out like solana-devnet is.

examples/typescript/clients/axios/multi-network-signer.ts
```diff
@@ -12,27 +12,31 @@ config();
...
 
 /**
- * This example shows how to use the x402-axios package to make a request to a resource server that requires a payment.
+ * This example shows how to use the x402-axios package with a multi-network signer that supports EVM, SVM, and Hedera.
  *
  * To run this example, you need to set the following environment variables:
- * - PRIVATE_KEY: The private key of the signer
+ * - EVM_PRIVATE_KEY: The EVM private key (hex format, with or without 0x)
+ * - SVM_PRIVATE_KEY: The Solana private key (base58 format)
+ * - HEDERA_PRIVATE_KEY: The Hedera private key + account ID (format: "privateKey|accountId")
```
rbair23, line 30: Thought: for ECDSA keys on Hedera where the alias is derived from the public ECDSA key, we can just get away with EVM_PRIVATE_KEY. The reason I make this distinction is that for users who use Hedera but may not know their account ID, they can still work with Hedera native tokens if they use EVM_PRIVATE_KEY.

rbair23: Also below it says "HEDERA_ACCOUNT_ID" is a different environment flag. Maybe what we do is say EMV_PRIVATE KEY | ED25519_PRIVATE_KEY and HEDERA_ACCOUNT_ID.

if the user supplies only the EVM_PRIVATE_KEY then we figure out the public key & EVM address, and look up the account based on EVM address. If they supply HEDERA_ACCOUNT_ID, then we use that and then use the EVM_PRIVATE_KEY for ECDSA or ED25519_PRIVATE_KEY for ed25519.

examples/typescript/servers/express/index.ts
```diff
@@ -42,6 +42,20 @@ app.use(
         // network: "solana" // uncomment for Solana mainnet
         network: "base-sepolia",
       },
+      // Hedera examples
+      "GET /hedera-usdc": {
```
rbair23, line 46: Not sure if we need to list each stablecoin as a separate endpoint?

jaycoolslm: Yeah, we can remove. It will work just by swapping out the network: base sepolia -> hedera-testnet

examples/typescript/clients/axios/multi-network-signer.ts
jaycoolslm: I actually think passing a `Client` with Operator set may be the best way. Then we have a single argument, and its up to the user create the Client... they will know if ECDSA or ED25519.

This multi-network-signer is only relevant for some use cases as well... will also need to add an example which signs bytes via wallet connect

phdargen:
Hi @jaycoolslm @jaycoolh @rbair23, we really appreciate the contribution!
Now that v2 is released, the legacy packages are effectively frozen and we're closing out PRs targeting v1. 

The first step to add a new network would be to write a document that outlines the implementation for the `exact` payment scheme, see eg for EVM: https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md.
Once this is reviewed and approved, the actual implementation can begin.

There is already a PR up for Hedera here: https://github.com/coinbase/x402/pull/792. 
It would be great if you could review it and provide feedback

PR #416, opened 2025-09-26
feat: Enable Custom RPC URLs for Solana Payments

### LINKED ISSUE #415: Support Custom RPC URLs for Solana Payments
Currently, the Solana RPC is hardcoded to use public mainnet/devnet RPC endpoints in `typescript/packages/x402/src/schemes/exact/svm/client.ts` (`getRpcClient` does not pass in its optional URL argument). 

This creates issues for building clients with helpers like `wrapFetchWithPayment` (which by default use `typescript/packages/x402/src/schemes/exact/svm/client.ts`):
- **Local development**: Cannot use `solana-test-validator` running on `localhost:8899` limiting development optoins
- **Production reliability**: Cannot use dedicated/paid RPC for production applications -- this could add latency and also will likely run into rate public rate limits at any meaningful scale

amilz:
## Description
Enabled custom RPC usage in clients interacting with Solana Payments in x402. Currently hard-coded mainnet/devnet options restrict local development and prevent production applications for scale.

- Add `getRpcClientFromRequirements()` and `getRpcSubscriptionsFromRequirements()` helper functions to reduce repetitive RPC client creation code
- Update `createPaymentHeader()` to use custom RPC URLs from payment requirements
- Add `withRpcConfig()` and `withLocalRpc()` selector utilities for client-side RPC configuration
- Export new utilities in x402-fetch and x402-axios packages

```typescript
const selector = withRpcConfig({
  solanaDevnetRpc: "http://localhost:8899",
  solanaMainnetRpc: "https://custom-rpc.com"
});
// or   const localSelector = withLocalRpc();
const fetchWithPay = wrapFetchWithPayment(fetch, wallet, maxValue, selector);
```

Fixes #415

## Tests

New tests:
- `typescript/packages/x402/src/shared/svm/rpc.test.ts` 

Added relevant tests cases to:
- `typescript/packages/x402/src/client/selectPaymentRequirements.test.ts`
- `typescript/packages/x402/src/schemes/exact/svm/client.test.ts`
- `typescript/packages/x402/src/schemes/exact/svm/facilitator/settle.test.ts`
- `typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.test.ts`

All tests passings. Run `pnpm test` from the `/typescript` directory.

## Checklist

- [X] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

erikreppel-cb (dismissed):

erikreppel-cb:
This seems reasonable to me, @notorious-d-e-v want to lend a second pair of eyes?

erikreppel-cb:
May be unrelated but python also appears to be failing

notorious-d-e-v:
Agree that this is super useful! Nice one!

I don't think we want to pass the RPC URL through the `PaymentRequirements` though.

The `PaymentRequirements` are defined by the server/merchant. 
It doesn't make sense for them to define the RPC that the client and facilitator should use.

Imo this should probably be passed into the [createPaymentHeader](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/client/createPaymentHeader.ts#L14) function of the client and the [verify](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/facilitator/facilitator.ts#L28) and [settle](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/facilitator/facilitator.ts#L73) functions of the facilitator, as part of an optional config object. 

Then that optional config object can be passed down to the respective EVM and SVM functions, e.g. [createPaymentHeaderExactEVM](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/schemes/exact/evm/client.ts#L101), [createPaymentHeaderExactSVM](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/schemes/exact/svm/client.ts#L41), [verifyExactSVM](https://github.com/coinbase/x402/blob/main/typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts#L61), etc.

The benefit of setting up this optional object is that it may be used for other things in the future.

At the same time, this rpc url may not be needed on EVM.

On EVM (using viem) the signer and the rpc connection live together in the same object, whereas they are separate in SVM (using solana kit).

The alternative would be to create a `SolanaSignerWithRPC` type that encapsulates the `KeyPairSigner` and the rpc url.
And later on in the solana functions we can use the rpc url to create the rpc connection.

Would ask for input from @CarsonRoscoe here as well.

amilz:
Thanks, @notorious-d-e-v . I like the optional object argument. Happy to work on a refactor once we lock down a direction. 

A couple of thoughts:

### 1. Verify/Settle
Strong agree the facilitator needs an optional config. I think it's okay for this it be inherited but also makes sense for the facilitator to be able to specify their own defaults as well.

### 2. Client use of RPC
New thought on client-side, looking at what we _actually_ need RPC for, I feel like we could improve performance by removing a few RPC calls:

1. [simulate](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/client.ts#L104) for CU estimation
2. [fetchMaybeAccount](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/client.ts#L209) on the ATA to see if it exists or not
3. [fetchMint](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/client.ts#L138) to check mint's token program
4. [fetch blockhash](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/client.ts#L108) for final txn (still required)

At present the scope of the payment transaction generated by the client is very narrow & predictable. We arguably should only be making the blockhash fetch. I think we could do: 

1. *Static CU Configuration*: Since every transaction gets [verified](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts#L164) in a very predicitible format, it might make more sense to set CUs as const based on config, e.g., 

```ts
  const PAYMENT_COMPUTE_UNITS = {
    tokenProgram: {
      withAta: 20_000, // example placeholder values...would need to check these
      withoutAta: 15_000,
    },
    token2022: {
      withAta: 25_000,
      withoutAta: 20_000,
    },
  };
```

2. *Idemopotent ATA Ix*: Instead of fetching the ATA to see if it exists, I'd suggest using `getCreateAssociatedTokenIdempotentInstruction` which effectively handles the same check but during [program execution](https://github.com/solana-program/associated-token-account/blob/78788c1cd3a8bf6bce9558a45a75c9530b2bb64a/program/src/processor.rs#L94). 
3. *Token Program Server Response*: Token Program is static as well -- having the client helper check token program abstracts this from the client but also adds latency to making a payment. I think better approach here might be to use `TOKEN_PROGRAM` by default and add `tokenProgram` to the `SPLTokenAmount` [here](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/types/shared/middleware.ts#L50) which could be passed as an `extra` in the server response [like done for eip72](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402-express/src/index.ts#L144).

I still like your idea to allow client to pass an RPC and Facilitator to specify their own -- just think we could also improve processing speed and devex a little by nixing the client RPC calls. Maybe can address those in a separate issue/PR but including the thought here in case it helps shape the overall direction (and happy to address them in this PR if there's support for those changes)

Also @erikreppel-cb this PR shouldn't impact the Python tests. Looks like there was some [changes to the Python CI recently](https://github.com/coinbase/x402/commit/1b12c781a71ceded0ffae09954a0ca983298268b) so maybe an issue there.

notorious-d-e-v:
@amilz sounds good!

Lots of good ideas! 

Regarding the client use of RPC -- agree it's probably better as a separate PR.

> [simulate](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/client.ts#L104) for CU estimation. 

I really like saving an RPC call by having static compute unit estimates.
Is it possible that those numbers can change down the line due to solana-side upgrades/changes?
If not then seems like a great idea!

> [fetchMaybeAccount](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/schemes/exact/svm/client.ts#L209) on the ATA to see if it exists or not

The reason we chose to go with two branches (one to create the ATA if needed, and the other one to not create the ATA) was that in most cases you won't need to create an ATA, as the ATA only needs to be created once per (merchant address, token).

So we decided to prioritize a lower network fee in most cases, compared to the one-time setup of creating the ATA (and extra RPC call).

I believe the `CreateIdempotentATA` instruction added 30,000 or so compute units to each transaction.
[Table of CU costs](https://github.com/coinbase/x402/pull/283/files/9bb91d6e8142860d04e554a30c859bbee031b706#r2221597628) 

> Token Program Server Response: Token Program is static as well -- having the client helper check token program abstracts this from the client but also adds latency to making a payment. I think better approach here might be to use TOKEN_PROGRAM by default and add tokenProgram to the SPLTokenAmount [here](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402/src/types/shared/middleware.ts#L50) which could be passed as an extra in the server response [like done for eip72](https://github.com/coinbase/x402/blob/b1fb05b573da276b52a3b0441e489f091f3d38c2/typescript/packages/x402-express/src/index.ts#L144).  

Does that mean the server will end up doing the lookup? If this then we are just moving the rpc call somewhere else.

Or is the developer expected to type that by hand? If this then we are maybe making devex worse.

What do you think?

amilz:
@notorious-d-e-v  good thoughts. I made an issue (#421) for the client optimizations and replied there.

Regarding this original PR, just confirming: we are going to wait for @CarsonRoscoe  to weigh in on RPC config as an optional object OR a new `SolanaSignerWithRPC` or something else, correct?

notorious-d-e-v:
Yep I think that's a good idea.

CarsonRoscoe (commented):
Thanks for the contribution @amilz and @notorious-d-e-v for the early eyes,

I agree with notorious dev that the RPC url should be loaded into the config, rather than be loaded in from PaymentRequirements' `extra`. Extra is useful for when the server must provide additional information required to fulfill a specific implementation of the given scheme/network tuple. So for an EVM network, `extra` has EIP712 domain information, and for a Solana network, it has the feePayer information. This extra info is required for the clients to create the signed payment header.

This RPC concern is different from that. It does not need to come from the server and is not required to fulfill an x402 request. However, it is a critical piece of infrastructure that should to be configurable. For those reasons, I believe it should be added to a config and passed through, per Notorious DEV's suggestion

To be clear, I am specifically suggesting that [createPaymentHeader](https://github.com/coinbase/x402/blob/b311fc82a3f7a8795d85e2e9f42c3b2cbd267745/typescript/packages/x402/src/schemes/exact/svm/client.ts#L41) takes in a config that includes the rpc, and passes it along to `createAndSignPayment`, who passes it along to `createTransferTransactionMessage`/`partiallySignTransactionMessageWithSigners`/`getBase64EncodedWireTransaction`. Those three functions use the config from that to discern rpc rather than `getRpcClientFromRequirements`, but still falling back to the hardcoded mainnet/testnet values if none were provided via the config.

amilz:
@CarsonRoscoe @notorious-d-e-v Thanks for the feedback.
Just pushed a refactor: removing the `PaymentRequirements.extra` and selector functions (`withRpcConfig`, `withLocalRpc`) and replaced with an optional `X402Config` parameter through function calls.

#### Changes

1. Add New Config Types
    - Created `SvmConfig` interface with optional `rpcUrl` field
    - Created `X402Config` interface with optional `svmConfig` field (extensible for future EVM config)

2. Updated Function Signatures
All functions now accept an optional config?: `X402Config` parameter:
    - `createPaymentHeader()` (top-level and SVM-specific)
    - `createAndSignPayment()`
    - `createTransferTransactionMessage()`
    - `verify()` (top-level and SVM-specific)
    - `settle()` (top-level and SVM-specific)
    - `wrapFetchWithPayment()` (x402-fetch)
    - `withPaymentInterceptor()` (x402-axios)

3. Removed the `.extra` RPC implementation

4. Updated tests

#### Example

```ts
const config: X402Config = {
  svm: { rpcUrl: "http://localhost:8899" }
};
const fetchWithPay = wrapFetchWithPayment(fetch, wallet, maxValue, undefined, config);
```

#### Outstanding question
RN the client is only using the RPC to create the payment. The RPC url is not being passed as a header to the server (which it would then need to pass to the facilitator). Would you want to add that functionality to the client/middleware or leave as-is? (The facilitator can set its own custom RPC -- I've added this to the example facilitator).

typescript/packages/x402/src/schemes/exact/svm/facilitator/verify.ts
```diff
@@ -70,10 +70,10 @@ export async function verify(
     // decode the base64 encoded transaction
     const svmPayload = payload.payload as ExactSvmPayload;
     const decodedTransaction = decodeTransactionFromPayload(svmPayload);
-    const rpc = getRpcClient(payload.network);
```
notorious-d-e-v, line 73: Nice catch!

notorious-d-e-v:
Looks great!

CarsonRoscoe (approved):
Thank you for the contribution,

It looks good to me, and I can confirm that the e2e tests are also passing

Thank you for the contribution

PR #426, opened 2025-09-29
feat: add `deferred` payment scheme spec

tmigone:
Update March 2026:
- Scheme has been reworked to better integrate with x402 v2 primitives (extensions and sdk framework)
- Feedback and comments from previous version should be all addressed.
- The scheme has been redesigned with voucher portability in mind. Sellers will hold voucher state, but optionally can rely on facilitators to offload that work if they wanted. This is facilitated by a facilitator extension `deferred-voucher-store`.
- As a result of the rework, facilitator custom behavior was reduced greatly, now relying more on verify/settle and a few extra custom endpoints.
- A minor modification to the SDK is going to be needed for the implementation. The deferred scheme requires injecting per-buyer voucher state into `PaymentRequirements.extra`. Currently, the `SchemeNetworkServer.enhancePaymentRequirements()` [hook](https://github.com/coinbase/x402/blob/9f52f9cc0efb4f85ba70162092571e7dc0ed53dc/typescript/packages/core/src/types/mechanisms.ts#L174) doesn't receive transport context (HTTP headers, etc.), which prevents schemes from accessing request-specific data. In the deferred scheme case, this data is the payer address which allows retrieving voucher data, so what we propose is to add an optional `transportContext` parameter to `enhancePaymentRequirements()`. Alternatively, we could implement the deferred scheme as a server extension, but that greatly complicates ergonomics and set up for the seller 

--- 
At @edgeandnode, we’ve been working on a new payments scheme for x402. Over the past months, we’ve met with the Coinbase team leading the broader x402 effort to ensure our approach is aligned with the spirit of x402 and contributes meaningfully to its vision.

## Description

This PR introduces the `deferred` scheme specification. A followup PR will add the sdk implementation.

>`deferred` is a scheme designed to support trust minimized micro-payments. Unlike the `exact` scheme, which requires a payment to be executed immediately and fully on-chain, `deferred` allows clients to issue signed vouchers (IOUs) off-chain, which can later be aggregated and redeemed by the seller. This scheme enables payments smaller than the minimum feasible on-chain transaction cost.

The following diagram represents the sequence diagram for a payment using the `deferred` scheme. For a detailed explanation on how the scheme works please refer to the specification documents in `spec/` directory and the project's README. The very quick TL;DR is:

- Buyer deposits funds into an escrow contract for a specific seller (can be done via signed authorization as well)
- Buyer sends signed vouchers off-chain for each payment (like writing checks)
- Vouchers aggregate - each new payment increases the total owed on the same voucher ID
- Seller collects accumulated vouchers on-chain when it's worth the gas cost

More diagrams - v3 (current)

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

desimmons:
This is a great proposal. At [Dhali](https://dhali.io/), we use XRPL payment channels to support API monetisation. We would definitely consider adopting something like this proposal within our stack. FWIW - we also tweeted about [this PR](https://x.com/dhali_io/status/1973121734330163300)

Can this PR be updated to allow support for arbitrary chains?

ukstv:
This looks suspiciously close to a unidirectional payment channel design.

tmigone:
> Can this PR be updated to allow support for arbitrary chains?

@desimmons, as long as we are talking about EVM chains it should be trivial. It's just a matter of deploying the escrow contract to the target chain. Glad to hear this could be useful for you! API monetisation seems like a good use case for this type of scheme. Curious to know if there are any pain points or missing features you see in the design that you could benefit from.

desimmons:
> > Can this PR be updated to allow support for arbitrary chains?
> 
> @desimmons, as long as we are talking about EVM chains it should be trivial. It's just a matter of deploying the escrow contract to the target chain. Glad to hear this could be useful for you! API monetisation seems like a good use case for this type of scheme. Curious to know if there are any pain points or missing features you see in the design that you could benefit from.

For us at Dhali, x402 would ideally support non-EVM chains like XRPL too.

XRPL supports payment channels at the protocol level rather than as deployed smart contracts, so I am reasonably confident this extension of x402 _could_ support something like XRPL. I don't have a good feel at all about the effort involved though.

tmigone:
> For us at Dhali, x402 would ideally support non-EVM chains like XRPL too.

> XRPL supports payment channels at the protocol level rather than as deployed smart contracts, so I am reasonably confident this extension of x402 could support something like XRPL. I don't have a good feel at all about the effort involved though.

My bad, I missed the obvious which is you were asking about XRPL 😄 The biggest hurdle I think for XRPL support would be replacing the Escrow smart contract functionality we have. I'm not super knowledgeable on XRPL, perhaps this is something that we could replicate with the built in escrow types. Definitely something to look into once the scheme and it's EVM version is mature enough.

Dhruv-2003:
interesting take, this seems similar to payment channel and i've built out that scheme as well, have it live on testnet since Jan & been collecting feedback. V1 wasn't compatible with x402, but V2 is.

if u wanna give it a read -https://docs.pipegate.xyz/methods/payment-channels

specs/schemes/deferred/scheme_deferred_evm.md
```diff
@@ -0,0 +1,314 @@
+# Scheme: `deferred` on `EVM`
+
+## Summary
+
+The `deferred` scheme on EVM chains uses `EIP-712` signed vouchers to represent payment commitments from a buyer to a seller. Before issuing vouchers, the buyer deposits funds—denominated in a specific `ERC-20` token—into an on-chain escrow earmarked for the seller. Each voucher authorizes a payment against that escrow balance, and explicitly specifies the asset being used.
```
erikreppel-cb, line 5: How are funds deposited into escrow? Always deposit authorization?

specs/schemes/deferred/scheme_deferred_evm.md
```diff
@@ -0,0 +1,314 @@
...
+- `id`: Unique identifier for the voucher (bytes32)
+- `buyer`: Address of the payment initiator (address)
+- `seller`: Address of the payment recipient (address)
+- `valueAggregate`: Total outstanding amount in the voucher, monotonically increasing (uint256)
+- `asset`: ERC-20 token address (address)
+- `timestamp`: Last aggregation timestamp (uint64)
+- `nonce`: Incremented with each aggregation (uint256)
+- `escrow`: Address of the escrow contract (address)
+- `chainId`: Network chain ID (uint256)
+- `expiry`: Expiration timestamp after which voucher cannot be collected (uint64)
```
erikreppel-cb: Who owns the logic of making sure expiry isn't used adversarially? I'd consider not have an expiry

specs/schemes/deferred/scheme_deferred_evm.md
```diff
@@ -0,0 +1,314 @@
...
+
+### Common Fields
+- `type`: Indicates whether this is a `"new"` voucher or an `"aggregation"` of an existing voucher
+- `account` (optional): Current escrow account details for the buyer-seller-asset tuple
+  - `balance`: Current escrow balance in atomic token units
+  - `assetAllowance`: Current token allowance for the escrow contract
+  - `assetPermitNonce`: Current permit nonce for the token contract
+  - `facilitator`: Address of the facilitator managing the escrow
+
+### For New Vouchers (`type: "new"`)
```
erikreppel-cb: In the event of new, if you're using a deposit authorization you likely need to include token name, decimals, etc to ensure the client can sign an authorization

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+**Response (400 Bad Request):**
+```json
+{
+  "error": "Invalid parameters"
+}
+```
+
+### POST /vouchers
```
erikreppel-cb: How is this protected? One could spam "valid" voucher signatures

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
```diff
@@ -0,0 +1,454 @@
+# Deferred Facilitator Specification
+
+## Summary
+
+This specification defines the REST API endpoints that facilitators must implement to support the `deferred` payment scheme. These endpoints enable sellers to store, retrieve, and settle vouchers through the facilitator's voucher store infrastructure. Vouchers are identified by a unique combination of `id` (64-character hex string) and `nonce` (non-negative integer).
```
erikreppel-cb: In V2 we're formalizing the concept of extensions, I'd consider this a facilitator extension. My question is can we come up with a version where someone can use deferred, while maintaining facilitator portability (not requiring state on the facilitator), and not requiring the deferred storage extension?

Based on my understanding, I think we could accomplish this via the following:
1. Make it clear servers can store their own vouchers, in which case they do not need to call the facilitator extension
2. Facilitators that want to support deferred implement the verify / settle logic, but not the storage logic
3. Settle takes extra with flags to `deposit` or `claim` for funding the escrow, or claiming funds in a voucher

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
```diff
@@ -0,0 +1,454 @@
...
+**Response (400 Bad Request):**
+```json
+{
+  "isValid": false,
+  "invalidReason": "invalid_deferred_evm_payload_signature",
+  "payer": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C"
+}
+```
+
+### POST /vouchers/:id/:nonce/settle
```
erikreppel-cb: I would consider just putting this in `/settle`, perhaps with `"extra": {"claim": true | false}`

specs/schemes/deferred/scheme_deferred_evm_voucher_store.md
```diff
@@ -0,0 +1,168 @@
+# Voucher Store Specification
+
+## Summary
+
+The Voucher Store is a critical component of the X402 `deferred` payment scheme that manages the persistence and retrieval of signed payment vouchers and their settlement records. It serves as the data layer for sellers and facilitators to track off-chain payment obligations and their eventual on-chain settlements.
```
erikreppel-cb: nit: lowercase x in x402

specs/schemes/deferred/scheme_deferred_evm_voucher_store.md
```diff
@@ -0,0 +1,168 @@
+# Voucher Store Specification
```
erikreppel-cb, line 1: I think this is more prescriptive than it needs to be and should be marked "recommended". Ultimately there is no way to enforce what happens beyond the API interface for a facilitator. A given implementation should be able to store data in which ever way they choose.

erikreppel-cb (commented):
Thanks for putting this together! Its off to a great start, I like the vouchers concept. I've left some comments on the specs to start, but this is a lot of code. Generally we prefer to have the spec PR'd first, discuss that, then implement as efforts can be wasted and large PRs are hard to review.

I think we should consider rejecting large deposits into escrow. For anything thats >$0.01 I think its safe to assume they'll use exact, and because funding is part of flow (nice work) funding an escrow should be low friction

solidity/deferred-escrow/src/DeferredPaymentEscrow.sol
```diff
@@ -0,0 +1,799 @@
...
+
+    MainStorage storage $ = _getMainStorage();
+    IDeferredPaymentEscrow.EscrowAccount storage account = $.accounts[msg.sender][seller][asset];
+
+    // Check if the requested thaw amount can be accommodated
+    uint256 newThawingAmount = account.thawingAmount + amount;
+    require(account.balance >= newThawingAmount, InsufficientBalance(account.balance, newThawingAmount));
+
+    _thaw(msg.sender, seller, asset, amount);
+  }
```
erikreppel-cb: Who calls this? this kind of functionality should also be called out in the top level spec as it affects when withdrawal can happen

specs/schemes/deferred/scheme_deferred_evm.md
```diff
@@ -0,0 +1,314 @@
...
+1. **Signature validation**: Verify the EIP-712 signature is valid
+2. **Payment requirements matching**:
+    - Verify scheme is `"deferred"`
+    - Verify `paymentPayload.network` matches `paymentRequirements.network`
+    - Verify `paymentRequirements.payTo` matches `paymentPayload.voucher.seller`
+    - Verify `paymentPayload.voucher.asset` matches `paymentRequirements.asset`
+    - Verify `paymentPayload.voucher.chainId` matches the chain specified by `paymentRequirements.network`
+3. **Voucher aggregation validation** (if aggregating an existing voucher):
+    - Verify `nonce` equals the previous `nonce + 1`
+    - Verify `valueAggregate` is equal to the previous `valueAggregate + paymentRequirements.maxAmountRequired`
```
erikreppel-cb: what happens in a race?

specs/schemes/deferred/scheme_deferred_evm_escrow_contract.md
```diff
@@ -0,0 +1,225 @@
...
+- Voucher remains valid for future collection of remaining amount
+- Prevents voucher failures due to temporary fund shortages
+
+**Note for Sellers**: Before accepting a voucher off-chain, sellers should verify that the escrow account has sufficient balance to cover the voucher amount. This can be checked using `getOutstandingAndCollectableAmount(voucher)` which returns both the outstanding amount owed and the amount that can actually be collected immediately.
+
+## Withdrawal Protection
+
+The thawing mechanism protects sellers from sudden fund withdrawals:
+
+1. **Thaw Initiation**: Buyer calls `thaw(seller, asset, amount)` (calling `thaw()` multiple times will add to the thawing amount and reset the timer)
```
erikreppel-cb, line 130: We should try to find a way to make this feel less disjoint from the rest of the flow. Suddenly a buyer needs gas to withdraw funds. minimum we should have a signature based variant of this.

specs/schemes/deferred/scheme_deferred_evm.md
tmigone: Signed deposit authorization should be the main deposit mechanism. Buyers could also do a direct interaction with the contract if they desired, but that's not the expectation. I'll clarify this in the text.

specs/schemes/deferred/scheme_deferred_evm.md
tmigone: Yes good observation, it actually applies to both new and aggregation types, it depends on the state of the buyer's balance. 

The additional data required is token domain name and version for the permit, given it's mostly static data we initially thought this could be provided client side to avoid cluttering too much the payment requirements. The client library would match the payment requirement's asset against a set of defaults (USDC) and any user provided ones and take it from there.

On hindsight this creates a few other issues like having to keep client libraries in sync with any token changes and the extra steps required client side. So I'll change it to be part of the payment requirements as well, seems more correct.

---

Another thing that this comment made me realize though is that the deposit authorization flow only works for tokens that support ERC-2612 permits and this is not made clear in the spec, need to fix that.

specs/schemes/deferred/scheme_deferred_evm.md
tmigone: So the idea of the expiry was to provide more predictable balances for the buyer, a way to ensure there are no unexpected collections happening way down the line for very old vouchers. Considering we are dealing with small amounts and we added the deposit with auth flow I think it's sensible to remove the expiry and reduce the surface for abuse.

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
tmigone: You could spam it as long as the voucher is valid, which means the voucher passes /verify logic. So you'd need to commit funds to post the vouchers.

specs/schemes/deferred/scheme_deferred_evm_voucher_store.md
tmigone: Sounds good. Would you rather remove this spec or just tag as recommended?

specs/schemes/deferred/scheme_deferred_evm_escrow_contract.md
tmigone: The flush mechanism (which was a recommendation you made) allows for a gasless thawing/withdrawing. A buyer can initiate it with a signed message. The document here is probably not making this clear enough, i'll adjust it.

solidity/deferred-escrow/src/DeferredPaymentEscrow.sol
tmigone: See my reply below, but the buyer can use the flush mechanism for a gasless version of these thaw/withdraw primitives.

I'll make sure to note this in the top level spec, ty.

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
tmigone: Just caught up with the extensions concept from v2, neat idea seems like it gives all the flexibility required. 

So if I'm getting this right, what you are proposing is:
- Formalize server storage for vouchers
- Bring back the deferred lifecycle to /verify, do work, /settle. Settle would in this case execute different actions depending on the contents of extra properties:
   - `deposit`: fund the escrow if there is a deposit authorization attached to the payment requirements
   - `store`: For facilitators with voucher store extension: store a voucher without claiming it.
   - `claim`: execute the onchain settlement for a voucher. The voucher must be provided with the payment payload (this would be used by servers storing their own).
- Servers relying on the facilitator voucher store would call /settle at any time to trigger the `claim` process described above, except they just need to provide a reference to the voucher in the payment payload, not the entire voucher object.

I think this would work and simplify things a bit but it does change a bit the meaning of the /settle endpoint. If that is OK with you we can definitely do this, we actually considered something quite similar (without the extensions part) but felt like it was modifying the intent of /settle a bit too much.

There would still be a problem however. The settle endpoint requires providing both the payment payload and payment requirements. A server relying on facilitator voucher store would not have those available when calling the endpoint in a deferred way. Payment requirements are used for verification, a voucher on the facilitators store has already passed those, so we could skip the verification steps that require access to the payment requirements and keep only those that relate to voucher validity and on-chain state (that's actually what we are doing in our current implementation with our custom endpoint to settle vouchers). 
But this would only work if payment requirements are optional for the settle endpoint. Is that something you'd be open to do?

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
tmigone: Good idea, see my long response on the other comment: https://github.com/coinbase/x402/pull/426#discussion_r2449426118

specs/schemes/deferred/scheme_deferred_evm.md
tmigone: I assume you are referring to parallel requests? In that case the current design doesn't really support it as is. The verification step will only accept nonce + 1, so any subsequent voucher reusing a nonce will get rejected.

Our assumption was that agentic usage, the main use case we see for this scheme, will not typically require parallelism, though it could be supported by generating new voucher ids. It does have it's challenges though, do you think this would be a deal breaker?

tmigone:
> I think we should consider rejecting large deposits into escrow. For anything thats >$0.01 I think its safe to assume they'll use exact, and because funding is part of flow (nice work) funding an escrow should be low friction

I like this idea! I think it could be "enforced" at the implementation level and not restricted at protocol/scheme level. For example, the [selectPaymentRequirements](https://github.com/coinbase/x402/blob/7a7e676f42839585511f136502ed717c0edd5bd7/typescript/packages/x402/src/client/selectPaymentRequirements.ts#L15) function could be wired to ignore deferred schemes if the amount to pay exceeds $0.01. wdyt?

tmigone:
@erikreppel-cb thanks for the review! Appreciate your feedback, seems like we could simplify this quite a bit if we agree on how to solve a few things. In line with this I've left some comments/questions for you on how we can move forward. 

In the meantime I have extracted the implementation from this PR so we can just focus on the spec. I'll submit another PR once we are good with this one.

specs/schemes/deferred/scheme_deferred_evm_facilitator.md
erikreppel-cb: Not quite what I was getting at.

The things I'm trying to prevent is faciliator lock in. If 1 facilitator holds all your vouchers, you are forcibly locked into using them until you clear you vouchers. What happens if that facilitator goes offline?

The voucher concept in no way necessitates that the facilitator stores the vouchers. A server could easily store as well.

By separating storage from settlement, you get the benefit where theres a clear path to facilitator portability for any server than wants to store vouchers themselves. To accomplish this you'd make the paths for claiming the voucher part of standard flow, and storage of vouchers be an optional extension.

specs/schemes/deferred/scheme_deferred_evm_voucher_store.md
erikreppel-cb: I think just tag recommended

specs/schemes/deferred/scheme_deferred_evm.md
erikreppel-cb: A lot of agents leverage parallelism for handling multiple tool calls at once. I don't think its a deal breaker but it is worth accounting for.

Ctree35:
## Aligning FluxA’s ZK‑batched model with `deferred`

Great proposal — we’ve been building **FluxA**, a trust‑minimized, open deferred settlement layer for x402 that lets clients commit now, settle later via EIP‑712 signed commitments that are batched and proven with a ZK‑SNARK, then settled on‑chain in bulk to multiple sellers. We’d like to make FluxA fully compatible with this `deferred` scheme.

Concretely, the PR defines a `deferred` flow with buyer escrow, off‑chain signed vouchers that aggregate over time, and settlement when it’s gas‑efficient. That’s conceptually aligned with FluxA; our main delta is that we batch many commitments (potentially across multiple sellers) and settle with an on‑chain ZK proof rather than posting per‑voucher proofs. Could we extend the spec to allow a “batched/zk” variant of `deferred` with minimal changes?  ￼

Key places we’d propose small additions:
- 402 “Payment Required Response” → add new type in `extra` field: allow a discriminator so the server/facilitator advertises whether it supports _single‑voucher increment_ or _ZK‑batched commitments_.
- /verify and /deferred/vouchers flow: keep as‑is for single vouchers, and add an optional **bundle submission** path for ZK batches at settlement time.  ￼

Happy to open a follow‑up PR with these minimal extensions. 

—Chris from FluxA

desimmons:
We've put in a substantial amount of work to make [Dhali](https://dhali.io/) EVM compatible. We now support streaming USDC, RLUSD, and ETH on Ethereum (the primary use-case obviously being API monetisation). To achieve this, we developed this [smart contract](https://github.com/Dhali-org/payment-channel-evm). If anyone has any feedback on what we've done, that would be great!

phdargen:
Closed by https://github.com/x402-foundation/x402/pull/2051 
Thanks a lot for all your input!

PR #537, opened 2025-10-28
Add `channel` scheme (ledger-agnostic payment channel receipts)

jolestar:
## Summary
This PR introduces a new payment scheme `channel` to x402, defining a chain-agnostic, receipt-based payment flow that supports high-frequency, usage-based, and streaming monetization. It decouples per-request authorization from on-ledger settlement, enabling asynchronous and batched settlement while preserving cryptographic verifiability at request time.

**Documents**:
- Core spec: `scheme_channel.md` - Chain-agnostic protocol (normative)
- Rooch binding: `scheme_channel_rooch.md` - Reference implementation integration
- EVM binding: `scheme_channel_evm.md` - Experimental proposal

## Motivation
The existing `exact` scheme (EVM/EIP-3009) is ideal for fixed-amount, per-request payments with immediate on-ledger settlement. However, many modern use cases (AI agents, LLM streaming, pay-per-token, toolchains) require:
- High-frequency micropayments with minimal latency impact
- Usage-based pricing determined post-flight (after work is done)
- Asynchronous/batched settlement to reduce on-chain overhead

A channel/receipt model addresses these needs by exchanging signed receipts per request, while allowing settlement to be deferred and aggregated.

## Compatibility
- Non-breaking: Does not modify the `exact` scheme or core types.
- Facilitators can advertise support for `(scheme="channel", network=...)` alongside existing `(scheme="exact", ...)` pairs.
- Networks/ledgers integrate via binding documents; they can evolve independently.

### Core spec (`scheme_channel.md`)
- **§2 "How it works"**: Sequence diagram front-loaded for quick understanding of the protocol flow
- **Core concepts (§3-6)**: Focus on protocol understanding (~120 lines)
  - Channel, sub-channel, epoch, receipt, proposal, delta
  - Data model and invariants
  - Security considerations
- **Implementation details (Appendices A-D)**: For implementers (~90 lines)
  - State management, verification protocol, settlement specification
  - Detailed interaction flows

### Network bindings
- **Rooch binding** (`scheme_channel_rooch.md`): Reference implementation validated on testnet
- **EVM binding** (`scheme_channel_evm.md`): Experimental proposal for discussion

See "Network bindings status" section below for details.

## Alternatives considered
- Extending `exact` to support usage-based payments: this reintroduces ledger specifics and does not naturally model accumulated receipts with sub-stream isolation.
- Pushing everything into `deferred`: strong for card/bank rails, but a cryptographic, chain-friendly receipt format is still valuable for ledgers and hybrid deployments.

## Comparisons
- vs `exact` / EIP-3009: `exact` encodes a specific amount and expects immediate settlement; `channel` encodes accumulated totals and supports deferred/batched settlement.
- vs bidirectional/multiparty channels: Unidirectional simplifies lifecycle and disputes, which fits provider billing; bindings can still implement richer lifecycles as needed.

## Next steps
- Community feedback on binding structure and receipt canonicalization rules.
- Additional bindings (EVM implementation, Solana, etc.) contributed by respective ecosystems.
- **EVM binding**: Seeking feedback from EVM developers on Hub security model, gas optimization strategies, and contract architecture.

### Rooch binding (`scheme_channel_rooch.md`)
- **Status**: Reference implementation available and validated on Rooch testnet
- **Context**: Maps existing Rooch payment channels to x402 protocol
- **Key features**:
  - DID-based identity (`did:rooch`) with verification method fragments
  - BCS-serialized SubRAV for on-chain verification
  - Type-safe asset binding via Move generics `<CoinType>`
  - Dual operation modes: client-initiated and facilitator-proxied (x402 aligned)
- **Resources**:
  - Contracts: [payment_channel.move](https://github.com/rooch-network/rooch/blob/main/frameworks/rooch-framework/sources/payment_channel.move)
  - SDK: [payment-kit](https://github.com/nuwa-protocol/nuwa/tree/main/nuwa-kit/typescript/packages/payment-kit)

### EVM binding (`scheme_channel_evm.md`)
- **Status**: Early-stage experimental proposal (no implementation or testing)
- **Purpose**: Spark discussion and gather feedback from EVM developers
- **Architecture sketch**:
  - EIP-712 typed structured data for signatures
  - Settlement contract + Payment Hub (pooled collateral)
  - EOA/EIP-1271 support with optional DID
  - Lazy channel open pattern
- **Open questions**: Gas optimization, Hub security model, key management, dispute resolution, cross-L2 channels


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

MicahZoltu (commented):
In the sequence diagram and description, there is no mention of when/where the client signs a transaction destined for the ledger that actually puts money into the payment channel.

This feels like it should happen as the first ledger transaction, as the server/facilitator are reputation owning entities while the client has no reputation, so they should be the one who bears the initial cost of instantiation of the channel.

----

I wonder if there would be value in standardizing a "rate limit" on payment channel draw down?  If a person is watching streaming videos, or using an LLM, there is presumably a reasonable estimate on how quickly the client can burn through the credits.  If this rate limit were codified into the payment channel, the user could be protected against a bug, malicious app, or backend hacker instantly draining the funds.  The client's tooling (e.g., browser) would present this rate limit to the user up front letting them know the fastest that their assets can be burned down so the user can have some confidence/protection that their whole channel won't automatically be liquidated, while simultaneously avoiding the need for the end-user to manually approve every micro transaction.

As a user of a third party sketchy LLM, I want confidence in a few things:
1. If the page isn't active (tab in foreground), then I'm not paying anything.  My browser can provide this guarantee for me.
2. I don't have to click a button every time I receive a token from the LLM.  My browser can automatically sign each payment channel update, as fast as round trip time to the server.
3. The app cannot instantly drain my channel once it is opened by sending a payment update that is unreasonable (which my browser would auto-sign).  My browser doesn't have enough information with the current spec to know what a "reasonable" rate limit is, but the server could communicate that information during channel setup so the browser can present it to the user and correctly rate limit to whatever I configured.

jolestar:
Thanks @MicahZoltu

1) Channel funding / who pays the first on-chain tx
- We plan to support an “exact + channel” bootstrap: the very first interaction can use `exact` to pay the one-off cost to instantiate/fund (ledger-dependent), then switch to `channel` receipts for all subsequent requests. This avoids a separate funding tx flow and makes “who pays the first gas” explicit (client signs once via `exact`).
- I’ll add a “Setup phase” before §2 in the core spec noting this optional bootstrap path, leaving details to bindings.

2) Avoiding fragmented balances across channels
- Bindings will recommend a pooled balance (“micropayment wallet/Hub”) per payer+asset. Multiple channels draw from the same pool, so no stranded funds across channels.
- Rooch binding (implemented) and EVM binding (proposal) both describe pooling semantics. I’ll surface this in the core doc as a binding-level recommendation and keep normative details in bindings.

3) On “rate limit” (raising the problem first)

Rate limiting is tricky to standardize across apps. Different workloads burn at very different rates (e.g., LLMs can vary a lot by model/price). 

Two distinct knobs (client vs server):
- client.maxAmount (client-side): a local per-receipt cap. The client simply refuses to sign a receipt above this value.
- paymentRequirements.maxAmountRequired (server-side): a server-declared upper bound for a single step that verifiers enforce.

If we add a “rate limit”, it should be client-side (like client.maxAmount), not a server-provided config. A browser/SDK MAY expose something like:
- client.rateLimit = { maxPerWindow, windowSeconds }  // purely local throttling

For LLM scenarios, clients MAY do scenario-specific checks (e.g., local token counting vs server-quoted cost) to catch obvious spikes, but this is application-specific and better documented as best practices rather than standardized.

MicahZoltu:
When you say "client" here, do you mean the trusted browser or do you mean the untrusted app that is served to the browser for rendering and local execution?

I think you mean "client" as in a trusted agent operating on behalf of the user (e.g., a browser, curl, etc.).  If so, then the client doesn't have enough information to suggest reasonable rate limits for a given payment request.  As you mentioned, every LLM likely charges a different amount per token, same with video streaming services per second, etc.  In order for the client (generic trusted agent working on behalf of the user) to present the user with a meaningful payment channel request, they need to let the user know the expected burn rate.  The client, without input from the counterparty, cannot guess what a reasonable rate limit would be.

phdargen:
Closed by https://github.com/x402-foundation/x402/pull/2051

PR #648, opened 2025-11-17
Merge cardano x402 into v2-development

fabianbormann:
_We are adding this as a draft PR first. The draft status will be removed after internal alignments._

## Description

This PR adds the exact schema for Cardano to allow an integration of x402 with the Cardano blockchain.

## Checklist

- [X] My commits are signed (required for merge)


fabianbormann:
Hi x402 team, should we create the PR against `main` rather than `v2-development`? Also, we’re happy to jump on a quick call to walk you through the schema and Cardano dynamics to help with the review process. Let us know if that would be useful!

cc: @erikreppel-cb @CarsonRoscoe @phdargen

phdargen:
Hi @fabianbormann, yes please target `main` and update your specs to align with the [v2 specs](https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md), also see the migration guide here: https://docs.x402.org/guides/migration-v1-to-v2

fabianbormann:
@phdargen thank you for the quick response! I've created a new [PR](https://github.com/coinbase/x402/pull/1093) and therefore I will close this one. I've applied the migrations for v2 already, but a few things may work different with Cardano compared to EVM or SVM. The user will send the transaction already pre-built to the server and it already includes signature and validity intervals. We don't need to grant access to contracts as the assets are native assets under Cardano which behave like Ada and can freely circulate and doesn't need contract executions. Contracts under Cardano are validators that just validate if a given transaction is valid or not according to some logic, which removes a bit of workload from the facilitator. Feel free reaching out in case anything remains unclear. Thank you!

PR #659, opened 2025-11-20
fix: streaming responses in express middleware

phdargen:
## Description

**Problem:**
The Express payment middleware only intercepts `res.end()` (internally called by `res.send()`), which allows streaming responses like `res.write()` to send data to the client before payment settlement completed

**Fix:**
- Intercept all response methods that can commit data to the client: writeHead, write, end, and flushHeaders
- Buffer all calls to these methods until payment settlement completes
- After successful settlement, replay all buffered calls in the original order
- Updated test mocks to include the newly intercepted methods

## Tests

Tested with modified servers/express example:

```
// Normal endpoint as sanity check
app.get("/weather", (req, res) => {
  res.send({
    report: {
      streaming: "false",
      weather: "sunny",
      temperature: 70,
    },
  });
});

// Streaming endpoint using res.write() and res.writeHead() 
app.get("/streaming-weather", (req, res) => {
  res.writeHead(200, { "Content-Type": "application/json" });
  res.write(
    JSON.stringify({
      report: {
        streaming: "true",
        weather: "sunny",
        temperature: 70,
      },
    }),
  );
  res.end();
});

```

Fetch client output before:

```
{ report: { streaming: 'false', weather: 'sunny', temperature: 70 } }
{
  success: true,
  transaction: '0x8dc480d3e238eb2d77c28a8ecf5dabe37a70f718755501734edc159a7709658f',
  network: 'base-sepolia',
  payer: '0xE33A295AF5C90A0649DFBECfDf9D604789B892e2'
}

{ report: { streaming: 'true', weather: 'sunny', temperature: 70 } }
SyntaxError: Unexpected token 'e" is not valid JSON
```
-> client receives no payment header

And server throws: `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`

Fetch client output after fix:

```
{ report: { streaming: 'false', weather: 'sunny', temperature: 70 } }
{
  success: true,
  transaction: '0x150565ae9a8af4170b6328bd4661e89233d0a7ea5ed3b6a94e48a7a87ef0af07',
  network: 'base-sepolia',
  payer: '0xE33A295AF5C90A0649DFBECfDf9D604789B892e2'
}

{ report: { streaming: 'true', weather: 'sunny', temperature: 70 } }
{
  success: true,
  transaction: '0x3f4a942271d23d713b31d60ef2deafe570ddb579ef4e17bf4086f6053fff62df',
  network: 'base-sepolia',
[...]

PR #687, opened 2025-11-26
tests:  adding tests to verify that the middleware correctly delivers content based on settlement results

souhailaS:
## Description

In the FastAPI payment middleware (i haven't tried the other ones) there is an issue that makes  response content being sent to clients before payment settlement completed, which can allow clients to receive paid content for free if settlement failed for whatever reason. The `response = await call_next(request)`returned  a `StreamingResponse` that begins transmitting data immediately. So, by the momen the settlement successes is being being verified, it is already too late. the response headers and body have already been sent to the client.

In this fix I suggest to buffering the response body in memory after verification instead and to only `return Response(content=response_body, ...)` after verification of settlement is succeful, then if settlement fails, a 402 Payment Required error is returned without delivering any content. 

There are no breaking changes in this fix

## Tests

I added the three tests i used in the test_middleware.py
## Checklist

- [x]  My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

phdargen:
Hi @souhailaS, thanks a lot for looking into this! We observed similar issues for the `express` and `flask` middlewares (https://github.com/coinbase/x402/pull/659, https://github.com/coinbase/x402/pull/693), so its possible that `fastapi` is impacted as well.

However, it is my understanding that the `StreamingResponse` object returned by `response = await call_next(request)`
is only consumed when the response is actually being sent, which happens after the middleware returns.

I tested the following protected endpoints and forcing settle_response.success = False in the fastapi middleware:
```
@app.get("/weather")
async def get_weather() -> Dict[str, Any]:
    return {
        "report": {
            "weather": "sunny",
            "temperature": 70,
        }
    }

@app.get("/stream")
async def paid():
    async def generate():
        yield b"chunk 1\n"
        await asyncio.sleep(0.1)
        yield b"chunk 2\n"
        await asyncio.sleep(0.1)
        yield b"final chunk\n"

    return StreamingResponse(generate(), media_type="text/plain")
```
I could not observe any leaked content at the client and instead received the 402 error response as intended.

If you’ve actually observed leaked content, it would be much appreciated if you could provide a minimal example that shows a client receiving any part of the paid response before settlement is completed.

Even if such a vulnerability does exist, just buffering the response would likely not be sufficient to fix it and one would likely need to intercept response methods and replay after settlement as is done in the express/flask middleware.

phdargen:
I also tested your new tests and they pass with the middleware unchanged.

In any case, your added unit tests fill an important gap in coverage for the payment settlement logic and we would be happy to merge those.

souhailaS:
@phdargen  thanks for looking into it, and you're absolutely right! Sorry for the wrong alarm,  I did run a few tests and indeed the `call_next` is not like in express where the route handler can immediately start writing response data during next() before the middleware regains control .. I fixed removed the buffering from the PR and kept the tests

PR #721, opened 2025-12-05
fix(java): skip settlement for error responses (4xx/5xx)

madisoncarter1234:
## Summary

Adds response status code check before settlement in Java PaymentFilter. If the handler returns a 4xx or 5xx status code, settlement is skipped to prevent charging users for failed requests.

## Problem

Currently the Java middleware settles payments regardless of whether the handler succeeded or failed. If the handler returns a 500 error, the user is still charged even though they didn't receive the content they paid for.

Go middleware already checks `writer.statusCode >= 400` before settlement.
Python middleware already checks `response.status_code >= 200 and response.status_code < 300` before settlement.

## Changes

- Added `response.getStatus() >= 400` check before calling `facilitator.settle()`
- Added unit tests for 4xx and 5xx response scenarios
- Updated existing tests to mock `resp.getStatus()` appropriately

## Related

- #720 - Go equivalent fix
- #707 - TypeScript settlement failure handling


phdargen:
Duplicate of https://github.com/coinbase/x402/pull/723

PR #732, opened 2025-12-08
Add scheme for ERC-7710

danfinlay:
## Description

Adds a new payment scheme utilizing ERC-7710 account permissions. Allows payment in any token from accounts that expose this simple and flexible interface. Also allows payments to be made by permitted agents, including AI agents.

## Tests

Just a specification for now, but tests can be added in the implementation phase.

## Checklist

### DIFF
--- a/specs/schemes/exact/scheme_exact_evm.md
+++ b/specs/schemes/exact/scheme_exact_evm.md
@@ -6,14 +6,15 @@ The `exact` scheme on EVM executes a transfer where the Facilitator (server) pay
 
 This is implemented via one of two asset transfer methods, depending on the token's capabilities:
 
-| AssetTransferMethod | Use Case                                                     | Recommendation                                 |
-| :------------------ | :----------------------------------------------------------- | :--------------------------------------------- |
-| **1. EIP-3009**     | Tokens with native `transferWithAuthorization` (e.g., USDC). | **Recommended** (Simplest, truly gasless).     |
-| **2. Permit2**      | Tokens without EIP-3009. Uses a Proxy + Permit2.             | **Universal Fallback** (Works for any ERC-20). |
+| AssetTransferMethod | Use Case                                                     | Recommendation                                 | Usage Semantics                     |
+| :------------------ | :----------------------------------------------------------- | :--------------------------------------------- | :---------------------------------- |
+| **1. EIP-3009**     | Tokens with native `transferWithAuthorization` (e.g., USDC). | **Recommended** (Simplest, truly gasless).     | One-time use                        |
+| **2. Permit2**      | Tokens without EIP-3009. Uses a Proxy + Permit2.             | **Universal Fallback** (Works for any ERC-20). | One-time use                        |
+| **3. ERC-7710**      | Smart accounts with delegation support.                              | **Smart Account Option** (Paid from ERC-7710 compatible account). | One-time use and multi-use |
 
 If no `assetTransferMethod` is specified in the payload, the implementation should prioritize `eip3009` (if compatible) and then `permit2`.
 
-In both cases, the Facilitator cannot modify the amount or destination. They serve only as the transaction broadcaster.
+In all cases, the Facilitator cannot modify the amount or destination. They serve only as the transaction broadcaster.
 
 ---
 
@@ -201,6 +202,114 @@ Settlement is performed by calling the `x402ExactPermit2Proxy`.
 
 ---
 
+## 3. AssetTransferMethod: `ERC-7710`
+
+This asset transfer method uses [ERC-7710](https://eips.ethereum.org/EIPS/eip-7710) smart contract delegation to authorize transfers from accounts that support the standard. It is particularly suited for smart contract accounts (e.g., ERC-4337 accounts, ERC-7579 modular accounts) that have enabled delegation capabilities.
+
+### Prerequisites
+
+For ERC-7710 to work, the following must be true:
+
+1. **Delegator Account**: The payer's account must be a smart contract that supports ERC-7710 delegation (e.g., a modular smart account with delegation capabilities).
+2. **Delegation Manager**: A `DelegationManager` contract implementing the `ERC7710Manager` interface must be deployed on the network.
+3. **Active Delegation**: The payer must have created a delegation authorizing the delegate to execute token transfers on their behalf, with appropriate caveats (amount limits, recipient restrictions, etc.).
+
+### Phase 1: Obtaining a Delegation
+
+The process of obtaining a delegation is outside the scope of x402. Delegations may be obtained through:
+
+- [ERC-7715](https://eips.ethereum.org/EIPS/eip-7715) permission requests
+- Direct wallet interactions
+- Pre-configured session keys
+- Other delegation protocols
+
+The key requirement is that the client is able to issue a delegation to the facilitator that permits the required token transfer.
+
+### Phase 2: `PAYMENT-SIGNATURE` Header Payload
+
+The `payload` field must contain:
+
+- `delegationManager`: The address of the ERC-7710 Delegation Manager contract.
+- `permissionContext`: The delegation proof/context required by the specific Delegation Manager implementation.
+- `delegator`: The address of the account that created the delegation.
+
+**Example PaymentPayload:**
+
+```json
+{
+  "x402Version": 2,
+  "resource": {
+    "url": "https://api.example.com/premium-data",
+    "description": "Access to premium market data",
+    "mimeType": "application/json"
+  },
+  "accepted": {
+    "scheme": "exact",
+    "network": "eip155:84532",
+    "amount": "10000",
+    "asset": "0x036CbD53842c5426634e7929541eC2318f3dCF7e",
+    "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
+    "maxTimeoutSeconds": 60,
+    "extra": {
+      "assetTransferMethod": "erc7710",
+      "name": "USDC",
+      "version": "2"
+    }
+  },
+  "payload": {
+    "delegationManager": "0xDelegationManagerAddress",
+    "permissionContext": "0x...",
+    "delegator": "0x857b06519E91e3A54538791bDbb0E22373e36b66"
+  }
+}
+```
+
+**Note:** The structure of `permissionContext` is determined by the specific Delegation Manager implementation. Common implementations (e.g., MetaMask Delegation Framework) use EIP-712 signed delegation chains.
+
+### Phase 3: Verification Logic
+
+Unlike EIP-3009 and Permit2, ERC-7710 verification is performed entirely through simulation. The `permissionContext` is opaque to the facilitator but verifiable by simulating the intended action.
+
+The facilitator:
+
+1. **Constructs** the `executionCallData` encoding an ERC-20 `transfer(payTo, amount)` call for the required payment.
+
+2. **Constructs** the `mode` appropriate for the execution (typically `0x00...` for single call mode per ERC-7579).
+
+3. **Simulates** `delegationManager.redeemDelegations([permissionContext], [mode], [executionCallData])` to verify:
+   - The delegation is valid and authorizes the intended transfer.
+   - The delegator has sufficient balance of the asset.
+   - The transaction will succeed when executed.
+
+If the simulation succeeds, the payment is considered valid. The simulation serves as the sole verification mechanism—no trusted list of Delegation Manager implementations is required.
+
+**Security Considerations**:
+
+1. **Race Condition Risk**: A facilitator may be vulnerable to a race condition where the client invalidates their delegation between simulation and transaction execution, causing the facilitator to pay gas for a failed transaction. This risk can be mitigated by:
+   - Submitting transactions via a private mempool to reduce the window for front-running.
+   - Building trust signals for client accounts (e.g., reputation systems) that can be used to flag or ban abusive behavior.
+
+2. **Malicious Delegation Manager Gas Consumption**: A malicious or poorly implemented Delegation Manager could attempt to consume excessive gas during execution. To mitigate this risk:
+   - Facilitators should always set an explicit gas limit on their `redeemDelegations` call, as is standard practice for all Ethereum transactions.
+   - Pre-execution simulation helps identify whether a transaction is likely to use a reasonable amount of gas.
+   - If simulation reveals unexpectedly high gas consumption, this may indicate a "trap door" implementation designed to drain facilitator funds, and the transaction should be rejected.
+
+### Phase 4: Settlement Logic
+
+Settlement is performed by calling `redeemDelegations` on the Delegation Manager:
+
+```solidity
+delegationManager.redeemDelegations(
+    [permissionContext],  // bytes[] - delegation proof
+    [mode],               // bytes32[] - execution mode
+    [executionCallData]   // bytes[] - encoded transfer call
+);
+```
+
+The Delegation Manager validates the delegation authority and calls the delegator account to execute the token transfer. The delegator account then performs `token.transfer(payTo, amount)`.
+
+---
+
 ## Implementer Notes
 
 - **Permit2 Dependency:** Both the Permit2 contract and the x402ExactPermit2Proxy are audited, battle-tested contracts. However, integrators inherit their security properties and any future vulnerabilities discovered in either dependency.
@@ -208,6 +317,14 @@ Settlement is performed by calling the `x402ExactPermit2Proxy`.
 ---
 
 ## Annex
+
+### ERC-7710 Delegation Managers
+
+ERC-7710 does not define a canonical Delegation Manager. Implementations may vary in their delegation structure, caveat enforcement, and permission context format. Notable implementations include:
+
+- **MetaMask Delegation Framework**: A full-featured implementation supporting EIP-712 signed delegation chains, caveat enforcement, and batch processing. See [gator.metamask.io](https://gator.metamask.io/) for documentation.
+
+Since verification is performed entirely through simulation, facilitators do not need to maintain a trusted list of Delegation Manager implementations.
 
 ### Canonical Permit2
 


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

danfinlay:
What’s “from codeflow.yml” mean as a Heimdall requirement?

IMG_1992

phdargen:
Hi @danfinlay, thanks for putting this proposal together! Could you please update this to v2 (https://docs.cdp.coinbase.com/x402/migration-guide)?

As an initial comment, I think this needs a better separation between the `schema` defining a logical way of moving money (eg exact or upto) and the mechanism via which this is implemented. Imo, ERC-7710 is not a schema but an alternative mechanism.

Please have a look at a recent similar PR (https://github.com/coinbase/x402/pull/769) suggesting `Permit2` as an additional settlement mechanism for the `exact` scheme.

danfinlay:
I just updated per x402v2 per your private message, but now I see you also suggest the proposal be in the form of an extension to the `exact` scheme.

If this is the preferred format for alternate authorization formats, I would probably recommend updating the `exact` scheme to not so heavily declare its dependence on ERC-3009. The scheme's evm definition reads like an ERC-3009 payment format, not as a general purpose envelope for expressing multiple payment schemes. If it is merely an envelope, maybe ERC-3009 support should itself be an extension of the `exact` scheme.

If you insist on this format, I'll oblige, but the current organization makes it very easy for potential contributors to choose this same approach.

danfinlay:
Opened an alternate PR that formats this proposal as an extension to the `exact` scheme: https://github.com/coinbase/x402/pull/807

phdargen:
Thanks a lot for the prompt update @danfinlay! 

Agreed that the `exact` evm scheme needs to be refined to remove the ERC-3009 bias, this is started in https://github.com/coinbase/x402/pull/769, see https://github.com/coinbase/x402/blob/1f1dc9f1651244513929dc1237a679f00ffa3dbd/specs/schemes/exact/scheme_exact_evm.md. 

V2 just launched and with that we are beginning to explore additional schemes and mechanisms, any feedback is much appreciated 🙏

Looping in @fabrice-cheng (author of https://github.com/coinbase/x402/pull/769) for further guidance and review

danfinlay:
Nice to meet you, @fabrice-cheng! Happy new year! Let's make x402 even better!

phdargen:
Hi @danfinlay, thanks for considering my comment in https://github.com/coinbase/x402/pull/807. 

However, I believe there was a misunderstanding. I did not recommend to implement `ERC-7710` as an extension to the `exact` scheme but as an additional mechanism for the `exact` scheme similar to the `Permit2` mechanism in https://github.com/coinbase/x402/pull/769. 

https://github.com/coinbase/x402/pull/769 introduces `Permit2` as a new `AssetTransferMethod` besides `EIP-3009` which is specified in the payload: 
```
{
  "x402Version": 2,
  "accepted": {
    "scheme": "exact",
    "network": "eip155:84532",
    "amount": "10000",
    "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
    "maxTimeoutSeconds": 60,
    "asset": "0x036CbD53842c5426634e7929541eC2318f3dCF7e",
    "extra": {
      "assetTransferMethod": "permit2",
      "name": "USDC",
      "version": "2"
    }
  }
  "payload": {
    "signature": "0x2d6a7588d6acca505cbf0d9a4a227e0c52c6c34008c8e8986a1283259764173608a2ce6496642e377d6da8dbbf5836e9bd15092f9ecab05ded3d6293af148b571c",
    "permit2Authorization": {
      "permitted": {
        "token": "0x036CbD53842c5426634e7929541eC2318f3dCF7e",
        "amount": "10000"
      },
      "from": "0x857b06519E91e3A54538791bDbb0E22373e36b66",
      "spender": "0xx402Permit2ProxyAddress",
      "nonce": "0xf3746613c2d920b5fdabc0856f2aeb2d4f88ee6037b8cc5d04a71a4462f13480",
      "deadline": "1740672154",
      "witness": {
        "to": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
        "validAfter": "1740672089",
        "extra": {}
      }
    }
  },
}
```
https://github.com/coinbase/x402/pull/769 then also introduces 2 extensions, [erc20ApprovalGasSponsoring](https://github.com/coinbase/x402/blob/4698b6f88e09ccdb6695fa69660baf1f121e8964/specs/extensions/erc20_gas_sponsoring.md) and [eip2612GasSponsoring](https://github.com/coinbase/x402/blob/4698b6f88e09ccdb6695fa69660baf1f121e8964/specs/extensions/eip2612_gas_sponsoring.md) but these are optional for the Permit2 mechanism to sponsor the approval transaction on the user's behalf.

My recommendation for your proposal would be to closely follow https://github.com/coinbase/x402/pull/769 and modify scheme_exact_evm.md with ERC-7710 as new AssetTransferMethod, see https://github.com/coinbase/x402/blob/4698b6f88e09ccdb6695fa69660baf1f121e8964/specs/schemes/exact/scheme_exact_evm.md

I am thus closing https://github.com/coinbase/x402/pull/807, please refrain from opening new PRs so we can consolidate the discussion around this proposal in a single place

specs/schemes/7710/scheme_7710_evm.md
```diff
@@ -0,0 +1,460 @@
...
+
+## Payment Requirements Format
+
+### 402 Response with 7710 Scheme
+
+The server MAY offer the `7710` scheme alongside other schemes (like `exact`) to provide maximum client compatibility.
+
+```http
+HTTP/1.1 402 Payment Required
+X-PAYMENT-REQUIRED: <base64-encoded-json>
```
phdargen: `X-PAYMENT-REQUIRED` -> `PAYMENT-REQUIRED` see https://github.com/coinbase/x402/blob/main/specs/transports-v2/http.md#header-summary

specs/schemes/7710/scheme_7710_evm.md
```diff
@@ -0,0 +1,460 @@
...
+    participant Server as Server<br/>(Payee)
+    participant Facilitator as Facilitator<br/>(Redeemer)
+    participant DM as Delegation<br/>Manager
+
+    Client->>Server: 1. GET /resource
+    Server-->>Client: 2. 402 + accepts [7710]
+
+    Note over Client: 3. Client creates delegation<br/>via signer (EOA, smart wallet, etc.)
+
+    Client->>Server: 4. GET /resource + X-PAYMENT<br/>(7710 payload)
```
phdargen: `X-PAYMENT` -> `PAYMENT-SIGNATURE`

danfinlay:
I've revised the PR to follow the `AssetTransferMethod `style established in #769.

Since that PR also significantly revises the organization of the `exact` scheme spec, rather than duplicate that format, I forked off that PR, since it sounds like it's pretty sure to be merged. Let me know if you'd rather I remove it, that would be easy, but will make merging both harder later on.

fabrice-cheng:
Nice to meet you @danfinlay .

Would you mind merging main? We just added Permit2 to EvmScheme.

I'll be reviewing your PR in a bit, appreciate your contribution

danfinlay:
Merged in main, ready for review.

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,124 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+### Phase 1: Obtaining a Delegation
+
+The process of obtaining a delegation is outside the scope of x402. Delegations may be obtained through:
+
+- [ERC-7715](https://eips.ethereum.org/EIPS/eip-7715) permission requests
+- Direct wallet interactions
+- Pre-configured session keys
+- Other delegation protocols
+
+The key requirement is that the delegate (typically the x402 client or facilitator) holds a valid delegation that permits the required token transfer.
```
fabrice-cheng: ideally, the x402 client will hold the valid delegation and pass it to the Facilitator.
This ensures that the x402 client can use any facilitator interchangeably.

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,124 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+}
+```
+
+**Note:** The structure of `permissionContext` is determined by the specific Delegation Manager implementation. Common implementations (e.g., MetaMask Delegation Framework) use EIP-712 signed delegation chains.
+
+### Phase 3: Verification Logic
+
+The verifier must execute these checks in order:
+
+1. **Verify** the `delegationManager` is a known, trusted ERC-7710 implementation.
```
fabrice-cheng: could you clarify how we define "known & trusted"?

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,124 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+
+2. **Verify** the `delegator` account supports ERC-7710 (is a compatible smart account).
```
fabrice-cheng: What's the best practice to verify this? AFAIK, there's no `supportsInterface`?

fabrice-cheng (changes requested):
Amazing PR, thank you @danfinlay.

Could you also add the specs for the Facilitators extensions to support 7710?

cb-heimdall:
Review Error for fabrice-cheng @ 2026-01-13 14:25:51 UTC

specs/schemes/exact/scheme_exact_evm.md
danfinlay: Yeah, I think I see how this line is not great. How about:

```suggestion
The key requirement is that the client is able to issue a delegation to the facilitator that permits the required token transfer.
```

fabrice-cheng: @danfinlay it reads as the client needs to delegate to the facilitator's wallet address?

fabdarice:
Following up on this @danfinlay if that's still something you'd like to push forwarad

ukstv:
@danfinlay how can a client do a delegation within an x402 voucher/transaction?

Ben playing with exact Permit2 variant. Having an onchain tx from a client defeats the original x402 experience. Gasless approvals are definitely a way to go there.

One silly approach would be to have a 7710 delegation from a client to a sort of MetaTx delegate contract. That though would kill existing 7710 delegation for the user. So, if the client could command a delegator manager as part of x402 meta tx, that could be a boon for ux.

danfinlay:
> Following up on this @danfinlay if that's still something you'd like to push forwarad

Yes, will update in response to the feedback items very soon.

> how can a client do a delegation within an x402 voucher/transaction?

Hi @ukstv, nice to see you! Long time no collab on MetaTxs :)

I'm not 100% sure I understand the question, but let me try:

The current spec proposes the client passing a delegation for the cost of the web request to the server, allowing it to withdraw the funds however they choose, including:
1. They can either use the delegation to crate their own transaction (paying their own gas).
2. They can create a UserOp and ask another bundler to craft the transaction that initiates the withdraw.

It is possible to also construct other methods using this protocol:
3. The server could grant a delegation specifying themselves as recipient to someone else, like a MetaTx contract, allowing someone _else_ to submit the transaction some other time (I'm not sure if this has an advantage).
4. The authorized agent could use their own delegation to perform the payment before submitting the web request, and so be able to provide a tx hash that included the payment, similar to the simple scheme.

Does that cover your question?

specs/schemes/exact/scheme_exact_evm.md
danfinlay: Yes, in this arrangement, the facilitator now has the right to perform the withdraw (or assign someone else that right and responsibility). I'm accepting this change since it seems to have conveyed the right idea.

specs/schemes/exact/scheme_exact_evm.md
danfinlay: This item was incorrect. This protocol does not depend on knowing and trusting 7710 implementations.

danfinlay:
Alright @fabdarice , I corrected the verification section to better reflect the correct usage of the protocol.

danfinlay:
Had to do an ammend & force push to get the signed commit up, sorry about the noise.

danfinlay:
Hi @fabrice-cheng , I hope you enjoyed EthDenver. Let's get some agents authorized to spend funds without holding them!

fabrice-cheng (dismissed):

fabrice-cheng:
Approved, but you need to updates your commit to have verified signature.
cc: @danfinlay.

Could you also kickoff a follow-up PR with updates to the SDKs?

danfinlay:
Alright, I've signed the latest commit. I'm happy to follow up with SDK additions. Thank you!

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,120 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+
+**Note:** The structure of `permissionContext` is determined by the specific Delegation Manager implementation. Common implementations (e.g., MetaMask Delegation Framework) use EIP-712 signed delegation chains.
+
+### Phase 3: Verification Logic
+
+Unlike EIP-3009 and Permit2, ERC-7710 verification is performed entirely through simulation. The `permissionContext` is opaque to the facilitator but verifiable by simulating the intended action.
+
+The facilitator:
+
+1. **Constructs** the `executionCallData` encoding an ERC-20 `transfer(payTo, amount)` call for the required payment.
```
fabrice-cheng, line 275: out of curiosity, what's the point of reconstructing executionCalldata and mode if it's being passed via the client?
I'm no seeing any verification that is checking that input.executionCalldata == constructed.executionCalldata

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,120 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+delegationManager.redeemDelegations(
+    [permissionContext],  // bytes[] - delegation proof
+    [mode],               // bytes32[] - execution mode
+    [executionCallData]   // bytes[] - encoded transfer call
+);
+```
+
+The Delegation Manager validates the delegation authority and calls the delegator account to execute the token transfer. The delegator account then performs `token.transfer(payTo, amount)`.
+
+---
```
fabrice-cheng, line 311: Any 7710 specific error codes? if so, let's list them here.

see [upto](https://github.com/coinbase/x402/pull/1074/changes#diff-d8f6bae9ab396753a0cbb198751ee39db07cec4ffb8fc6969dd62abe8badfcabR244) as an example

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,120 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+
+The key requirement is that the client is able to issue a delegation to the facilitator that permits the required token transfer.
+
+### Phase 2: `PAYMENT-SIGNATURE` Header Payload
+
+The `payload` field must contain:
+
+- `delegationManager`: The address of the ERC-7710 Delegation Manager contract.
+- `permissionContext`: The delegation proof/context required by the specific Delegation Manager implementation.
+- `mode`: The ERC-7579 execution mode (typically single call mode: `0x0000...`).
```
fabrice-cheng: just to confirm, this should support any type of mode?

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,120 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+- Other delegation protocols
+
+The key requirement is that the client is able to issue a delegation to the facilitator that permits the required token transfer.
+
+### Phase 2: `PAYMENT-SIGNATURE` Header Payload
+
+The `payload` field must contain:
+
+- `delegationManager`: The address of the ERC-7710 Delegation Manager contract.
+- `permissionContext`: The delegation proof/context required by the specific Delegation Manager implementation.
```
fabrice-cheng, line 233: how are we handling replay protection?

AFAIK, delegation allows multiple redemptions. This would be more akin to the "[upto](https://github.com/coinbase/x402/pull/1074/changes#diff-f1eab07f1a490ce98f105d18ed5d229b6139a8f4305e830676dc32a1f0b69914)" scheme

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,120 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+
+The facilitator:
+
+1. **Constructs** the `executionCallData` encoding an ERC-20 `transfer(payTo, amount)` call for the required payment.
+
+2. **Constructs** the `mode` appropriate for the execution (typically `0x00...` for single call mode per ERC-7579).
+
+3. **Simulates** `delegationManager.redeemDelegations([permissionContext], [mode], [executionCallData])` to verify:
+   - The delegation is valid and authorizes the intended transfer.
+   - The delegator has sufficient balance of the asset.
```
fabrice-cheng, line 281: Is this worthwhile performing those checks outside of the simulation to get as accurate error message?

fabrice-cheng (changes requested):

cb-heimdall:
Review Error for fabrice-cheng @ 2026-03-04 00:17:50 UTC

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -201,8 +202,120 @@ Settlement is performed by calling the `x402Permit2Proxy`.
...
+
 ## Annex
 
+### ERC-7710 Delegation Managers
+
+ERC-7710 does not define a canonical Delegation Manager. Implementations may vary in their delegation structure, caveat enforcement, and permission context format. Notable implementations include:
+
+- **MetaMask Delegation Framework**: A full-featured implementation supporting EIP-712 signed delegation chains, caveat enforcement, and batch processing. See [delegator.xyz](https://delegator.xyz/) for documentation.
+
+Since verification is performed entirely through simulation, facilitators do not need to maintain a trusted list of Delegation Manager implementations.
```
fabrice-cheng, line 327: On top of the race condition described above, there's also a risks of malicious delegationManager who could aim to drain as much gas as possible as the facilitator is the one bearing the cost.

How could we mitigate this? Should facilitator set a gas limit on redeemDelegations for example?

specs/schemes/exact/scheme_exact_evm.md
danfinlay: These delegations can be redeemed in a single exact redemption, or multiple redemptions. They can also be used to issue payments that aren't in token transfers, like ether transfer, or the right to call a mint function.

This peculiar open-endedness is part of why [I initially submitted this as its own scheme type](https://github.com/coinbase/x402/commit/7beefe3cee8ad7dacca16a600494db1dacb3413b). [This was discouraged](https://github.com/coinbase/x402/pull/732#issuecomment-3657857471), and I was asked to refactor the contribution as a new method on the `exact` scheme.

To be clear: You're now asking me to make this a new method of the `upto` scheme? Are you sure that's a good fit? I'd love to get a "once and for all" blessing on the basic format that this proposal should take, as this will be the second full refactor of the form of the proposal, while the content has been basically the same for the three months the PR has been open.

How about, since we agree ERC-7710 can be used for _either_ `exact` or `upto` payments, we say that it _could_ live in both? And for now, let's get this first, basic, `exact` one merged, to show we are able to get anything merged?

specs/schemes/exact/scheme_exact_evm.md
danfinlay: This is a good point. The facilitator should form the request, and so having the user submit these parameters is needless complexity. Revising.

specs/schemes/exact/scheme_exact_evm.md
danfinlay: There are no ERC-7710 specific error codes at this time.

specs/schemes/exact/scheme_exact_evm.md
danfinlay: It can, but for the `exact` scheme it realistically will always be 0. Maybe for the simplicity of this scheme extension we should just specify the exact value.

specs/schemes/exact/scheme_exact_evm.md
danfinlay: Because ERC-7710 is unopinionated on the form of the delegation, a generic client-side validation is not feasible. Known 7710 implementations can be attempted, and a facilitator could choose to only accept delegations from known implementations that can be verified without simulation, but even then, the form of the delegation is not proof that the issuer’s implementation is compliant, and so the ability to actually move the funds is the ultimate test of whether the funds will move.

specs/schemes/exact/scheme_exact_evm.md
danfinlay: The facilitator will naturally set a gas limit on their call to redeemDelegations as every Ethereum account does on every transaction. Simulating in advance can help identify if the transaction is likely to use a reasonable amount of gas, and can be used to set a limit in case of a “trap door” malicious implementation. I will add that specificity to the security section.

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: Thanks for your feedback, and appreciate your patience.
We're discovering things alongside your spec.

Can we enforce single-redemption? My preference would be for all exact payment methods to behave the same way and have the same properties.

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: Got it. Thank you.

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: Appreciate it. Thanks.

danfinlay:
Alright, I've implemented some improvements, mostly from the feedback provided.

What I haven't done is refactor the entire proposal into any new format (like as an extension for the `upto` scheme), and [I'd like some strong confidence that a new refactor of format is likely to be in the correct format for the intended feature](https://github.com/coinbase/x402/pull/732#discussion_r2885473598).

specs/schemes/exact/scheme_exact_evm.md
danfinlay: It is possible for an ERC-7710 delegation issuer to enforce that the permission be used only once, and its value not be split among multiple withdraws, although there isn't really a way to enforce that they use it in that way, and I'm not sure why anyone paying for a service would care whether the payment was withdrawn in multiple transactions (as long as the gas was covered by the withdrawer).

fabrice-cheng: There is an inherent property shared across all exact assetTransferMethods: once an x402 payment is settled, it is final.

In other words, the same signature or permission cannot later be used to withdraw funds again. A settled payment fully consumes the authorization.

We expect multi-use payment flows to be handled by a separate payment scheme (e.g., subscriptions, deferred payments, or other recurring mechanisms).

Let me circle back shortly with a proposal for what we believe is the best path forward here.
Apologies again for the back and forth.

fabrice-cheng:
Hi @danfinlay

Thank you again for putting this together, and for iterating on the proposal so quickly. I appreciate the thought behind enabling delegated execution and agent-driven payment flows.

After reviewing ERC-7710 more closely, I’m hesitant to support it as part of `exact`, and more broadly I’m not yet convinced it belongs in x402 as a payment mechanism.

My main concern is that ERC-7710 appears to define a delegation / authorization model, not a payment primitive. In the PR description, the proposal is framed around “account permissions” and allowing agents to act on behalf of accounts, which matches that interpretation.

For `exact` specifically, one of the key properties we rely on today is single-use settlement: once a payment is redeemed, that authorization is effectively spent. With ERC-7710, that property is not guaranteed. A delegation may be redeemable multiple times unless the DelegationManager enforces nonce consumption or equivalent replay protection. The problem is that x402 cannot assume or deterministically verify that behavior across arbitrary DelegationManager / DelegatorAccount implementations without introducing whitelisting or other trust assumptions, which would weaken the guarantees of the scheme.

More broadly for x402, the facilitator should be able to reason about the payment from the payload itself with strong, minimal-trust verification. ERC-7710 seems weaker on that front:

it does not define a canonical fund movement primitive;

the intended transfer cannot be deterministically inferred from the payload alone when authorization and validation logic are delegated to arbitrary implementations;

`/verify` would likely depend on simulation; and
`/settle` would require post-settlement inspection of logs / effects to determine whether the expected ERC-20 transfer actually occurred.

That starts to look less like a payment scheme and more like a general-purpose execution relay. I’m also not sure what net-new payment capability this adds to x402 relative to existing schemes/mechanisms, aside from possibly long-lived delegated agent permissions for use cases like subscriptions, deferred execution, or automation. If that is the target, then I think we should evaluate it explicitly as a multi-use / recurring payment model rather than fit it into exact, whose guarantees are much stricter.

So my current view is:
1. ERC-7710 should not be included in exact as currently proposed.
2. I’m not yet convinced it should be added to x402 at all unless we are comfortable with the weaker trust / verification model. The core issue is that `permissionContext` and the `DelegationManager` make the validation / authorization semantics too opaque for a facilitator that is supposed to provide strong payment assurances.

Let me know your thoughts on this, and make sure I am understanding the standard accurately.

danfinlay:
> After reviewing ERC-7710 more closely, I’m hesitant to support it as part of exact, and more broadly I’m not yet convinced it belongs in x402 as a payment mechanism.

Thank you for reviewing the ERC closely. I'm a little surprised you hadn't yet, considering [you just "approved" this PR](https://github.com/coinbase/x402/pull/732#issuecomment-3993260309) days ago.

> My main concern is that ERC-7710 appears to define a delegation / authorization model, not a payment primitive. In the PR description, the proposal is framed around “account permissions” and allowing agents to act on behalf of accounts, which matches that interpretation.

Payment primitives are a subset of account permissions, as token transfers are a subset of the computation on the EVM. You might as well say you don't want to allow contract calls because they aren't exclusively financial (and ban token transfers).

> For exact specifically, one of the key properties we rely on today is single-use settlement: once a payment is redeemed, that authorization is effectively spent. With ERC-7710, that property is not guaranteed. A delegation may be redeemable multiple times unless the DelegationManager enforces nonce consumption or equivalent replay protection. 

x402 may not be able to ensure that a 7710 delegation can only be spent once, but a wallet issuing a 7710 delegation can be expected to provide that guarantee for itself. Would it satisfy your concern if we added to the specification that the issuer MUST ensure their permission can only be redeemed once?  It is the user's agent, and so it is in the user's own interest to ensure that any key security properties are upheld.

> The problem is that x402 cannot assume or deterministically verify that behavior across arbitrary DelegationManager / DelegatorAccount implementations without introducing whitelisting or other trust assumptions, which would weaken the guarantees of the scheme.

Much of Ethereum has this security assumption. The 4337 EntryPoint limits _how much_ state can be depended on, to reduce gas-griefing, but the possibility is still present. Even your standard `exact` scheme doesn't touch on how many confirmations to finality. Both your ERC-3009 and permit2 type `exact` schemes are vulnerable to the same kind of "sender front running a different spend that invalidates this authorization" issue. A simulation or trusted validator, or some layer of "block an ip and account if they ever gas grief you" is just reality, and it's not a unique concern to this proposal.

> More broadly for x402, the facilitator should be able to reason about the payment from the payload itself with strong, minimal-trust verification. ERC-7710 seems weaker on that front:

This is obviously untrue. A 3009 or permit2 signature payload may be easy to verify as a signature, but cannot be verified as a reliable source of payment until it lands onchain. You are holding this proposal to a double standard.

> I’m also not sure what net-new payment capability this adds to x402 relative to existing schemes/mechanisms

This would be the first x402 scheme that allows multi-hop authorized x402 payments. This is actually significant, so since this is something you haven't understood yet, please take some time to consider the implications: Under this 7710 extension of x402, authorized agents can pay for requests without the user having to lock funds in the agent. This is great in a world of many short-lived agents. Those short-lived agents still need to be able to issue individual requests for web resources, and I think focusing on the `exact` scheme as a first supported method for 7710 is a good place for us to build this common understanding exactly _because it is the simplest scenario_ (Not all resources may require subscriptions, keep it simple!).

For example: If you had a StableDiffusion API and you charged 1 USDC to use it, I might want my agent to be able to pay for it _when it decides it has a good enough joke_, but just today, and if it doesn't, I don't want to bother with withdrawing payments again. These short-lived, opportunistic, financially-empowered agents are probably actually a majority of use cases. One big wallet for one big agent with tons of context sounds like a recipe for context creep and [dangerous mistakes](https://www.anthropic.com/research/agentic-misalignment).

Even if you don't think that's likely to be a big use case, you should be able to recognize that it's a use case that the existing schemes are unable to satisfy:

3009 and permit2 can only be issued from an account holder itself, with funds in its account. It's unable to be sent from someone's agent (in the broker/representative sense of the word) without that representative having funds transferred to them first.

> If that [subscriptions] is the target, then I think we should evaluate it explicitly as a multi-use / recurring payment model rather than fit it into exact, whose guarantees are much stricter.

Like I said, I believe permissions and authorization is a superset of financial payments, so I see `exact` payments to be simply the smallest use case for applying 7710 to x402, and so while I do strongly suspect there are opportunities to extend the other schemes with 7710, I think we should figure out what the simple case looks like first, because it won't get easier than this.

So in summary, my opinion of your stated view:

1. I hope I've changed your mind.
2. The trust assumptions aren't actually weaker than your existing schemes (3009, permit2), as any of those payloads, as "inspectable" as they are, are not actually able to convey credible payment until finality is achieved, same as any scheme like this one.

fabrice-cheng:
Thanks for the detailed response, I understand the frustration. Please know I’m not trying to introduce unnecessary friction into the review process, and I appreciate you helping clarify that point.

I had misunderstood `mode` as something that could enforce single-use.

To summarize the core intent of my response for further discussion: I wanted to call out the current inability to reliably distinguish or enforce multi-use versus single-use semantics, which is an important property for all `exact` EVM `assetTransferMethods` today.

The agentic payment use case is compelling: giving an agent a scoped delegation to spend from a user's account,  bounded by amount, time, single-use, etc.. without requiring the user to pre-fund agent wallets. That's a capability gap in x402 today.
That said, this flexibility is also what makes the fit within the current x402 payment scheme non-obvious, since they live within the DelegationManager, and cannot be enforced by the Server and/or Facilitator. That's not necessarily a blocker, but it does raise the question of where this best fits within the spec.

I'll circle back with next steps after syncing with the team. Thank you for your patience and for pushing this forward.

fabrice-cheng:
Thank you for your patience. After reviewing this further with the team, we agree that `7710` is a great new payment capabilities for x402. We can start by adding it to the `exact` payment scheme and decided loosen up the requirement that all `assetTransferMethods` must be globally single-use within `exact`.

Could you please update the PR to add a column for each `assetTransferMethod`, indicating its usage semantics? Specifically:
- `3009` and `Permit2` should be marked as one-time use
- `7710` should be marked as supporting both one-time and multi-use

danfinlay:
Absolutely. Thank you!

fabrice-cheng (dismissed):

danfinlay:
@fabrice-cheng ready to go, rebased with a signed commit!

fabrice-cheng:
@danfinlay merge conflicts :)

danfinlay:
Alright, sorry about that @fabrice-cheng . I took a much cleaner approach to condensing the commit history to sign it in one recent commit. I hope this does it!

fabrice-cheng (approved):

LembaGang:
Hi @danfinlay @fabrice-cheng — wanted to share a live data point relevant to this discussion.
Headless Oracle (headlessoracle.com) is an x402-payable signed market state oracle for autonomous financial agents. We issue Ed25519-signed receipts confirming whether an exchange (NYSE, LSE, CME, Coinbase, etc.) is OPEN, CLOSED, or HALTED — with a 60-second TTL and fail-closed semantics.
Our receipt model maps naturally onto the single-use ERC-7710 delegation flow described here. An autonomous trading agent holds a delegation from the user's wallet. Before executing a trade, it calls our /v5/status endpoint via x402. The facilitator redeems the delegation, the signed receipt comes back as cryptographic proof of market state — delegation consumed exactly once, at exactly the moment the agent needed it. The agent never holds USDC. The user's funds stay in their wallet until the market check is needed.
This is precisely Dan's short-lived agent use case. Happy to build a reference implementation demonstrating this end-to-end once SDK support lands. Congratulations on the merge.

PR #769, opened 2025-12-12
Feature: Add Permit2 Support to Exact EVM Scheme

fabrice-cheng:
### Summary
This PR introduces **Permit2** as a settlement mechanism for the `exact` scheme on EVM. This enables standardized gasless transfers for any ERC-20 token (even those without native EIP-3009 support) by utilizing a `x402Permit2Proxy` and the canonical Permit2 contract.

### Motivation
To maximize the "gasless" payment coverage for the `exact` scheme. While EIP-3009 is preferred for supported tokens (like USDC), Permit2 provides a universal fallback for the vast majority of ERC-20 tokens, ensuring users can pay with any asset without holding native gas.

#### 1. New Mechanism: Permit2
Updated `specs/schemes/exact/scheme_exact_evm.md` to include **Mechanism 2: Permit2**.
- **Proxy Pattern**: Defines `x402Permit2Proxy` which uses the "Witness" pattern (`permitWitnessTransferFrom`) to securely bind the transfer to a specific destination (`to`), preventing malicious facilitator.
- **Settlement Logic**: Detailed steps for verifying signatures, allowances, and executing the settlement via the proxy.

#### 2. New Extensions
Added two extensions to support different gasless approval flows for Permit2 (one-time approval):
- **`specs/extensions/erc20_gas_sponsoring.md`** (`erc20ApprovalGasSponsoring`):
  - Allows a Facilitator to sponsor the gas for the user's initial `approve(Permit2)` transaction.
  - Useful for users with no native gas tokens.
- **`specs/extensions/eip2612_gas_sponsoring.md`** (`eip2612GasSponsoring`):
  - Allows tokens with EIP-2612 support to authorize Permit2 via an off-chain signature.
  - Provides a fully gasless experience without on-chain approval transactions.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

specs/extensions/eip2612_gas_sponsoring.md
```diff
@@ -0,0 +1,157 @@
...
+          },
+          "asset": {
+            "type": "string",
+            "pattern": "^0x[a-fA-F0-9]{40}$",
+            "description": "The address of the ERC-20 token contract."
+          },
+          "spender": {
+            "type": "string",
+            "pattern": "^0x[a-fA-F0-9]{40}$",
+            "description": "The address of the spender (Canonical Permit2)."
```
brendanjryan, line 55: is this the canonical `permit2` contract or the `x402permitProxy`? in which cases should we use one vs. the other?

specs/extensions/erc20_gas_sponsoring.md
```diff
@@ -0,0 +1,213 @@
...
+Because these tokens lack native gasless approvals:
+
+- The **Client** must sign a normal EVM transaction calling `approve(Permit2, amount)`.
+- The **Facilitator** agrees to:
+
+  - Fund the Client’s wallet with enough native gas token.
+  - Broadcast the Client’s signed approval transaction.
+  - Immediately perform settlement via [`x402Permit2Proxy`](../schemes/exact/scheme_exact_evm.md#reference-implementation-x402permit2proxy) after the approval confirms.
+
+This flow is typically executed using an **atomic batch transaction**.
```
brendanjryan: is this left open to the facilitator to interpret? Should this be a "should" do or is it a "must" do

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -63,50 +62,253 @@ Full `PaymentPayload` object:
...
+
+### Phase 3: Settlement Logic
+
+Settlement is performed via the facilitator calling the `transferWithAuthorization` function on the `EIP-3009` compliant contract with the `payload.signature` and `payload.authorization` parameters from the `PAYMENT-SIGNATURE` header.
+
+---
+
+## 2\. Mechanism: `Permit2`
+
+This mechanism uses the `permitWitnessTransferFrom` from the canonical **Permit2** contract combined with a `x402Permit2Proxy` to enforce receiver address security via the "Witness" pattern.
```
brendanjryan: I assume the canonical permit2 is the one defined at [0]?

[0] https://docs.uniswap.org/contracts/v4/deployments

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -63,50 +62,253 @@ Full `PaymentPayload` object:
...
+    }
+
+    function _settleInternal(
+        ISignatureTransfer.PermitTransferFrom calldata permit,
+        uint256 amount,
+        address owner,
+        Witness calldata witness,
+        bytes calldata signature
+    ) internal {
+        require(block.timestamp >= witness.validAfter, "Too early");
```
brendanjryan, line 303: why check these in this method instead of upstream in `settleWith2612`? Could this result in permits to be submitted but the parameters of the witness not be valid?  or is this reference implementation open for interpretation and ordering is an implementation detail and readers could optimize gas consumption / external contract calls?

specs/schemes/exact/scheme_exact_evm.md
```diff
@@ -63,50 +62,253 @@ Full `PaymentPayload` object:
...
+
+**Important Logic:** The `spender` in the signature is the **x402Permit2Proxy**, not the Facilitator. This Proxy enforces that funds are only sent to the `witness.to` address.
+
+**Example PaymentPayload:**
+
+```json
+{
+  "x402Version": 2,
+  "accepted": {
+    "scheme": "permit2",
```
phdargen: I think we need a better separation between scheme and mechanism here as schemes other than `exact` might also be implemented via the permit2 mechanism. Ideally sth like:
```suggestion
    "scheme": "exact",
    "mechanism": "permit2"
```

phdargen: or alternatively without introducing a new field: `"scheme": "exact:permit2"`

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: That's correct.

specs/extensions/erc20_gas_sponsoring.md
fabrice-cheng: To mitigate the risk of fund theft, it is **highly recommended** to ensure atomicity. Without atomicity, malicious actors could front-run the transaction, intercepting funds between the Facilitator funding step and the final settlement operation.

Updated the text for clarity

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: Great catch. scheme remains `exact`.
Will add an extra property called `permitType`.

Updated

specs/extensions/eip2612_gas_sponsoring.md
fabrice-cheng: This is the canonical Permit2 contract.
It should always be Permit2 contract, so we can reuse the same flow as the permit2 flow.

2612 doesn't contain any enforceable `to`

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: 1. This line exists we perform this check also for the standard `settle`
2. settleWith2612 would revert if this line fails.
3. The x402Permit2Proxy will be the canonical contract for all x402 via Permit2, and is for the time being purely illustrative. Following its implementation, there will be rounds of optimization + audits 
Note: Facilitator typically perform offchain validation & simulation. I don't anticipate a lot of transactions that would revert on this line (causing the gas inefficiency that you highlighted) 

if you have any suggestions, please do not hesitate

specs/extensions/eip2612_gas_sponsoring.md
brendanjryan: 👍 agree -- for this spec I think it would be useful to be explicit as to where this address should be found. I imagine both clients and servers should ensure that a canonical address is used

specs/extensions/erc20_gas_sponsoring.md
brendanjryan: ack -- thanks! Main concern here was openness of the language, as we should expect facilitators will need to follow this spec very closely

specs/schemes/exact/scheme_exact_evm.md
brendanjryan: 🙇

specs/schemes/exact/scheme_exact_evm.md
brendanjryan: >  and is for the time being purely illustrative

👍 thanks! I think this high level shape makes sense, don't need to focus on granular details if this will come in a further PR

specs/extensions/eip2612_gas_sponsoring.md
```diff
@@ -0,0 +1,157 @@
...
+{
+  "x402Version": "2",
+  "accepts": [
+    {
+      "scheme": "permit2",
+      "network": "eip155:84532",
+      "amount": "10000",
+      "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
+      "maxTimeoutSeconds": 60,
+      "asset": "0x036CbD53842c5426634e7929541eC2318f3dCF7e"
```
brendanjryan: nit: for completeness should this be a token that the exact flow would not work for?

or would we expect servers to offer multiple schemes for USDC with both the permit and the exact flow

specs/extensions/eip2612_gas_sponsoring.md
```diff
@@ -0,0 +1,157 @@
...
+## `PaymentRequired`
+
+A Facilitator advertises support for this extension by including the `eip2612GasSponsoring` key in the `extensions` object of the `402 Payment Required` response.
+
+```json
+{
+  "x402Version": "2",
+  "accepts": [
+    {
+      "scheme": "permit2",
```
brendanjryan: I don't _love_ `permit2` as the scheme name here. 

Ultimately the scheme defines the rules of money movement, and in this case isn't the behavior the same as the 3009 flow? `permit2` is an implementation detail that could apply to multiple different money movement flows (schemes).

Curious for @erikreppel-cb 's thoughts here, but it seems to me like there may need to be a `mechanism` field added to the `(scheme, network)` tuple in order to disambiguate situations like this.

In this flow we would have:

* Scheme: defines rules for money movement, and high level flow
* Network: defines the network that funds settle on.
* Mechanism: defines the implementation details for implementing the scheme on the given network, in case there are multiple supported flows. 

e.g. 

```bash
(scheme, mechanism, network)
(exact, eip3009, base) 
(exact, permit2, base)
(exact, delegate, solana) // mechanisms may be network specific
```

specs/extensions/erc20_gas_sponsoring.md
```diff
@@ -0,0 +1,219 @@
...
+
+### 3. Check User Balance
+
+- Check if the user (`from`) has enough native gas tokens to cover the transaction cost.
+- If the user has enough balance, the Facilitator skips the funding step.
+- If the user lacks balance, the Facilitator calculates the deficit.
+
+### 4. Simulate the full execution sequence
+
+The Facilitator must simulate in a single atomic batch transaction:
```
brendanjryan, line 215: given that we say _must ... in a single atomic batch transaction_ here 

Why do we leave the above open to interpretation? I think this would be clearer to facilitators if we made the language as explicit as possible here.

https://github.com/coinbase/x402/pull/769/changes#diff-dfa7ac90a6df4c00164411eb53aa4cc0470674106355d99f12762f011dd924b8R16

brendanjryan (commented):
Left some more comments after thinking about this for a bit!

In particular, i'm not sure if this strictly requires a new scheme and we may be overloading that primitive. 

This flow also seems like it pushes a lot of complexity to _both_ nodes of the network (facilitator, client) to juggle payment methods and more complicated approval flows. Is there a way to push more of this complexity to the facilitator where it can be more easily abstracted?

specs/extensions/eip2612_gas_sponsoring.md
fabrice-cheng: Apologies, I forgot to update the payload in extensions.

`Permit2` isn't a mechanism. the scheme is still exact, and decide to add a permitType to `extra`.But this field could be optional. 

AFAIK, the desire is that exact (scheme,network) would be supported in order of priority as part of the spec
i.e: for (exact,evm):
1. if the token supports 3009, both client and facilitators would prioritize 3009.
2. If not, Permit2.

specs/extensions/erc20_gas_sponsoring.md
fabrice-cheng: If the simulation doesn't happen in batch, it will fail, because the token hasn't been approved.

specs/extensions/eip2612_gas_sponsoring.md
fabrice-cheng: Good callout, will update

specs/extensions/eip2612_gas_sponsoring.md
brendanjryan: ack -- thanks!

This makes sense, in that there could be multiple `accepts` items for the same (scheme, network, currency) pair.

Is this "preference cascade" explicitly encoded in the core spec? If not, I think this would be a useful thing to define to disambiguate

specs/extensions/erc20_gas_sponsoring.md
brendanjryan: Correct. If we require everything to be batched for simulation why leave it open for facilitators to pull apart this batch for submission ("should")? 

I think this opens the door for more undefined behavior / bugs and doesn't simplify the spec at all since the txn needs to be atomically simulated as is

specs/extensions/eip2612_gas_sponsoring.md
fabrice-cheng: That's correct.

The priority cascade is specified at the very top of the exact scheme.
https://github.com/coinbase/x402/blob/4d9ab4de91fc0aa743dcd52401c6de16b40075c8/specs/schemes/exact/scheme_exact_evm.md#summary

specs/schemes/exact/scheme_exact_evm.md
erikreppel-cb: `permitType` should be `extra`

fabrice-cheng: Yep, it's in the `extra`

specs/extensions/eip2612_gas_sponsoring.md
```diff
@@ -0,0 +1,162 @@
...
+
+To utilize this extension, the client must generate a valid EIP-2612 signature and include it in the `eip2612GasSponsoring` under the key `extensions`.
+
+### Example PaymentPayload
+
+```json
+{
+  "x402Version": "2",
+  "accepted": {
+    "scheme": "permit2",
```
erikreppel-cb: scheme should remain `exact`

specs/extensions/eip2612_gas_sponsoring.md
```diff
@@ -0,0 +1,162 @@
...
+}
+```
+
+## Verification Logic
+
+When the Facilitator receives a payload containing `eip2612GasSponsoring` data, they must verify the following:
+
+1.  **Verify** the `asset` address actually implements `IERC20Permit`.
+2.  **Verify** the `signature` was signed for the `spender` and recovers to `from`.
+    - **Note**: The Facilitator MUST verify that `spender` matches the expected contract (e.g., Canonical Permit2).
```
erikreppel-cb: facilitator probably wants to do some basic transaction validity checks prior to broadcasting. Some that jump to mind:
- `from` actually  does not have gas (prevent gassing unneeded wallets)
- `tx` `nonce` is valid
could argue these get caught implicitly by simulating, but may be a good idea to have lower latency checks that early return before simulating

specs/extensions/erc20_gas_sponsoring.md
```diff
@@ -0,0 +1,224 @@
...
+```
+extensions.erc20ApprovalGasSponsoring
+```
+
+### Client Implementation Note
+
+The Client must ensure:
+
+- `maxFee` & `maxPriorityFee` are aligned with the current network prices.
+- `nonce` matches the current on-chain nonce of the Client wallet
```
erikreppel-cb, line 123: these adds the requirement of the client having an RPC endpoint, potentially this is fine because public RPCs should suffice, but we should consider if there is a way to prevent needing the client to access RPC

specs/extensions/eip2612_gas_sponsoring.md
fabrice-cheng: ops, missed this one. Updated

al24826 (approved):

cb-heimdall:
Review Error for al24826 @ 2025-12-15 20:59:32 UTC

al24826 (approved):

cb-heimdall:
Review Error for al24826 @ 2025-12-15 21:00:44 UTC

al24826 (approved):

cb-heimdall:
Review Error for al24826 @ 2025-12-15 21:07:33 UTC

al24826 (approved):

cb-heimdall:
Review Error for al24826 @ 2025-12-15 21:11:35 UTC

specs/schemes/exact/scheme_exact_evm.md
ryanRfox: I don't love `permitType` but do agree a new field should be introduced within `extra` and not add a new `mechanism` field (at this time). Suggesting something more descriptive such as `assetTransferMethod` to mitigate overloading the term "permit" here in favor of something describing the method used to move the asset, either within the token contract (`EIP3009`) or canonical proxy contract (`permit2`).

specs/extensions/eip2612_gas_sponsoring.md
fabrice-cheng: in 2612, the `from` doesn't need gas since everything is gasless.

As far as the `nonce`, this would also require an rpc call to test for its validity.
Unsure how much "faster", we would be failing fast, versus the added latency to perform multiple rpc calls that a single `simulate` would catch.

Let me know if you disagree.

specs/extensions/erc20_gas_sponsoring.md
fabrice-cheng: Yep, I highlighted the tradeoff with @CarsonRoscoe.
Client would likely need an rpc regardless to detect if they already had approved Permit2 contract or not.
At first glance, hard to avoid an rpc on the client side for this flow, unless we want to expose another Facilitator extensions that can retrieve those values.

specs/schemes/exact/scheme_exact_evm.md
fabrice-cheng: Thanks @ryanRfox . Great suggestion, updated `permitType` to `assetTransferMethod`

specs/extensions/erc20_gas_sponsoring.md
fabrice-cheng: Updated to "must be batch". Thanks for the feedback.

erikreppel-cb (approved):

cb-heimdall:
Review Error for erikreppel-cb @ 2026-01-07 18:55:54 UTC

erikreppel-cb (approved):

PR #790, opened 2025-12-15
feat(go): add transaction idempotency for facilitator settle

### LINKED ISSUE #786: Enhance transaction idempotency in facilitator for duplicate settle calls
**Description:**
Currently, when the `settle` method calls the facilitator, there are cases where the transaction may have been broadcast by the facilitator, but the client does not receive the `txhash` in response. This makes it difficult for the application layer to ensure idempotency when retrying or re-initiating the same operation.

To improve reliability and avoid duplicate transactions, we suggest implementing a deduplication mechanism inside the facilitator. The proposal is to use a composite key composed of:
- `from address`
- `nonce`
- `amount`

When the facilitator receives multiple requests with the same key and the corresponding transaction has already been processed, it should return the `txhash` from the previously executed transaction instead of reprocessing or rebroadcasting.

**Expected Behavior:**
- If a transaction with the same `from address`, `nonce`, and `amount` has already been broadcast, the facilitator should respond with the existing `txhash`.
- This will help the upstream systems handle retries safely and maintain idempotency.

**Additional Context:**
This enhancement would reduce unnecessary network traffic, prevent potential duplicate transfers, and provide a more predictable experience for client applications integrating with the facilitator.

madisoncarter1234:
## Summary

Adds application-level idempotency to the facilitator's `Settle()` method to prevent duplicate transaction submissions when clients retry after timeouts or network failures.

- Add `SettlementCache` with TTL-based expiration and in-flight request tracking
- Atomic `CheckAndMark()` operation prevents race conditions between concurrent requests
- Context-aware `WaitForResult()` respects cancellation while waiting for in-flight settlements
- Cache successful settlements for 10 minutes
- Failed settlements are not cached to allow retries

## Problem

The current implementation has a race condition: blockchain-level nonce protection (via EIP-3009 `authorizationState`) only works after a transaction is confirmed on-chain. During the pending period, duplicate settle calls can slip through because `checkNonceUsed()` returns false until confirmation.

**Scenario:**
1. Client calls `Settle()` → transaction submitted to blockchain
2. Client times out before receiving transaction hash
3. Client retries → `Verify()` passes (tx still pending)
4. Second transaction submitted → duplicate

## Solution

Add an in-memory deduplication cache that:
- Uses SHA256 hash of payload bytes as the deduplication key
- Atomically checks cache and marks in-flight in a single operation (no race condition)
- Tracks in-flight requests with channels to coordinate concurrent calls
- Respects context cancellation while waiting for in-flight requests
- Returns cached transaction hash for duplicate requests
- Expires entries after 10 minutes (covers typical confirmation times)

## Test plan

- [x] `TestSettlementCache_CheckAndMark_Cached` - cached results returned
- [x] `TestSettlementCache_CheckAndMark_InFlight` - in-flight detection works
- [x] `TestSettlementCache_Expiry` - TTL expiration works
- [x] `TestSettlementCache_Fail` - failed settlements allow retry
- [x] `TestSettlementCache_WaitForResult_Success` - waiting for in-flight works
- [x] `TestSettlementCache_WaitForResult_ContextCancelled` - context cancellation respected
- [x] `TestSettlementCache_ConcurrentWaiters` - multiple waiters get same result
- [x] `TestSettlementCache_AtomicCheckAndMark` - only 1 of 10 concurrent requests gets the slot
- [x] All existing tests pass (60+)

Closes #786


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

0x402ai:
Thanks for your work on this. Would you consider letting developers implement the specific cache location (e.g., in-memory, database-backed, or Redis-based)?

go/settlement_cache.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// SettlementStatus represents the result of checking the cache.
+type SettlementStatus int
+
+const (
+	// StatusNotFound means no cached result and no in-flight request.
+	StatusNotFound SettlementStatus = iota
+	// StatusCached means a cached result was found.
+	StatusCached
```
madisoncarter1234: Yeah you're right this only works for single-instance deployments.

For load-balanced setups we would need sticky sessions or swap in a Redis/DB backend. Could abstract the cache interface if that's needed. On restart the blockchain nonce check catches duplicates anyway

are facilitators deployed usually single instance or load-balanced?

madisoncarter1234:
> Thanks for your work on this. Would you consider letting developers implement the specific cache location (e.g., in-memory, database-backed, or Redis-based)?

Can add a cache interface so folks can swap in Redis or whatever for multi-instance setups. In-memory would be the default.

do we add that here or keep this PR simple and do it separately?

CarsonRoscoe (commented):
Hey @madisoncarter1234, appreciate the contribution!

My hesitation with this PR is it adds the requirement of state tracking to facilitators.

The core protocol should remain stateless. We cannot assume whether a facilitator is being run in a lambda dealing with cold starts and no persistance, whether a server is using a `x402Facilitator` directly to facilitate locally, whether it's hosted as a single instance, or whether it's load-balanced

I do however appreciate the in-memory default with interfacing for Redis/dbs. That makes it much simpler

I would prefer to implement idempotency as an extension facilitators can opt-into, and have the reference implementation of the extension leverage the `x402Facilitator`'s hooks, rather than implement the logic directly in the shared pure settle

madisoncarter1234:
> Hey @madisoncarter1234, appreciate the contribution!
> 
> My hesitation with this PR is it adds the requirement of state tracking to facilitators.
> 
> The core protocol should remain stateless. We cannot assume whether a facilitator is being run in a lambda dealing with cold starts and no persistance, whether a server is using a `x402Facilitator` directly to facilitate locally, whether it's hosted as a single instance, or whether it's load-balanced
> 
> I do however appreciate the in-memory default with interfacing for Redis/dbs. That makes it much simpler
> 
> I would prefer to implement idempotency as an extension facilitators can opt-into, and have the reference implementation of the extension leverage the `x402Facilitator`'s hooks, rather than implement the logic directly in the shared pure settle

thanks for the feedback, 100%
                                                                                                     
I've refactored this to move idempotency into extensions/idempotency/ as an opt-in wrapper:        
                                                                                                     
   // Users explicitly opt-in                                                                         
   facilitator := idempotency.Wrap(baseFacilitator)                                                   
                                                                                                     
   // Supports custom backends for distributed deployments                                            
   facilitator := idempotency.Wrap(baseFacilitator,                                                   
      idempotency.WithStore(myRedisStore),                                                           
  )                                                                                                  
                                                                                                     
  Changes:                                                                                           
  - Core x402Facilitator is now fully stateless (removed cache field entirely)                       
  - Added SettlementStore interface so users can plug in Redis/DB for load-balanced setups           
  - Provided InMemoryStore as reference implementation                                               
  - All existing tests moved and passing                                                             
                                                                                                     
 Let me know if this is more in line with what you had in mind.

PR #900, opened 2026-01-03
feat(v2-sdk): support async dynamic hooks for price and metadata

shuhei0866:
# feat(v2-sdk): Support Async Dynamic Hooks for Price and Metadata

## 1. Problem Statement
The current implementation of `DynamicPrice` and `DynamicPayTo` in the Python v2 SDK only supports synchronous callables. Real-world use cases—such as CMS integrations (e.g., Notion), adaptive pricing engines, or user-specific discounts—require looking up data in a database or calling an external API to determine the price or metadata. These operations must be performed asynchronously to maintain server responsiveness.

Additionally, metadata fields like `description` and `resource` were static, preventing developers from dynamically identifying the specific resource being sold at runtime (e.g., specific article titles in a CMS).

## 2. Proposed Solution: Enhanced Dynamic Hooks
This PR significantly matures the v2 SDK's dynamic capabilities by introducing robust, non-blocking hooks across all key route parameters.

### Architecture
```mermaid
sequenceDiagram
    participant Client as Client
    participant MW as x402 Middleware (FastAPI/Flask)
    participant Hook as Dynamic Hook (User Logic)
    
    Client->>MW: Request (GET /contents/123)
    Note over MW, Hook: Lazy Evaluation at Runtime
    MW->>Hook: await/call execute(context)
    Hook-->>MW: price, description, resource, etc.
    MW->>MW: Generate PaymentRequirements
    MW-->>Client: 402 Payment Required (Dynamic Info)
```

### Key Technical Changes
- **Unified Sync/Async Support**: Seamlessly handles both synchronous and asynchronous callables using `asyncio.iscoroutine` detection. This ensures 100% backward compatibility with initial v2 designs while unlocking full async power.
- **Metadata Dynamism**: Added `DynamicString` support for `description` and `resource` fields in `RouteConfig`.
- **Per-Route Timeout Configuration**: Introduced `hook_timeout_seconds` in `RouteConfig` (defaults to 5.0s), allowing developers to tune resilience based on their hook's expected latency.
- **Robust Flask/WSGI Bridge**: Enhanced event loop management in the Flask middleware to prevent loop-nesting issues in synchronous environments, ensuring production-grade reliability.
- **Sanitized Error Handling**: Hook failures and timeouts now return a standardized 500 error, ensuring internal stack traces are not exposed to clients.

## 3. Testing Performed
- ✅ **Async Integration**: Verified true async resolution with `asyncio.sleep`.
- ✅ **Sync Backward Compatibility**: Confirmed that legacy synchronous hooks still resolve perfectly.
- ✅ **Timeout Verification**: Confirmed that hooks exceeding the customized timeout return a proper 500 error instead of hanging.
- ✅ **Concurrency**: Verified that concurrent requests with different pricing logic maintain isolation.
- ✅ **Regression**: Updated all 16 existing integration tests to be async-compatible; all passed.

## 4. Why this matters for v2
This contribution brings the Python v2 SDK to functional parity with the advanced features of the TypeScript SDK, enabling the sophisticated "Advanced Patterns" (Dynamic Pricing) envisioned in the v2 specification.

### DIFF
--- a/python/x402/http/middleware/flask.py
+++ b/python/x402/http/middleware/flask.py
@@ -1,6 +1,7 @@
 """Flask middleware for x402 payment handling.
 
 Provides payment-gated route protection for Flask applications.
+Uses x402HTTPResourceServerSync for synchronous request processing without asyncio overhead.
 """
 
 from __future__ import annotations
@@ -23,7 +24,7 @@ from ..types import (
     RouteConfig,
     RoutesConfig,
 )
-from ..x402_http_server import PaywallProvider, x402HTTPResourceServer
+from ..x402_http_server import PaywallProvider, x402HTTPResourceServerSync
 
 if TYPE_CHECKING:
     from ...server import x402ResourceServer
@@ -315,7 +316,7 @@ class PaymentMiddleware:
             _register_bazaar_extension(server)
 
         self._app = app
-        self._http_server = x402HTTPResourceServer(server, routes)
+        self._http_server = x402HTTPResourceServerSync(server, routes)
         self._paywall_config = paywall_config
         self._sync_on_start = sync_facilitator_on_start
         self._init_done = False
@@ -362,7 +363,7 @@ class PaymentMiddleware:
                 self._http_server.initialize()
                 self._init_done = True
 
-            # Process payment request
+            # Process payment request synchronously (no asyncio overhead)
             result = self._http_server.process_http_request(context, self._paywall_config)
 
             if result.type == "no-payment-required":
--- a/python/x402/http/types.py
+++ b/python/x402/http/types.py
@@ -133,14 +133,10 @@ class PaywallConfig:
     testnet: bool = False
 
 
-# Dynamic function types for async server (supports both sync and async callbacks)
+# Dynamic function types (supports both sync and async callbacks)
 # Keep Awaitable support for Price and PayTo to achieve parity with TS
 DynamicPayTo = Callable[[HTTPRequestContext], "str | Awaitable[str]"]
 DynamicPrice = Callable[[HTTPRequestContext], "Price | Awaitable[Price]"]
-
-# Dynamic function types for sync server (sync callbacks only)
-SyncDynamicPayTo = Callable[[HTTPRequestContext], str]
-SyncDynamicPrice = Callable[[HTTPRequestContext], "Price"]
 
 
 @dataclass
@@ -177,7 +173,7 @@ class RouteConfig:
     custom_paywall_html: str | None = None
     unpaid_response_body: UnpaidResponseBody | None = None
     extensions: dict[str, Any] | None = None
-    hook_timeout_seconds: float = 5.0
+    hook_timeout_seconds: float | None = None
 
 
 RoutesConfig = dict[str, RouteConfig] | RouteConfig
--- a/python/x402/http/types.py
+++ b/python/x402/http/types.py
@@ -133,9 +133,14 @@ class PaywallConfig:
     testnet: bool = False
 
 
-# Dynamic function types
-DynamicPayTo = Callable[["HTTPRequestContext"], str]
-DynamicPrice = Callable[["HTTPRequestContext"], "Price"]
+# Dynamic function types for async server (supports both sync and async callbacks)
+# Keep Awaitable support for Price and PayTo to achieve parity with TS
+DynamicPayTo = Callable[[HTTPRequestContext], "str | Awaitable[str]"]
+DynamicPrice = Callable[[HTTPRequestContext], "Price | Awaitable[Price]"]
+
+# Dynamic function types for sync server (sync callbacks only)
+SyncDynamicPayTo = Callable[[HTTPRequestContext], str]
+SyncDynamicPrice = Callable[[HTTPRequestContext], "Price"]
 
 
 @dataclass
@@ -146,7 +151,7 @@ class UnpaidResponseResult:
     body: Any
 
 
-UnpaidResponseBody = Callable[["HTTPRequestContext"], UnpaidResponseResult]
+UnpaidResponseBody = Callable[[HTTPRequestContext], UnpaidResponseResult]
 
 
 @dataclass
@@ -172,6 +177,7 @@ class RouteConfig:
     custom_paywall_html: str | None = None
     unpaid_response_body: UnpaidResponseBody | None = None
     extensions: dict[str, Any] | None = None
+    hook_timeout_seconds: float = 5.0
 
 
 RoutesConfig = dict[str, RouteConfig] | RouteConfig
--- a/python/x402/http/types.py
+++ b/python/x402/http/types.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import re
-from collections.abc import Callable
+from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Literal, Protocol
 
--- a/python/x402/http/x402_http_server.py
+++ b/python/x402/http/x402_http_server.py
@@ -174,7 +174,7 @@ class _x402HTTPServerBase:
                 "unpaidResponseBody", config.get("unpaid_response_body")
             ),
             extensions=config.get("extensions"),
-            hook_timeout_seconds=config.get("hook_timeout_seconds", 5.0),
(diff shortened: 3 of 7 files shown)


phdargen:
Hi @shuhei0866, thanks a lot for your contribution!

The async support for the dynamic route config looks good to me and brings the python v2 implementation in line with TS v2. 

Not sure about making `description` and `resource` dynamic as well. This goes beyond the current TS/Go implementations and might have some implications for the discovery layer. In my opinion would be better for the python v2 release to have feature-parity with the TS/Go packages. As a follow-up we could then introduces dynamic `description` and `resource` for all packages simultaneously if desired. Looping in @CarsonRoscoe for feedback

python/x402/http/middleware/flask.py
```diff
@@ -300,7 +301,24 @@ def _wsgi_middleware(
                 self._init_done = True
 
             # Process payment request
-            result = self._http_server.process_http_request(context, self._paywall_config)
+            try:
+                # Try to get existing event loop
+                loop = asyncio.get_event_loop()
+                if loop.is_running():
+                    # Loop is already running, cannot nest - use asyncio.run()
+                    result = asyncio.run(
```
phdargen: I think this creates a new event loop and cannot be called when another loop is running, might need another workaround or careful testing that this does not fail at runtime

CarsonRoscoe:
Great catch, yes fully agree on the hooks being async.

Also +1 to @phdargen's comment regarding resource/description. For now, lets aim for feature parity between Typescript/Go/Python. Once we have the Python SDK in with parity, we can start a Github issue to discuss generally across the SDKs what enhancements we want, and make those improvements to all SDKs. For now though, it's a race for Python parity

python/x402/http/x402_http_server.py
```diff
@@ -209,7 +210,7 @@ def register_paywall_provider(self, provider: PaywallProvider) -> x402HTTPResour
     # Request Processing
     # =========================================================================
 
-    def process_http_request(
+    async def process_http_request(
```
CarsonRoscoe: I do have concerns regarding making the root functions async by default.

As a rule of thumb, it's easy to add sync code to an async codebase. It's hard to add async code to a sync codebase. By making the root methods async, we're severely limiting the portability of the SDK.

However, the callbacks should be able to be async for the reasons you described.

I'm debating what the best course of action is. Maybe it's exporting two sets of wrapper classes, `x402HTTPServer` and `x402HTTPServerAsync`, or two sets of functions, `process_http_request` and `process_http_request_async`?

CarsonRoscoe: Right now I'm leaning on two classes. `x402HTTPServer` is the async default one with the best features (i.e. async callbacks). If someone can't use async in their codebase, we can provide a secondary `x402HTTPServerSync` class who allows the functions to remain sync, but is limited by having sync callbacks?

phdargen:
Please also rebase against [feat/python-v2-sdk](https://github.com/coinbase/x402/tree/feat/python-v2-sdk) @shuhei0866

shuhei0866:
Hey @phdargen @CarsonRoscoe!

Thanks so much for the detailed feedback — really appreciate you taking the time to explain the design considerations around sync/async handling!

## Summary of Changes

Based on your suggestions, I've implemented the following:

### Removed dynamic `description`/`resource`
Keeping only dynamic `price`/`payTo` for TS/Go parity, as discussed.

### Implemented dual class design
Following Carson's suggestion:
- **`x402HTTPResourceServer`** — async version for FastAPI/Starlette
- **`x402HTTPResourceServerSync`** — sync version for Flask/Django

Both classes share common logic via `_x402HTTPServerBase`, keeping the codebase DRY.

### Updated Flask middleware
Now uses `x402HTTPResourceServerSync` directly — no more `asyncio.run()` workarounds! This resolves the event loop concerns @phdargen raised.

```python
# Before (hacky)
result = asyncio.run(self._http_server.process_http_request(...))

# After (clean)
result = self._http_server.process_http_request(...)
```

### Integrated with `enrich_extensions`
Both async and sync versions properly call `enrich_extensions` for bazaar and other extensions.

### All tests passing
Rebased onto `feat/python-v2-sdk` and verified all tests pass.

## Design Highlights

- **Clear error handling**: `x402HTTPResourceServerSync` raises a helpful `TypeError` if someone accidentally passes an async callback
- **Timeout support**: Per-route `hook_timeout_seconds` configuration for async hooks
- **Backward compatible**: Existing code using `x402HTTPResourceServer` continues to work

---

Ready for review! Happy to adjust anything based on your feedback.

python/x402/http/types.py
```diff
@@ -133,9 +133,14 @@ class PaywallConfig:
...
-DynamicPayTo = Callable[["HTTPRequestContext"], str]
-DynamicPrice = Callable[["HTTPRequestContext"], "Price"]
+# Dynamic function types for async server (supports both sync and async callbacks)
+# Keep Awaitable support for Price and PayTo to achieve parity with TS
+DynamicPayTo = Callable[[HTTPRequestContext], "str | Awaitable[str]"]
+DynamicPrice = Callable[[HTTPRequestContext], "Price | Awaitable[Price]"]
+
+# Dynamic function types for sync server (sync callbacks only)
+SyncDynamicPayTo = Callable[[HTTPRequestContext], str]
+SyncDynamicPrice = Callable[[HTTPRequestContext], "Price"]
```
phdargen: SyncDynamicPayTo/SyncDynamicPrice aren't used anywhere. For type checking I think we would also need to define a SyncPaymentOption with SyncDynamicPayTo/SyncDynamicPrice but I would prefer a simpler API keeping an unified PaymentOption and simply removing SyncDynamicPayTo/SyncDynamicPrice. Having separate x402HTTPResourceServer and x402HTTPResourceServerSync plus clear runtime errors, should be enough

phdargen:
Please add `x402HTTPResourceServerSync` to the exports in [http/__init__.py](https://github.com/coinbase/x402/blob/feat/python-v2-sdk/python/x402/http/__init__.py)

python/x402/http/types.py
```diff
@@ -172,6 +177,7 @@ class RouteConfig:
     custom_paywall_html: str | None = None
     unpaid_response_body: UnpaidResponseBody | None = None
     extensions: dict[str, Any] | None = None
+    hook_timeout_seconds: float = 5.0
```
phdargen: The default timeout of 5s seems a bit low. 

Would also consider removing the timeout entirely as the TS implementation doesn't have it and to me this is more of a service infrastructure concern, rather than for the core SDK.

If it's handled at the server implementation level, timeouts could also be recovered with a fallback, eg sth like this:
```
async def get_dynamic_price(context: HTTPRequestContext) -> str:
    try:
        return await asyncio.wait_for(fetch_price_from_db(), timeout=3.0)
    except asyncio.TimeoutError:
        return "$0.01"  # fallback price
```

An alternative would be making `hook_timeout_seconds` None by default (no timeout) with an opt-in

shuhei0866:
Thanks for the feedback @phdargen! Addressed all points:

- Exported `x402HTTPResourceServerSync` 
- Removed `SyncDynamicPayTo`/`SyncDynamicPrice` (kept unified `PaymentOption`)
- Changed `hook_timeout_seconds` default to `None` (opt-in)

All tests passing.

phdargen (approved):

phdargen:
Looks great @shuhei0866, thanks again for your contribution!

phdargen:
We require commit signing @shuhei0866, happy to merge once you rebase with signed commits.
See here for instructions: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

shuhei0866:
Rebased with signed commits. Ready to merge!

phdargen:
> Rebased with signed commits. Ready to merge!

Thanks @shuhei0866, all your commits must be verified. Could you please sign the remaining ones as well?
Screenshot 2026-01-14 at 10 49 51

shuhei0866:
Sorry for the oversight! I wasn't aware that all commits needed to be signed.

Rebased and signed all 3 commits. All should now show as Verified.

phdargen:
> Sorry for the oversight! I wasn't aware that all commits needed to be signed.
> 
> Rebased and signed all 3 commits. All should now show as Verified.

Thanks, all good now 🚀

phdargen (approved):

PR #915, opened 2026-01-07
feat: added contracts/evm for evm permit2 update

CarsonRoscoe:
**_THE CONTRACTS IN THIS PR HAVE YET TO BE AUDITED. ONLY DEPLOY TO TEST NETWORKS AT THIS TIME_**

## Description

In order to implement #769, an `x402Permit2Proxy` contract must be written and deployed.

This PR adds a foundry project under `contracts/evm`, which both acts as a foundry project for smart contract development

[Base Sepolia Deployment](https://sepolia.basescan.org/address/0x40203F636c4EDFaFc36933837FFB411e1c031B50)

## Tests

Added united tests & base-sepolia forked integration tests

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

contracts/evm/script/ComputeAddress.s.sol
```diff
@@ -0,0 +1,76 @@
...
+    /// @notice Canonical Permit2 address
+    address constant PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
+
+    /// @notice Arachnid's deterministic CREATE2 deployer
+    address constant CREATE2_DEPLOYER = 0x4e59b44847b379578588920cA78FbF26c0B4956C;
+
+    /// @notice Salt for deterministic deployment
+    bytes32 constant SALT = 0x62bb59fa735c572ac45816aa0f1e00b2de3c4671993a9147999a3808c574240e;
+
+    function run() public view {
```
stevieraykatz: You can add arguments to arbitrary script methods and then invoke them like so: 
`forge script script/ComputeAddress.s.sol --sig "myMethod(bytes32)" SALT`

This way you make a generic method that doesn't rely on editing in-line constants.

contracts/evm/src/interfaces/ISignatureTransfer.sol
stevieraykatz, line 1: The file name and in the interface it defines should match.

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+ * @notice Trustless proxy for x402 payments using Permit2
+ * @dev This contract acts as the authorized spender in Permit2 signatures.
+ *      It uses the "witness" pattern to cryptographically bind the payment destination,
+ *      preventing facilitators from redirecting funds.
+ *
+ * Security Properties:
+ * - Immutable: No upgrade mechanism
+ * - No custody: Never holds tokens
+ * - Destination locked: Witness pattern enforces payTo address
+ * - Reentrancy protected: Uses OpenZeppelin's ReentrancyGuard
```
stevieraykatz: This is vibey cruft

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import {ISignatureTransfer} from "./interfaces/IPermit2.sol";
```
stevieraykatz: Add line between external vs. internal imports: [style ref](https://github.com/coinbase/solidity-style-guide?tab=readme-ov-file#c-group-imports-by-external-and-local-with-a-new-line-in-between)

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+    function permit(
+        address owner,
+        address spender,
+        uint256 value,
+        uint256 deadline,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) external;
+}
```
stevieraykatz: Move to own file and import `IERC20Permit.sol`. This iface is also available to vendor from OZ: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/IERC20Permit.sol

contracts/evm/src/x402UptoPermit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import {ISignatureTransfer} from "./interfaces/IPermit2.sol";
+
+/**
+ * @title x402Permit2Proxy
+ * @notice Trustless proxy for x402 payments using Permit2
+ * @dev This contract acts as the authorized spender in Permit2 signatures.
+ *      It uses the "witness" pattern to cryptographically bind the payment destination,
+ *      preventing facilitators from redirecting funds.
```
stevieraykatz, line 13: Newlines between tag types: [style ref](https://github.com/coinbase/solidity-style-guide?tab=readme-ov-file#3-newlines-between-tag-types)

contracts/evm/src/x402UptoPermit2Proxy.sol
stevieraykatz, line 1: Run forge fmt to fix lint. Can add a github action CI step to check forge fmt returns 0 like: https://github.com/coinbase/smart-wallet/blob/main/.github/workflows/test.yml

contracts/evm/src/x402UptoPermit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+        Witness calldata witness,
+        bytes calldata signature
+    ) internal {
+        // Validate addresses
+        if (owner == address(0)) revert InvalidOwner();
+        if (witness.to == address(0)) revert InvalidDestination();
+
+        // Validate time window
+        if (block.timestamp < witness.validAfter) revert PaymentTooEarly();
+        if (block.timestamp > witness.validBefore) revert PaymentExpired();
```
stevieraykatz: Natspec above says `witness.validAfter` is the earliest timestamp that is acceptable. Thus,  control should be expressed using `<=`. Same for `validBefore`.

contracts/evm/src/x402UptoPermit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+
+        // Validate amount
+        if (amount > permit.permitted.amount) revert AmountExceedsPermitted();
```
stevieraykatz: Should be `amount >= permit.permitted.amount`

stevieraykatz:
Missing dep on OZ's reentrancy guard. Need to `forge install OpenZeppelin/openzeppelin-contracts`

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+        // Reconstruct witness hash to enforce integrity
+        bytes32 witnessHash = keccak256(
+            abi.encode(WITNESS_TYPEHASH, keccak256(witness.extra), witness.to, witness.validAfter, witness.validBefore)
+        );
+
+        // Execute transfer via Permit2
+        PERMIT2.permitWitnessTransferFrom(permit, transferDetails, owner, witnessHash, WITNESS_TYPE_STRING, signature);
+
+        // Emit event for observability
+        emit X402PermitTransfer(owner, transferDetails.to, transferDetails.requestedAmount, permit.permitted.token);
```
stevieraykatz: I think all the data emitted here is redundant considering the events that will be emitted by the ERC20 contract. Is there a specific reason we need to emit this data ourselves?

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+
+    /// @notice EIP-712 type string for witness data
+    /// @dev Must match the exact format expected by Permit2
+    /// Types must be in ALPHABETICAL order after the primary type (TokenPermissions < Witness)
+    string public constant WITNESS_TYPE_STRING =
+        "Witness witness)TokenPermissions(address token,uint256 amount)Witness(bytes extra,address to,uint256 validAfter,uint256 validBefore)";
+
+    /// @notice EIP-712 typehash for witness struct
+    bytes32 public constant WITNESS_TYPEHASH =
+        keccak256("Witness(bytes extra,address to,uint256 validAfter,uint256 validBefore)");
```
stevieraykatz: Should be: 
```solidity
keccak256("Witness(address to,uint256 validAfter,uint256 validBefore,bytes extra)");
```
provided the type def below.

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+ */
+contract x402Permit2Proxy is ReentrancyGuard {
+    /// @notice The canonical Permit2 contract address
+    ISignatureTransfer public immutable PERMIT2;
+
+    /// @notice EIP-712 type string for witness data
+    /// @dev Must match the exact format expected by Permit2
+    /// Types must be in ALPHABETICAL order after the primary type (TokenPermissions < Witness)
+    string public constant WITNESS_TYPE_STRING =
+        "Witness witness)TokenPermissions(address token,uint256 amount)Witness(bytes extra,address to,uint256 validAfter,uint256 validBefore)";
```
stevieraykatz: Should be:
```solidity
"Witness witness)TokenPermissions(address token,uint256 amount)Witness(address to,uint256 validAfter,uint256 validBefore,bytes extra)" 
```
provided the type def below.

contracts/evm/src/x402Permit2Proxy.sol
```diff
@@ -0,0 +1,226 @@
...
+        // Validate amount
+        if (amount > permit.permitted.amount) revert AmountExceedsPermitted();
+
+        // Prepare transfer details with destination from witness
+        ISignatureTransfer.SignatureTransferDetails memory transferDetails =
+            ISignatureTransfer.SignatureTransferDetails({to: witness.to, requestedAmount: amount});
+
+        // Reconstruct witness hash to enforce integrity
+        bytes32 witnessHash = keccak256(
+            abi.encode(WITNESS_TYPEHASH, keccak256(witness.extra), witness.to, witness.validAfter, witness.validBefore)
```
stevieraykatz: Make sure this matches WITNESS_TYPEHASH's ordering given comments above.

contracts/evm/src/x402UptoPermit2Proxy.sol
CarsonRoscoe: I believe the current code if (amount > permit.permitted.amount) is correct. This allows amount == permitted.amount (exact permitted amount) and only reverts when exceeding. Using >= would reject exact matches, breaking valid use cases where users want to transfer exactly what they permitted.

contracts/evm/src/x402UptoPermit2Proxy.sol
CarsonRoscoe: I think there was slight confusion in reviewing this, please verify my understanding.

`if (block.timestamp < witness.validAfter) revert PaymentTooEarly()` means that if the current timestamp is earlier than the earliest valid timestamp, revert

Similarly, `if (block.timestamp > witness.validBefore) revert PaymentExpired()` means that if the current timestamp is later than the last valid timestamp, revert

I believe that is correct, as `<=` and `>=` would then revert when equal, becoming exclusive. We want it to be inclusive

contracts/evm/src/x402Permit2Proxy.sol
CarsonRoscoe: Good points. Reached out to the data scientist who will eventually be observing these contracts to get his opinion on if he wants/needs any events or not, given the redundancy. Will let his decision guide what we emit

contracts/evm/src/x402UptoPermit2Proxy.sol
stevieraykatz: Oh man, yep you got it right. We're describing the revert edges. My bad!

contracts/evm/src/x402UptoPermit2Proxy.sol
stevieraykatz: same as above

contracts/evm/foundry.toml
```diff
@@ -0,0 +1,51 @@
...
+base_sepolia = "${BASE_SEPOLIA_RPC_URL}"
+optimism_sepolia = "${OPTIMISM_SEPOLIA_RPC_URL}"
+arbitrum_sepolia = "${ARBITRUM_SEPOLIA_RPC_URL}"
+
+[etherscan]
+mainnet = { key = "${ETHERSCAN_API_KEY}" }
+base = { key = "${BASESCAN_API_KEY}" }
+base_sepolia = { key = "${BASESCAN_API_KEY}" }
+optimism = { key = "${OPTIMISM_ETHERSCAN_API_KEY}" }
+arbitrum = { key = "${ARBISCAN_API_KEY}" }
```
stevieraykatz: Etherscan's v2 api uses the same key for all networks now

contracts/evm/src/x402BasePermit2Proxy.sol
```diff
@@ -0,0 +1,151 @@
...
+
+    /// @notice EIP-712 typehash for witness struct
+    bytes32 public constant WITNESS_TYPEHASH =
+        keccak256("Witness(address to,uint256 validAfter,uint256 validBefore,bytes extra)");
+
+    /// @notice Emitted when settle() completes successfully
+    event Settled();
+
+    /// @notice Emitted when settleWithPermit() completes successfully
+    event SettledWithPermit();
```
fabrice-cheng, line 37: Do we really see a value of being able to distinguish Settled vs SettledWithPermit()?

contracts/evm/src/x402BasePermit2Proxy.sol
```diff
@@ -0,0 +1,151 @@
...
+        uint8 v;
+    }
+
+    /**
+     * @notice Initializes the proxy with the Permit2 contract address
+     * @param _permit2 Address of the canonical Permit2 contract
+     * @dev Reverts if _permit2 is the zero address
+     */
+    constructor(
+        address _permit2
```
fabrice-cheng, line 86: could you double check that adding this to the constructor with potential diff values won't affect the deployed contract address?

Unsure if the bytecode for CREATE2/CREATE3 is affected by constructor payload

contracts/evm/src/x402BasePermit2Proxy.sol
```diff
@@ -0,0 +1,151 @@
...
+        Witness calldata witness,
+        bytes calldata signature
+    ) internal {
+        // Validate addresses
+        if (owner == address(0)) revert InvalidOwner();
+        if (witness.to == address(0)) revert InvalidDestination();
+
+        // Validate time window
+        if (block.timestamp < witness.validAfter) revert PaymentTooEarly();
+        if (block.timestamp > witness.validBefore) revert PaymentExpired();
```
fabrice-cheng: fyi, Permit2 has a deadline in it, so this field would be redundant
`

struct PermitTransferFrom {
        TokenPermissions permitted;
        // a unique value for every token owner's signature to prevent signature replays
        uint256 nonce;
        // deadline on the permit signature
        uint256 deadline;
    }
 `

contracts/evm/src/x402BasePermit2Proxy.sol
CarsonRoscoe: I don't believe it's strictly necessary, however I do prefer having it to future proof our observability.

Without these events, there is no event driven approach for a data scientist to determine which transactions are flowing through the 2612 path, or through a separate erc20 approval. I foresee us wanting to know the breakdowns of transfer methods in the future, which is why I added this

contracts/evm/src/x402BasePermit2Proxy.sol
CarsonRoscoe: Good catch, a change in constructor arguments would indeed change the deployed address

CarsonRoscoe: The constructor parameter I am planning to use is Uniswap's canonical Permit2 contract address. Based on [these docs](https://api-docs.uniswap.org/guides/supported_chains), the address should exist on 16 different mainnets, which is a decent range but also limiting.

We could resolve with a separate `initialize(_permit2)` function gated by an initializer modifier to ensure it's only called once, and add a initialized() modifier on the settle calls to ensure initialize() was called before the contract gets used

Thoughts on that @fabrice-cheng ?

contracts/evm/src/x402BasePermit2Proxy.sol
CarsonRoscoe: Great catch, addressing now

CarsonRoscoe: https://github.com/coinbase/x402/pull/915/commits/082614ea084ee20d88e52a85ab8ec6278536e53f

contracts/evm/src/x402BasePermit2Proxy.sol
apmcdermott: 100% agreed on this take ^

PR #921, opened 2026-01-07
feat(extensions): SIWX Extension

sragss:
Implements CAIP-122 compliant Sign-In-With-X (SIWX) extension for x402 v2 protocol aligned with [CHANGELOG-v2.md](https://github.com/coinbase/x402/blob/main/typescript/site/CHANGELOG-v2.md) and [sign-in-with-x/TODO.md](https://github.com/coinbase/x402/blob/main/typescript/packages/extensions/src/sign-in-with-x/TODO.md). 

# Supported Signature Schemes
- `eip191` - personal_sign (EOAs)
- `eip1271` - Smart contract wallets
- `eip6492` - Counterfactual smart wallets
- `siws` - Sign-In-With-Solana

Missing: EIP-712 + Stellar.

Note the interface for using Smart Wallets is a bit odd, didn't want to bake the RPC into a generic extension.

```
  // EOA only
  const result = await verifySIWxSignature(payload);

  // Smart wallet support (EIP-1271/EIP-6492)
  const publicClient = createPublicClient({ chain: base, transport: http()
  });
  const result = await verifySIWxSignature(payload, {
    evmVerifier: publicClient.verifyMessage,
  });

```

# Update 1/17/25
- Added `spec.md`
- Added `wrapFetchWithSIWx` exported from `@x402/extensions/siwx` to match `wrapFetchWithPay`
- Added `examples/typescript/{client,server}/sign-in-with-x`

Implementation note: we use conditional middleware in the `server` example. I considered putting special logic into `paymentMiddleware` but decided that was against the premise of optional extensions. Another pattern that could be clean would be to put a `skipIf: (req) => bool` lambda as an optional param in the `paymentMiddleware` which takes the simple SIWx verification function.

**Example logs**
*Server*
```
Server running at http://localhost:4021
Routes: GET /weather, GET /joke
Payment recorded: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /weather
SIWX auth: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /weather
Payment recorded: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /joke
SIWX auth: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /joke
```

*Client*
```
Client address: 0x6d157C00d034541020fF73A989960D22b299d8e1
Server: http://localhost:4021

--- /weather ---
1. First request (paying)...
   Response: { weather: 'sunny', temperature: 72 }
2. Second request (SIWX auth)...
   Response: { weather: 'sunny', temperature: 72 }

--- /joke ---
1. First request (paying)...
   Response: {
  joke: 'Why do programmers prefer dark mode? Because light attracts bugs.'
}
2. Second request (SIWX auth)...
   Response: {
  joke: 'Why do programmers prefer dark mode? Because light attracts bugs.'
}

Done. Each resource required payment once, then SIWX auth worked.
```


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

CarsonRoscoe (commented):
Amazing work @sragss, thank you, will give this a review shortly

phdargen:
Thanks for your contribution @sragss, excited for this!

@CarsonRoscoe shall we have a spec document for extensions as well, like we have for schemes and transports? 
The TODO.md deleted in this PR might be a good starting point for that

There is none for the discovery extension, but might be worth to add, especially for 3rd party sdk devs

CarsonRoscoe:
@phdargen Agreed, we should have a top level /specs/extensions folder with specs per extension

CarsonRoscoe (commented):
Hey @sragss!

This is looking really good. Very happy with it from my readings,

My two asks are:
1. Can you add a `/specs/extensions/sign-in-with-x.md` file outlining the spec as implemented? This will become the north start to ensure Go/Python implementations remain compatible
2. Can you integrate these into an example for testing. For example, a `/examples/typescript/clients/sign-in-with-x` (using fetch) and `/examples/typescript/servers/sign-in-with-x` (using express) example that can be run to showcase how a client/server could leverage this extension to prove a client already paid? I'm imagining when running these examples, the flow would be that the server caches payments in memory, and is able to skip payments for clients who already paid. Then the first client run would make payment, while the second one would get the resource without needing payment due to siwx?

Please include a run of the client & server examples in your PR afterwards, we'll consider it additional testing

Great work, I think the examples & spec are all that I need to feel comfortable merging :)

sragss:
Done. 


**Update 1/17/25**
- Added `spec.md`
- Added `wrapFetchWithSIWx` exported from `@x402/extensions/siwx` to match `wrapFetchWithPay` and simplify client experience
- Added `examples/typescript/{client,server}/sign-in-with-x`

Implementation note: we use conditional middleware in the `server` example. I considered putting special logic into `paymentMiddleware` but decided that was against the premise of optional extensions. Another pattern that could be clean would be to put a `skipIf: (req) => bool` lambda as an optional param in the `paymentMiddleware` which takes the simple SIWx verification function. Punting on that for now.

Example logs
*Server*
```
Server running at http://localhost:4021
Routes: GET /weather, GET /joke
Payment recorded: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /weather
SIWX auth: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /weather
Payment recorded: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /joke
SIWX auth: 0x6d157C00d034541020fF73A989960D22b299d8e1 for /joke
```

*Client*
```
Client address: 0x6d157C00d034541020fF73A989960D22b299d8e1
Server: http://localhost:4021

--- /weather ---
1. First request (paying)...
   Response: { weather: 'sunny', temperature: 72 }
2. Second request (SIWX auth)...
   Response: { weather: 'sunny', temperature: 72 }

--- /joke ---
1. First request (paying)...
   Response: {
  joke: 'Why do programmers prefer dark mode? Because light attracts bugs.'
}
2. Second request (SIWX auth)...
   Response: {
  joke: 'Why do programmers prefer dark mode? Because light attracts bugs.'
}

Done. Each resource required payment once, then SIWX auth worked.
```

phdargen:
Hi @sragss, thanks a lot for adding the examples. I tested them and it works well! The building blocks are all there, but the examples still include a lot of custom code. I have some suggestions to improve devx and make it easier for existing projects to add the extension

examples/typescript/servers/sign-in-with-x/siwx-middleware.ts
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Checks if an address has paid for a resource.
+ *
+ * @param resource - The resource path
+ * @param address - The address to check
+ * @returns True if the address has paid
+ */
+export function hasPaid(resource: string, address: string): boolean {
+  return paidAddresses.get(resource)?.has(address.toLowerCase()) ?? false;
+}
```
phdargen: Suggest to define a `SIWxStorage` interface in `package/extensions/sign-in-with-x`:
```
export interface SIWxStorage {
  hasPaid(resource: string, address: string): boolean | Promise<boolean>;
  recordPayment(resource: string, address: string): void | Promise<void>;
}
```
and provide an in-memory storage implementation `InMemorySIWxStorage` that can be imported in examples

examples/typescript/servers/sign-in-with-x/index.ts
```diff
@@ -0,0 +1,83 @@
...
+const facilitatorClient = new HTTPFacilitatorClient({ url: facilitatorUrl });
+
+const resourceServer = new x402ResourceServer(facilitatorClient)
+  .register(NETWORK, new ExactEvmScheme())
+  .onAfterSettle(async ctx => {
+    const payload = ctx.paymentPayload.payload as { authorization: { from: string } };
+    const address = payload.authorization.from;
+    const resource = new URL(ctx.paymentPayload.resource.url).pathname;
+    recordPayment(resource, address);
+  });
```
phdargen: This hook could also be implemented in the extension package for convenience
```suggestion
  .onAfterSettle(createSIWxSettleHook({ storage }))
  });
```

phdargen:
> 
> Implementation note: we use conditional middleware in the `server` example. I considered putting special logic into `paymentMiddleware` but decided that was against the premise of optional extensions. Another pattern that could be clean would be to put a `skipIf: (req) => bool` lambda as an optional param in the `paymentMiddleware` which takes the simple SIWx verification function. Punting on that for now.

Concerning the conditional middleware hack, I propose a `onProtectedRequest` hook in https://github.com/coinbase/x402/pull/1010 that goes in the direction of your `skipIf` suggestion but is more general-purpose and aligns well with our existing lifecycle hook structure. `onProtectedRequest` would run in `x402HTTPResourceServer` on every request to a protected route, before payment processing. It can grant access without payment, deny the request or continue the payment flow.

The hook could then be implemented in the siwx package: 
```
export function createSIWxRequestHook(options: CreateSIWxRequestHookOptions): ProtectedRequestHook {
  const { storage } = options;

  return async (
    context: HTTPRequestContext,
    _routeConfig: RouteConfig,
  ): Promise<void | { grantAccess: true } | { abort: true; reason: string }> => {
    // Check for SIWX header (case-insensitive)
    const header = context.adapter.getHeader(SIGN_IN_WITH_X.toLowerCase());
    if (!header) {
      return; // No SIWX header, continue to payment flow
    }

    try {
      // Parse the SIWX payload from the header
      const payload = parseSIWxHeader(header);

      // Get the full resource URI for validation
      const resourceUri = context.adapter.getUrl();

      // Validate the SIWX message 
      const validation = await validateSIWxMessage(payload, resourceUri);
      if (!validation.valid) {
        // Invalid SIWX message, continue to payment flow
        return;
      }

      // Verify the signature and recover the address
      const verification = await verifySIWxSignature(payload);
      if (!verification.valid || !verification.address) {
        // Invalid signature, continue to payment flow
        return;
      }

      // Check if this address has previously paid for this resource
      const hasPaid = await storage.hasPaid(context.path, verification.address);
      if (hasPaid) {
        console.log(`SIWX: Access granted for ${verification.address} on ${context.path}`);
        return { grantAccess: true };
      }

      // Address hasn't paid, continue to payment flow
    } catch (error) {
      // Error processing SIWX, continue to payment flow
      console.warn("SIWX: Error processing header:", error);
    }
  };
}
```

phdargen:
All together this would simplify the server example significantly:
```
import {
  declareSIWxExtension,
  createSIWxRequestHook,
  createSIWxSettleHook,
  InMemorySIWxStorage,
} from "@x402/extensions/sign-in-with-x";

const storage = new InMemorySIWxStorage();

const resourceServer = new x402ResourceServer(facilitatorClient)
  .register(NETWORK, new ExactEvmScheme())
  .onAfterSettle(createSIWxSettleHook({ storage }))

const httpServer = new x402HTTPResourceServer(resourceServer, routes)
  .onProtectedRequest(createSIWxRequestHook({ storage }));

const app = express();

app.use(paymentMiddlewareFromHTTPServer(httpServer));
```

phdargen:
On the client side, the current SIWX client implementation requires two separate wrappers: 
- wrapFetchWithPayment: handles payment
- wrapFetchWithSIWx: handles SIWX authentication

Users must understand which wrapper to use when and each client needs its own SIWX wrapper implementation (currently only fetch implemented, axios missing).

In https://github.com/coinbase/x402/pull/1012 I propose a new `onPaymentRequired` hook in `x402HTTPClient` mirroring the server-side `onProtectedRequest` hook in `x402HTTPResourceServer` discussed above. On 402 it can first retry with authentication headers before trying payment. 

SIWX package would just implement the hook (no client changes or additional fetch/axios wrappers needed):
```
export function createSIWxClientHook(signer: SIWxSigner): PaymentRequiredHook {
  return async ({ paymentRequired }) => {
    const siwxExtension = paymentRequired.extensions?.[SIGN_IN_WITH_X] as
      | { info: SIWxExtensionInfo }
      | undefined;

    if (!siwxExtension?.info) {
      return; // No SIWX support, continue to payment
    }

    const payload = await createSIWxPayload(siwxExtension.info, signer);
    const siwxHeader = encodeSIWxHeader(payload);

    return {
      headers: { [SIGN_IN_WITH_X]: siwxHeader },
    };
  };
}

```

Example usage would simplify to:
```
import { createSIWxClientHook } from "@x402/extensions/sign-in-with-x";

const client = new x402Client();
registerExactEvmScheme(client, { signer });

const httpClient = new x402HTTPClient(client)
  .onPaymentRequired(createSIWxClientHook(signer));

const fetchWithPayment = wrapFetchWithPayment(fetch, httpClient);
```
An onPaymentRequired hook for a svm signer could be easily chained

phdargen:
The new hooks in https://github.com/coinbase/x402/pull/1010 and https://github.com/coinbase/x402/pull/1012 have been merged @sragss

sragss:
Thanks for the thorough review team!

Addressed DevX across the board.

  - Added `SIWxStorage` interface with `hasPaid()` / `recordPayment()` methods
  - Added `InMemorySIWxStorage` default implementation
  - Added `createSIWxSettleHook` for `onAfterSettle` - records payment after successful settlement
  - Added `createSIWxRequestHook` for `onProtectedRequest` - validates SIWX header and grants access if
  previously paid
  - Added `createSIWxClientHook` for `onPaymentRequired` - automatically sends SIWX auth header before
  paying

  Multi-chain support: settle hook extracts payer from `payload.authorization.from` (EVM) or
  `payload.payer` (Solana).

  Simplified examples now use hooks directly:
  ```ts
  // Server
  const storage = new InMemorySIWxStorage();
  const resourceServer = new x402ResourceServer(facilitatorClient)
    .register(NETWORK, new ExactEvmScheme())
    .onAfterSettle(createSIWxSettleHook({ storage }));

  const httpServer = new x402HTTPResourceServer(resourceServer, routes)
    .onProtectedRequest(createSIWxRequestHook({ storage }));

  // Client
  const httpClient = new x402HTTPClient(client)
    .onPaymentRequired(createSIWxClientHook(signer));
```

**Comment by phdargen** [comment]
Thanks a lot for the quick turnaround @sragss, much appreciated! Looks great! 

Could you please fix the failing unit tests, retesting the examples now

File: examples/typescript/clients/sign-in-with-x/index.ts [hunk 3]
```diff
@@ -0,0 +1,57 @@
...
+  // First request: pays for access
+  console.log("1. First request (paying)...");
+  const response1 = await fetchWithPayment(url);
+  console.log("   Response:", await response1.json());
+
+  // Second request: SIWX hook automatically proves we already paid
+  console.log("2. Second request (SIWX auth)...");
+  const response2 = await fetchWithPayment(url);
+  console.log("   Response:", await response2.json());
+}
```
phdargen, line 57: Would be nice to explicitly check for a PAYMENT-RESPONSE header here and print it if available. Also if running the client example twice, the first request on the second run should also be unpaid (if signature still valid), so better to check for payment than assuming a certain flow

phdargen:
The `declareSIWxEntension()` function generates `issuedAt`, `exiprationTime` and `nonce` only once at server startup. After 5mins (current default `exiprationTime`) of running the server, SIWX validation will always fail.

These properties need to be refreshed for each 402 response. Fortunately we can reuse the `enrichDeclaration` hook that the bazaar extension uses to select the http request method. The enrichDeclaration function is automatically called when the extension is registered via .registerExtension(). It's invoked for each 402 response in processHTTPRequest at 
https://github.com/coinbase/x402/blob/428647f2fcad17c00c6e9a97e4f99d4857ce19b2/typescript/packages/core/src/http/x402HTTPResourceServer.ts#L432-L434

An implementation of the `enrichDeclaration` in the extension package could then override the `info.issuedAt` etc values of the extension declaration. The absolute `exiprationTime` property in the 402 response is fine but a relative parameter in `declareSIWxExtension` might be more straightforward.

Server example would change to sth like this:
```
const resourceServer = new x402ResourceServer(facilitatorClient)
  .register(NETWORK, new ExactEvmScheme())
  .registerExtension(siwxResourceServerExtension)
  .onAfterSettle(createSIWxSettleHook({ storage }));
```

phdargen:
> The `declareSIWxEntension()` function generates `issuedAt`, `exiprationTime` and `nonce` only once at server startup. After 5mins (current default `exiprationTime`) of running the server, SIWX validation will always fail.
> 

Hm this was working within the 5mins timeframe implying that there is currently no nonce validation. 
I'd suggest to extend the `SIWxStorage` as follows:
```
export interface SIWxStorage {
  // Payment tracking (required)
  hasPaid(resource: string, address: string): boolean | Promise<boolean>;
  recordPayment(resource: string, address: string): void | Promise<void>;
  
  // Nonce tracking (optional - "Server SHOULD track used nonces to prevent replay attacks" per spec)
  hasUsedNonce?(nonce: string): boolean | Promise<boolean>;
  recordNonce?(nonce: string): void | Promise<void>;
}
```
and implement a simple nonce tracking in `InMemorySIWxStorage`.

If `storage` implements `hasUsedNonce/recordNonce`, nonce validation is automatically enabled and the `createSIWxRequestHook` passes a `checkNonce` fct (already defined in `SIWxValidationOptions`) to `validateSIWxMessage`

phdargen:
Testing a svm example made me realise that a server currently can only advertise a single authentication method since
the spec has a single info object with fixed chainId, type and signatureScheme
https://github.com/coinbase/x402/blob/76307be3672c9893d01baad3eb3889a71b0411be/specs/extensions/sign-in-with-x.md?plain=1#L30-L44

A clean option for multiple auth support could look like this:
```
"sign-in-with-x": {
  "info": {
    "domain": "api.example.com",
    "uri": "https://api.example.com/premium-data",
    "version": "1",
    "nonce": "a1b2c3d4e5f67890...",
    "issuedAt": "2024-01-15T10:30:00.000Z",
    "expirationTime": "2024-01-15T10:35:00.000Z",
    "statement": "Sign in to access premium data"
  },
  "supportedChains": [
    { "chainId": "eip155:8453", "type": "eip191", "signatureSchemes": ["eip191", "eip1271", "eip6492"] },
    { "chainId": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp", "type": "ed25519", "signatureScheme": "siws" }
  ]
}
```
but not sure what implications this would have on the implementation. I guess clients would then need to try all auth methods. Looping in @CarsonRoscoe for some guidance.

sragss:
Appreciate the thorough review! Nice catch on the expirationTime. I believe I've addressed all the comments and implemented multi-siwe provider as well as tests. New Storage interface is clean and should work well with Redis which I expect will be typical.

I don't think the client looping through auth methods will be an issue, can just match on chainId / extension.

Summary from Claude:

  - Implemented `enrichDeclaration` hook to refresh time-based fields per-request (fixes 5-minute expiration bug)
  - Extended `SIWxStorage` with optional `hasUsedNonce()` / `recordNonce()` for replay attack prevention
  - Added multi-chain support - `network` accepts string or array
  - Changed API: `expirationTime` (absolute) → `expirationSeconds` (relative) to support infinite expiration
  - Added server-side logging via `onEvent` callbacks

  Server example simplified to:
  ```ts
  const storage = new InMemorySIWxStorage();

  const resourceServer = new x402ResourceServer(facilitatorClient)
    .register(NETWORK, new ExactEvmScheme())
    .registerExtension(siwxResourceServerExtension)
    .onAfterSettle(createSIWxSettleHook({ storage }));

  const httpServer = new x402HTTPResourceServer(resourceServer, routes)
    .onProtectedRequest(createSIWxRequestHook({ storage }));

```

  Breaking changes:
  - expirationTime?: string → expirationSeconds?: number in DeclareSIWxOptions
  
  Multi-chain:
  // Single-chain
  `declareSIWxExtension({ network: 'eip155:8453', ... })`

  // Multi-chain (EVM + Solana)
`declareSIWxExtension({ network: ['eip155:8453', 'solana:...'], ... })`

**Review by CarsonRoscoe — COMMENTED** [review]
Hey @sragss and @phdargen,

I am aligned with the assessment that the current SIWX spec shape implicitly limits a server to advertising a single auth configuration, which becomes problematic for multi-chain servers (e.g. EVM + SVM)

Conceptually, separating message-level metadata (domain, uri, nonce, issuedAt, expirationTime, statement) from authentication capabilities makes sense. A structure like a top-level info plus a supportedChains[] (chainId + signature schemes) feels like the right long-term direction

I don't have a strong opinion on how clients determine which to use. My rule of thumb is to filter down the list based on clients' registered support, and then iterate through the remaining options in the order presented, which effectively makes the order as declared by the server their preference. However open to ideas

**Review by CarsonRoscoe — COMMENTED** [review]
Ah great timing @sragss, see you added multi-network while I was typing! Will give it a look 🙏

**Comment by sragss** [comment]
Ah @CarsonRoscoe @phdargen my current version duplicates the `info` field and all children `domain` `uri` etc. I think your version

```
  {
    "sign-in-with-x": {
      "info": {
        "domain": "...",
        "uri": "...",
        "nonce": "...",
        "issuedAt": "...",
        "expirationTime": "...",
        "statement": "..."
      },
      "supportedChains": [
        { "chainId": "eip155:8453", "type": "eip191", "signatureSchemes": ["eip191", "eip1271", "eip6492"] },
        { "chainId": "solana:5eykt...", "type": "ed25519", "signatureScheme": "siws" }
      ]
    }
  }
```

is cleaner. Let me take a stab at that tm.

**Comment by sragss** [comment]
I've updated to the new schema for multi-chain support. 
  - Migrated to supportedChains[] architecture - separates shared info (domain, uri, nonce, issuedAt) from per-chain auth options
  - Added siwxResourceServerExtension that auto-derives fields from context:
    - network → from accepts[].network in route config
    - resourceUri → from request URL
    - domain → parsed from resourceUri
    - Time-based fields (nonce, issuedAt, expirationTime) refreshed per-request (fixes 5-min expiration bug)
  - Simplified declareSIWxExtension() - most configs just need statement, everything else derived
  - Client hook now matches payment network to supportedChains instead of detecting signer chain

Server example is now minimal:
```
  const routes = {
    "GET /weather": {
      accepts: [{ scheme: "exact", price: "$0.001", network: NETWORK, payTo }],
      extensions: declareSIWxExtension({ statement: "Sign in to access" }),
    },
  };

  const resourceServer = new x402ResourceServer(facilitator)
    .registerExtension(siwxResourceServerExtension)
    .onAfterSettle(createSIWxSettleHook({ storage }));

  Extension response shape:
  {
    "sign-in-with-x": {
      "info": { "domain": "...", "uri": "...", "nonce": "...", "issuedAt": "...", ... },
      "supportedChains": [
        { "chainId": "eip155:8453", "type": "eip191" },
        { "chainId": "solana:5eykt...", "type": "ed25519" }
      ]
    }
  }
```

  Updated spec, README, and tests.


---

On nonce tracking: I'm not sure I understand the use-case. I know it's a SIWE suggestion, but when would you want pay once, access once? 

I understand pay once, fixed duration access (solved by `expirationSeconds`). 
And I understand pay once, access forever (solved by `expirationSeconds=undefined`).

It's quite easy to add via the SIWxStorageInterface but I think I'm missing the point in the x402 context.

File: typescript/packages/extensions/src/sign-in-with-x/hooks.ts [hunk 4]
```diff
@@ -0,0 +1,211 @@
...
+  const { storage, onEvent } = options;
+
+  return async (ctx: {
+    paymentPayload: { payload: unknown; resource: { url: string } };
+    result: { success: boolean };
+  }): Promise<void> => {
+    // Only record payment if settlement succeeded
+    if (!ctx.result.success) return;
+
+    const address = extractPayerAddress(ctx.paymentPayload.payload);
```
phdargen: The SVM payload is just `{ transaction: string }`, there's no `payer` field.  Suggest to use `result` instead
```suggestion
    result: { success: boolean; payer?: string };
  }): Promise<void> => {
    // Only record payment if settlement succeeded
    if (!ctx.result.success) return;

    const address = ctx.result.payer;
```

phdargen:
Thanks for the update @sragss, changes look great! I think we are almost there, just one minor fix to extract the svm payer. 

Concerning the nonce check, its not about one-time access but signature reuse. Without the nonce check a valid signature could be intercepted or shared such that the resource could be accessed by people who haven't paid. With nonce check, each access needs a fresh signature. 
Imo no need to actually implement the nonce tracking in`InMemorySIWxStorage`, we can leave it to servers if and how they want to implement it. Just defining optional `hasUsedNonce/recordNonce` in the interface and calling them in the hook (if implemented) would be sufficient

phdargen:
Also please run the linter before you commit

phdargen:
We are having [changelogs](https://github.com/coinbase/x402/pull/1036) now!

Please rebase against main and add an entry as described here: https://github.com/coinbase/x402/blob/main/typescript/CONTRIBUTING.md#changelog

You might get the honour of the first entry 🙏

sragss:
Nice catch again on the Solana payer. Suppose I should've tried a Solana example as well just didn't have a test key setup there :)

Solana fix and tests are in. Nonce tracking set up in the interface: optionally called if implemented. And change log is in as well.

phdargen (approved):

phdargen:
Awesome @sragss, we are good to go 🚀  Thanks again for this contribution!

sragss:
Appreciate the thorough review!

Issue #496, opened 2025-10-25
[Propospal] Verifiable Receipts for x402 Payment Flows


junbeomlee:
## Context

x402 already makes it easy for agents, apps, and APIs to send and receive payments in a standard way.
As more people start using it, we’re seeing a new need:
sometimes it’s not enough to know that a payment happened — we also want to prove what it was for and what was delivered in return.

## Idea

Add an optional verifiable receipt that links the full flow: `request ↔ payment ↔ result`
Each receipt is a small signed object that connects these parts together:

```
{
  "version": "x402-1",
  "request_hash": "0x...",
  "payment_ref": "0x...",
  "output_hash": "0x...",
  "payer": "0x...",
  "payee": "0x...",
  "timestamp": "2025-10-25T12:34:56Z",
  "signature": "0x..."
}
```

This lets anyone verify that:
- A certain request was made
- A payment was sent for it
- A specific result or file was delivered
- The payee signed to confirm delivery

akangaziz:
we need this very hard.

rajpatil7322:
Strongly agree

kladkogex:
I think this can be partially addressed in the current spec by using nonce as a merkle root of arbitrary info. 

In the simplest example nonce can be hash of resource URL  concatenated with a random number. Then, the sender will be able to later relate payment and the resource.

alftom:
OMA3 has put together an initial draft of an x402 extension for signed offers and receipts, which we believe addresses the gap described in this issue:

[https://github.com/oma3dao/x402/blob/spec/extension-receipt/specs/extensions/extension-offer-and-receipt.md](https://github.com/oma3dao/x402/blob/spec/extension-receipt/specs/extensions/extension-offer-and-receipt.md)

This extension grew out of our work on OMATrust, an attestation-based trust framework for the open internet. In permissionless attestation systems, anyone can submit a review for a service, but the ones that should matter most are those backed by a cryptographically verifiable proof of real interaction.

What makes x402 especially powerful in this context is that these proofs can be generated automatically as part of the protocol flow. Unlike reviews or attestations about arbitrary websites or APIs, where it is difficult to prove that an interaction actually occurred, x402 payments create a tight, machine-verifiable linkage between an explicit service offer, a concrete on-chain payment, and a specific requester identity. This gives x402 additional utility beyond payments alone.

This extension proposes two optional artifacts that can be used in both x402 v1 and v2:

- signed payment offer (server to client)
[...]

PR #847, opened 2025-12-22
ecosystem: add PEAC Protocol

jithinraj:
## Description

Add PEAC Protocol to the x402 ecosystem page.

**PEAC Protocol adds verifiable interaction records to x402 payment flows.**

After an HTTP 402 payment succeeds, the server can return a signed PEAC receipt that binds the payment proof, request/response, and policy terms into a portable record. Anyone can verify it independently (including offline) for reconciliation, disputes, and audits—without relying on either party's internal logs.

**Project Details:**
- **Website:** https://peacprotocol.org
- **Category:** Infrastructure & Tooling
- **GitHub:** https://github.com/peacprotocol/peac
- **npm:** [@peac/rails-x402](https://www.npmjs.com/package/@peac/rails-x402)
- **Live Demo:** https://x402.peacprotocol.org

## Tests

This PR only adds a new partner metadata JSON file to the ecosystem page. No code logic changes.

- [x] Verified the metadata.json follows the standard template format
- [x] Confirmed the logo file exists at `/logos/peac-protocol.svg`

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge)

PR #935, opened 2026-01-09
feat(extensions): add optional signed Offer & Receipt extension (draft, addresses #496)

alftom:
## Description

This PR proposes and implements an **optional Offer & Receipt extension** for x402, along with a reference TypeScript implementation and documentation.  It is intended as a response to the discussion in **https://github.com/coinbase/x402/issues/496**.

### What this introduces

- **Signed Offers**: optional, signed metadata attached to `402 Payment Required` responses (accepts[])
- **Signed Receipts**: optional, signed acknowledgements returned on successful payment and service delivery
- **Extension-only design**:
  - fully opt-in
  - no behavior change unless explicitly enabled
  - no impact on existing x402 flows
- **TypeScript reference implementation** under `typescript/packages/extensions/`
- **Server and client usage examples**
- A written extension spec:
  - `specs/extensions/extension-offer-and-receipt.md`

The goal is to enable **portable, verifiable proofs of paid service** that can be reused outside the original request/response flow (e.g. disputes, auditing, or trust signals).  One concrete use case: the [OMATrust](https://github.com/oma3dao/omatrust-docs) project (part of [OMA3](https://www.oma3.org)) plans to use these receipts as proofs for verified user-review attestations.

### What this PR is *not*

- It does **not** require signed offers or receipts
- It does **not** mandate key-binding, identity, or attestation systems

Key authorization (e.g. whether a signing key is bound to a service identity) is intentionally treated as **verifier policy**, not a protocol requirement. The extension only defines the data model and signing semantics.

### Intent of this PR

This PR is opened as a **Draft** to request maintainer and community feedback on the *concept* and *integration surface* before additional hardening.

In particular, feedback is requested on:
1. Whether signed offers and/or receipts belong as an x402 extension
2. Whether the framework-adapter integration surface is appropriate
3. Whether the extension spec and field placement make sense
4. Whether this functionality should live in x402 or remain out-of-band

If there is alignment on the concept, follow-up work can add:
- E2E extension tests
- additional adversarial / tamper test coverage
- expanded examples across frameworks
- JSDoc documentation
- 
## Tests

- **TypeScript unit tests** for the Offer/Receipt extension  
  (`typescript/packages/extensions/test/offer-receipt.test.ts`)
- All existing repository tests pass
- Build and typecheck are clean across the monorepo

No E2E extension test is included in this PR; this is intentional for the draft stage and can be added once maintainers confirm interest in the extension.

## Checklist

- [x] I have formatted my code
- [ ] Lint passes (JSDoc documentation pending - will add before marking ready)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

alftom:
To close the trust loop around this PR, we’ve been integrating these proofs into existing trust and identity standards rather than designing something from the ground up in x402.  [Here's a PR](https://github.com/erc-8004/erc-8004-contracts/pull/20) that's an example of integrating into ERC-8004, leveraging EAS. Of course the receipt model is platform-agnostic so it can be integrated into other ecosystems in the same way.

phdargen:
Hi @alftom, thanks a lot for putting this interesting proposal together! Verifiable receipts would be a valuable addition and a good fit for the new extensions.

We recently had an ecosystem addition from a team working on a project called PEAC that goes into a similar direction, see https://github.com/coinbase/x402/pull/847. @jithinraj, would be great if your team could provide some feedback on this proposal as well. Ideally we could structure this extension such that independent teams could easily build their custom solutions on top of it

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,451 @@
...
+**4.5.1 Signer Authorization**
+
+Verifiers MUST confirm that the signing key is authorized to act on behalf of the service identified by `resourceUrl`. This specification does not mandate a specific authorization mechanism. Common approaches include:
+
+- **`payTo` address signing**: The simplest approach — the service signs with the private key corresponding to the `payTo` address. Verifiers accept the signature if the recovered signer matches `payTo`.
+- **External key registry**: An external system (e.g., DID documents, on-chain attestations, or other key binding mechanisms) maps the signing key or `kid` to the service identity.
+
+**4.6 Offer Expiration**
+
+If `issuedAt` and `maxTimeoutSeconds` are both present and non-zero, verifiers MAY reject offers where:
```
phdargen: Shouldn't this be MUST/SHOULD?
```suggestion
If `issuedAt` and `maxTimeoutSeconds` are both present and non-zero, verifiers MUST reject offers where:
```

phdargen: Is it required for the server to check offer expiration before accepting payment?

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,451 @@
...
+
+**5.2 Receipt Payload Fields**
+
+The canonical receipt payload contains the following fields:
+
+| Field         | Type   | Required | Description                                      |
+| ------------- | ------ | -------- | ------------------------------------------------ |
+| `resourceUrl` | string | Yes      | The paid resource URL                            |
+| `payer`       | string | Yes      | Payer identifier (commonly a wallet address)     |
+| `issuedAt`    | number | Yes      | Unix timestamp (seconds) when receipt was issued |
```
phdargen: Without network in the receipt, a 3rd party verifier can't reconstruct the domain to verify the signature.
Suggest adding it aligning with the offer payload.
```suggestion
| `issuedAt`    | number | Yes      | Unix timestamp (seconds) when receipt was issued |
| `network`           | string | Yes                     | Blockchain network identifier             |
```

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,451 @@
...
+}
+```
+
+**6.3 Success Response with Receipt (x402 v1)**
+
+```json
+{
+  "success": true,
+  "transaction": "0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef",
+  "network": "eip155:8453",
```
phdargen, line 768:
```suggestion
  "network": "base,
```

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,451 @@
...
+          "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
+          "amount": "10000",
+          "maxTimeoutSeconds": 60,
+          "issuedAt": 1703123456
+        },
+        "signature": "0x1234567890abcdef..."
+      }
+    }
+  ]
+}
```
phdargen, line 594: `accepts` should remain unchanged. Extension specific info goes into `'extension'` field, see eg bazaar extension in https://github.com/coinbase/x402/pull/956 or https://github.com/coinbase/x402/blob/main/specs/extensions/erc20_gas_sponsoring.md
```suggestion
{
  "x402Version": 2,
  "resource": {
    "url": "https://api.example.com/premium-data",
    "description": "Access to premium market data",
    "mimeType": "application/json"
  },
  "accepts": [
    {
      "scheme": "exact",
      "network": "eip155:8453",
      "amount": "10000",
      "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
      "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
      "maxTimeoutSeconds": 60,
      "extra": {
        "name": "USDC",
        "version": "2"
      }
    }
  ],
  "extensions": {
    "offer-receipt": {
      "info": {
        "offers": [
          {
            "format": "eip712",
            "payload": {
              "resourceUrl": "https://api.example.com/premium-data",
              "scheme": "exact",
              "settlement": "txid",
              "network": "eip155:8453",
              "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
              "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
              "amount": "10000",
              "maxTimeoutSeconds": 60,
              "issuedAt": 1703123456
            },
            "signature": "0x1234567890abcdef..."
          }
        ]
      },
      "schema": {
        "$schema": "https://json-schema.org/draft/2020-12/schema",
        "type": "object",
        "properties": {
          "offers": {
            "type": "array",
            "description": "Array of signed offers, one per accepts[] entry (same index)",
            "items": {
              "type": "object",
              "properties": {
                "format": { "type": "string", "enum": ["eip712", "jws"] },
                "payload": { "type": "object" },
                "signature": { "type": "string" }
              },
              "required": ["format", "signature"]
            }
          }
        },
        "required": ["offers"]
      }
    }
  }
}
```

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,451 @@
...
+When constructing the EIP-712 domain, `chainId` MUST be derived from the `network` field. For CAIP-2 identifiers of the form `eip155:<id>`, the numeric `<id>` value is used as the EIP-712 `chainId`.
+
+**3.2.1 EIP-712 Schema Is Normative and Not Transmitted**
+
+For `format = "eip712"`, the signing digest is computed using the EIP-712 domain, the message (the artifact payload), and the canonical `types` and `primaryType` defined in this specification.
+
+- The canonical `types` and `primaryType` definitions MUST NOT be included in transmitted x402 messages (offers/receipts).
+- Signers MUST use the canonical `types` and `primaryType` definitions from this specification when producing EIP-712 signatures.
+- Verifiers MUST obtain and use the same canonical `types` and `primaryType` definitions from this specification when verifying EIP-712 signatures.
+- Because EIP-712 hashes the schema into the signature, any change to the canonical `types` or `primaryType` constitutes a breaking change and MUST be accompanied by explicit versioning (e.g., bumping the EIP-712 domain `version` or publishing a new spec version).
```
phdargen, line 92: Should offer/receipts include a version string then?

phdargen:
The biggest difference wrt PEAC implementation is omitting txHash, amount in the receipt. 

While I appreciate the privacy-minimal approach, the `payer`, `issuedAt`, `resourceUrl` (which implies `payTo`) info should be sufficient to query the blockchain and find the exact transaction.
So I'd argue that the privacy benefit is limited. 
Adding txHash, amount in the receipt would have the benefit that 3rd parties could easily verify that payment actually happened.  Otherwise, a dishonest server could issue receipts without actual payments and spoof reviews, etc. 

Could consider making these fields at least optional

phdargen:
As a general comment about the implementation, the core packages should stay lean and new extensions should ideally not modify them at all. Otherwise core will get bloated quickly and loose interoperability with other sdks that might only implement the core specs. 

Currently this PR:
* defines OfferReceiptSigner in @x402/core/types
* adds offerReceiptConfig parameter to every middleware
* bakes sign/receipt logic into x402HTTPResourceServer

If the implementation lived entirely in extensions/offer-receipt, it could even work as a third-party plugin outside the official SDK

phdargen:
> As a general comment about the implementation, the core packages should stay lean and new extensions should ideally not modify them at all. Otherwise core will get bloated quickly and loose interoperability with other sdks that might only implement the core specs.
> 
> Currently this PR:
> 
> * defines OfferReceiptSigner in @x402/core/types
> * adds offerReceiptConfig parameter to every middleware
> * bakes sign/receipt logic into x402HTTPResourceServer
> 
> If the implementation lived entirely in extensions/offer-receipt, it could even work as a third-party plugin outside the official SDK

I see the difficulty is that you need to intercept the flow before headers are created in `x402HTTPResourceServer`. I'd suggest to introduce new lifecycle hooks that extend the interface. 

```
// @x402/core/types/extensions.ts 
export interface ResourceServerExtension {
  key: string;
  enrichDeclaration?: (declaration: unknown, transportContext: unknown) => unknown;
  
  enrichPaymentRequirements?: (
    requirements: PaymentRequirements[],
    resourceUrl: string,
    context: unknown
  ) => Promise<PaymentRequirements[]>;
  
  enrichSettlementResponse?: (
    response: SettleResponse,
    paymentPayload: PaymentPayload,
    requirements: PaymentRequirements
  ) => Promise<SettleResponse>;
}
```
where `x402HTTPResourceServer` calls the hook eg after settlement:
```
// @x402/core/src/http/x402HTTPResourceServer.ts: processHTTPRequest()
let settleResponse = await this.ResourceServer.settlePayment(...);
for (const ext of this.ResourceServer.getExtensions()) {
  if (ext.enrichSettlementResponse) {
    settleResponse = await ext.enrichSettlementResponse(settleResponse, paymentPayload, requirements);
  }
}
const headers = this.createSettlementHeaders(settleResponse, requirements, settleResponse.receipt);
```

Then the implementation of enrichPaymentRequirements/enrichSettlementResponse goes into @x402/extensions/offer-receipt. Will discuss this with the team

jithinraj:
> Hi @alftom, thanks a lot for putting this interesting proposal together! Verifiable receipts would be a valuable addition and a good fit for the new extensions.
> 
> We recently had an ecosystem addition from a team working on a project called PEAC that goes into a similar direction, see #847. @jithinraj, would be great if your team could provide some feedback on this proposal as well. Ideally we could structure this extension such that independent teams could easily build their custom solutions on top of it

Thanks for the draft @alftom, and thanks @phdargen for looping PEAC in. Portable, verifiable proofs of paid service are a real missing primitive.

A few interop-focused suggestions (aligned with the review comments above):

1) Extension-only + lean core
Strong +1 on keeping `@x402/core` stable so core stays interoperable across SDKs. The `ResourceServerExtension` hook surface you sketched (`enrichPaymentRequirements` + `enrichSettlementResponse`) looks like the right pattern and keeps offer/receipt logic entirely under `extensions/` without per-extension config threading through core/middleware.

2) Keep `accepts[]` unchanged
Agree `accepts[]` should remain the stable base contract. Extension-specific data belongs under the standard `extensions` surface (consistent with other x402 extensions).

3) Expiration semantics
Server SHOULD reject expired offers before accepting/settling, and MUST NOT sign/return receipts for expired offers. Verifiers MAY also reject based on policy, but a deterministic default helps avoid inconsistent behavior across clients.

4) Third-party verification completeness
For trust-signal use cases (reviews/attestations/audits), receipts should support an optional settlement reference when available (txHash/transferId/etc.), and optionally a commitment variant (sha256:…) for privacy-sensitive flows. Also +1 that receipt should include `network` consistent with the offer.

5) Explicit versioning
+1 on including an explicit `schemaVersion` (or strict domain version discipline) in signed payloads since signature verification is brittle across type changes; especially if canonical EIP-712 types/primaryType are kept out-of-band.

If useful, we can share (a) a minimal mapping showing how PEAC records embed/reference the x402 offer/receipt proof (x402 = handshake proof; PEAC = audit/dispute record when needed), and (b) a small tamper/replay/expiry conformance vector set.

slavik0329:
Nice integration. The receipt -> attestation flow for verified reviews is a solid EAS use case.

Following the thread on privacy vs verifiability, it's worth considering offchain attestations for this. Users may not want purchase history publicly indexed. The receipt could stay private and only be disclosed selectively when needed for verification.

specs/extensions/extension-offer-and-receipt.md
alftom: Good catch.  First of all, the language is incorrect.  "verifiers" should be "resource servers", as servers are the ones doing the rejecting.  We could probably add that clients SHOULD check the expiration before accepting an offer and paying (but perhaps this belongs in the base specs since it's independent of signed offers).  

Second, my thinking on using MAY is that the offer is like a coupon with an expiration.  If the coupon has expired the merchant that offered the coupon can still elect to honor it even though it expired.

specs/extensions/extension-offer-and-receipt.md
alftom: Another good catch.  There are actually two ways to solve this:  A) your suggestion (mention that 'network' needs to be a CAIP-2 ID), or B) remove chainId from the domain.  It seems A is preferable since chainId is commonly used in EIP712Domain.  If we go with A we'd need to change 5.2 and 5.3 (add 'network' to both).

specs/extensions/extension-offer-and-receipt.md
alftom: Got it.  I considered using the extensions field as well, so this makes sense.  The client will need to compare the object fields rather than relying on array ordering with multiple offers though.  So, the tradeoff is keeping the accepts array untouched in exchange for more client complexity.

specs/extensions/extension-offer-and-receipt.md
alftom: That makes a lot of sense. We can add another row to 4.2 and 5.2, and percolate those changes throughout the rest of the document.

alftom:
> The biggest difference wrt PEAC implementation is omitting txHash, amount in the receipt.
> 
> While I appreciate the privacy-minimal approach, the `payer`, `issuedAt`, `resourceUrl` (which implies `payTo`) info should be sufficient to query the blockchain and find the exact transaction. So I'd argue that the privacy benefit is limited. Adding txHash, amount in the receipt would have the benefit that 3rd parties could easily verify that payment actually happened. Otherwise, a dishonest server could issue receipts without actual payments and spoof reviews, etc.
> 
> Could consider making these fields at least optional

I agree that making txHash optional is the best approach.  Leave it up to the servers whether they want to include it.  I do want to say that the reasons for including it are very complex though.  The privacy benefits need to be explored in more detail depending on the use case (e.g.- if an agent is hitting dozens of APIs in a small timeframe can you really find the exact transaction?).  Verifiers can filter out user reviews that come from an attester that only reviews one resource (those look fraudulent).  We can figure out policy recos later.

alftom:
> > As a general comment about the implementation, the core packages should stay lean and new extensions should ideally not modify them at all. Otherwise core will get bloated quickly and loose interoperability with other sdks that might only implement the core specs.
> > Currently this PR:
> > 
> > * defines OfferReceiptSigner in @x402/core/types
> > * adds offerReceiptConfig parameter to every middleware
> > * bakes sign/receipt logic into x402HTTPResourceServer
> > 
> > If the implementation lived entirely in extensions/offer-receipt, it could even work as a third-party plugin outside the official SDK
> 
> I see the difficulty is that you need to intercept the flow before headers are created in `x402HTTPResourceServer`. I'd suggest to introduce new lifecycle hooks that extend the interface.
> 
> ```
> // @x402/core/types/extensions.ts 
> export interface ResourceServerExtension {
>   key: string;
>   enrichDeclaration?: (declaration: unknown, transportContext: unknown) => unknown;
>   
>   enrichPaymentRequirements?: (
>     requirements: PaymentRequirements[],
>     resourceUrl: string,
>     context: unknown
>   ) => Promise<PaymentRequirements[]>;
>   
>   enrichSettlementResponse?: (
>     response: SettleResponse,
>     paymentPayload: PaymentPayload,
>     requirements: PaymentRequirements
>   ) => Promise<SettleResponse>;
> }
> ```
> 
> where `x402HTTPResourceServer` calls the hook eg after settlement:
> 
> ```
> // @x402/core/src/http/x402HTTPResourceServer.ts: processHTTPRequest()
> let settleResponse = await this.ResourceServer.settlePayment(...);
> for (const ext of this.ResourceServer.getExtensions()) {
>   if (ext.enrichSettlementResponse) {
>     settleResponse = await ext.enrichSettlementResponse(settleResponse, paymentPayload, requirements);
>   }
> }
> const headers = this.createSettlementHeaders(settleResponse, requirements, settleResponse.receipt);
> ```
> 
> Then the implementation of enrichPaymentRequirements/enrichSettlementResponse goes into @x402/extensions/offer-receipt. Will discuss this with the team

New extension hooks would be great.  Touching core packages is a burden on extension designers as well.  I didn't expect it to take this long to put together this PR for example.  If you implement it I can change the submission.

For the other comments on this PR (versioning, network in receipt, offer expiration wording, optional txHash, moving offers to extensions field), let me know if you want me to update the PR or you prefer to make the changes yourself.

jithinraj:
Thanks @alftom. +1 on keeping core lean: the hook approach (`enrichPaymentRequirements` / `enrichSettlementResponse`) feels like the right surface so offer/receipt can live entirely under `extensions/offer-receipt` (and even work as a third-party plugin).

One interop detail to keep verification deterministic while preserving `accepts[]`:
- Include an `acceptIndex` (0-based) inside the *signed* offer/receipt payload. Verifiers select `accepts[acceptIndex]` and confirm terms match; out-of-range or mismatch should be treated as invalid. This makes reordering/misalignment tamper-evident.

On sequencing (cc @phdargen): if hooks are the direction, it may be simplest to land the hook surface first as a small core PR, then have this PR focus on the extension spec + examples + implementation on top.

phdargen:
Thanks for the follow-up @alftom!

I will implement the hooks in a separate PR and let the team review it.
In the meanwhile, it would be great if you could update just the specs doc with what was discussed above. 

Once we have the hooks and the specs ready, we can discuss strategy about the implementation

specs/extensions/extension-offer-and-receipt.md
phdargen: I think its recommended to include chainID in the EIP712Domain

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,451 @@
...
+| Field               | Type   | Required                | Description                               |
+| ------------------- | ------ | ----------------------- | ----------------------------------------- |
+| `resourceUrl`       | string | Yes                     | The paid resource URL                     |
+| `scheme`            | string | Yes                     | Payment scheme identifier (e.g., "exact") |
+| `settlement`        | string | Yes (v2), Optional (v1) | Settlement type (e.g., "txid")            |
+| `network`           | string | Yes                     | Blockchain network identifier             |
+| `asset`             | string | Yes                     | Token contract address or "native"        |
+| `payTo`             | string | Yes                     | Recipient wallet address                  |
+| `amount`            | string | Yes                     | Required payment amount                   |
+| `maxTimeoutSeconds` | number | Server optional         | Offer validity window in seconds          |
```
phdargen: `maxTimeoutSeconds` is an existing payload field of core and shouldn't be repurposed for the extension, please rename to avoid confusion

alftom: It's the same value as the `maxTimeoutSeconds` in the existing payload field.  The resource server just puts the same value in the offer payload, just like the other fields in the offer (e.g.- asset).  I'll change the description to "Maximum time allowed for payment completion" to mimic the v1 and v2 spec.

alftom:
> Thanks @alftom. +1 on keeping core lean: the hook approach (`enrichPaymentRequirements` / `enrichSettlementResponse`) feels like the right surface so offer/receipt can live entirely under `extensions/offer-receipt` (and even work as a third-party plugin).
> 
> One interop detail to keep verification deterministic while preserving `accepts[]`:
> 
> * Include an `acceptIndex` (0-based) inside the _signed_ offer/receipt payload. Verifiers select `accepts[acceptIndex]` and confirm terms match; out-of-range or mismatch should be treated as invalid. This makes reordering/misalignment tamper-evident.
> 
> On sequencing (cc @phdargen): if hooks are the direction, it may be simplest to land the hook surface first as a small core PR, then have this PR focus on the extension spec + examples + implementation on top.

I think this could be a good addition, but a client should still confirm with other field(s) as well.  Thoughts @phdargen?  My first spec revision won't include it since it's more for convenience, but I'd be happy to add it later.

alftom:
@phdargen I have a question about extensions and versions.  The `SettlementResponse` message doesn't define an `extensions` field, which is why the extension spec adds a new `receipt` field.  Should we stick with this or add the `extensions` field at the top level and use the same structure as PaymentRequired?

phdargen:
> @phdargen I have a question about extensions and versions. The `SettlementResponse` message doesn't define an `extensions` field, which is why the extension spec adds a new `receipt` field. Should we stick with this or add the `extensions` field at the top level and use the same structure as PaymentRequired?

Good question @alftom, I would prefer the same structure as PaymentRequired ( SettlementResponse.extensions.{extension-name} ) such that metadata associations are clear when multiple extensions are used

alftom:
> Hi @alftom, thanks a lot for putting this interesting proposal together! Verifiable receipts would be a valuable addition and a good fit for the new extensions.
> 
> We recently had an ecosystem addition from a team working on a project called PEAC that goes into a similar direction, see #847. @jithinraj, would be great if your team could provide some feedback on this proposal as well. Ideally we could structure this extension such that independent teams could easily build their custom solutions on top of it

@phdargen I updated the pull request with a new ext spec.  @jithinraj PEAC looks really interesting.  Let me know if you have any more suggestions.

> Thanks for the follow-up @alftom!
> 
> I will implement the hooks in a separate PR and let the team review it. In the meanwhile, it would be great if you could update just the specs doc with what was discussed above.
> 
> Once we have the hooks and the specs ready, we can discuss strategy about the implementation

@phdargen I updated the pull request with a new ext spec.

alftom:
> Thanks @alftom. +1 on keeping core lean: the hook approach (`enrichPaymentRequirements` / `enrichSettlementResponse`) feels like the right surface so offer/receipt can live entirely under `extensions/offer-receipt` (and even work as a third-party plugin).
> 
> One interop detail to keep verification deterministic while preserving `accepts[]`:
> 
> * Include an `acceptIndex` (0-based) inside the _signed_ offer/receipt payload. Verifiers select `accepts[acceptIndex]` and confirm terms match; out-of-range or mismatch should be treated as invalid. This makes reordering/misalignment tamper-evident.
> 
> On sequencing (cc @phdargen): if hooks are the direction, it may be simplest to land the hook surface first as a small core PR, then have this PR focus on the extension spec + examples + implementation on top.

@jithinraj PEAC looks really interesting.  Thanks for your contributions on this.  Let me know if you have any more suggestions.

phdargen:
Thanks for the update @alftom, I will have a look at it tmr. 
Have a draft PR ready with the new hooks here: https://github.com/coinbase/x402/pull/1003

specs/extensions/extension-offer-and-receipt.md
phdargen: The change you made here is good. 

But I was referring more to 4.6 Offer Expiration, where you define: `now > issuedAt + maxTimeoutSeconds`. `maxTimeoutSeconds` is a property of core and should not be repurposed in the extension. I'd suggest to add a new `offerValiditySeconds` property to offers. If set and `now > issuedAt + offerValiditySeconds`, the server may reject the service

specs/extensions/extension-offer-and-receipt.md
```diff
@@ -0,0 +1,849 @@
...
+      "amount": "10000",
+      "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
+      "payTo": "0x209693Bc6afc0C5328bA36FaF03C514EF312287C",
+      "maxTimeoutSeconds": 60
+    }
+  ],
+  "extensions": {
+    "offer-receipt": {
+      "info": {
+        "version": 1,
```
phdargen: Should version be included in the payload rather than top-level?

phdargen:
Hi @alftom, I am mostly good with the specs now. I'd suggest to add optional `offerValiditySeconds` and `metadata` (eg for TS or privacy policy) properties and to move version inside the payload. 

PR with the new hooks is here: https://github.com/coinbase/x402/pull/1003
As a test I made a quick prototype for the `offer-receipt `extension here: https://github.com/phdargen/x402/tree/offer-receipt
As far as I can tell, the hooks are sufficient to fully implement it as a standalone package

The extension just needs to implement the hook:
```
enrichSettlementResponse: async (declaration: unknown, context: SettleResultContext) => {

        ...

        const signedReceipt: SignedReceipt = await signer.signReceipt(
          resourceUrl,
          payer,
          network,
          txHashToInclude,
          config?.metadata,
        );
        return {
          info: {
            version: 1,
            receipt: signedReceipt,
          },
        };
}
```
see https://github.com/phdargen/x402/blob/offer-receipt/typescript/packages/extensions/src/offer-receipt/extension.ts

Server config would work like this:
```
  const resourceServer = new x402ResourceServer(facilitatorClient)
    .register("eip155:84532", new ExactEvmScheme())
    // Register the offer-receipt extension
    .registerExtension(createOfferReceiptExtension(signer))
    // Enable server-side offer validation (validates acceptIndex match and validUntil expiration)
    .onBeforeVerify(offerValidationHook);

  const app = express();
  app.use(
    paymentMiddleware(
      {
        "GET /weather": {
          accepts: {
            scheme: "exact",
            price: "$0.001",
            network: "eip155:84532",
            payTo: evmAddress,
          },
          description: "Weather data with signed offer/receipt",
          mimeType: "application/json",
          // Enable offer-receipt extension for this route
          extensions: {
            ...declareOfferReceipt({  includeTxHash: false, validitySeconds: 60, metadata: {
                tos: "https://example.com/terms-of-service",
                version: "1.0",
                provider: "Example Corp",
              }, }),
        },
      resourceServer,
    ),
  );
```
see https://github.com/phdargen/x402/blob/offer-receipt/examples/typescript/servers/offer-receipt/index.ts

phdargen:
For the prototype implementation I also found @jithinraj `acceptIndex` idea quite convenient and suggest we add that to the specs

alftom:
> For the prototype implementation I also found @jithinraj `acceptIndex` idea quite convenient and suggest we add that to the specs

I agree that the convenience is worth achieving, but there are actually two ways to solve it:  1) `acceptIndex` and 2) require that the `offers` array order match the `accepts` array order.  Both rely on the server to implement correctly, and if they client wants to make sure they need to match the fields.  `acceptIndex` is probably more convenient than ordering, but it adds a new field to the signed offer that has nothing to do with the offer itself.  You could remove it before signing the object, but that adds complexity.  2) seems almost as convenient (the client implementation has to keep track of the index) and keeps the signed offer more pure cryptographically.  My preference (putting my "standards hat" on) is to go with 2) but I haven't done the implementation yet.

specs/extensions/extension-offer-and-receipt.md
alftom: Yes, this makes sense because an offer will have value outside of the x402 protocol, and consumers of the offer need to know the version of the offer structure. I'll start working on the update.

alftom:
> Hi @alftom, I am mostly good with the specs now. I'd suggest to add optional `offerValiditySeconds` and `metadata` (eg for TS or privacy policy) properties and to move version inside the payload.

Agree with moving version inside the payload (see prev comment). I'm not exactly clear about your suggestion on `offerValiditySeconds` and `metadata` though.

- Are you suggesting replacing `maxTimeoutSeconds` with `offerValiditySeconds`, or adding an additional `offerValiditySeconds`? If the latter, how would it be different from `offerValiditySeconds`?

- Would `metadata` also go inside of the signed payload? If so, what would be the structure? We could leave it for a future version now that we're adding version inside the signed payload. 

> 
> PR with the new hooks is here: #1003 As a test I made a quick prototype for the `offer-receipt `extension here: https://github.com/phdargen/x402/tree/offer-receipt As far as I can tell, the hooks are sufficient to fully implement it as a standalone package
> 

I will take a look. Do you want me to wait for PR #1003 to be merged into main before changing the code in this PR?

jithinraj:
Thanks @alftom and @phdargen, this is converging. Quick thoughts:

- **acceptIndex vs ordering**  
An optional `acceptIndex` inside the *signed* offer/receipt payload seems like a clean binding to a specific `accepts[]` entry. Relying on “offers[] order matches accepts[]” is an implicit coupling that can be reordered (or accidentally mismatched) without anything inside the signed object indicating it. If `acceptIndex` is present, verifiers should still confirm `accepts[acceptIndex]` terms match (so the index is never trusted alone).

- **offer validity**  
This would be **add** an offer-expiry field in the signed offer payload, not repurpose core `maxTimeoutSeconds` (which stays the payment-completion window). `validUntil` (unix seconds) is the clearest; `offerValiditySeconds` (+ `issuedAt`) also works if you prefer relative semantics.

- **metadata**  
If included in v1: inside the signed payload, kept minimal and URL-shaped (e.g., `tos`, `privacy`, `provider`). Anything richer can wait for a later version.

re #1003 / implementation: Updating the spec in this PR now makes sense. For the TS reference implementation, it should reduce rework to port it once the #1003 hooks are merged, so the extension can stay fully standalone under `extensions/`.

alftom:
> * **acceptIndex vs ordering**
>   An optional `acceptIndex` inside the _signed_ offer/receipt payload seems like a clean binding to a specific `accepts[]` entry. Relying on “offers[] order matches accepts[]” is an implicit coupling that can be reordered (or accidentally mismatched) without anything inside the signed object indicating it. If `acceptIndex` is present, verifiers should still confirm `accepts[acceptIndex]` terms match (so the index is never trusted alone).

As you say, `acceptIndex` doesn't solve the ordering problem either because it fails if the accepts[] array re-orders, which is why you correctly recommend verifying the other terms match. `acceptIndex` is a convenient suggestion.  My bigger worry is that once the signed payload leaves that specific x402 session (e.g.- it's used in PEAC infra) the `acceptIndex` now becomes meaningless and potentially misleading.  

> * **offer validity**
>   This would be **add** an offer-expiry field in the signed offer payload, not repurpose core `maxTimeoutSeconds` (which stays the payment-completion window). `validUntil` (unix seconds) is the clearest; `offerValiditySeconds` (+ `issuedAt`) also works if you prefer relative semantics.

Any field name works for me.  

> * **metadata**
>   If included in v1: inside the signed payload, kept minimal and URL-shaped (e.g., `tos`, `privacy`, `provider`). Anything richer can wait for a later version.

OK. Should be a solid, well thought out proposal to include it in v1.  Otherwise we can intro it later.

> re #1003 / implementation: Updating the spec in this PR now makes sense. For the TS reference implementation, it should reduce rework to port it once the #1003 hooks are merged, so the extension can stay fully standalone under `extensions/`.

Agree.  I'll add a comment to the #1003 PR to show the changes we would make.

alftom:
> > * **acceptIndex vs ordering**
> >   An optional `acceptIndex` inside the _signed_ offer/receipt payload seems like a clean binding to a specific `accepts[]` entry. Relying on “offers[] order matches accepts[]” is an implicit coupling that can be reordered (or accidentally mismatched) without anything inside the signed object indicating it. If `acceptIndex` is present, verifiers should still confirm `accepts[acceptIndex]` terms match (so the index is never trusted alone).
> 
> As you say, `acceptIndex` doesn't solve the ordering problem either because it fails if the accepts[] array re-orders, which is why you correctly recommend verifying the other terms match. `acceptIndex` is a convenient suggestion. My bigger worry is that once the signed payload leaves that specific x402 session (e.g.- it's used in PEAC infra) the `acceptIndex` now becomes meaningless and potentially misleading.

If we want to include `acceptsIndex` in each `offers[]` entry one approach would be to:

* keep `acceptsIndex` out of the signed payload and place it alongside `format` and `signature`
* define the signature as covering only the payload
* allow `acceptsIndex` to be stripped when the signed offer is reused in other contexts (e.g. attestations or PEAC)

This adds a bit more complexity but addresses my above concern.

BTW- I started to use `validUntil` in the spec update and I'm looking into the `metadata` field based on @phdargen's implementation.

jithinraj:
Good point on portability.

One concern with keeping `acceptIndex` *outside* the signature is that it becomes a hint only (it can be modified in transit), so verifiers can’t rely on it for binding and still end up doing full term-matching anyway.

If the goal is a binding that remains meaningful once the signed object is reused elsewhere, a more portable option could be an optional `acceptRef`/`acceptDigest` *inside* the signed payload (e.g., hash of canonicalized accept terms as defined by the spec). Then `acceptIndex` can stay purely as an in-band convenience selector (and be safely ignored/stripped downstream).

Either way, keeping the binding inside what’s signed preserves tamper-evidence.

alftom:
I looked at the metadata implementation from @phdargen's [code](https://github.com/phdargen/x402/tree/offer-receipt) and I have a few questions:

- What is the use case for each of the fields in `metadata`?
- What happens when the content of the URLs change (which is often)?  Does that undermine the purpose of the field?
- Is `provider` expected to have a different base URL than `resource.url`?  If yes, how do we prevent a random resource server from claiming `coinbase.com`?  If not, why is this field necessary?

Answering these questions will help us build a solid spec or decide if we should push it to version 2.

phdargen:
Hi @alftom @jithinraj, the hook PR https://github.com/coinbase/x402/pull/1003 is merged

phdargen:
> I looked at the metadata implementation from @phdargen's [code](https://github.com/phdargen/x402/tree/offer-receipt) and I have a few questions:
> 
> * What is the use case for each of the fields in `metadata`?
> * What happens when the content of the URLs change (which is often)?  Does that undermine the purpose of the field?
> * Is `provider` expected to have a different base URL than `resource.url`?  If yes, how do we prevent a random resource server from claiming `coinbase.com`?  If not, why is this field necessary?
> 
> Answering these questions will help us build a solid spec or decide if we should push it to version 2.

I just put in some random fields for quick testing, so yeah 'provider' is not needed. The use case I had in mind was to include a term of service or similar in the offer but you are right that it wouldn't mean much as content can change (unless its a ipfs hash). Another thing to consider is that headers can not be arbitrarily large. 

I am good with keeping it simple for version 1 and reduce to the strictly necessary, we can see what use cases develop out of the extension and iterate

alftom:
> Hi @alftom @jithinraj, the hook PR #1003 is merged

I just updated the spec adding `validUntil` and `acceptIndex`, removing `issuedAt` and `maxTimeoutSeconds`, and moving `version` to the signed payload as per @jithinraj reaction above.  `metadata` will be left out of this version as per @phdargen's comment above.  If you're good with the spec I can start on the code changes.

jithinraj:
Thanks for the updates @alftom and @phdargen, this is getting close.

One spec-language note: since `acceptIndex` is outside the signed payload, verifiers MUST NOT rely on it as binding. They MUST verify that `accepts[acceptIndex]` matches the signed payload terms (and reject on out-of-range / mismatch).

Otherwise +1 to proceed with the hook-based refactor now that #1003 is merged.

alftom:
> Thanks for the updates @alftom and @phdargen, this is getting close.
> 
> One spec-language note: since `acceptIndex` is outside the signed payload, verifiers MUST NOT rely on it as binding. They MUST verify that `accepts[acceptIndex]` matches the signed payload terms (and reject on out-of-range / mismatch).
> 
> Otherwise +1 to proceed with the hook-based refactor now that #1003 is merged.

Thanks @jithinraj. The language is actually in the spec at line 130:

```
**Within the x402 session (clients):**

When `acceptIndex` is present, clients SHOULD:
- Check that `acceptIndex` is in-range for the `accepts[]` array
- Validate that `accepts[acceptIndex]` terms match the signed payload fields (`network`, `asset`, `payTo`, `amount`, etc.)

Clients MUST NOT treat `acceptIndex` as authoritative — field matching against the signed payload is the source of truth.
```


@phdargen let me know when you've signed off on the spec.

phdargen:
Hi @alftom, thanks for the update. I am good with the specs!

alftom:
> Hi @alftom, thanks for the update. I am good with the specs!

@phdargen I updated the PR over the weekend with all the code.  Two things I was unsure about:  1) does wrapFetchWithPayment support the extension hooks yet?  The client example uses the manual flow for now.  2) does the e2e mock server support the extension hooks yet?  If it does I will work on e2e tests next.

@jithinraj I added a comment for PEAC integration in the client example.

phdargen:
> @phdargen I updated the PR over the weekend with all the code. Two things I was unsure about: 1) does wrapFetchWithPayment support the extension hooks yet? The client example uses the manual flow for now. 2) does the e2e mock server support the extension hooks yet? If it does I will work on e2e tests next.

Thanks for the update @alftom!

1) The new hooks I added are server-side and were needed to keep the core clean. As far as I can tell, getting the offer and receipt from the server responses should work without modifying core already. What hooks are you missing?

2) The e2e servers are real (not mock). They use `workspace:*` dependencies, so you can just import what you need from the extension package. But actually not sure how we handle e2e tests for the extensions (the discovery extension might be a bit of a special case), let me look into this. For now, I think it would be best just to add a server example in `examples/typescript/servers/offer-receipt/` that shows how to setup the extension. Please keep it as close as possible to the existing basic [express example](https://github.com/coinbase/x402/tree/main/examples/typescript/servers/express) and only add the few lines to register/declare the extension

examples/typescript/clients/receipt-attestation/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. Install dependencies from the typescript examples root:
+
+```bash
+cd ../../
+pnpm install && pnpm build
+cd clients/receipt-attestation
+```
+
+2. Copy `.env.example` to `.env` and configure:
```
phdargen: Nit: rename to be consistent with other examples
```suggestion
2. Copy `.env-local` to `.env` and configure:
```

examples/typescript/clients/receipt-attestation/index.ts
```diff
@@ -0,0 +1,246 @@
...
+  //      a) did:web document at https://<domain>/.well-known/did.json
+  //      b) DNS TXT record binding the DID to the domain
+  //      c) On-chain attestation (e.g., OMATrust key binding attestation)
+  //
+  // This verification is typically performed by downstream trust systems
+  // (OMATrust, PEAC) when you submit the receipt as proof. However, clients
+  // can also verify directly if they need immediate trust decisions.
+  //
+  // See: specs/extensions/extension-offer-and-receipt.md §4.5.1
+  // =========================================================================
```
phdargen, line 196: Please keep text in examples minimal and move all this additional info (including Integration Point below) to the README

examples/typescript/clients/receipt-attestation/index.ts
```diff
@@ -0,0 +1,246 @@
...
+ * x402 Receipt Attestation Client Example
+ *
+ * Demonstrates extracting signed offers and receipts from x402 payment flows.
+ * Uses the raw flow for visibility into what's happening at each step.
+ *
+ * Use Cases for Signed Receipts/Offers:
+ * - Verified user reviews ("Verified Purchase" badges)
+ * - Audit trails and compliance records
+ * - Dispute resolution evidence
+ * - Agent memory (AI agents proving past interactions)
```
phdargen: move to README

examples/typescript/clients/receipt-attestation/index.ts
```diff
@@ -0,0 +1,246 @@
...
+//    In production, use secure key management (HSM, KMS, hardware wallets).
+//
+// 2. KEY SEPARATION
+//    The payment signing key SHOULD be different from keys controlling
+//    wallets with significant funds.
+//
+// 3. KEY-TO-DOMAIN BINDING (for servers)
+//    See specs/extensions/extension-offer-and-receipt.md §4.5.1
+//
+// ============================================================================
```
phdargen: think this would also be sufficient in the readme

examples/typescript/clients/receipt-attestation/index.ts
```diff
@@ -0,0 +1,246 @@
...
+  //
+  // IMPORTANT: Key Binding Verification
+  // ------------------------------------
+  // The extraction functions (extractReceiptPayload, extractOfferPayload) do
+  // NOT verify signatures. They only decode the payload. Before trusting a
+  // receipt or offer, you should verify:
+  //
+  // 1. SIGNATURE VALIDITY
+  //    - JWS: Use jose.compactVerify() with the public key
+  //    - EIP-712: Use viem's recoverTypedDataAddress() to recover the signer
```
phdargen: Could we add some utility functions in the extension that implement this?

phdargen:
The client example looks good to me, I think we could leave manual flow here. If you'd like we could abstract the offer capture from the end-user by using the existing `onPaymentRequired` hook. Flow could look like this: 
```
const httpClient = new x402HTTPClient(client);
const offerCapture = createOfferCaptureHook();
httpClient.onPaymentRequired(offerCapture.hook);

const fetchWithPay = wrapFetchWithPayment(fetch, httpClient);
const response = await fetchWithPay(url);

// Get offers and receipt
const offers = offerCapture.getOffers();
const receipt = extractReceiptFromResponse(response);
```
with a hook interface like:
```
export interface OfferCaptureHook {
  /** The hook to register with httpClient.onPaymentRequired() */
  hook: PaymentRequiredHook;
  /** Get the captured offers from the last 402 response */
  getOffers: () => DecodedOffer[];
  /** Clear the captured offers */
  clear: () => void;
}
```

phdargen:
I don't see a meaningful devx improvement with an `onPaymentComplete` hook. The client already receives the signed-receipt with the `PAYMENT-HEADER` in the `const response = await fetchWithPay(url);` call which can then passed through to utility functions to extract/verify the receipt

typescript/packages/extensions/src/offer-receipt/server.ts
```diff
@@ -0,0 +1,381 @@
...
+    return ctx.requestUrl;
+  }
+
+  // Try request.url
+  if (ctx?.request?.url) {
+    return ctx.request.url;
+  }
+
+  return undefined;
+}
```
phdargen: The `requestUrl/request` fallback fields dont exist and can be removed
```suggestion
function extractResourceUrl(transportContext: unknown): string | undefined {
  const ctx = transportContext as HTTPRequestContext | undefined;
  return ctx?.adapter?.getUrl();
}
```
with `import type { HTTPRequestContext } from "@x402/core/server";`.  `HTTPTransportContext` can be removed

typescript/packages/extensions/src/offer-receipt/server.ts
```diff
@@ -0,0 +1,381 @@
...
+    async signReceipt(resourceUrl: string, payer: string, network: string, transaction?: string) {
+      const chainId = extractEIP155ChainId(network);
+      return createReceiptEIP712(
+        { resourceUrl, payer, network, transaction },
+        chainId,
+        signTypedData,
+      );
+    },
+  };
+}
```
phdargen: ChainId for the EIP712 signature must correspond to the server signer, not the payment network. 
Would break for a server supporting eg solana payments otherwise.

Also suggest to derive the kid internally given server network, address.  

```suggestion
export function createEIP712OfferReceiptSigner(
  network: string,
  address: string,
  signTypedData: SignTypedDataFn,
): OfferReceiptSigner {
  const chainId = extractEIP155ChainId(network);
  const kid = `did:pkh:${network}:${address}`;

  return {
    kid,
    format: "eip712",

    async signOffer(resourceUrl: string, input: OfferInput) {
      return createOfferEIP712(resourceUrl, input, chainId, signTypedData);
    },

    async signReceipt(resourceUrl: string, payer: string, network: string, transaction?: string) {
      return createReceiptEIP712(
        { resourceUrl, payer, network, transaction },
        chainId,
        signTypedData,
      );
    },
  };
}
```

typescript/packages/extensions/src/offer-receipt/signing.ts
```diff
@@ -0,0 +1,714 @@
...
+ * EIP-712 types for Offer (§4.3)
+ */
+export const OFFER_TYPES = {
+  Offer: [
+    { name: "version", type: "uint256" },
+    { name: "resourceUrl", type: "string" },
+    { name: "scheme", type: "string" },
+    { name: "network", type: "string" },
+    { name: "asset", type: "string" },
+    { name: "payTo", type: "address" },
```
phdargen: That's another issue for a server accepting solana payments, need to be changed to "string" here and elsewhere including specs
```suggestion
    { name: "payTo", type: "string" },
```

typescript/packages/extensions/src/offer-receipt/server.ts
```diff
@@ -0,0 +1,381 @@
...
+ * @param config - Optional configuration for the extension
+ * @returns Extension declaration object to spread into route config
+ */
+export function declareOfferReceiptExtension(
+  config?: OfferReceiptDeclaration,
+): Record<string, OfferReceiptDeclaration> {
+  return {
+    [OFFER_RECEIPT]: {
+      includeTxHash: config?.includeTxHash,
+      maxTimeoutSeconds: config?.maxTimeoutSeconds,
```
phdargen: Please rename to offerValiditySeconds (or whatever you prefer) to avoid confusion with core's `maxTimeoutSeconds`
```suggestion
      offerValiditySeconds: config?. offerValiditySeconds,
```

alftom:
@phdargen thanks — your comments make sense and I’ll update the PR.  I do want your opinion on the EIP-712 issue you surfaced though. This feels like an ambiguity in the spec, not just an implementation detail. To resolve it, there are a few questions I'd like your opinion on:

1. Should we constrain EIP-712 to EVM offers/receipts only, and require JWS for non-EVM payment networks?  If so, deriving `chainId` from `network` (e.g. `eip155:*`) would be acceptable and unambiguous, since we’re using EIP-712 as an off-chain attestation format — not something intended to be submitted to a blockchain.  If we do this, `payTo` can remain typed as `address`.

2. Should we require a `kid` for EIP-712 payloads, just like we already do for JWS?  The `kid` (DID URL) would identify the server signer. Then `chainId` could be derived from the `kid` instead of the payment `network`. If we still allow non-EVM offers/receipts in this model, we’d also need to change `payTo` from `address` to `string` as you pointed out.

We could also do both of the above.  And there’s a third question that reflects a simpler solution:

3. Should we eliminate EIP-712 entirely and rely on JWS only, given that Ethereum keys can sign JWS (ES256K) and it avoids these cross-domain ambiguities altogether?

Let me know your thoughts.

examples/typescript/clients/receipt-attestation/index.ts
alftom: Yes, but this depends on how we answer the questions about EIP-712 spec changes in my previous comment.

phdargen:
> 1. Should we constrain EIP-712 to EVM offers/receipts only, and require JWS for non-EVM payment networks?  If so, deriving `chainId` from `network` (e.g. `eip155:*`) would be acceptable and unambiguous, since we’re using EIP-712 as an off-chain attestation format — not something intended to be submitted to a blockchain.  If we do this, `payTo` can remain typed as `address`.

No, that would be confusing and an unnecessary constrain. EIP-712 is just used as a signing method here, shouldn't be concerned about the data it signs. 

> 2. Should we require a `kid` for EIP-712 payloads, just like we already do for JWS?  The `kid` (DID URL) would identify the server signer. Then `chainId` could be derived from the `kid` instead of the payment `network`. If we still allow non-EVM offers/receipts in this model, we’d also need to change `payTo` from `address` to `string` as you pointed out.

Yes that would work. 

ChainId in the EIP-712 domain is mainly for replay protection for onchain tx. As we are not submitting anything onchain, maybe we can remove it? Only concern is that some implementations (eg smart wallets) might require it.
An alternative to kid, imo we could just define a canonical default value of chainID=1

> 3. Should we eliminate EIP-712 entirely and rely on JWS only, given that Ethereum keys can sign JWS (ES256K) and it avoids these cross-domain ambiguities altogether?

I like that we have a crypto-native and a web2 signature method, prefer not to mix them up

alftom:
> > 2. Should we require a `kid` for EIP-712 payloads, just like we already do for JWS?  The `kid` (DID URL) would identify the server signer. Then `chainId` could be derived from the `kid` instead of the payment `network`. If we still allow non-EVM offers/receipts in this model, we’d also need to change `payTo` from `address` to `string` as you pointed out.
> 
> Yes that would work.
> 
> ChainId in the EIP-712 domain is mainly for replay protection for onchain tx. As we are not submitting anything onchain, maybe we can remove it? Only concern is that some implementations (eg smart wallets) might require it. An alternative to kid, imo we could just define a canonical default value of chainID=1

Let's go with requiring chainId=1.  I'll update the PR.

alftom:
@phdargen I updated the PR to address all the comments except this one:

> Also suggest to derive the kid internally given server network, address.

Is it possible for you to clarify this comment?

phdargen:
> @phdargen I updated the PR to address all the comments except this one:
> 
> > Also suggest to derive the kid internally given server network, address.
> 
> Is it possible for you to clarify this comment?

you can ignore, should no longer be relevant with the chainId=1 approach

examples/typescript/servers/offer-receipt/README.md
```diff
@@ -0,0 +1,120 @@
...
+Required variables:
+- `FACILITATOR_URL` — Facilitator endpoint
+- `EVM_ADDRESS` — Address to receive payments
+- `SIGNING_PRIVATE_KEY` — Base64-encoded PKCS#8 private key (ES256/P-256)
+- `SERVER_DOMAIN` — Domain for DID construction
+
+2. Generate a signing key:
+
+```bash
+openssl ecparam -genkey -name prime256v1 -noout | openssl pkcs8 -topk8 -nocrypt | base64 -w0
```
phdargen: Didnt work on mac, I believe this is more platform agnostic
```suggestion
openssl ecparam -genkey -name prime256v1 -noout | openssl pkcs8 -topk8 -nocrypt -outform DER | openssl base64 -A
```

typescript/packages/extensions/src/offer-receipt/server.ts
```diff
@@ -0,0 +1,344 @@
...
+      // Get network and transaction from settlement result
+      const network = context.result.network;
+      if (!network) {
+        console.warn("[offer-receipt] No network available for signing receipt");
+        return undefined;
+      }
+      const transaction = context.result.transaction;
+
+      // Get resource URL from captured URL during declaration enrichment
+      const resourceUrl = currentResourceUrl;
```
phdargen: This is kind of a hack. I am working on an update of the `enrichPaymentRequiredResponse/enrichSettleResponse` hooks to give them additional transport context 
in https://github.com/coinbase/x402/pull/1101. 

The capture var could then be removed and be replaced by
```
      const httpCtx = context.transportContext as HTTPRequestContext | undefined;
      const resourceUrl = httpCtx?.adapter?.getUrl();
```

typescript/packages/extensions/src/offer-receipt/signing.ts
```diff
@@ -0,0 +1,844 @@
...
+ * Sign a payload using JWS Compact Serialization
+ *
+ * @param payload - The payload object to sign
+ * @param signer - The JWS signer
+ * @returns The JWS compact serialization string
+ */
+export async function signJWS<T extends object>(payload: T, signer: JWSSigner): Promise<string> {
+  const canonical = canonicalize(payload);
+  const payloadBytes = new TextEncoder().encode(canonical);
+  return signer.sign(payloadBytes);
```
phdargen: Had to make these changes to get the server example running, please double check
```suggestion
  const headerObj = { alg: signer.algorithm, kid: signer.kid };
  const headerB64 = jose.base64url.encode(new TextEncoder().encode(JSON.stringify(headerObj)));

  const canonical = canonicalize(payload);
  const payloadB64 = jose.base64url.encode(new TextEncoder().encode(canonical));

  const signingInput = new TextEncoder().encode(`${headerB64}.${payloadB64}`);
  const signatureB64 = await signer.sign(signingInput);

  return `${headerB64}.${payloadB64}.${signatureB64}`;
```

phdargen:
Thanks for the update @alftom!

I tested the server+client example and encountered several issues that prevented it from working end-to-end:
- `signJWS` doesn't seem to produce valid JWS compact serialization (see comment above)
- Server example doesn't serve a DID document. For did:web JWS verification, the server needs to expose /.well-known/did.json with the public key
- local DID resolution fails since resolveDidWeb always uses HTTPS, even for localhost

phdargen:
Please also add the eip712 signing method to the server example with a env var `SIGNER_TYPE="jws" || "eip712"` to choose the option

alftom:
@phdargen PR has been updated.  In addition to addressing the above issues I changed the test script so it no longer masks the signing bug you uncovered.

typescript/packages/extensions/src/offer-receipt/server.ts
phdargen: https://github.com/coinbase/x402/pull/1101 has been merged @alftom, please rebase and use the transport context as suggested above

phdargen:
Hi @alftom, sorry for the delay as I was on holidays. Could you please implement the one change suggested above, then I will test again

typescript/packages/extensions/src/offer-receipt/server.ts
alftom: @phdargen the PR has been updated.  I decided not to do a rebase yet because there were some conflicts with early commits on this branch and I'd like to keep the history in case there are more changes needed.  Happy to squash everything down to a clean commit once we've confirmed the integration looks good on your end.

phdargen: Thanks for the update @alftom! Merging the upstream changes in is also fine, there should be no need to do a rebase or squash

typescript/packages/extensions/src/offer-receipt/server.ts
```diff
@@ -0,0 +1,336 @@
...
+      // Get network and transaction from settlement result
+      const network = context.result.network;
+      if (!network) {
+        console.warn("[offer-receipt] No network available for signing receipt");
+        return undefined;
+      }
+      const transaction = context.result.transaction;
+
+      // Get resource URL from transport context
+      const resourceUrl = extractResourceUrl(context.transportContext);
```
phdargen: Retesting the examples this fails to retrieve the url and I am getting `[offer-receipt] No resource URL available for signing receipt`

phdargen: Suggest to remove extractResourceUrl and do this inline:
```suggestion
      const resourceUrl =(context.transportContext as HTTPTransportContext)?.request?.adapter?.getUrl?.();
```
with `import type { HTTPTransportContext } from "@x402/core/http";` instead of `import type { HTTPRequestContext } from "@x402/core/http";`

typescript/packages/extensions/src/offer-receipt/server.ts
```diff
@@ -0,0 +1,336 @@
...
+    enrichPaymentRequiredResponse: async (
+      declaration: unknown,
+      context: PaymentRequiredContext,
+    ): Promise<unknown> => {
+      const config = declaration as OfferReceiptDeclaration | undefined;
+
+      // Get resource URL from transport context or payment required response
+      const resourceUrl =
+        context.paymentRequiredResponse.resource?.url ||
+        extractResourceUrl(context.transportContext);
```
phdargen: same change here
```suggestion
        (context.transportContext as HTTPTransportContext)?.request?.adapter?.getUrl?.();
```

specs/extensions/README.md
```diff
@@ -0,0 +1,31 @@
+# Extensions
```
phdargen: This file can be removed

phdargen:
Please also add a changelog entry https://github.com/coinbase/x402/blob/main/typescript/CONTRIBUTING.md#changelog

examples/typescript/clients/offer-receipt/index.ts
```diff
@@ -0,0 +1,245 @@
...
+  });
+
+  if (!paidResponse.ok) {
+    console.error(`Payment failed: ${paidResponse.status}`);
+    return;
+  }
+
+  const responseBody = await paidResponse.json();
+  console.log("Response:", responseBody);
+
```
phdargen, line 150: Please add:
```
  const paymentResponse = httpClient.getPaymentSettleResponse(name =>
    paidResponse.headers.get(name),
  );
  console.log("\nPayment response:", JSON.stringify(paymentResponse, null, 2));
```

typescript/packages/extensions/src/offer-receipt/server.ts
alftom: @phdargen PR has been updated to fix this and address your other comments.  I'll need to set up e2e to catch these issues in the future...

phdargen:
Thanks for the update @alftom, I am happy with the implementation now and retested without issues!

phdargen:
One more question on verifiability:  An external verifier can confirm a eip712 signature is valid and recover the signer address but nothing binds it to the actual server. For example in the edge case of a server that only accepts svm payments but signs with a evm key, there is no mapping to the `payTo` address. Would it make sense to add an optional `signer` field (or kid ?) field to the 402 response to give verifiers a hint?

alftom:
> One more question on verifiability: An external verifier can confirm a eip712 signature is valid and recover the signer address but nothing binds it to the actual server. For example in the edge case of a server that only accepts svm payments but signs with a evm key, there is no mapping to the `payTo` address. Would it make sense to add an optional `signer` field (or kid ?) field to the 402 response to give verifiers a hint?

Yes, this is worthwhile addressing.  Keeping the `payTo` wallet separate from the receipt-signing wallet is good security practice, so it's not just the SVM/EVM use case.  In the future a resource server could have multiple `payTo` addresses for privacy (e.g.- Koharu wallet).  

If a `signer` field is added to the 402 response the client can verify the signer wallet's signature, but third parties are not privy to the 402 response or the session so they can't verify in the same way the client can.  There needs to be a way for third parties verify that the signing wallet is authorized by the resource URL.  

There are existing solutions.  The resource owner can put the signer wallet address in a DNS TXT entry or host a .well-known JSON object off the resource URL domain (e.g.- did.json) that contains the signer wallet address.  This shows the base URL authorizes the signer wallet.  

To make the system even more bullet proof a trusted third party can submit a witness attestation (e.g.- Ethereum Attestation Service) to anchor the authorization.  This prevents the resource from removing the address leaving verifiers with no authorization confirmation. However, this is not required.

JWKS and DID have standardized the .well-known mechanism.  ERC-8004 can be extended to have signer authorization contained in the agent.json object.  OMATrust is in the process of standardizing the DNS TXT mechanism and building a [witness system](https://docs.omatrust.org/reputation/attestation-types). 

Bottom line is that the signer authorization problem can be solved outside of x402.  x402 offer and receipt does everything else (integrity + signer recoverability).

A1igator:
Hey all! Just wanted to say we at https://www.x402r.org/ would love to see this proposal get merged soon as it would be great for dispute resolution to have a receipt attached to a merchant payload. Thanks for the great work!

phdargen:
Thanks for the details @alftom, I agree that the signer authorization problem is out of scope of this extension. 

Still I was wondering if it would be helpful for 3rd parties to include an optional kid in the signed EIP-712 payload to tell them if they should look eg for a .well-known or agent.json etc. Otherwise they need to try all common solutions.

But this is more of a convenience, not a necessity and I trust your judgement as someone already building on top of this

phdargen:
There is also a merge conflict, could you please resolve? Probably just in the lock files

alftom:
> Thanks for the details @alftom, I agree that the signer authorization problem is out of scope of this extension.
> 
> Still I was wondering if it would be helpful for 3rd parties to include an optional kid in the signed EIP-712 payload to tell them if they should look eg for a .well-known or agent.json etc. Otherwise they need to try all common solutions.
> 
> But this is more of a convenience, not a necessity and I trust your judgement as someone already building on top of this

@phdargen I appreciate your thoughtfulness on this.  The issue with including a hint is that the resource owner might later decide to change their authorization mechanism. If that happens, the hint in previously issued receipts could become misleading.  An alternative would be to define a verification order outside of the protocol, but this is outside the scope of this extension.  My recommendation is to leave the hint out. That keeps the extension focused on integrity and signer recoverability while leaving signer authorization to the identity layer.  I'm open to other opinions though.

alftom:
> There is also a merge conflict, could you please resolve? Probably just in the lock files

Resolved.

alftom:
> > There is also a merge conflict, could you please resolve? Probably just in the lock files
> 
> Resolved.

@phdargen I resolved the lockfile conflicts, but main is moving quickly and now it's out of date again. Would you prefer I keep syncing with main, or would it be easier to resolve on your end right before approving?

phdargen (approved):

phdargen:
Thanks again for your contribution @alftom, this is a great addition!

We would also be happy to add your project to our ecosystem page, see https://github.com/coinbase/x402/blob/main/typescript/site/README.md#adding-your-project-to-the-ecosystem

alftom:
> Thanks again for your contribution @alftom, this is a great addition!
> 
> We would also be happy to add your project to our ecosystem page, see https://github.com/coinbase/x402/blob/main/typescript/site/README.md#adding-your-project-to-the-ecosystem

Thank you @phdargen!  The first of many contributions hopefully.  Thanks to @CarsonRoscoe, @jithinraj, @slavik0329, and @A1igator for your support as well.  I'll get started on the ecosystem page now.

alftom:
> Thanks again for your contribution @alftom, this is a great addition!
> 
> We would also be happy to add your project to our ecosystem page, see https://github.com/coinbase/x402/blob/main/typescript/site/README.md#adding-your-project-to-the-ecosystem

@phdargen PR #1616

Issue #1011, opened 2026-01-23
[Proposal] Escrow Scheme for x402 using Base Commerce Payments Protocol


A1igator:
# [Proposal] Escrow Scheme for x402 using Base Commerce Payments Protocol

> Based on [Agentokratia's escrow proposal (#834)](https://github.com/coinbase/x402/issues/834) and its reference implementation: [x402-escrow](https://github.com/Agentokratia/x402-escrow) (MIT).

## Summary

We at [x402r.org](https://x402r.org) propose an `escrow` scheme for x402 v2 that leverages the [Base Commerce Payments Protocol](https://github.com/base/commerce-payments) to enable refundable payments and conditional fund handling. The client signs once to authorize a maximum amount, and the facilitator settles through the commerce-payments operator — routing funds into escrow or directly to the receiver.

Built on audited on-chain escrow contracts deployed on Base and other EVM chains, this scheme adds the **auth/capture pattern** as a building block for x402.

## Problem

The `exact` scheme works well for immediate-delivery payments, but creates friction for:

- **High-value transactions** — No recourse if service fails after payment
- **Variable pricing** — Usage-based billing (LLM tokens, compute time, API calls)
- **Long-running tasks** — Work that takes hours or days to complete

## Solution

The `escrow` scheme supports two settlement paths through the commerce-payments operator:

| Method      | Function      | Behavior                                                     |
| :---------- | :------------ | :----------------------------------------------------------- |
| `authorize` | `authorize()` | Funds held in escrow. Can be captured, refunded, or voided.  |
| `charge`    | `charge()`    | Funds sent directly to receiver. Refundable post-settlement. |

Both methods share identical function signatures and use the same operator, fee system, and token collector infrastructure.

### Lifecycle

**Authorize (default):**
```
SIGN → AUTHORIZE → RESOURCE DELIVERED
```
1. Client signs an ERC-3009 `receiveWithAuthorization` for the maximum amount
2. Facilitator calls `authorize()` on the operator — funds locked in escrow
3. Server returns the resource (HTTP 200)

Post-settlement, the commerce-payments contracts enable capture, refund, void, or reclaim.

**Charge:**
```
SIGN → CHARGE → RESOURCE DELIVERED
```
1. Client signs an ERC-3009 authorization (same as above)
2. Facilitator calls `charge()` on the operator — funds go directly to receiver
3. Server returns the resource (HTTP 200)

[...]

PR #1074, opened 2026-02-03
Add upto payment scheme specification for EVM

fabrice-cheng:
## Summary

Introduces the `upto` payment scheme specification for EVM chains. This scheme enables usage-based payments where clients authorize a **maximum amount** and servers settle for the **actual amount used** at the end of the request.

## Key Features

- **Permit2 only**: Uses `permitWitnessTransferFrom` exclusively (EIP-3009 doesn't support variable amounts)
- **Flexible settlement**: Server can settle for any amount from 0 to the authorized maximum
- **New `settledAmount` field**: Added to SettlementResponse to report actual charge
- **Zero settlement allowed**: Unused authorizations expire naturally without on-chain transactions
- **Reuses existing infrastructure**: Uses the same `x402Permit2Proxy` contract

## Use Cases

- LLM token generation (charge per token)
- Bandwidth/data transfer metering (charge per byte)
- Time-based API access (charge per minute)
- Dynamic compute pricing

## Files Added

- `specs/schemes/upto/scheme_upto.md` - Base scheme overview
- `specs/schemes/upto/scheme_upto_evm.md` - Full EVM implementation specification

## Notes

This is a specification document only. Implementation will follow in a separate PR.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>2/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

notorious-d-e-v:
Woohoo, it's time for `upto`! Thanks for putting this spec together @fabrice-cheng!

Questions for Solana: Permit2 gives `upto` three important properties that aren’t native to SPL Token approvals:

- recipient binding (witness.to == payTo)

- explicit time bounds (validAfter/deadline)

- single‑use / replay protection (nonce consumed)

On Solana, the closest primitive is SPL Token / Token‑2022 delegate allowance (`ApproveChecked` + `TransferChecked` via delegate), which can express “up to cap” but:

- has no on‑chain expiry

- has no nonce/single‑use semantics (delegate could transfer multiple times up to the cap)

So before we can write an SVM `upto` spec/implementation, we should clarify what the core scheme MUST guarantee across networks?


1) **Are time limits (validAfter/deadline) REQUIRED properties of `upto`?**
   - If REQUIRED, SVM likely needs a small on‑chain program (Permit2‑like) to enforce expiry.

2) **Is single‑use authorization REQUIRED?**
   - Permit2’s nonce makes each authorization settle at most once. If `upto` requires this, SPL delegate alone can’t provide it without an on‑chain program.

   - For the Solana upto scheme, should an authorization be single-use per request (one settlement transfer total), or should it behave like a spend budget / allowance that can be settled multiple times (streaming/chunking) until the max cap (amount) is exhausted?

      - imo: It should allow multiple settlements so servers can support streaming and chunked delivery (e.g., pay-per-chunk), where each settlement is <= remaining cap and total settled across all chunks is <= amount.


Happy to draft the SVM `upto` scheme doc once we have alignment here. My current read is:
- If (1) and (2) are MUSTs, SVM needs a dedicated program and the program should be the approved delegate.
- If they’re SHOULDs, we can rely on SPL Token and Token 2022 "approve" functions `ApproveChecked` and then trigger a `TransferChecked` after.

fabdarice:
Thanks @notorious-d-e-v for your comment.

My take: `upto` should be single-use + time-bound for security and consistency across all the networks.
The multiple-settlement pattern you describe is valuable but feels like a different payment scheme, like `subscription`, where the agent isn't required in the loop for each payment.

Happy to discuss if there are strong reasons to relax these constraints.

notorious-d-e-v:
Thanks @fabdarice !

On single-use vs multi-use -- the client is already trusting the server not to take more than what's actually owed. So whether the server settles once or multiple times, the trust assumption is the same, imo.

Multi-use feels more practical. Think streaming video or downloading a large file -- we'd want to settle in chunks rather than waiting until the very end. Whether we do that under `upto` or a separate scheme is a design choice, but the flexibility seems worth having.

Either way it seems like for Solana we'll need to deploy a program to enforce expiry and single-use/replay protection, since SPL Token delegates don't have that natively.

On the implementation side, I've been looking at the [Commerce Payments Protocol](https://github.com/base/commerce-payments). It does authorize-and-capture with modular token collectors, configurable fees, time-based expiry, and reclaim mechanisms -- basically a superset of what we need. 

Might be worth deploying something similar for Solana rather than building a narrow single-purpose program. That way we'd have a foundation that could technically support `exact`, `upto`, and whatever comes next.

Any thoughts around that?

specs/schemes/upto/scheme_upto.md
```diff
@@ -0,0 +1,36 @@
...
+
+This scheme is ideal for usage-based pricing models where the final cost is not known until after the resource has been consumed.
+
+## Example Use Cases
+
+- Paying for LLM token generation (charge per token generated)
+- Bandwidth or data transfer metering (charge per byte)
+- Time-based API access (charge per minute/second of usage)
+- Dynamic compute pricing (charge based on actual resources consumed)
+- Streaming media consumption (charge based on content delivered)
```
phdargen: I think streaming isn't really possible because all content is send to client after settlement

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,275 @@
...
+
+> **Note**: EIP-3009 (`transferWithAuthorization`) is **not supported** for the `upto` scheme because it requires exact amounts at signature time.
+
+---
+
+## Use Cases
+
+- **LLM Token Generation**: Client authorizes up to $5, actual charge based on tokens generated
+- **Bandwidth/Data Transfer**: Pay per byte transferred, up to a cap
+- **Time-Based Access**: Authorize hourly max, charge for actual minutes used
```
phdargen: Time based authorization seems out of scope for this scheme

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,275 @@
...
+- **API Rate Metering**: Pay for actual API calls made, up to a budget
```
phdargen: This implies multiple-settlement pattern which requires another scheme per your comment

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,275 @@
...
+    - if ERC20.allowance(from, Permit2_Address) < amount:
+      - Check for **Sponsored ERC20 Approval** (Extension): Refers to [`erc20ApprovalGasSponsoring`](../../extensions/erc20_gas_sponsoring.md).
+      - Check for **EIP2612 Permit** (Extension): Refers to [`eip2612GasSponsoring`](../../extensions/eip2612_gas_sponsoring.md).
+      - **If neither exists:** Return `412 Precondition Failed` (Error Code: `PERMIT2_ALLOWANCE_REQUIRED`). This signals the client that a one-time Direct Approval transaction is required before retrying.
+
+3.  **Verify** the `client` has sufficient balance of the `asset` to cover `amount`.
+
+4.  **Verify** the `permit2Authorization.permitted.amount` equals the `amount` from requirements.
+
+5.  **Verify** the `deadline` (not expired) and `witness.validAfter` (active).
```
phdargen, line 156: I anticipate with `upto` we will get some longer running, compute heavy tasks. Currently verification only checks that deadline > now, but doesn't ensure sufficient runway for API execution + settlement.
Could we add `minTimeoutSeconds` to PaymentRequirements? This would let servers specify the minimum remaining validity they need at verification time and reject authorizations early that would likely expire before API + settlement completes, wasting their resources

phdargen: Not really `upto` specific, would be useful for `exact` as well but think the problem is elevated for `upto`

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,275 @@
...
+| `maxTimeoutSeconds` | `number` | Required | Maximum time allowed for payment completion                                   |
+| `extra`             | `object` | Optional | Scheme-specific additional information (must include `name` and `version`)    |
+
+> **Note**: In the `upto` scheme, the `amount` field represents the **maximum** amount the client authorizes. The actual settled amount may be less than or equal to this value.
+
+---
+
+## 3. SettlementResponse Schema Extension
+
+The `upto` scheme extends the base [`SettlementResponse`](../../x402-specification-v2.md#53-settlementresponse-schema) with the actual settled amount:
```
phdargen, line 231: Suggest to update the base `SettlementResponse` in `x402-specification-v2.md` directly to include `settledAmount` (with `settledAmount=amount` for `exact`) to avoid scheme specific response formats. 
I think even for `exact` its a useful info to get as payment receipt. Clients might use auto-approve for low amounts and could easily see the spent amount in the response without querying the info from the txHash etc

fabrice-cheng: Thanks @phdargen - I'm reluctant to include any x402 main spec change as part of this PR, but this is a valid point and worth addressing as a distinct PR for spec changes.
cc: @erikreppel-cb @CarsonRoscoe thoughts on this? Happy to submit one.

specs/schemes/upto/scheme_upto_evm.md
fabrice-cheng: Thanks @phdargen - I'm reluctant to include any x402 main spec change as part of this PR, but this is a valid point and worth addressing as a distinct PR for spec changes.

cc: @erikreppel-cb @CarsonRoscoe thoughts on this? Happy to submit one.

specs/schemes/upto/scheme_upto_evm.md
fabrice-cheng: fixed, thanks

specs/schemes/upto/scheme_upto_evm.md
fabrice-cheng: fixed, thanks

specs/schemes/upto/scheme_upto.md
fabrice-cheng: fixed, thanks

fabrice-cheng:
> On single-use vs multi-use -- the client is already trusting the server not to take more than what's actually owed. So whether the server settles once or multiple times, the trust assumption is the same, imo.

This is true from a trust perspective, but single-use has other benefits:
- Simpler mental model (one authorization = one settlement)
- No ambiguity about "remaining balance"
- Higher security properties (against the server going rogue in the future or compromised)

> Multi-use feels more practical. Think streaming video or downloading a large file -- we'd want to settle in chunks rather than waiting until the very end. Whether we do that under `upto` or a separate scheme is a design choice, but the flexibility seems worth having.

I think multi-use is very practical as you suggest, but possibly worthwhile having as a distinct payment scheme (i.e: subscription) to clearly segregated the distinct mental models. I'll sync with some other contributors to get their takes.

> On the implementation side, I've been looking at the [Commerce Payments Protocol](https://github.com/base/commerce-payments). It does authorize-and-capture with modular token collectors, configurable fees, time-based expiry, and reclaim mechanisms -- basically a superset of what we need.

Yes, I'm also part of the team that released it. We are deploying similar concepts in for EVM, but AFAIK, the direction is to have a distinct smart contract per payment method / scheme. Regardless, it'll be hard to future proof all needs so it's expected for new contracts to emerge. Happy to discuss more.

notorious-d-e-v:
Agree @fabrice-cheng, sounds good to me!

phdargen:
There is a proposal for an escrow scheme using the Base commerce protocol in https://github.com/coinbase/x402/issues/1011 that could be more suitable for a multi-use pattern

specs/schemes/upto/scheme_upto_evm.md
```diff
@@ -0,0 +1,274 @@
...
+| ------------------- | -------- | -------- | ----------------------------------------------------------------------------- |
+| `scheme`            | `string` | Required | Must be `"upto"`                                                              |
+| `network`           | `string` | Required | Blockchain network identifier in CAIP-2 format (e.g., "eip155:84532")         |
+| `amount`            | `string` | Required | Maximum payment amount in atomic token units                                  |
+| `asset`             | `string` | Required | Token contract address                                                        |
+| `payTo`             | `string` | Required | Recipient wallet address                                                      |
+| `maxTimeoutSeconds` | `number` | Required | Maximum time allowed for payment completion                                   |
+| `extra`             | `object` | Optional | Scheme-specific additional information (must include `name` and `version`)    |
+
+> **Note**: In the `upto` scheme, the `amount` field represents the **maximum** amount the client authorizes. The actual settled amount may be less than or equal to this value.
```
ukstv: The spec as it exists right now is ambiguous with regards to how the amount-to-settle is communicated from the server to the facilitator (even if it is optional to the protocol).

By x402 spec the verify and settle payloads are the same. Here with `upto` the payment requirements for both payloads have slightly different semantics though. `amount` for verify-time payment requirements is the max amount, this is clear. But for settle-time, (my current hunch, and as implemented in Daydreams facilitator and x402-rs facilitator code), we should repurpose `amount` of payment requirements to be the amount-to-settle. Hence suggestion:

```
**Note**: In the `upto` scheme, the `amount` field of `PaymentRequirements` is phase-dependent for server-to-facilitator communication:

- At _verification_ time, `amount` represents the **maximum** amount the client authorizes.
- At _settlement_ time, `amount` represents the **actual amount to settle**, which MUST be less than or equal to the previously authorized maximum.

The actual settled amount is communicated by the resource server to the
facilitator via the `amount` field in the settlement-time payment requirements.
```

ukstv:
Hey, do we need name and version here in extra considering they are not used by the proxy contracts?

specs/schemes/upto/scheme_upto_evm.md
CarsonRoscoe: Good call out,

@fabdarice he's saying when the server calls `facilitatorClient.settle(paymentPayload, paymentRequirements)`, where is the **actual** amount passed in? The amount in `paymentPayload` and `paymentRequirements` both refer to the **maximum** amount, but there is no field for the **actual** amount. Adding a new field would be a breaking change.

Since `PaymentRequirements` is the server sharing what payment it expects to receive from the client, we should override `amount` on `PaymentRequirements` to be the actual amount before calling the facilitator. Then facilitators will refer to `paymentPayload` for the authorization, but `paymentRequirements.amount` for the actual amount

I agree

CarsonRoscoe (commented):
> Hey, do we need name and version here in extra considering they are not used by the proxy contracts?

They may not be. Let me investigate if they are used internally by the eip2612 tokens when validating the eip2612 signatures, but you are correct that our new contracts don't directly take these params in

EDIT: They will likely be required. The proxy contract calls the `permit2` contract, and it reads the `name`/`version` of the asset in question when reconstructing the expected signed payload

specs/schemes/upto/scheme_upto_evm.md
fabrice-cheng: Discussed with @CarsonRoscoe.

`maxTimeoutSeconds` purpose serves what meant by `minTimeoutSeconds`. This params naming is confusing and I think we'll have to rethink it and likely change it for V3 (breaking change).

For the time being, what you are referring to can be set via `maxTimeoutSeconds`

specs/schemes/upto/scheme_upto_evm.md
fabrice-cheng: Discussed with @CarsonRoscoe - Will circle back. Either way, this would be a spec change and will leave outside of the `upto` spec

specs/schemes/upto/scheme_upto_evm.md
fabrice-cheng: Thank you both. Updated.

erikreppel-cb (approved):

specs/schemes/upto/scheme_upto_evm.md
phdargen: `maxTimeoutSeconds` is only used client-side, not enforced in `verify/`. Anyways, will discuss with Carson outside this PR

murrlincoln (approved):

PR #1081, opened 2026-02-03
Add Concordium exact scheme for Concordium blockchain integration

oleh-drozd-eth:


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

CarsonRoscoe (commented):
Hey @oleh-drozd-eth, appreciate the contribution!

I have some concerns regarding the approach of clients submitting the transaction. 

In order for x402 to truly become the internet standard for digital payments, ux and devx are key.
* Clients should have confidence they will only pay for resources they receive
* Servers should have confidence they will only work for payments that are addressed to them and have a high probability of landing
* Clients IDEALLY shouldn't need gas. The facilitator (or a paymaster/sponsor) should be who spends the gas

In this spec's approach,
* Clients' ALWAYS send funds regardless of a servers' behaviour
* Servers' cannot have confidence in verify, as it cannot verify many structural details of the transaction, such as the `payTo`. Verify is effectively a no-op
*  Clients pay for gas.

I would encourage you to look into the Solana/Aptos/Stellar/Sui specs regarding sending signed encoded transactions, rather than transaction hashes. Specifically Solana, as it is most similar with a proven approach. 

Following their approach, in the `extra` field of `PaymentRequirements` you can put a `feePayer`/`sponsor` property which represents the Facilitator address who will sponsor the transaction. Have the client sign a sponsored transaction with the facilitators' address as the authorized user, and then the facilitator is expected to sponsor it and put it on-chain.

If the facilitator puts the transaction on-chain, rather than the client, that means:
* Verify has visibility into the transaction structure and can validate that `payTo`, `amount`, and `asset` (among others) are correct, and that the transaction should land. It also opens up the possibility of facilitators' simulating in verify.
* Clients have confidence that they will not be charged for failing endpoints. They only get charged on successful `settlement`, right before they receive the resource they paid for
* Clients do not pay gas

Let me know if you have any thoughts, and thank you for the spec!

andynelson:
Hi @CarsonRoscoe 

Andy from Concordium here:

In our P10 chain update, which goes live next week, Concordium will support fully sponsored transactions at the protocol level.

These are based on the transaction containing the normal transaction payload, with an extra header field specifying the account which will pay the fees, and both parties signing the transaction before it is submitted.

Our normal flow for these is that the sponsor (which is expected to be a merchant of some kind) constructs the sponsored transaction, signs it, then sends it to the effective (end-user) account's wallet to sign & submit.
(Sponsored transactions increment the sequence number for the effective account, but not the sponsor account, so this ensures the account that could otherwise be disrupted by late or out-of-order transaction submission retains control)  

I think this approach matches well with the requirements you state above in that:

* Since the Server/Facilitator creates the transaction payload, they have all the information they need to verify `payTo`, `amount` and `asset` etc.
* Clients have confidence that they are only paying for something they're going to receive, and won't get hit with gas fees, since the merchant/facilitator already co-signed the transaction agreeing the details and that they will pay the chain fee
* (additionally the client can make a final decision on whether to go-ahead, since they are the last to sign, and control submission of the transaction)
* Since the transaction cannot be changed between the facilitator signing it, and it being submitted, the facilitator only needs to inspect the successful transaction to know that they can deliver, and since it is co-signed by them, it is easily discoverable)
* Clients are do not pay gas. 

Would a spec/implementation based on this flow be acceptable?

oleh-drozd-eth:
Hi,  @CarsonRoscoe 
I updated the Concordium scheme to be compatible with the desired x402 architecture, including sponsored transactions.
Please, lmk if you have any questions. 
Thanks!

CarsonRoscoe (commented):
Hey @oleh-drozd-eth, appreciate the update,

Giving it another review, and a couple thoughts:

#### 1.  Amount check uses `>=` instead of `==`, which violates the intention of `exact`

Verification rule 5 states:

> the transfer `amount` MUST be ≥ `PaymentRequirements.amount`

Every other `exact` scheme spec requires the amount to match exactly

#### 2. Asset identified by empty string

This spec defines:

> `asset`: Empty string `""` for native CCD

However the x402 spec defines [asset](https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md?plain=1#L120)  as `"Token contract address or ISO 4217 currency code for fiat"`. 

That said, I understand why empty string is necessary here, as any other string such as "CCD" itself might collide with a token identifier.

I am fine with empty string, as long as we update the x402 spec to match.

Please include re-defining asset the x402-specification-v2.md doc to `"Token contract address or ISO 4217 currency code for fiat; an empty string ("") MAY represent the network's native asset."` so that your addition here is valid.

#### 3. Signature verification is ambiguously SHOULD vs MUST

Rule 8:

> The sender signature SHOULD be verified cryptographically against the sender's on-chain account credentials

This should be **MUST**, not SHOULD

#### 4 SDK-specific method references

The spec references `Transaction.signableFromJSON()` and `Transaction.Signable.verifySignature()`. Specs should be implementation-agnostic. Please describe the operations abstractly (e.g., "the transaction MUST deserialize successfully from JSON into a valid V1 sponsored transaction").


------

After these changes I am happy to merge!

Apologies for the delays, I appreciate your patience 🙏

andynelson:
For 'asset' would namespacing in a similar way to CAIP-19 work to avoid the empty string?

(Something along the lines of "ccd" for native ccd, "PLT:token" for PLT & contract address for smart contract based tokens)?

It feels like this field being based on CAIP-19 could be the way to go in a future iteration of the standard, since that would cover all of these in a consistent manner.

oleh-drozd-eth:
Hey @CarsonRoscoe , appreciate the detailed review!
I made suggested changes to the Concordium exact spec - strict equality for amount checks, MUST for signature verification, and removed SDK-specific method references. Feel free to check it in the last commits.

Also, I made a change to the x402-v2 spec in the asset section ( added `an empty string ("") MAY represent the network's native asset.`). Could move x402 v2 spec change to a separate PR if you wish, or merge as is.

oleksandradidukh:
Hi @CarsonRoscoe , I hope you are doing great.  Could you please review the PR of @oleh-drozd-eth  https://github.com/x402-foundation/x402/pull/1895 which includes all your comments and suggestions? Thank you in advance

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+```json
+{
+  "scheme": "exact",
+  "network": "ccd:9dd9ca4d19e9393877d2c44b70f89acb",
+  "amount": "1000000",
+  "asset": "",
+  "payTo": "4FmiTW2L4RvCsSVTjFAavYvrgnPLGNj43eiwPYmbhNqtAcMbWW",
+  "maxTimeoutSeconds": 60,
+  "extra": {
+    "sponsorAddress": "4FmiTW2L4RvCsSVTjFAavYvrgnPLGNj43eiwPYmbhNqtAcMbWW"
```
phdargen: Let's align this with SVM/algorand/hedera:
```suggestion
    "feePayer": "4FmiTW2L4RvCsSVTjFAavYvrgnPLGNj43eiwPYmbhNqtAcMbWW"
```

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+  }
+}
+```
+
+- `asset`: Identifies the payment asset for this transaction.
+  - `""` (empty string): Native CCD. This is the default.
```
phdargen, line 50: Not a fan of this notation, would prefer sth more explicit. Either asset="native" or sth like asset=0x0000000000000000000000000000000000000000 

cc @CarsonRoscoe

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+  - Token name (e.g., `"EURR"`): PLT/CIS-2 token identified by its registered symbol.
```
phdargen: Token name doesnt seem to be unambiguous, should use address

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+| PLT Token  | Token name   | `"EURR"` |
+
+### Amount Format
+
+All amounts are expressed in the smallest unit (atomic):
+
+| Asset Type | Unit | Decimals | Example: 10 CCD / 5 EURR |
+|------------|------|----------|---------------------------|
+| Native CCD | microCCD | 6 | `"10000000"` |
+| PLT Token | Smallest subunit | 6 | `"5000000"` |
```
phdargen:
```suggestion
| PLT Token | Smallest subunit | depends on token? | `"5000000"` |
```

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+
+## PaymentPayload `payload` Field
+
+The `payload` field of the `PaymentPayload` contains:
+
+```json
+{
+  "signedTransaction": { ... },
+  "sender": "3kBx2h5Y2veb4hZvAE2c1Zr6DYJwWbPr9xQJJBPWyFnXHF9UuN"
```
phdargen: sender is redundant, already included in signedTransaction.header.sender
```suggestion
  "signedTransaction": { ... }
```

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+  },
+  "accepted": {
+    "scheme": "exact",
+    "network": "ccd:9dd9ca4d19e9393877d2c44b70f89acb",
+    "amount": "1000000",
+    "asset": "",
+    "payTo": "4FmiTW2L4RvCsSVTjFAavYvrgnPLGNj43eiwPYmbhNqtAcMbWW",
+    "maxTimeoutSeconds": 60,
+    "extra": {
+      "sponsorAddress": "4FmiTW2L4RvCsSVTjFAavYvrgnPLGNj43eiwPYmbhNqtAcMbWW"
```
phdargen: nit: choose random address different from `payTo`

specs/schemes/exact/scheme_exact_concordium.md
phdargen: PR https://github.com/x402-foundation/x402/pull/1004 proposed "native"

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+
+6. Asset type
+
+- For native CCD: the transaction MUST be a `SimpleTransfer` or `SimpleTransferWithMemo` type.
+- For PLT tokens: the transaction MUST be a `TokenUpdate` type, and the `tokenId` MUST correspond to `PaymentRequirements.asset`.
+
+7. Transaction expiry
+
+- `transaction.header.expiry` MUST be in the future.
+- The expiry SHOULD NOT exceed 10 minutes from the current time.
```
phdargen: 10mins seems somewhat arbitrary, why not tie to `maxTimeoutSeconds`?

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+  "success": true,
+  "transaction": "a1b2c3d4e5f6789012345678901234567890abcdef1234567890abcdef123456",
+  "network": "ccd:9dd9ca4d19e9393877d2c44b70f89acb",
+  "payer": "3kBx2h5Y2veb4hZvAE2c1Zr6DYJwWbPr9xQJJBPWyFnXHF9UuN"
+}
+```
+
+## Facilitator Verification Rules (MUST)
+
+A facilitator verifying an `exact`-scheme Concordium payment MUST enforce all of the following checks before sponsoring and broadcasting the transaction:
```
phdargen, line 205: Verification checks are so far only structural payload checks but should also check if the tx would actually succeed onchain, ideally with a full tx simulation. Or if thats not possible, at least some targeted checks like sufficient client token balance, nonce unused etc

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,251 @@
...
+
+- `transaction.header.expiry` MUST be in the future.
+- The expiry SHOULD NOT exceed 10 minutes from the current time.
+
+8. Sender signature
+
+- `transaction.signatures.sender` MUST contain at least one credential signature.
+  The sender's credential signature(s) MUST be verified cryptographically against the sender's on-chain account credentials. The facilitator MUST fetch the sender's account information from the Concordium network and validate that the signature was produced by a key authorized for the sender account.
+
+9. Transaction payload safety
```
phdargen, line 248: When constructing the payload, who specifies the gas fees? Client or facilitator? If its the client, there should be an additional check bounding the transaction fee below a reasonable level

phdargen:
Thanks for the update @oleh-drozd-eth @oleksandradidukh, please have a look at my additional comments

CarsonRoscoe (commented):
Please add a PR description/overview as well 🙏

specs/schemes/exact/scheme_exact_concordium.md
andynelson: "native" is not a reserved word in Concordium PLT, so although there is not currently a PLT with the symbol `native`, that could happen.

specs/schemes/exact/scheme_exact_concordium.md
andynelson: On Concordium, PLT Tokens are not contract defined and have unique registered symbols... there is no ambiguity.

The rest of the document refers to PLT and not CIS-2, which leads me to think that this scheme is intended only for CCD & PLT... a scheme that was intended to support both PLT & CIS-2 should probably differentiate between the two, with appropriate identifiers for each.

specs/schemes/exact/scheme_exact_concordium.md
andynelson: agreed - Decimals on PLT are defined at the token level, so may be more or less than six

specs/schemes/exact/scheme_exact_concordium.md
andynelson: Since Concordium gas fees for transfers of either CCD or PLTs are charged at a fixed fiat-pegged rate, even if `energy_amount` for a transaction was set too high, the actual maximum cost for the facilitator sponsoring a transfer wouldn't change.

specs/schemes/exact/scheme_exact_concordium.md
andynelson: Sponsor is the correct term for the chain - see https://docs.concordium.com/concordium-grpc-api/#concordium.v2.AccountTransactionHeaderV1

specs/schemes/exact/scheme_exact_concordium.md
andynelson: (and strictly speaking both `CCD` and *any* `PLT` are native, since PLTs are defined at the protocol level rather than being contract based tokens)

pavelhbl:
Update pushed to `BoostyLabs/x402-boosty:feature/concordium-schema-proposal` (commit `67adea5b`) addressing inline review comments on `scheme_exact_concordium.md`:

- Align `PaymentRequirements.extra` with other exact schemes: use `extra.feePayer` (documented as Concordium sponsor).
- Clarify asset + PLT token identification, and that PLT decimals depend on token.
- Remove redundant `payload.sender` from the examples (still documented as optional; `signedTransaction.header.sender` is source of truth).
- Tie `transaction.header.expiry` bound to `PaymentRequirements.maxTimeoutSeconds`.
- Add a recommendation for preflight checks / simulation to ensure on-chain success likelihood.

Core spec `asset` definition already updated to allow `asset=""` for native assets.

pavelhbl:
Opened a replacement PR with a clearer summary and the latest spec review fixes: https://github.com/x402-foundation/x402/pull/2389

I’ll be continuing this work going forward (original author is no longer available). Please leave any additional requested changes on the new PR and I’ll address them there.

phdargen:
Superseded by https://github.com/x402-foundation/x402/pull/2389

PR #1093, opened 2026-02-05
Cardano x402 Protocol Specification

fabianbormann:
## Description

This pull request adds a specification document for the `exact` payment scheme on Cardano within the x402 protocol. The document details protocol flows, schema definitions, and payload formats for various payment flavors (address-to-address, Masumi smart protocol, and script payments).

[Link to the rendered version of the cardano exact schema 👀 ](https://github.com/cardano-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_cardano.md)

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

fabianbormann:
Dear x402 team, please let us know if we can support you by reviewing the schema. Sometimes it's much faster to quickly jump on a call to discuss potential design issues, but I'm also happy to answer questions here in the PR if anything comes up during the review process. Thank you for your support! @CarsonRoscoe @phdargen @erikreppel-cb

fabianbormann:
Dear x402 team, I don't want to push too hard on this, but I’m looking to start assigning people to implement the individual packages, which I know depends on the schema being merged. Do you have a rough timeline for reviewing the schema pull requests? If there’s currently a hold on schemas and you don't see them being reviewed within the next three months, that would be very helpful to know for our internal planning. No rush at all. I’m just trying to get a better sense of the timeline so I can plan accordingly. Thanks! 🙏 

cc: @phdargen @erikreppel-cb @CarsonRoscoe

phdargen:
Thanks a lot for putting this together @fabianbormann and sorry for the delay, we currently have a lot of pending/incoming requests.

I will bring it up with the team to check if I can give you an estimate on the timeline, but I think we should definitely be able to review it within the next 3 months

fabianbormann:
@phdargen Thank you for the answer! We are prepared and looking forward to getting our hands dirty 😊

CarsonRoscoe (commented):
Hey @fabianbormann, thank you for the spec!

A few things to discuss,

#### 1. `maxAmountRequired` is the v1 field, the v2 field is `amount`
Every PaymentRequirements example in this spec uses `maxAmountRequired`.

#### 2. The "flavors" concept, let's discuss scope

I've been going back and forth on this, and here's where I landed:

**`default`** clearly belongs in the core spec, straightforward address-to-address exact payment, same as every other chain.

**`script`** I'd like to inquire more about before deciding. Does it still enforce `payTo`, or does it become a blackbox that can pay out in arbitrary ways? That answer determines whether it belongs here or warrants its own scheme.

**`masumi`** I'd suggest should be an x402 *extension* rather than part of the core Cardano spec. It appears to be a third-party smart contract protocol, and there's an important implication to baking third-party implementations into a core scheme spec: if a feature is in the core spec, the expectation is that **all** clients, servers and facilitators that claim to support the scheme must support it. This would mean every Cardano x402 facilitator is required to implement the Masumi protocol. Extensions, by contrast, are optional, facilitators declare which extensions they support via `/supported`, and clients/servers only use them when both sides opt in. That makes extensions the right home for third-party integrations. It's ultimately up to the foundation whether Masumi deserves to be baked into the Cardano spec, but I'd encourage thinking carefully about what that means for the ecosystem.

Also, is "flavor" a standard Cardano term, or something you coined for this spec? If it's not an established term, how would you feel about aligning with the EVM spec's `assetTransferMethod` naming?

#### 3. Please include a Facilitator Verification Rules section
Every other chain spec has a dedicated section of explicit, numbered verification rules that a facilitator MUST enforce. This section is entirely absent from the Cardano spec.

#### 4. Accepting mempool inclusion as a valid settlement outcome is dangerous
The spec says the server MAY grant access upon mempool inclusion (`status: "mempool"`) and returns HTTP 200. Cardano uses Ouroboros Praos, which has probabilistic finality, a transaction in the mempool or even in a recent block can be rolled back.

#### 5. Script payment security model is entirely unspecified
Similar to topic 2, the script flavor specifically introduces significant attack surface with zero security analysis. For example, if script code is provided inline (`extra.script.code`), how is it evaluated safely?

Please elaborate on script , what you can and can't do with it, etc.

#### 6. Nonce field in payload is EVM-specific
The `payload` object in all three flavor examples includes `"0xf3746613c2d920b5fdabc0856f2aeb2d4f88ee6037b8cc5d04a71a4462f13480"`, which appears to be copied from the EVM examples. Please review.

#### 7. `success` field type is wrong
The `PAYMENT-RESPONSE` schema shows `success` equaling the string `"true"`, rather than the boolean `true`.

#### 8. `payer` field is missing from `PAYMENT-RESPONSE`
The standard `SettleResponse` schema includes a `payer` field (the address that authorized the payment).

#### 9. `extra.network` inside Masumi flavor duplicates and potentially conflicts with top-level `network`
The Masumi flavor has `extra.network: "Mainnet | Preprod"` alongside the standard top-level `network: "cardano:mainnet"`. These serve the same purpose and could diverge. Remove `extra.network`, implementations should derive network from `PaymentRequirements.network`.

#### 10. `X-SESSION-TOKEN` is out of scope for this spec
The JWT session token section, including the full JWT structure, claims, signing algorithm, and session management semantics, has nothing to do with the x402 payment protocol. x402 is a payment authorization protocol, not a session management protocol.

Session tokens are an application-layer concern. If a resource server wants to issue sessions after payment, that is their implementation detail. Including a full JWT spec here conflates the payment scheme with server-side session design.

If session management is considered important enough to standardize, it belongs as a separate x402 extension spec (similar to `sign-in-with-x`), not embedded in a chain-specific payment spec.

Thank you again for the spec! Excited to see this come to life :)

fabianbormann:
Hey @CarsonRoscoe, thank you for the review and your suggestions!

I changed the spec according to your feedback and quickly want to go through your points one by one:

1. **Field Naming**: `maxAmountRequired` has been updated to `amount` across all examples to align with the v2 schema.

2. I changed "flavor" to "assetTransferMethod". It makes sense from our perspective to keep Masumi in the core spec, as it will be our main use-case for x402 and it is already using agent-to-agent payments. In fact, it was the reason why we started to look into x402, and every facilitator should be aware of and implementing its protocol. Thanks for highlighting the extensions!

3. Facilitator Verification Rules have been included. In Cardano, the concept of a sender is quite difficult as different sender addresses could sign and unlock UTXOs that are holding the required assets and pay together or even share tx fees. That is why, especially for Cardano, it is actually good that the `payer` is optional, if I saw this correctly in the `x402-specification-v2`. As the assets on Cardano are native assets that behave like Ada and are not smart contracts, it needs a signature from the owner wallet to unlock the specific UTXOs they are sitting on and transferring them as part of the transaction. Therefore, the transaction would fail automatically with a missing witness.

4. I added a strong warning regarding probabilistic finality. While we want to keep the `mempool` status as an option for low-value/low-risk services (like a simple weather API), the spec now explicitly discourages its use for resources with real economic value.

5. In Cardano, we actually have no **active** smart contracts. Our smart contracts are validators. This means a contract is a script with an address and, for different transaction types, a different validation logic could kick in. This can replace, for example, the spending behavior of a UTXO that belongs to the script (**not to the user's wallet**). The user and submitter need to build the transaction actively, providing the full context and local states, while the `extra.script.code` is actually the code that validates if the transaction can pass or not. It won't actively "call" other contracts or transfer anything; it is really just a validator that checks the given transaction. A use-case could be a smart wallet that does not require the user's signature for spending a UTXO, but rather a given logic being fulfilled. At the end of the day the transaction needs to be provided by the user and therefore should be in the interest of the user. I hope this addresses all concerns. :)

6. I have changed the nonce field to a UTXO that is part of the transaction and should be spent within it. This is ideal to prevent replay attacks, as a UTXO is unique and cannot be spent twice, and creates a new UTXO instead.

7. I have changed the success type to the boolean `true`.

8. I hope the payer field is optional in the PAYMENT-RESPONSE because, as I mentioned above, the concept is not straightforward in Cardano.

9. I removed the `extra.network` inside Masumi's `assetTransferMethod`.

10. I removed the section about the `X-SESSION-TOKEN` entirely.

I am happy to dive deeper into any of these points if anything remains unclear or if you see further room for alignment. Looking forward to your thoughts!

CarsonRoscoe (approved):
LGTM! Thank you @fabianbormann 🙏

PR #1177, opened 2026-02-13
Specification for Hyperliquid scheme

andrewg-nm:
## Description

Adds hypercore scheme to perform x402 payments directly on Hyperliquid (which is not EVM-compatible).

## Tests

I have added tests to the PR which run in CI, as well as locally produced examples of all combinations of {client, server, facilitator} x {Go, Typescript, Python} from the SDK in this PR and verified that a full end-to-end x402 request worked.

## Checklist

- [X] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [X] I added a changelog fragment for user-facing changes (docs-only changes can skip)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

phdargen:
Hi @andrewg-nm, thanks a lot for your contribution! 

Commenting exclusively on the specs in the following, I think it would be better to remove the implementation for now, see also the contributing guide here: https://github.com/coinbase/x402/blob/main/CONTRIBUTING.md#new-chains

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+
+### 6. Amount Validation
+
+- The `action.amount` MUST be a decimal string representation with exactly 8 decimals (e.g., `"0.01000000"` for 0.01 USD).
+- When converted to raw units (multiply by 10^8 and truncate decimals), the value MUST be greater than or equal to `PaymentRequirements.amount`.
+
+### 7. Asset/Token Validation
+
+- The `action.token` MUST equal the `PaymentRequirements.asset` exactly.
+- The token format MUST be `USDH:0x{identifier}`.
```
phdargen: What about non-USDH assets?
```suggestion
- The token format MUST be `{tokenName}:0x{identifier}`.
```

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+- The recovered address will be used for ledger queries after settlement.
+
+### 5. Destination Validation
+
+- The `action.destination` MUST equal the `PaymentRequirements.payTo` address exactly (case-insensitive).
+
+### 6. Amount Validation
+
+- The `action.amount` MUST be a decimal string representation with exactly 8 decimals (e.g., `"0.01000000"` for 0.01 USD).
+- When converted to raw units (multiply by 10^8 and truncate decimals), the value MUST be greater than or equal to `PaymentRequirements.amount`.
```
phdargen: Later in the spec you say: ". **Amount Precision**: 8-decimal string format prevents floating-point precision issues" while using floating-point multiplication here. We should require strict equality of the amounts

```suggestion
- The `action.amount` MUST be a decimal string representation with exactly 8 decimals (e.g., `"0.01000000"` for 0.01 USD). The conversion MUST be performed using string operations only (no floating-point arithmetic).
- The resulting action.amount MUST equal the expected string exactly.
```

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+    { name: "destinationDex", type: "string" },
+    { name: "token", type: "string" },
+    { name: "amount", type: "string" },
+    { name: "fromSubAccount", type: "string" },
+    { name: "nonce", type: "uint64" }
+  ]
+}
+```
+
+## Facilitator Verification Rules (MUST)
```
phdargen, line 184: I suppose the API doesn's support tx simulation? 

Verification should at the very least include a balance check of the client

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+| Error Code                    | Description                                               | HTTP Status |
+| ----------------------------- | --------------------------------------------------------- | ----------- |
+| `invalid_network`             | Network is not `hypercore:mainnet` or `hypercore:testnet` | 400         |
+| `invalid_action_type`         | Action type is not `sendAsset`                            | 400         |
+| `destination_mismatch`        | Destination doesn't match `payTo`                         | 400         |
+| `insufficient_amount`         | Amount is less than required                              | 400         |
+| `token_mismatch`              | Token doesn't match required asset                        | 400         |
+| `nonce_too_old`               | Nonce is more than 1 hour old                             | 400         |
+| `invalid_signature_structure` | Signature is missing r, s, or v                           | 400         |
+| `invalid_signature`           | Signature verification failed                             | 400         |
```
phdargen: nit: error codes should follow the convention in https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md, eg

```suggestion
| `invalid_exact_hypercore_payload_signature `           | Signature verification failed                             | 400         |
```

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
+# Scheme: `exact` on `Hyperliquid` (Hypercore L1)
+
+## Summary
+
+The `exact` scheme on Hyperliquid (Hypercore L1) uses EIP-712 signed `SendAsset` actions combined with API-based settlement. Unlike traditional blockchains that require on-chain transactions, Hyperliquid's approach allows for stateless facilitators with no gas fees, where the client's signature authorizes a direct transfer via the Hyperliquid API.
```
phdargen: Is the Hyperliquid API the only way to submit txs on hypercore? Are there any rate limits or other constrains?

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+2. **Amount Conversion**: Always use string arithmetic for 8-decimal conversions to avoid precision loss
+3. **Ledger Querying**: Wait 1.5 seconds after submission before querying ledger for transaction hash
+4. **Retry Logic**: Implement retry with exponential backoff for ledger queries (recommended: 2 retries with 1s delay)
+5. **Error Handling**: Distinguish between client errors (400) and settlement errors (500) for proper user feedback
+
+### Future Extensions
+
+Potential future enhancements to the Hyperliquid exact scheme:
+
+1. **Multi-Asset Support**: Extend beyond USDH to support other Hyperliquid spot tokens
```
phdargen: Why limit to USDH? It appears the spec would already support any non-native tokens

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+
+## Appendix
+
+### Key Differences from EVM/SVM
+
+| Feature                  | EVM                  | SVM                  | Hyperliquid       |
+| ------------------------ | -------------------- | -------------------- | ----------------- |
+| **Settlement**           | On-chain transaction | On-chain transaction | API submission    |
+| **Facilitator Wallet**   | Required + gas       | Required + SOL       | Not required      |
+| **Gas Fees**             | ETH/token            | SOL                  | None              |
```
phdargen:
```suggestion
| **Gas Fees**             | ETH           | SOL                  | None              |
```

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+     },
+     "signature": {
+       "r": "0x...",
+       "s": "0x...",
+       "v": 27
+     }
+   }
+   ```
+
+2. **API Response**: The Hyperliquid API returns:
```
phdargen, line 282: What does the api return in case of concurrent requests with the same payload? If it would return 'ok' for all requests, we will need a similar mitigation strategy as for SVM, see https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact_svm.md#duplicate-settlement-mitigation-recommended

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+
+| Feature                  | EVM                  | SVM                  | Hyperliquid       |
+| ------------------------ | -------------------- | -------------------- | ----------------- |
+| **Settlement**           | On-chain transaction | On-chain transaction | API submission    |
+| **Facilitator Wallet**   | Required + gas       | Required + SOL       | Not required      |
+| **Gas Fees**             | ETH/token            | SOL                  | None              |
+| **Signature Type**       | EIP-3009 / Permit2   | Ed25519 SPL          | EIP-712 SendAsset |
+| **Nonce Format**         | Sequential           | Recent blockhash     | Timestamp (ms)    |
+| **Confirmation**         | Block inclusion      | Block inclusion      | Ledger query      |
+| **Stateful Facilitator** | Yes (nonce tracking) | Yes (blockhash)      | No (stateless)    |
```
phdargen: The facilitator doesn't track payment nonces for EVM. Nonce uniqueness is enforced at the smart contract level. SVM uses a short-term cache to track deduplications but no persistent storage, see https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact_svm.md#duplicate-settlement-mitigation-recommended
```suggestion
| **Stateful Facilitator** | No | Deduplication cache   | TBD?   |
```

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+   }
+   ```
+
+   Or an error response if submission fails.
+
+3. **Transaction Hash Retrieval**: After 1.5 seconds (to allow ledger indexing), the facilitator queries the ledger:
+   - Endpoint: `POST /info` with `{"type": "userNonFundingLedgerUpdates", "user": "<payer_address>"}`
+   - Matches the ledger entry by comparing `nonce` and `destination`
+   - Extracts the transaction hash from the matched entry
+   - Retries up to 2 times with 1-second delays if not found
```
phdargen, line 301: What if not found, should return success: false?

specs/schemes/exact/scheme_exact_hyperliquid.md
```diff
@@ -0,0 +1,381 @@
...
+       "type": "default"
+     }
+   }
+   ```
+
+   Or an error response if submission fails.
+
+3. **Transaction Hash Retrieval**: After 1.5 seconds (to allow ledger indexing), the facilitator queries the ledger:
+   - Endpoint: `POST /info` with `{"type": "userNonFundingLedgerUpdates", "user": "<payer_address>"}`
+   - Matches the ledger entry by comparing `nonce` and `destination`
```
phdargen: Couldn't find the response type of `/info` with `userNonFundingLedgerUpdates` specified in the docs (https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/info-endpoint/perpetuals#id-200-ok-successful-response-first-perp-dex-1). Seems it returns time, hash and type specific delta? Are `nonce` and `destination` included in delta? Please double-check that they are actually returned here

phdargen:
Could you please also have a look at the alternative HL proposal from @janklimo here https://github.com/coinbase/x402/pull/1581?

I found the following differences:
- CAIP-2 namesspace: hypercore:* vs hyperliquid:*
- EIP-712 chainID: 421614 vs 999
- default asset: USDH vs USDC

specs/schemes/exact/scheme_exact_hyperliquid.md
andrewg-nm: Yeah, this was wrong, any token should just work already

specs/schemes/exact/scheme_exact_hyperliquid.md
andrewg-nm: Main advantage is that the facilitator is gasless but needing gas doesn't exactly make a facilitator "stateful", so I'll pare down the mentions of statelessness generally.

specs/schemes/exact/scheme_exact_hyperliquid.md
andrewg-nm: I checked and the nonce/destination are included with the the API response for `send` actions (which are applicable here), but not all types. I'll clarify this in the spec.

Example:
```
curl -s -X POST https://api.hyperliquid.xyz/info \
  -H "Content-Type: application/json" \
  -d '{"type": "userNonFundingLedgerUpdates", "user": "0x8bfcf9e2764bc84de4bbd0a0f5aaf19f47027a73", "startTime": 0}' | python3 -m json.tool
```
One output from the list:
```
    {
        "time": 1773668257934,
        "hash": "0x9d457a8dd7c6cb869ebf04372b2188020b0c007372c9ea58410e25e096caa571",
        "delta": {
            "type": "send",
            "user": "0x8bfcf9e2764bc84de4bbd0a0f5aaf19f47027a73",
            "destination": "0x8bfcf9e2764bc84de4bbd0a0f5aaf19f47027a73",
            "sourceDex": "",
            "destinationDex": "spot",
            "token": "USDC",
            "amount": "1.0",
            "usdcValue": "1.0",
            "fee": "0.0",
            "nativeTokenFee": "0.0",
            "nonce": 1773668251489,
            "feeToken": ""
        }
    }
```

specs/schemes/exact/scheme_exact_hyperliquid.md
andrewg-nm: Hyperliquid's API handles this, it rejects on the nonce, e.g.
`{'status': 'err', 'response': 'Invalid nonce: duplicate nonce 1773765118832'}`

specs/schemes/exact/scheme_exact_hyperliquid.md
andrewg-nm: There may be third party APIs for submitting transactions to Hypercore, though I think the official Hyperliquid API is dominant. Rate limits are described at https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/api/rate-limits-and-user-limits (will link this in the spec)

andrewg-nm:
> Could you please also have a look at the alternative HL proposal from @janklimo here #1581?
> 
> I found the following differences:
> 
> * CAIP-2 namesspace: hypercore:* vs hyperliquid:*
> * EIP-712 chainID: 421614 vs 999
> * default asset: USDH vs USDC

CAIP naming convention: I don't believe there is a canonical answer for this as Hypercore is not an EVM chain, but after asking the HL team they suggested `hyperliquid` over `hypercore` for the chain name. I've swapped it out in the spec.

Chain ID: After more careful examination it looks like we do not need a `chainID` for `sendAsset` actions, somehow ended up including it in my implementations but we should just remove it to avoid confusion. AFAIK there is no canonical chain ID for hypercore but 999 for HyperEVM is often used.

specs/schemes/exact/scheme_exact_hyperliquid.md
andrewg-nm: Yeah, there is no transaction simulation. I added a recommendation to query spot balances (just a small tradeoff of making an extra API call)

janklimo:
@andrewg-nm thank you for your work on this! 

@phdargen thank you for your wonderfully thorough review so far 🙏

Based on my research in #1581 here are some observations.

### `chainId`
I strongly suggest we include `chainId` in the EIP-712 domain. I'm not 100% sure frontend implementations need it but I remember when I hardcoded 421614, signing in browser wallets would fail with a chain mismatch. Perhaps omitting it causes the same problem. Feel free to borrow `getEIP712Domain` logic from my PR.

### Amount representation inconsistency

The spec uses amounts + decimals representation which is something commonly used on the EVM. On HL L1, amounts are passed as formatted strings. Recommendation: `"0.01"` instead of `1000000` in payment requirements to keep this consistent between the implementation and HL API. Dual representation would be a cause of bugs.

You also mention

> MUST be a decimal string representation with exactly 8 decimals

it's at most 8 decimals. so "0.1" is also valid. It's good to have logic enforcing this 👍

### Balance Verification (SHOULD)

I believe this is a MUST. I only dug deeper into how Next.js implementation works and it returns the resource after `/verify` succeeds. So you could fetch any resource with a completely empty wallet.

### Rigid Fixed Fields

  The spec requires:
  sourceDex MUST be "spot"
  destinationDex MUST be "spot"
  fromSubAccount MUST be ""

I discussed my proposal briefly with the team and they suggested using `sendAsset` as a transaction type so that users can transfer out of their spot or perps balance. I recommend we let the server specify destination dex in the extra field (please see my PR if you agree). The client should decide source dex.

### USDH vs USDC as Default Token

I understand USDH is the premier aligned quote asset but given that: a) there will be others and b) USDC is used for every other chain implementation in this repo, I suggest we also use USDC, meaning when the server requires "$0.1" it is interpreted as 0.1 USDC.

### Settlement timing baked into the spec

I am still split on this one... it feels brittle. Fills can be returned in the API with arbitrary delay and I'm concerned this makes settlement slower and less reliable (failures despite payments succeeding). I let @phdargen weigh in here since I don't have the full picture. My approach assumed sufficient balance in spot/perps and valid signature guarantee that the payment succeeds.

andrewg-nm:
> ### Balance Verification (SHOULD)
> I believe this is a MUST. I only dug deeper into how Next.js implementation works and it returns the resource after `/verify` succeeds. So you could fetch any resource with a completely empty wallet.

Are you sure? I believe that the content is returned only when `/settle` succeeds, e.g. based on https://github.com/coinbase/x402/blob/6a90fe93c57fe20104f175149747ed399d491192/typescript/packages/http/next/src/utils.ts#L182-L189

> ### Settlement timing baked into the spec
> I am still split on this one... it feels brittle. Fills can be returned in the API with arbitrary delay and I'm concerned this makes settlement slower and less reliable (failures despite payments succeeding). I let @phdargen weigh in here since I don't have the full picture. My approach assumed sufficient balance in spot/perps and valid signature guarantee that the payment succeeds.

Yeah, I agree there is a tradeoff here. It would be ideal if the transaction information (e.g. hash) were available from the API call where it is submitted but that's just not how the Hyperliquid API works right now. I thought it would be useful to have the transaction information recorded when it does go through but polling the ledger isn't perfectly reliable.

andrewg-nm:
> ### `chainId`
> I strongly suggest we include `chainId` in the EIP-712 domain. I'm not 100% sure frontend implementations need it but I remember when I hardcoded 421614, signing in browser wallets would fail with a chain mismatch. Perhaps omitting it causes the same problem. Feel free to borrow `getEIP712Domain` logic from my PR.

This part of the Hyperliquid API is a bit confusing. It seems like you do need a `chainId` to submit to the Hyperliquid API, but the value is kind of meaningless. I tested out my implementation a bit with a minimal browser experience, and I had no issues making an x402 payment while signing from any EVM chain. So I don't think we need to set a particular value for it.

phdargen:
> > ### Balance Verification (SHOULD)
> > I believe this is a MUST. I only dug deeper into how Next.js implementation works and it returns the resource after `/verify` succeeds. So you could fetch any resource with a completely empty wallet.
> 
> Are you sure? I believe that the content is returned only when `/settle` succeeds, e.g. based on
> 
> https://github.com/coinbase/x402/blob/6a90fe93c57fe20104f175149747ed399d491192/typescript/packages/http/next/src/utils.ts#L182-L189

The flow is: verify -> resource handler executes -> settle -> client receives resource + PAYMENT-RESPONSE

Balance verification should be required to protect the server from doing unnecessary work for invalid requests

> > ### Settlement timing baked into the spec
> > I am still split on this one... it feels brittle. Fills can be returned in the API with arbitrary delay and I'm concerned this makes settlement slower and less reliable (failures despite payments succeeding). I let @phdargen weigh in here since I don't have the full picture. My approach assumed sufficient balance in spot/perps and valid signature guarantee that the payment succeeds.
> 
> Yeah, I agree there is a tradeoff here. It would be ideal if the transaction information (e.g. hash) were available from the API call where it is submitted but that's just not how the Hyperliquid API works right now. I thought it would be useful to have the transaction information recorded when it does go through but polling the ledger isn't perfectly reliable.

Hm just an API call returning 'OK' seems like a fundamentally weaker settlement guarantee than the onchain confirmations we have for all the other networks. Do you (or the HL team) have any numbers how long it typically takes for the tx to appear in the info API? And what would be an extreme delay? Whats the failure rate of fills for which the API returned OK?

phdargen:
> > ### `chainId`
> > I strongly suggest we include `chainId` in the EIP-712 domain. I'm not 100% sure frontend implementations need it but I remember when I hardcoded 421614, signing in browser wallets would fail with a chain mismatch. Perhaps omitting it causes the same problem. Feel free to borrow `getEIP712Domain` logic from my PR.
> 
> This part of the Hyperliquid API is a bit confusing. It seems like you do need a `chainId` to submit to the Hyperliquid API, but the value is kind of meaningless. I tested out my implementation a bit with a minimal browser experience, and I had no issues making an x402 payment while signing from any EVM chain. So I don't think we need to set a particular value for it.

Could you please check this with the HL team?

hl-core-dev:
> > > ### `chainId`
> > > I strongly suggest we include `chainId` in the EIP-712 domain. I'm not 100% sure frontend implementations need it but I remember when I hardcoded 421614, signing in browser wallets would fail with a chain mismatch. Perhaps omitting it causes the same problem. Feel free to borrow `getEIP712Domain` logic from my PR.
> > 
> > 
> > This part of the Hyperliquid API is a bit confusing. It seems like you do need a `chainId` to submit to the Hyperliquid API, but the value is kind of meaningless. I tested out my implementation a bit with a minimal browser experience, and I had no issues making an x402 payment while signing from any EVM chain. So I don't think we need to set a particular value for it.
> 
> Could you please check this with the HL team?

The "chainId" set in the signature can be arbitrary, because HyperCore is not an EVM chain and the transactions are not valid EVM transactions. They are a specific EIP-712 payload that cannot be replayed on other chains, because they include `"hyperliquidChain": "Mainnet" | "Testnet"` in the payload. The EVM chain ID is redundant and therefore can be arbitrary.

I'm not too familiar with x402, but perhaps it's worth treating HyperCore as a standalone chain instead of assigning it an EVM chain ID.

phdargen:
> > > > ### `chainId`
> > > > I strongly suggest we include `chainId` in the EIP-712 domain. I'm not 100% sure frontend implementations need it but I remember when I hardcoded 421614, signing in browser wallets would fail with a chain mismatch. Perhaps omitting it causes the same problem. Feel free to borrow `getEIP712Domain` logic from my PR.
> > > 
> > > 
> > > This part of the Hyperliquid API is a bit confusing. It seems like you do need a `chainId` to submit to the Hyperliquid API, but the value is kind of meaningless. I tested out my implementation a bit with a minimal browser experience, and I had no issues making an x402 payment while signing from any EVM chain. So I don't think we need to set a particular value for it.
> > 
> > 
> > Could you please check this with the HL team?
> 
> The "chainId" set in the signature can be arbitrary, because HyperCore is not an EVM chain and the transactions are not valid EVM transactions. They are a specific EIP-712 payload that cannot be replayed on other chains, because they include `"hyperliquidChain": "Mainnet" | "Testnet"` in the payload. The EVM chain ID is redundant and therefore can be arbitrary.
> 
> I'm not too familiar with x402, but perhaps it's worth treating HyperCore as a standalone chain instead of assigning it an EVM chain ID.

Thanks for the clarification @hl-core-dev!

x402 uses CAIP-2 namespaces, no need to assign a EVM chain ID

@andrewg-nm I suggest to use whatever chainID you prefer in the example payload etc shown in the spec and add the statement of @hl-core-dev somewhere

phdargen:
> > > ### Settlement timing baked into the spec
> > > I am still split on this one... it feels brittle. Fills can be returned in the API with arbitrary delay and I'm concerned this makes settlement slower and less reliable (failures despite payments succeeding). I let @phdargen weigh in here since I don't have the full picture. My approach assumed sufficient balance in spot/perps and valid signature guarantee that the payment succeeds.
> > 
> > 
> > Yeah, I agree there is a tradeoff here. It would be ideal if the transaction information (e.g. hash) were available from the API call where it is submitted but that's just not how the Hyperliquid API works right now. I thought it would be useful to have the transaction information recorded when it does go through but polling the ledger isn't perfectly reliable.
> 
> Hm just an API call returning 'OK' seems like a fundamentally weaker settlement guarantee than the onchain confirmations we have for all the other networks. Do you (or the HL team) have any numbers how long it typically takes for the tx to appear in the info API? And what would be an extreme delay? Whats the failure rate of fills for which the API returned OK?

Hi @hl-core-dev, could you please also comment on this issue? Is there any reliable and fast way to get the transaction hash?

phdargen:
> > > > ### Settlement timing baked into the spec
> > > > I am still split on this one... it feels brittle. Fills can be returned in the API with arbitrary delay and I'm concerned this makes settlement slower and less reliable (failures despite payments succeeding). I let @phdargen weigh in here since I don't have the full picture. My approach assumed sufficient balance in spot/perps and valid signature guarantee that the payment succeeds.
> > > 
> > > 
> > > Yeah, I agree there is a tradeoff here. It would be ideal if the transaction information (e.g. hash) were available from the API call where it is submitted but that's just not how the Hyperliquid API works right now. I thought it would be useful to have the transaction information recorded when it does go through but polling the ledger isn't perfectly reliable.
> > 
> > 
> > Hm just an API call returning 'OK' seems like a fundamentally weaker settlement guarantee than the onchain confirmations we have for all the other networks. Do you (or the HL team) have any numbers how long it typically takes for the tx to appear in the info API? And what would be an extreme delay? Whats the failure rate of fills for which the API returned OK?
> 
> Hi @hl-core-dev, could you please also comment on this issue? Is there any reliable and fast way to get the transaction hash?

also pinging @andrewg-nm

phdargen:
Any updates @andrewg-nm @hl-core-dev? Would be nice to get this in soon

PR #1311, opened 2026-02-23
feat: add spec exact scheme on btc Lightning Network

ocknamo:
### Overview

Adds `specs/schemes/exact/scheme_exact_lightning.md`, defining the `exact` payment scheme for the Bitcoin Lightning Network using BOLT11 invoices.

#### No facilitator required.

Lightning payments are routed and settled off-chain instantly through the Lightning Network. Unlike EVM/SVM schemes that require on-chain broadcasting, the resource server verifies payment directly via its own Lightning node or wallet API — no intermediary facilitator is needed.

#### The BOLT11 invoice string is  used as the `transaction` identifier.

Lightning Network has no on-chain transaction hash equivalent. The BOLT11 invoice contains the payment hash, which uniquely identifies the payment to a Lightning node. A note has been added to the spec clarifying that if integrators need the payment hash for use with Lightning tooling, it can be extracted by decoding the invoice string.

#### Can payment be verified if it happens out-of-band (outside the x402 protocol)?

Yes. The resource server queries its own Lightning node or wallet API for the settlement status of the invoice (Verification step 11). Because the server generates the invoice itself, it can detect unpaid or unrecognized invoices.

#### What unit is `amount` in?

Millisatoshis (msats). BOLT11 invoices also express amounts in millisatoshis, so no unit conversion is needed during verification (Verification step 8).

#### Compared to L402

Compared to[ L402](https://github.com/lightninglabs/L402/tree/master), this specification simplifies the process by omitting the return of the preimage, enabling complete separation between the client and the wallet. It also simplifies client implementation, allowing for operations such as settling payments on a different device using a payment QR code displayed on the client.

However, this comes with a security trade-off for convenience.
Since it simply treats possession of the paid invoice (payment hash) as proof, implementation-level countermeasures are required against replay attacks.
And payment verification necessitates a server-side wallet query.

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

clankwright:
This is a strong spec and the right direction for x402. I run [satring](https://github.com/toadlyBroodle/satring), an L402 service directory, and the ability for Lightning services to be accessible through both protocols on the same payment rail is something I'm interested in actively building against. Having Lightning as a first-class x402 scheme lowers the barrier for the existing Lightning API ecosystem to adopt x402 alongside L402.

The design decision to omit the preimage return is well-reasoned. Decoupling the client from the wallet is a real usability win, especially for browser-based and mobile clients that can scan a QR to pay from a separate device. L402's preimage requirement ties payment and authentication to the same client, which is more restrictive.

A couple of minor notes that could be addressed in follow-ups (not blockers for this PR):

- **`extra.paymentHash` in SettlementResponse**: Most Lightning tooling (LND, CLN, LDK) indexes by payment hash rather than invoice string. Including it as an explicit field would be a convenience for implementors doing reconciliation or analytics, saving them from decoding the invoice. Small addition, easy follow-up.

- **In-flight payment handling**: The settlement section could note that servers should expect occasional in-flight states (HTLC locked but not yet settled) at verification step 11, and recommend a consistent response (e.g. `Retry-After`). This is a documentation detail, not an architectural issue.

Very interested in building a reference implementation or test suite against this spec once it lands.

ocknamo:
> A couple of minor notes that could be addressed in follow-ups (not blockers for this PR):
> 
> * **`extra.paymentHash` in SettlementResponse**: Most Lightning tooling (LND, CLN, LDK) indexes by payment hash rather than invoice string. Including it as an explicit field would be a convenience for implementors doing reconciliation or analytics, saving them from decoding the invoice. Small addition, easy follow-up.
> * **In-flight payment handling**: The settlement section could note that servers should expect occasional in-flight states (HTLC locked but not yet settled) at verification step 11, and recommend a consistent response (e.g. `Retry-After`). This is a documentation detail, not an architectural issue.
> 
> Very interested in building a reference implementation or test suite against this spec once it lands.

@toadlyBroodle 
Thank you for your thoughtful feedback.
Just to clarify your suggestion regarding extra.paymentHash:
Are you proposing that we keep the invoice as the canonical field and include paymentHash as an additional field (either optional or required)?
From your perspective, would you recommend making paymentHash mandatory in the response, or keeping it optional for implementors?

Regarding the in-flight payment handling note, that makes sense. I’ll consider adding clarification in the settlement section.

clankwright:
Yes, keep the invoice as canonical. The paymentHash would just be an additional convenience field in the `extra` object of the SettlementResponse, alongside `extra.invoice` and `extra.settledAt`.

I'd suggest optional. It's a derived value, so making it required adds a validation burden without much benefit. Optional lets implementors include it when their stack already surfaces it (most Lightning wallet APIs return the payment hash alongside invoice status), and clients that need it for indexing or reconciliation can use it without pulling in a BOLT11 decoder.

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,195 @@
...
+
+1. **Extract requirements**: Use `payload.accepted` to get the payment requirements being fulfilled.
+2. Verify `x402Version` is `2`.
+3. Verify the network matches the agreed upon chain.
+4. Verify `payload.invoice` matches `payload.accepted.extra.invoice` exactly (prevents invoice substitution attacks).
+5. Verify the invoice in `payload.invoice` was issued by this resource server (i.e., exists in the server's invoice database).
+6. Decode the BOLT11 invoice to extract payment details (payment hash, amount, timestamp, expiry).
+7. Verify the invoice has not expired (check current time against invoice timestamp + expiry).
+8. Verify the invoice amount matches `requirements.amount` exactly.
+9. Verify `payTo`:
```
reneaaron: The node `pubkey` is always derivable from the invoice, so there's no real case where it's unknowable. Since the `pubkey` is encoded in the invoice, how would it be possible this `pubkey` is different from the one in the invoice? What attack vectors does checking this field against the `pubkey` from the invoice prevent?

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,195 @@
...
+## SettlementResponse
+The `PAYMENT-RESPONSE` header is base64 encoded and returned to the client from the resource server.
+
+Once decoded, the `PAYMENT-RESPONSE` is a JSON string following the standard `SettlementResponse` schema:
+
+```json
+{
+  "success": true,
+  "transaction": "lnbc1u1p3...",
+  "network": "bip122:000000000019d6689c085ae165831e93",
```
reneaaron, line 139: Every other x402 scheme I've seen routes purely on `(scheme, network)` tuple. Clients and middleware can decide whether to handle a payment requirement by looking at those two fields alone. This spec requires also inspecting `extra.paymentMethod`, which is a layer deeper and easy to miss.

I am wondering whether `lightning:000000000019d6689c085ae165831e93` would be a better network identifier? Analogous to how Solana got `solana:` rather than being stuffed into a Bitcoin or EVM namespace. That would let `(scheme, network)` remain the canonical routing key and makes Lightning vs. on-chain Bitcoin a clean branch at the network level rather than inside the `extra` field.

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,195 @@
...
+}
+```
+
+### Field Descriptions
+
+- `scheme`: Always `"exact"` for this scheme
+- `network`: CAIP-2 network identifier - `bip122:000000000019d6689c085ae165831e93` (Bitcoin mainnet) or `bip122:000000000933ea01ad0ee984209779ba` (Bitcoin testnet)
+- `amount`: The exact amount to transfer in **millisatoshis** (1 satoshi = 1000 millisatoshis)
+- `asset`: Always `"BTC"` for Bitcoin Lightning Network
+- `payTo`: Required for schema consistency with other x402 schemes. In Lightning, the actual payment destination is encoded in the invoice itself, so this field is informational. This can be either:
```
reneaaron: The receiving node `pubkey` is encoded in the invoice and should always be available. When exactly would it be set to `anonymous`?

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,195 @@
...
+    "settledAt": 1739116800,
+    "paymentHash": "a]1b2c3d4e5f..."
+  }
+}
+```
+
+### Field Descriptions
+
+- `success`: Boolean indicating whether the payment settlement was successful
+- `transaction`: The BOLT11 invoice string (serves as the transaction identifier). Note: Unlike blockchain transaction hashes, this is the full invoice string. If integration with Lightning tooling requires a payment hash, it can be extracted by decoding this BOLT11 invoice, or use the optional `extra.paymentHash` field if provided.
```
reneaaron: Isn't the `paymentHash` the semantically more correct value to use here? What's the reason you directly use the BOLT11 invoice here?

reneaaron (commented):
Thanks for taking the initiative to get a Lightning spec for x402 fleshed out! 🙌

We recently shipped a facilitator for NWC-based Lightning payments at [x402.albylabs.com](https://x402.albylabs.com/) and noticed some differences with this draft, so having a formalized spec to align around is very much needed!

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: I completely understand what you're saying, and I actually considered that kind of intuitive design myself at first. However, it's simply because v2 is specified to use the CAIP-2 format.

https://github.com/coinbase/x402/blob/main/specs/x402-specification-v2.md

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: "payTo" This is included solely to maintain consistency with the x402 specification and is an unnecessary value. This is because Lightning has no equivalent to an on-chain address and is always anonymous.

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: That's correct.
As I mentioned in another comment, since the payTo pubKey is not required, this validation is likely unnecessary.
To avoid confusion, it might be better to use a fixed value such as "anonymous".

specs/schemes/exact/scheme_exact_lightning.md
reneaaron: You are probably right, makes sense :+1:

specs/schemes/exact/scheme_exact_lightning.md
reneaaron: Okay, so how do we deal with it? Always set it to `anonymous` and remove any references to the node pubkey?

specs/schemes/exact/scheme_exact_lightning.md
reneaaron: Fine for me :+1: Let's not forget to remove the verification step from this list here as well.

reneaaron:
@ocknamo Thanks for your feedback! :raised_hands: 

Will you find the time to update the spec anytime soon? Otherwise I am happy to pick it up from here, just let me know.

Let's try to get this to a merge-able state and I am happy to adjust our current implementation according to our discussions here :raised_hands:

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: Originally, I used the invoice string here to avoid the need to decode it to extract the payment hash. However, since the server likely already has the payment hash, it makes more sense to use paymentHash as the transaction identifier.
This also makes the extra.paymentHash field redundant, so we'll remove it.
Will update.

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: Yap. I'll always set it to "anonymous" and remove the node pubkey references. Let me update the spec.

ocknamo:
@reneaaron  Thank you. I have some time now, so I'll update the spec. Commits welcome too!

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: Updated here. https://github.com/coinbase/x402/pull/1311/commits/89404c40feeece5ecf3abd156ca3f418c98ea83a

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: Updated here. 　https://github.com/coinbase/x402/pull/1311/commits/7f8586503029b217782fa04c5359de0e8f10349a

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: Updated here. https://github.com/coinbase/x402/pull/1311/commits/7f8586503029b217782fa04c5359de0e8f10349a

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,185 @@
...
+    "invoice": "lnbc1u1p3..."
+  }
+}
+```
+
+### Field Descriptions
+
+- `scheme`: Always `"exact"` for this scheme
+- `network`: CAIP-2 network identifier - `bip122:000000000019d6689c085ae165831e93` (Bitcoin mainnet) or `bip122:000000000933ea01ad0ee984209779ba` (Bitcoin testnet)
+- `amount`: The exact amount to transfer in **millisatoshis** (1 satoshi = 1000 millisatoshis)
```
bumi, line 55: I wish this would be sats.
there is so much confusion always around sats vs. millisats.
many tools also do not understand and support millisats.  
(and those also don't really exit)

ocknamo: I chose millisats because the design is based on the BOLT11 spec.
That said, I agree the unit isn’t very intuitive and could impact UX, so I’d also like to hear what others think.

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,185 @@
...
+8. Verify the invoice amount matches `requirements.amount` exactly.
+9. Verify the invoice has not already been used.
+10. Query the Lightning wallet API or Lightning node to verify the invoice has been paid. If the payment is still in-flight (HTLC locked but not yet settled), the server SHOULD respond with `402` and a `Retry-After` header indicating when the client should retry.
+
+## Settlement
+
+Settlement for Lightning Network payments is fundamentally different from blockchain schemes:
+
+1. **Payment is already settled**: Lightning payments are settled instantly off-chain when the client pays the invoice (before submitting the `PAYMENT-SIGNATURE` header).
+2. **Server verification only**: The resource server only needs to verify that the payment was received by querying their Lightning wallet API or node.
```
Bortlesboat, line 124: small consistency note: step 10 in Verification already accounts for in-flight states (locked but not yet settled), which seems right. the Settlement text saying "no risk of pending states" reads stronger than the rest of the spec. maybe soften that to near-instant finality under normal conditions.

specs/schemes/exact/scheme_exact_lightning.md
```diff
@@ -0,0 +1,185 @@
...
+- Gas fee sponsorship (Lightning fees are paid from payment amount)
+- Transaction broadcasting (handled by Lightning Network routing)
+- Replay protection infrastructure (invoices are inherently one-time use)
+
+This architectural simplification makes Lightning integration more straightforward for resource servers.
+
+## Recommendation
+
+- Use the spec defined above for the first version of the protocol
+- Implement replay protection using a key-value store or database
```
Bortlesboat, line 185: not for this PR, but is BOLT12 offers something we'd want to mention as a possible future extension? for high-traffic APIs, reusable payment endpoints could reduce per-request invoice generation overhead. even a brief note on forward compatibility might be enough.

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: That's right. 
updated it: https://github.com/coinbase/x402/pull/1311/commits/33cb714e07b425ee59cbd8da66c64bf90aa49aeb

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: Good point. I'd like to explore BOLT12 support in the future.
The current extra field is designed to be extensible, so BOLT12 can be accommodated by adding new fields (e.g., an offer field alongside the existing invoice field) without breaking the current schema.

junderw (approved):
LGTM 33cb714e07b425ee59cbd8da66c64bf90aa49aeb

cb-heimdall:
Review Error for junderw @ 2026-03-24 09:53:54 UTC
User failed mfa authentication, see [go/mfa-help](http://go/mfa-help)

reneaaron:
@junderw Something seemed to go wrong with your heimdall service: 

> Review Error for junderw @ 2026-03-24 09:53:54 UTC
User failed mfa authentication, see [go/mfa-help](http://go/mfa-help)

ocknamo:
This is my first working DEMO.

https://simpl402.ocknamo.com/

https://github.com/ocknamo/simpl402/tree/main/nostr-api

ocknamo:
I closed a few threads that appear to have been resolved.

specs/schemes/exact/scheme_exact_lightning.md
ocknamo: I realized the lack of support probably isn’t a real issue for this fixed-price use case.

Since BOLT11 is based on millisatoshis, I think keeping millisats is still the most reasonable choice.

ocknamo:
Dear x402 team, I know you have a lot going on, but please let me know if there’s anything I can do to help with this schema review and move it forward. :+1: 
cc: @phdargen @CarsonRoscoe

zekebuilds-lab:
Shipped a live implementation of the Lightning verification extension this PR proposes.

**Endpoint**: https://captcha-mcp.powforge.dev/mcp — MCP server that issues PoW CAPTCHA challenges with optional Lightning payment skip via an L402-style flow. The `verify` tool exercises the verifier path this PR describes.

**Package**: https://www.npmjs.com/package/@powforge/captcha-mcp (v0.2.0, stdio + HTTP Streamable transports)

One gap I noticed while implementing: the current draft doesn't include test vectors for the **invoice substitution attack** — where an agent swaps a legitimate invoice for one pointing to an attacker-controlled node before the payer signs. Without spec-level test vectors, two compliant implementations can both pass conformance and still be exploitable against each other.

Happy to put together a test vectors appendix (invoice substitution + replay/nonce + amount-tampering) if it would help land this PR. Will post a draft as a separate gist regardless so you can pull from it.

ocknamo:
@zekebuilds-lab I would really appreciate it if you could share your "gist".
I want to decide whether it should be incorporated into the specifications.

zekebuilds-lab:
Here's the test vectors draft: https://gist.github.com/zekebuilds-lab/0ca1bf8b4c24110402ed008fbbcaf19d

Three scenarios: invoice substitution (wrong destination node), replay (payment_hash already used), and amount tampering (invoice msat < required price). Each has a valid base case and an attack case with expected rejection reason.

Happy to revise based on what fits the spec format — especially around how the server-nonce binding for invoice substitution should be expressed (whether that's a challenge field in the 402 response, a session ID, or something else).

ocknamo:
@zekebuilds-lab 
Thanks for the review. That said, the receipt shape you've described doesn't match the contents of this PR. The payload here contains only invoice — no preimage — the network ID is CAIP-2 (bip122:...), and the design assumes no facilitator (see "No Facilitator Required"). A receipt with network: "bitcoin-lightning" and a preimage looks like a different design. Were you perhaps looking at an older revision or a different draft?

To clarify the verification flow: the server issues the invoice itself and confirms payment via step 5 (does it exist in the server's own invoice DB) and step 10 (querying its own node for receipt). Because these two checks do the work, Vector 1 (substitution) and Vector 3 (amount) are rejected without relying on preimage verification, and Vector 2 (replay) is already handled by step 9 plus the "Replay Attack Prevention" appendix. The substitution/amount concerns appear to assume a preimage-receipt model. Beyond that, Vector 1 presupposes tampering with the 402 response — but if TLS is intact, that tampering can't occur in the first place, so the impact is limited (it only becomes relevant if you separately assume a broken TLS chain, a malicious browser extension, or similar).

zekebuilds-lab:
Thanks for the clarification on the design — that changes the test vector shape considerably.

A few follow-up questions on the server-side verification model:

1. **How does the server confirm payment?** When the client pays the invoice, does the spec assume the server is polling the Lightning node's settled invoices, waiting for a webhook callback, or using a HODL invoice that the server can check atomically?

2. **Statelessness tradeoff**: The no-preimage design means the server needs to track invoice state (pending → paid) rather than deriving payment proof from the preimage alone. For `@powforge/x402-lightning`, I implemented preimage-based validation specifically to stay stateless — but I can adapt the reference implementation to match the spec's intent. Which verification method does the spec intend as canonical?

3. **TLS point**: Agreed that TLS addresses the MITM case on 402 responses. The preimage approach was defense-in-depth for environments with TLS termination at a CDN or reverse proxy where the payment response might traverse an internal hop. Worth a note in the spec or just rely on TLS?

Happy to update the test vectors once the verification flow is confirmed. The invoice substitution, replay, and amount tampering vectors I submitted apply to whichever server-side verification mechanism the spec chooses — just the proof mechanism differs.

zekebuilds-lab:
Thanks for the correction — I was reviewing against a preimage-receipt model, which clearly doesn't apply here. The server-issues + server-confirms flow (steps 5 and 10) is now clear.

One follow-up: at step 10, when the server queries its own node for the settled receipt, does the node return a preimage? If so, is exposing an optional `payment_preimage` field in the receipt response intentionally out of scope for this PR, or something worth documenting for integrators who want client-verifiable proof for auditing/dispute purposes? No concern if it's a deliberate design boundary — just want to understand whether that path is considered closed or deferred.

ocknamo:
@zekebuilds-lab 
Exposing a payment_preimage field is intentionally out of scope for this PR, as [noted at the top](https://github.com/x402-foundation/x402/pull/1311#issue-3979379002). It's a reasonable use case, but it's a deliberate trade-off to keep this distinct from L402, and there's no plan to follow up on it.

benthecarman:
@ocknamo is there a reason you aren't using the preimage for proof of payment and requiring them to look up the invoice? It's a lot better when you can make it stateless by just verifying sha256(preimage) == preimage without any extra requests

ocknamo:
@benthecarman Thank you for your comment.
The main reason is usability. Requiring the payer to obtain and send the preimage is inconvenient in many real-world scenarios.
For example:
- A user scans a QR code displayed on a website using a separate mobile wallet and expects to gain access to the resource.
- An AI agent makes a payment through a wallet API.
In the first case, the browser client typically has no access to the preimage because the payment is made on another device.
In the second case, exposing the preimage through a wallet API raises security concerns, and many wallet APIs do not expose it at all (unless they are tightly integrated with a Lightning node).

benthecarman:
> A user scans a QR code displayed on a website using a separate mobile wallet and expects to gain access to the resource.

this seems out of scope, x402 is for agents

> An AI agent makes a payment through a wallet API.
In the first case, the browser client typically has no access to the preimage because the payment is made on another device.
In the second case, exposing the preimage through a wallet API raises security concerns, and many wallet APIs do not expose it at all (unless they are tightly integrated with a Lightning node).

Basically every wallet exposes this nowdays and its not a security concern whatsover

ocknamo:
@benthecarman 
My understanding is that x402 is not limited to AI agents. As the README states, "x402 is an open standard for internet native payments."

Also, could you share which wallet APIs you're referring to? My understanding is that support for exposing the preimage varies across wallets.

benthecarman:
@phdargen  can you clarify here? Is supposed to be x402 agent native or are human users in scope?

ocknamo:
Closing in favor of #2861, which covers "exact" on Lightning.

The stateless design is clearly better in some respects — the facilitator verifies without receiver access — but it's a tradeoff: payer adapters that can't return the preimage are excluded.

That said, I broadly agree with the overall direction.

Thanks @benthecarman @phdargen .

PR #1266, opened 2026-02-20
feat: Aptos fast follow updates

CarsonRoscoe:
## Description

The new `@x402/aptos` PR followed the `registerAptosExactScheme` helper function pattern outlined in `@x402/evm` and `@x402/svm`

However, that pattern is not the preferred pattern - the preferred pattern is to follow the builder pattern. The helper function to wrap registering was explicitely added to EVM/SVM as they have backwards compatible v1 registering to do, AND those register functions required one call per network (do not support the CAIP-2 "eip155:*" regex). Therefore, the helper function was there to hide a more complex registry setup for mirroring legacy x402 SDK support.

New mechanism implementations only support `v2`, and therefore support the simple `{family}:*` registry approach, and therefore this function is redundant.

This PR:
* Removes the `registerAptosExactScheme` helpers from Aptos
* Updated `/aptos/src/exact/index.ts` to only re-export client rather than server/facilitators (mirrors EVM/SVM)
* Removes the usage of `registerEvmExactScheme` and `registerSvmExactScheme` from e2e & example clients/servers/facilitators so future contributors do not lean on this anti-pattern

## Tests

All unit tests, integration tests, and the full e2e test suite with just `--min` are all passing

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [x] I added a changelog fragment for user-facing changes (docs-only changes can skip)

PR #1360, opened 2026-02-26
feat(hedera): add initial @x402/hedera package

piotrswierzy:
add initial @x402/hedera package with exact payment scheme support
- Introduced the @x402/hedera package implementing the x402 v2 exact payment scheme for Hedera.
- Added client, server, and facilitator implementations along with necessary utilities and types.
- Included configuration files for ESLint, Prettier, and Vitest.
- Created documentation in README and CHANGELOG for package usage and features.
- Established a .prettierignore file to exclude specific directories and file types from formatting checks.

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

examples/typescript/clients/advanced/all_networks.ts
```diff
@@ -52,6 +61,15 @@ async function main(): Promise<void> {
     client.register("solana:*", new ExactSvmScheme(svmSigner));
     console.log(`Initialized SVM account: ${svmSigner.address}`);
   }
+  if (hederaAccountId && hederaPrivateKey) {
+    const hederaSigner = createClientHederaSigner(
+      hederaAccountId,
+      PrivateKey.fromStringECDSA(hederaPrivateKey),
+      { network: hederaNetwork },
+    );
```
jaycoolslm, line 89: Could we make this async and fetch Hedera Account Id from mirrornode?

examples/typescript/clients/advanced/package.json
```diff
@@ -18,7 +18,9 @@
     "@scure/base": "^1.2.6",
     "@x402/evm": "workspace:*",
     "@x402/svm": "workspace:*",
+    "@x402/hedera": "workspace:*",
     "@x402/fetch": "workspace:*",
+    "@hashgraph/sdk": "^2.75.0",
```
jaycoolslm: We should use @hiero-ledger/sdk

examples/typescript/clients/axios/index.ts
jaycoolslm: Not part of the review but are we awaiting something that a developer might not use?

piotrswierzy, line 34: I'm not sure, if I understand the question

examples/typescript/clients/advanced/all_networks.ts
skurzyp-blockydevs: I checked this and it is possible to derive the account ID by using the public key and querying the Mirror Node.

Example:

```ts
const publicKey = PrivateKey.toStringRaw(); 
// Note: the Mirror Node query fails if the public key is DER-encoded

const resp = await fetch(
  `https://testnet.mirrornode.hedera.com/api/v1/accounts?account.publickey=${publicKey}`
);

const data = await resp.json();
console.log(data.accounts[0].account); // prints the account ID
```

While this approach works, it is generally **not recommended**. On the Hedera network, multiple accounts can share the same public key. In such cases, querying by public key alone is not sufficient to uniquely determine the account ID.

examples/typescript/clients/advanced/package.json
skurzyp-blockydevs: Is there a difference? Is the `hashgraph/sdk` project being discontinued? I thought these SDKs were the same.

skurzyp-blockydevs: Also I think that the version should be locked

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -137,6 +144,27 @@ if (svmPrivateKey) {
...
+  const hederaKey = PrivateKey.fromStringECDSA(hederaPrivateKey);
+  const hederaSigner = toFacilitatorHederaSigner({
+    getAddresses: () => [hederaAccountId],
+    signAndSubmitTransaction: async (transactionBase64: string, _, network: string) => {
+      const tx = Transaction.fromBytes(Buffer.from(transactionBase64, "base64"));
+      if (!(tx instanceof TransferTransaction)) {
+        throw new Error("expected TransferTransaction");
+      }
+      const signed = await tx.sign(hederaKey);
+      const client = network === "hedera:mainnet" ? Client.forMainnet() : Client.forTestnet();
```
skurzyp-blockydevs: Just for maximum clarity we should mention that the `previewnet` is not supported.

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+    _tokenTransfers?: unknown[];
+  };
+  const isTransferTransaction =
+    Array.isArray(transaction._hbarTransfers) && Array.isArray(transaction._tokenTransfers);
+  const hbarTransfers = isTransferTransaction
+    ? (transaction._hbarTransfers ?? []).map(transfer => normalizeHbarTransfer(transfer))
+    : [];
+  const tokenTransfers = isTransferTransaction
+    ? normalizeTokenTransfers(transaction._tokenTransfers ?? [])
+    : {};
```
kierzniak: Instead of use dedicated type we can try to cast Transaction
```
const isTransferTransaction = transaction instanceof TransferTransaction;
```
also we should probably not use private properties like `_hbarTransfers` and `_tokenTransfers` but public methods.

https://github.com/hiero-ledger/hiero-sdk-js/blob/6579a696ed898f480a7deb8ebfae52bb68b32b11/src/account/TransferTransaction.js#L167
https://github.com/hiero-ledger/hiero-sdk-js/blob/6579a696ed898f480a7deb8ebfae52bb68b32b11/src/token/AbstractTokenTransferTransaction.js#L417

skurzyp-blockydevs (approved):

cb-heimdall:
Review Error for skurzyp-blockydevs @ 2026-03-12 14:43:04 UTC

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+ * @returns Atomic amount string
+ */
+export function convertToAtomicAmount(decimalAmount: string, decimals: number): string {
+  const [whole = "0", fraction = ""] = decimalAmount.trim().split(".");
+  if (!/^\d+$/.test(whole) || !/^\d*$/.test(fraction)) {
+    throw new Error(`Invalid decimal amount: ${decimalAmount}`);
+  }
+  const paddedFraction = fraction.padEnd(decimals, "0").slice(0, decimals);
+  return BigInt(`${whole}${paddedFraction || "0"}`).toString();
+}
```
kierzniak: I think we have a small bug here:
```
console.log(convertToAtomicAmount("5", 0)); // Produce 50 instead of 5.
```
also we do not really care about rounding. I we provide value 1.123456789 with 6 as decimals 789 suffix is just dropped.

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+ * Default network-specific HTS token config map key helper.
+ *
+ * @param network - CAIP-2 network
+ * @returns Network key
+ */
+export function getNetworkKey(network: string): string {
+  if (network === HEDERA_MAINNET_CAIP2) return HEDERA_MAINNET_CAIP2;
+  if (network === HEDERA_TESTNET_CAIP2) return HEDERA_TESTNET_CAIP2;
+  return network;
+}
```
kierzniak: This function doesn't make sense. It's essentially just returning the `network` argument as-is. Since it doesn't appear to be used anywhere, I'd suggest removing it entirely.

typescript/packages/mechanisms/hedera/.env.test
kierzniak, line 1: I don't see any other packages generating this file. They are just throw exception in integration test file that some of the environmental variables must be set. I think we should follow the established convention and remove it.

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,559 @@
...
+  }
+
+  /**
+   * Verify an exact Hedera payload.
+   *
+   * @param payload - Payment payload
+   * @param requirements - Matched payment requirements
+   * @returns Verification result
+   */
+  async verify(
```
kierzniak, line 91: I'm not sure about that but other implementation does a little bit more work during verification.

 1. Balance check
EVM checks balanceOf() in verify, SVM/Aptos catch it via simulation. Hedera does neither, so underfunded transactions only fail at settlement. Add an optional best-effort checkBalance?() hook to FacilitatorHederaSigner, matching EVM's try-catch pattern.

  2. Transaction simulation
SVM and Aptos simulate transactions during verify to catch bad signatures, insufficient funds, and invalid accounts in one call. Add an optional simulateTransaction?() hook to FacilitatorHederaSigner. This would also cover signature verification (point 3) for free.

  3. Signature verification
No check that the payer actually signed the partially-signed transaction. An unsigned payload passes verify() and only fails at settle(). Either implement via simulation (point 2) like SVM/Aptos, or add an explicit verifySignature?() hook like EVM's verifyTypedData().

  4. Transaction expiration check
Hedera transactions have a 180s validity window from validStart in the TransactionId, but the facilitator never checks it just like other implementation.

typescript/packages/mechanisms/hedera/src/signer.ts
```diff
@@ -0,0 +1,245 @@
...
+  nodeUrl?: string;
+};
+
+/**
+ * Identity helper for explicit client signer typing.
+ *
+ * @param signer - Client signer
+ * @returns Same signer
+ */
+export function toClientHederaSigner(signer: ClientHederaSigner): ClientHederaSigner {
```
kierzniak: Is this function necessary?

typescript/packages/mechanisms/hedera/src/signer.ts
```diff
@@ -0,0 +1,245 @@
...
+  if (nodeUrl) {
+    // A custom endpoint is mapped to account 0.0.3 by default.
+    // This can be overridden by constructing your own ClientHederaSigner.
+    return Client.forNetwork({ [nodeUrl]: AccountId.fromString("0.0.3") });
+  }
+  if (network === "hedera:mainnet") {
+    return Client.forMainnet();
+  }
+  return Client.forTestnet();
+}
```
kierzniak, line 252: We could use constants instead of strings like `hedera:mainnet` and `hedera:testnet`, this apply to all codebase. Also we could be more strict for network chose strategy `hedera:mainnet` or `hedera:testnet`or url else throw exception. What about `previewnet`?

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+    : {};
+
+  return {
+    transactionType: transaction.constructor?.name ?? "",
+    transactionId: transaction.transactionId?.toString?.() ?? "",
+    transactionIdAccountId: transaction.transactionId?.accountId?.toString?.() ?? "",
+    hasNonTransferOperations: !isTransferTransaction,
+    hbarTransfers,
+    tokenTransfers,
+  };
```
kierzniak, line 107: If something is wrong, we should return an error or throw an exception instead of falling back to empty strings.

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * Returns net transfer value received by an account.
+ *
+ * @param transfers - Transfer entries
+ * @param accountId - Account to calculate for
+ * @returns Net amount
+ */
+export function getNetForAccount(transfers: HederaTransferEntry[], accountId: string): bigint {
+  return transfers
+    .filter(entry => entry.accountId === accountId)
```
kierzniak: Can we use here `hederaAccountIdsEqual` function?

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+function normalizeHbarTransfer(transfer: unknown): HederaTransferEntry {
+  const candidate = transfer as {
+    accountId?: { toString?: () => string };
+    amount?: { toTinybars?: () => { toString?: () => string }; toString?: () => string };
+  };
+  return {
+    accountId: candidate.accountId?.toString?.() ?? "",
+    amount: candidate.amount?.toTinybars?.().toString?.() ?? candidate.amount?.toString?.() ?? "",
+  };
+}
```
kierzniak: A lot of silent fallback. If something is wrong, we should return an error or throw an exception instead of falling back to empty strings.

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+      accountId: candidate.accountId?.toString?.() ?? "",
+      amount: candidate.amount?.toString?.() ?? "",
+    };
+    if (!grouped[tokenId]) {
+      grouped[tokenId] = [];
+    }
+    grouped[tokenId].push(normalizedTransfer);
+  }
+  return grouped;
+}
```
kierzniak: A lot of silent fallback. If something is wrong, we should return an error or throw an exception instead of falling back to empty strings. `tokenId.length === 0` is also silent fallback.

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * @param payload - Hedera payload
+ * @returns Base64 transaction string
+ */
+export function decodeTransactionFromPayload(payload: ExactHederaPayloadV2): string {
+  if (!payload || typeof payload.transaction !== "string" || payload.transaction.length === 0) {
+    throw new Error("invalid_exact_hedera_payload_transaction");
+  }
+  return payload.transaction;
+}
```
kierzniak: This function do not decode transaction. We should change name or implement decoding.

typescript/packages/mechanisms/hedera/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,45 @@
...
+   * Creates a payment payload for exact Hedera payments.
+   *
+   * @param x402Version - x402 protocol version
+   * @param paymentRequirements - Selected payment requirements
+   * @returns Payload result
+   */
+  async createPaymentPayload(
+    x402Version: number,
+    paymentRequirements: PaymentRequirements,
+  ): Promise<Pick<PaymentPayload, "x402Version" | "payload">> {
```
kierzniak: Interface `SchemeNetworkClient`return different type from `createPaymentPayload` method. Can we make it consistent with interface?

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * @param network - Network identifier
+ * @returns Normalized network
+ */
+export function normalizeHederaNetwork(network: string): Network {
+  if (!SUPPORTED_HEDERA_NETWORKS.includes(network as (typeof SUPPORTED_HEDERA_NETWORKS)[number])) {
+    throw new Error(`Unsupported Hedera network: ${network}`);
+  }
+  return network as Network;
+}
```
kierzniak, line 41: This function is called multiple times in codebase for validation not normalization.  I think we are not normalizing anything here but narrowing `string` to `Network` type. Most of the time we even do not use a returned value. So this narrowing is not even used.

typescript/packages/mechanisms/hedera/src/exact/client/register.ts
```diff
@@ -0,0 +1,55 @@
...
+export interface HederaClientConfig {
+  /**
+   * The Hedera signer to use for creating payment payloads
+   */
+  signer: ClientHederaSigner;
+
+  /**
+   * Optional payment requirements selector function
+   */
+  paymentRequirementsSelector?: SelectPaymentRequirements;
```
kierzniak: This property is not used anywhere. Are we missing an implementation, or do we not need it?

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,559 @@
...
+    payload: PaymentPayload,
+    requirements: PaymentRequirements,
+  ): Promise<VerifyResponse> {
+    // Phase 1: requirements and accepted parity checks
+    const requirementsValidation = this.validateRequirements(payload, requirements);
+    if (requirementsValidation !== null) {
+      return requirementsValidation;
+    }
+    const feePayer = requirements.extra.feePayer as string;
```
kierzniak, line 100: I don't like this casting. Maybe we should extend PaymentRequirements with own type and guards to be sure that extra.feePayer exist?

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,559 @@
...
+  }
+
+  /**
+   * Validates payTo format based on alias policy.
+   *
+   * @param payTo - Destination account/alias value
+   * @returns True when payTo has acceptable format
+   */
+  private isValidPayToFormatForPolicy(payTo: string): boolean {
+    if (typeof payTo !== "string" || payTo.trim().length === 0) {
```
kierzniak, line 552: This part is redundant `typeof payTo !== "string"` We already check string in arguments.

typescript/packages/mechanisms/hedera/src/exact/facilitator/register.ts
```diff
@@ -0,0 +1,28 @@
...
+import type { FacilitatorHederaSigner } from "../../signer";
+import { ExactHederaScheme } from "./scheme";
+import type { HederaFacilitatorConfig } from "./scheme";
+
+/**
+ * Config for registering Hedera facilitator scheme.
+ */
+export interface HederaFacilitatorRegisterConfig extends HederaFacilitatorConfig {
+  signer: FacilitatorHederaSigner;
+  networks: Network | Network[];
```
kierzniak: In configuration interfaces sometimes I see `Network | Network[]` and sometimes just `Network[]` is it on purpose?

typescript/packages/mechanisms/hedera/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (typeof money === "number") {
+      return money;
+    }
+    const cleanMoney = money.replace(/^\$/, "").trim();
+    const amount = parseFloat(cleanMoney);
+    if (isNaN(amount)) {
+      throw new Error(`Invalid money format: ${money}`);
+    }
+    return amount;
+  }
```
kierzniak, line 139: This method exist in all mechanisms but it assume only using USD dollars? Does x402 standard only handle USD?

examples/typescript/clients/advanced/all_networks.ts
piotrswierzy: Also it's just in the example file

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
piotrswierzy: SVM does it the same way

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
piotrswierzy: 1. We are checking if the transaction is present in the transfer list, we are not verifying the on-chain balance itself. Hedera will enforce that at execution time, and checking a balance during verify() would be race-prone anyway because it can change before settle() submits.

2. For regular Hedera transactions like TransferTransaction, there is not a general built-in “simulate this exact transaction and tell me if it would succeed”

3. We can easily check for signature presence, but we cannot easily check “the customer really signed this payment correctly” without more account-key information and more protocol logic. So I wouldn't add it as it will be checked when the transaction is subbmited

4. Good catch. We currently only validate x402-level parity for maxTimeoutSeconds, but we do not check Hedera transaction freshness in facilitator verify(). At the moment this is enforced only by Hedera on submission via the transaction’s validStart/valid duration semantics, so an expired transaction should still fail at execution time

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
piotrswierzy: This is intentionally defensive at runtime. Even though isValidPayToFormatForPolicy() is typed as accepting a string, TypeScript types are erased at runtime and requirements can still come from untrusted input. The typeof check keeps this helper safe if it is ever called with malformed data.

typescript/packages/mechanisms/hedera/src/exact/facilitator/register.ts
piotrswierzy: Yes, this is on purpose. Public register/config APIs accept Network | Network[] as a convenience for callers, while the core facilitator normalizes that input to Network[] internally. So the union type is expected at the boundary; plain Network[] is used after normalization.

typescript/packages/mechanisms/hedera/src/exact/server/scheme.ts
piotrswierzy: This is only for a case where you want ot use stablecoin payments

examples/typescript/clients/advanced/package.json
piotrswierzy: done

typescript/packages/mechanisms/hedera/src/exact/client/register.ts
piotrswierzy: You're right, this property was unused here. In the current design paymentRequirementsSelector is configured on x402Client construction, not in per-scheme register helpers, and this Hedera helper only registers schemes/policies onto an already-created client. So there was no missing Hedera implementation; this was just redundant config surface. I removed it from the Hedera register config for clarity.

typescript/packages/mechanisms/hedera/src/exact/client/scheme.ts
piotrswierzy: Yes, we can make it consistent. The previous return type was a narrower compatible shape, but aligning the Hedera scheme with the shared SchemeNetworkClient signature is clearer. I updated it to use PaymentPayloadResult and the optional context parameter from the interface.

typescript/packages/mechanisms/hedera/src/signer.ts
piotrswierzy: You're right, this helper was not necessary. It was only an identity wrapper for explicit typing and did not add any behavior, unlike toFacilitatorHederaSigner(). To reduce redundant API surface, I removed it and updated the related tests.

typescript/packages/mechanisms/hedera/src/signer.ts
piotrswierzy: Agreed. I replaced the raw Hedera CAIP-2 strings with the shared constants and made the client selection logic strict. We now only map explicitly supported networks (hedera:mainnet / hedera:testnet), use a custom client when nodeUrl is provided, and throw for anything else instead of silently falling back to testnet. Previewnet is not currently supported by this mechanism, so it is rejected explicitly rather than inferred.

typescript/packages/mechanisms/hedera/src/utils.ts
piotrswierzy: Agreed. The previous helper name was misleading because it was not actually normalizing anything. I replaced it with two explicit helpers: isSupportedHederaNetwork() for conditional checks and assertSupportedHederaNetwork() for validation-plus-narrowing. The current call sites only need validation, so they now use the assert helper directly.

typescript/packages/mechanisms/hedera/src/utils.ts
piotrswierzy: Agreed. This helper was misnamed because it does not actually decode the transaction bytes; it only validates and extracts the base64 transaction string from the payload. I renamed it to extractTransactionFromPayload() to reflect the real behavior.

typescript/packages/mechanisms/hedera/src/utils.ts
piotrswierzy: Agreed. Returning empty-string fallbacks here can hide malformed transaction metadata and delay failure until later validation. I changed inspectHederaTransaction() to require transactionType, transactionId, and transactionIdAccountId, and to throw if any of them are missing instead of constructing a partially empty result

typescript/packages/mechanisms/hedera/src/utils.ts
piotrswierzy: Yes, that makes sense. I updated getNetForAccount() to use hederaAccountIdsEqual() so account matching is consistent with the rest of the Hedera validation logic instead of relying on raw string equality.

kierzniak (approved):

cb-heimdall:
Review Error for kierzniak @ 2026-03-16 11:45:39 UTC

phdargen:
Thanks a lot for the contribution @piotrswierzy, I'll begin reviewing shortly.

A few initial comments:
- please remove the registerExactHederaScheme helper functions in favor of builder pattern, see https://github.com/coinbase/x402/pull/1266
- revert changes in axios/fetch examples, hedera should ONLY be added to all_networks examples
- please add hedera to the e2e tests https://github.com/coinbase/x402/tree/main/e2e

piotrswierzy:
@phdargen 
Let me know if there is anything else I can do, to support you in the review process.

examples/typescript/clients/fetch/README.md
```diff
@@ -4,15 +4,18 @@ Example client demonstrating how to use `@x402/fetch` to make HTTP requests to e
 
 ```typescript
```
phdargen, line 5: please revert changes in this file, seems unrelated

examples/typescript/clients/axios/README.md
```diff
@@ -4,12 +4,19 @@ Example client demonstrating how to use `@x402/axios` to make HTTP requests to e
 
 ```typescript
```
phdargen, line 5: please revert

typescript/packages/mechanisms/hedera/test/integrations/exact-hedera.test.ts
```diff
@@ -0,0 +1,358 @@
+import { beforeEach, describe, expect, it } from "vitest";
```
phdargen, line 1: In the TS package root package.json please add hedera integration test here:
`"test:integration": "pnpm --filter @x402/core --filter @x402/evm --filter @x402/svm --filter @x402/aptos --filter @x402/stellar test:integration",`

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -37,13 +41,16 @@ if (!facilitatorUrl) {
...
+const HEDERA_HBAR_ASSET = "0.0.0" as const; // Native HBAR asset id
+const HEDERA_WEATHER_PRICE_TINYBARS = "100000" as const; // 0.001 HBAR
 const STELLAR_NETWORK = "stellar:testnet" as const; // Stellar Testnet
 
 // Build accepts array dynamically based on configured addresses
 const accepts: Array<{
   scheme: string;
-  price: string;
+  price: string | { amount: string; asset: string; extra?: Record<string, unknown> };
   network: `${string}:${string}`;
```
phdargen:
```suggestion
  price: Price;
  network: Network;
```
with import type { Network, Price } from "@x402/core/types";

examples/typescript/servers/advanced/.env-local
```diff
@@ -1,4 +1,5 @@
 EVM_ADDRESS=
 SVM_ADDRESS=
+HEDERA_ADDRESS=
```
phdargen: Took me a while to get the examples/e2e tests running, they fail when using an address like 
`HEDERA_ADDRESS=0x3f376c1e06b5495f2a015975671cc53f61e8e351`

This address was accepted by the faucet (https://portal.hedera.com/faucet), so seems to be a valid address. I created it with `PrivateKey.generateECDSA()` from `@hiero-ledger/sdk`

I then had claude write me a script to fetch the accountId from the API
`https://testnet.mirrornode.hedera.com/api/v1/accounts/` (any easier, more straightforward way to setup the accounts?)

With this the examples/e2e tests ran successfully
`HEDERA_ADDRESS=0.0.8450511`

Is `HEDERA_ADDRESS` supposed to accept both formats? Or just account ids? Then env vars should be renamed to `HEDERA_ACCOUNT_ID`

phdargen: Please also put a link to the faucet in the appropriate READMEs

phdargen:
There have been some changes to the e2e route structure, see https://github.com/coinbase/x402/pull/1779.
Could you please rebase against main and align your e2e tests?

Please also add e2e tests for the remaining server packages hono/next/fastify

typescript/packages/mechanisms/hedera/src/signer.ts
```diff
@@ -0,0 +1,239 @@
...
+        tx.addHbarTransfer(parsedAccountId, Hbar.fromTinybars((-amount).toString()));
+        tx.addHbarTransfer(payTo, Hbar.fromTinybars(amount.toString()));
+      } else {
+        const tokenId = TokenId.fromString(requirements.asset);
+        tx.addTokenTransfer(tokenId, parsedAccountId, -amount);
+        tx.addTokenTransfer(tokenId, payTo, amount);
+      }
+
+      tx.setTransactionId(TransactionId.generate(AccountId.fromString(feePayer)));
+      tx.setTransactionMemo(`x402:${Date.now().toString(36)}`);
```
phdargen: A transaction memo isnt mentioned in the specs, is this needed? Does it incur any additional tx costs? Should there be any constraints enforced by the facilitator on memo format and size?

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,559 @@
...
+    try {
+      inspected = inspectHederaTransaction(transactionBase64);
+    } catch {
+      return {
+        isValid: false,
+        invalidReason: "invalid_exact_hedera_payload_transaction_could_not_be_decoded",
+        payer: "",
+      };
+    }
+    const inspectedValidation = this.validateInspectedTransaction(inspected);
```
phdargen, line 125: I think we should check the client set txFee here and reject if above a reasonable threshold, similar to SVM. Otherwise malicious client could set it arbitrarily large and grief the facilitator

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,559 @@
...
+    const replayValidation = await this.validateReplay(inspected.transactionId);
+    if (replayValidation !== null) {
+      return replayValidation;
+    }
+
+    const payToValidation = await this.validatePayToPolicy(requirements);
+    if (payToValidation !== null) {
+      return payToValidation;
+    }
+
```
phdargen, line 147: Could we add tx simulation here or at least a balance check to protect the server from doing work for invalid requests?

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,559 @@
...
+    return null;
+  }
+
+  /**
+   * Runs optional replay protection hook.
+   *
+   * @param transactionId - Hedera transaction id
+   * @returns Failed verify response or null
+   */
+  private async validateReplay(transactionId: string): Promise<VerifyResponse | null> {
```
phdargen: Not sure about this replay protection mechanism. 

Could you please elaborate why this is needed in the first place? Is this security critical or an optimization? 

If a client sends the exact same tx multiple times, and the tx hasn't reached a confirmed state onchain yet, would settle return success for all even though only one actually settles and client pays only once?

If that's the case, could we align the implementation with the in-memory cache introduced for SVM in
https://github.com/coinbase/x402/pull/1468?

If a client sends a request and tx confirms onchain, then sends a request with exact same payload again, would settle fail? If we add simulation to `verify` would it fail for the duplicate tx? If so, do we still need the replay mechanism?

typescript/packages/mechanisms/hedera/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Default conversion when no custom parser handles the value.
+   *
+   * @param amount - Decimal amount
+   * @param network - Hedera network
+   * @returns AssetAmount in configured default HTS token
+   */
+  private defaultMoneyConversion(amount: number, network: Network): AssetAmount {
```
phdargen, line 148: could we define any default stablecoin for the moneyParser here as for all the other networks?

phdargen:
Hi @piotrswierzy, nice work! Made a first pass through the code and manually tested the examples/e2e tests, please find a few comments/suggestions above

typescript/packages/mechanisms/hedera/src/signer.ts
piotrswierzy: We should remove the transaction memo. According to the docs, it minimally increases the transaction cost, it's max length is 100 and it's enforced on the protocol level.

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
piotrswierzy: Hedera fees are protocol-fixed (e.g. ~$0.0001 for a CryptoTransfer), unlike SVM where the client can   
set an arbitrary priority fee via `SetComputeUnitPrice`, Hedera's `maxTransactionFee` is only an upper 
bound/cap, and the network always charges the standard fixed fee regardless. So the grief vector that exists on SVM (client inflates the fee the facilitator pays) doesn't apply here. Skipping this for now since the attack surface is negligible.

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
piotrswierzy: Hedera has no native transaction simulation API like Solana's simulateTransaction or Ethereum's `eth_call`, the sdk itself does some pre-checks internally.

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
piotrswierzy: After further analysis, Hedera has built-in protocol-level transaction deduplication: the network 
rejects any replayed transaction ID. This is the primary replay defense —      
double-spend is impossible. Unlike SVM, which uses an in-memory SettlementCache because Solana's replay 
protection is tied to blockhash expiry (~60-90s) and simulation catches concurrent duplicates, Hedera's   
native dedup makes the in-memory cache redundant. Removed `hasSeenTransaction`/`markTransactionSeen` hooks, `toFacilitatorHederaSigner`, and `DEFAULT_REPLAY_WINDOW_MS`

typescript/packages/mechanisms/hedera/test/integrations/exact-hedera.test.ts
piotrswierzy: done

typescript/packages/mechanisms/hedera/src/exact/server/scheme.ts
piotrswierzy: done

phdargen:
Thanks for the updates @piotrswierzy, looks good!

Could you please rebase against main as there are some merge conflicts and fix the following:
- one of the integration tests is failing
- facilitator example is not building as `@x402/hedera` doesn't export `toFacilitatorHederaSigner`
- when running the all-networks client example with Hedera configured, the process hangs after a successful payment as the underlying Hedera SDK Client stays open

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
phdargen: Could you please elaborate what these internal pre-checks do exactly?

Could we at least add a balance check if thats not covered by the pre-checks and simulation is not available?

piotrswierzy:
> Could you please rebase against main as there are some merge conflicts and fix the following:
> 
> * one of the integration tests is failing
> * facilitator example is not building as `@x402/hedera` doesn't export `toFacilitatorHederaSigner`
> * when running the all-networks client example with Hedera configured, the process hangs after a successful payment as the underlying Hedera SDK Client stays open

done

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
phdargen: These checks are still missing @piotrswierzy:
https://github.com/x402-foundation/x402/blob/2fda85136d7f1c7abfdcba86bde7f397c4976a2a/specs/schemes/exact/scheme_exact_hedera.md?plain=1#L157-L158

examples/typescript/servers/advanced/.env-local
phdargen: Please add the faucet info to the README @piotrswierzy

typescript/packages/mechanisms/hedera/src/signer.ts
```diff
@@ -0,0 +1,175 @@
...
+        tx.addHbarTransfer(payTo, Hbar.fromTinybars(amount.toString()));
+      } else {
+        const tokenId = TokenId.fromString(requirements.asset);
+        tx.addTokenTransfer(tokenId, parsedAccountId, -amount);
+        tx.addTokenTransfer(tokenId, payTo, amount);
+      }
+
+      tx.setTransactionId(TransactionId.generate(AccountId.fromString(feePayer)));
+      await tx.freezeWith(client);
+      client.close();
```
phdargen: this crashes at the second request:
```
  const response = await fetchWithPayment(url, { method: "GET" });
  const response2 = await fetchWithPayment(url, { method: "GET" });
```
with `Error: Failed to create payment payload: transaction must have been frozen before calculating the hash will be stable, try calling `freeze``

phdargen: The facilitator creates a fresh client inside each `signAndSubmitTransaction` call, uses it, and closes it. Suggest to follow the same strategy here, unless you have a better idea how to persist the client without reintroducing the process hangs

phdargen:
Thanks for the update @piotrswierzy! 

A few open issues, please see also see the open threads above:
- please fix client `close`
- please add missing fastify e2e test
- add faucet info to README
- add checks if transaction is expected to succeed onchain

piotrswierzy:
> Thanks for the update @piotrswierzy!
> 
> A few open issues, please see also see the open threads above:
> 
> * please fix client `close`
> * please add missing fastify e2e test
> * add faucet info to README
> * add checks if transaction is expected to succeed onchain

Thank you @phdargen 
I've applied the remarks

typescript/packages/mechanisms/hedera/README.md
```diff
@@ -0,0 +1,117 @@
...
+testnet HBAR for fees.
+
+- **Hedera Portal (faucet + account creation):** https://portal.hedera.com/
+  - Sign up, create a testnet account, and claim testnet HBAR.
+  - The portal also exposes your account id and ECDSA/ED25519 keys to plug into
+    `HEDERA_CLIENT_PRIVATE_KEY` / `HEDERA_FACILITATOR_PRIVATE_KEY`.
+
+For HTS token payments, both the payer and the recipient must have the token
+associated on their account before settlement. You can associate tokens via the
+portal or via `TokenAssociateTransaction` in the Hiero SDK.
```
phdargen: Didnt find how to do the association in the portal, so had to write a script. Could you please add a direct link where to do this?

phdargen: Please also add link to https://faucet.circle.com/ for getting USDC

phdargen:
Hi @piotrswierzy, I tested a USDC payment but it fails with:
`  error: Error: invalid_exact_hedera_payload_preflight_failed`
or when removing preflight checks:
```
    errorReason: 'transaction_failed',
    errorMessage: 't.startsWith is not a function',
```

phdargen:
> Hi @piotrswierzy, I tested a USDC payment but it fails with: ` error: Error: invalid_exact_hedera_payload_preflight_failed` or when removing preflight checks:
> 
> ```
>     errorReason: 'transaction_failed',
>     errorMessage: 't.startsWith is not a function',
> ```



> Hi @piotrswierzy, I tested a USDC payment but it fails with: ` error: Error: invalid_exact_hedera_payload_preflight_failed` or when removing preflight checks:
> 
> ```
>     errorReason: 'transaction_failed',
>     errorMessage: 't.startsWith is not a function',
> ```

This is what claude analyzed and the fix seems to work, but please double check:

Cause
The message t.startsWith is not a function comes from the Hiero SDK when something that is not a string is passed into hex.decode or AccountId.fromString (the minified parameter is often named t).

Your advanced facilitator and advanced client examples each listed @hiero-ledger/sdk in package.json and pulled it in again through @x402/hedera. With pnpm that produced two physical installs of the same version (different peer resolution, e.g. bn.js), so you had two different AccountId / Transaction class identities.

The client built the partially signed transfer with SDK copy A (via @x402/hedera). The facilitator did Transaction.fromBytes → sign → execute with SDK copy B (direct import). Inside the SDK, instanceof AccountId (and similar) can fail across those copies, so execution can fall through to code paths that call fromString with the wrong value, which surfaces as startsWith is not a function. Token (USDC) transfers are more likely to hit those branches than a simple HBAR transfer, which matches “HBAR worked, USDC failed.”

Fix
@x402/hedera now re-exports AccountId, Client, PrivateKey, Transaction, and TransferTransaction from @hiero-ledger/sdk so apps can use one SDK instance.

Removed the direct @hiero-ledger/sdk dependency from:

examples/typescript/facilitator/advanced/package.json
examples/typescript/clients/advanced/package.json
Both examples now import those symbols only from @x402/hedera.

e2e/servers/hono/index.ts
```diff
@@ -29,13 +30,17 @@ const EVM_NETWORK = (process.env.EVM_NETWORK || "eip155:84532") as `${string}:${
...
 const AVM_NETWORK = (process.env.AVM_NETWORK || "algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=") as `${string}:${string}`;
 const STELLAR_NETWORK = (process.env.STELLAR_NETWORK || "stellar:testnet") as `${string}:${string}`;
 const EVM_PAYEE_ADDRESS = process.env.EVM_PAYEE_ADDRESS as `0x${string}`;
 const SVM_PAYEE_ADDRESS = process.env.SVM_PAYEE_ADDRESS as string;
 const APTOS_PAYEE_ADDRESS = process.env.APTOS_PAYEE_ADDRESS as string;
+const HEDERA_PAYEE_ADDRESS = process.env.HEDERA_PAYEE_ADDRESS as string | undefined;
 const AVM_PAYEE_ADDRESS = process.env.AVM_PAYEE_ADDRESS as string;
 const STELLAR_PAYEE_ADDRESS = process.env.STELLAR_PAYEE_ADDRESS as string | undefined;
+const HEDERA_HBAR_ASSET = "0.0.0";
+const HEDERA_WEATHER_PRICE_TINYBARS = "100000"; // 0.001 HBAR
```
phdargen: Let's make these env vars here and in other server e2e tests so we can easily run test with USDC
```suggestion
const HEDERA_ASSET = process.env.HEDERA_ASSET ?? "0.0.0"; // 0.0.0 = HBAR or 0.0.429274 for USDC testnet
const HEDERA_AMOUNT = process.env.HEDERA_AMOUNT ?? "100000"; // price in smallest units (tinybars or token decimals), defaults to 0.0001 HBAR or 0.01 USDC
```

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,560 @@
...
+
+    const feePayer = requirements.extra?.feePayer as string;
+    const exactPayload = payload.payload as ExactHederaPayloadV2;
+    const transactionBase64 = extractTransactionFromPayload(exactPayload);
+    try {
+      const settled = await this.signer.signAndSubmitTransaction(
+        transactionBase64,
+        feePayer,
+        requirements.network,
+      );
```
phdargen, line 222: Got this in one of my tests:
```
Making request to: http://localhost:4021/weather

Response body: { report: { weather: 'sunny', temperature: 70 } }

Payment response: {
  "success": true,
  "payer": "0.0.8449447",
  "transaction": "0.0.8449469@1776757686.532272410",
  "network": "hedera:testnet"
}
```
even though the tx actually failed, see: https://hashscan.io/testnet/transaction/1776757694.619805037

phdargen: This shouldn't happen, we will need to add a check for tx success here

piotrswierzy: Wait, validating this, it should never happen

piotrswierzy: done, I didnt request to `getReceipt()` from the Hedera network

piotrswierzy:
> > Hi @piotrswierzy, I tested a USDC payment but it fails with: ` error: Error: invalid_exact_hedera_payload_preflight_failed` or when removing preflight checks:
> > ```
> >     errorReason: 'transaction_failed',
> >     errorMessage: 't.startsWith is not a function',
> > ```
> 
> > Hi @piotrswierzy, I tested a USDC payment but it fails with: ` error: Error: invalid_exact_hedera_payload_preflight_failed` or when removing preflight checks:
> > ```
> >     errorReason: 'transaction_failed',
> >     errorMessage: 't.startsWith is not a function',
> > ```
> 
> This is what claude analyzed and the fix seems to work, but please double check:
> 
> Cause The message t.startsWith is not a function comes from the Hiero SDK when something that is not a string is passed into hex.decode or AccountId.fromString (the minified parameter is often named t).
> 
> Your advanced facilitator and advanced client examples each listed @hiero-ledger/sdk in package.json and pulled it in again through @x402/hedera. With pnpm that produced two physical installs of the same version (different peer resolution, e.g. bn.js), so you had two different AccountId / Transaction class identities.
> 
> The client built the partially signed transfer with SDK copy A (via @x402/hedera). The facilitator did Transaction.fromBytes → sign → execute with SDK copy B (direct import). Inside the SDK, instanceof AccountId (and similar) can fail across those copies, so execution can fall through to code paths that call fromString with the wrong value, which surfaces as startsWith is not a function. Token (USDC) transfers are more likely to hit those branches than a simple HBAR transfer, which matches “HBAR worked, USDC failed.”
> 
> Fix @x402/hedera now re-exports AccountId, Client, PrivateKey, Transaction, and TransferTransaction from @hiero-ledger/sdk so apps can use one SDK instance.
> 
> Removed the direct @hiero-ledger/sdk dependency from:
> 
> examples/typescript/facilitator/advanced/package.json examples/typescript/clients/advanced/package.json Both examples now import those symbols only from @x402/hedera.

done

typescript/packages/mechanisms/hedera/README.md
piotrswierzy: done

typescript/packages/mechanisms/hedera/src/signer.ts
piotrswierzy: done

piotrswierzy:
@phdargen 
Done.
Hbar:
image

Token:
image

typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
phdargen: nice, retested and now correctly shows status=failed

phdargen (dismissed):
Thanks for the update @piotrswierzy, looks great 🚀

phdargen:
We need ALL your commits to be verified, could you please rebase and sign them all?

phdargen:
Would also be great if you could add optional Hedera support to the x402.org testnet facilitator here: https://github.com/x402-foundation/x402/blob/main/typescript/site/app/facilitator/index.ts

If the hedera team can provide sufficient testnet funds for ~10k txs, we'd be happy to include it there

This can be a follow-up PR

piotrswierzy:
@phdargen 
Done, thank you. I'll work on adding an optional Hedera support to the x402.org. To which accountId should we send the testnet hbar?

README.md
```diff
@@ -1,3 +1,7 @@
+> **Note:** We've moved the x402 repo under the x402 Foundation repo. All issues and PRs were transferred here: [github.com/x402-foundation/x402](https://github.com/x402-foundation/x402)
+>
+> Our repo ([coinbase/x402](https://github.com/coinbase/x402)) is now a development fork.
+
```
phdargen: Hey @piotrswierzy, seems like you rebased on coinbase/x402 (which is now a fork) not x402-foundation/x402. Could you please fix?

piotrswierzy: @phdargen missed this... it's fixed now

typescript/packages/mechanisms/hedera/src/utils.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Convert decimal string/number to atomic units.
+ *
+ * @param decimalAmount - Decimal amount (e.g. "1.25")
+ * @param decimals - Token decimals
+ * @returns Atomic amount string
+ */
+export function convertToAtomicAmount(decimalAmount: string, decimals: number): string {
```
phdargen: Please replace by `convertToAtomicAmount` from "@x402/core/utils", see https://github.com/x402-foundation/x402/pull/2053

phdargen:
We still need that verified signature @piotrswierzy 🙏

And please rebase again against x402-foundation/x402 main to resolve a conflict in the lock file

piotrswierzy:
> We still need that verified signature @piotrswierzy 🙏
> 
> And please rebase again against x402-foundation/x402 main to resolve a conflict in the lock file
@phdargen 
Commits signed and refactore to use: `convertToAtomicAmount ` is done

phdargen (approved):

piotrswierzy:
Thank you @phdargen for the support

phdargen:
> Thank you @phdargen for the support

Thank you for the contribution @piotrswierzy, great work!

jaycoolh:
LFG

kpachhai:
Amazing!!

PR #1408, opened 2026-03-02
Add files via uploadfeat: Add exact scheme specification for Tron (TVM) networks

EruditeIntelligence:
## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

EruditeIntelligence:
Hey team — this adds the exact scheme specification for Tron (TVM), the first x402 implementation on Tron. We have two confirmed on-chain USDT settlements running through our facilitator already. Ready for review whenever you get a chance. Thanks.

arjun215-eng:
Great to see Tron proposed here. A few things worth noting:

**1. Trust model worth calling out explicitly**
Unlike other chain specs that use a delegated authorization pattern, here the facilitator holds a fully-formed signed transaction. Worth a short note on what bounds the window between verification and broadcast — e.g. the expiration field — so the tradeoff is explicit for implementors.

**2. Address format comparison in ABI decoding**
Verification rule 3 says "the decoded recipient address MUST match `payTo` from `PaymentRequirements`" — but `payTo` is a base58check Tron address (starting with T), while the ABI-decoded address is 20 bytes (Tron strips the 0x41 network prefix for EVM ABI encoding). The spec doesn't define how this comparison should be performed, which is an easy implementation trap. Worth specifying the canonical conversion path explicitly.

**3. `raw_data` vs `raw_data_hex` — which is authoritative?**
The payload includes both `raw_data_hex` and `raw_data` (a JSON representation of the same data). The spec uses `raw_data_hex` for signature verification but reads field values (recipient, amount, contract address) from `raw_data`. A malicious client could provide inconsistent values between the two. Worth stating that all field extractions MUST be derived from parsing `raw_data_hex` directly, with `raw_data` treated as informational only.

EruditeIntelligence:
Thanks for the thorough review — all three points are solid catches. I've pushed an updated commit addressing each one:

**1. Trust model / broadcast window** — Added a dedicated subsection under Security Considerations ("Trust Model — Transaction Broadcast Window") that explicitly documents the `ref_block` + `expiration` mechanism that bounds the window between verification and broadcast. Recommended 60–120 second expiration for implementors.

**2. Address format comparison** — Expanded Verification Rule 3 to specify the canonical conversion path for comparing base58check `payTo` addresses against 20-byte ABI-decoded addresses. Includes both directions: strip the `0x41` prefix + checksum and compare raw bytes, or reconstruct base58check from the ABI bytes and compare strings.

**3. `raw_data` vs `raw_data_hex` consistency** — Added a preamble to the Verification Rules section declaring `raw_data_hex` as the authoritative data source. All field extractions during verification MUST be derived from parsing `raw_data_hex` directly. `raw_data` JSON is informational only. Also added a row to the EVM comparison table noting this distinction.

Appreciate the careful review. Let me know if anything else needs tightening up.

arjun215-eng:
@Sh11thead
@jiangyy0824
@Federico2014

Curious if Tron folks have thoughts about this approach!

EruditeIntelligence:
@arjun215-eng good points on both, let me address them.

**1. Trust model**

So this is actually already handled by how Tron transactions work but I agree it should be called out explicitly in the spec. Every signed Tron transaction has an `expiration` field in `raw_data` that the client sets at signing time (current time + `maxTimeoutSeconds`). If the facilitator sits on it and tries to broadcast after that window, the network rejects it outright — can't be mined. The client controls that value, not the facilitator.

On top of that the recipient address is baked into the signed transaction data (`transfer(address,uint256)` encoded in the contract call). You can't change where the funds go without breaking the signature. So even within the valid window, the facilitator can only send it where the client already agreed to.

I'll add a Trust Model section spelling this out.

**2. TOCTOU on balance**

Yeah this is a real window. Two things:

First — if the balance is gone by broadcast time the TRC-20 `transfer` just reverts on-chain. Facilitator eats the energy cost (~65k energy, roughly 27 TRX) but no funds move and no resource gets served. Fails closed.

Second — I'll add a SHOULD recommendation for facilitators to re-check `balanceOf` right before broadcasting. It's one read call, maybe 100ms on TronGrid, and it closes the gap for the normal case. The on-chain revert is still the backstop for actual race conditions.

I'll push both updates. Good catches.

________________________________

EruditeIntelligence:
Updated spec to address trust model and TOCTOU concerns — added explicit Trust Model section documenting expiration control and recipient lock guarantees, TOCTOU mitigation via balanceOf re-check before broadcast, and settlement failure modes table.

Issue #1247, opened 2026-02-19
`Task delivery escrow for agents — complement to x402 (looking for feedback)`


matteo-michele-bianchini:
x402 is great for instant pay-for-service — API calls, per-token billing, micropayments. But there's a different problem I haven't seen solved cleanly: what happens when an agent hires another agent for a task that takes 30 minutes and costs $50 USDC? Something where you don't know the quality of delivery upfront.

I see the community is already thinking about this — the escrow proposals in #839 and #834 both address the "pay now, settle later" pattern for metered usage. But there's a gap for **task delivery**: one task, one payment, uncertain quality.

I built [trust-escrow](https://www.npmjs.com/package/trust-escrow) — a smart contract on Base Sepolia (v1.2) + npm package that handles:
- Hold-until-delivery payment for agent tasks (not micropayments)
- Delivery verification window + dispute resolution via configurable arbiters
- Three independent time limits (delivery, payment release, arbiter response)
- 0.3% protocol fee, MIT licensed

```
npm install trust-escrow
```

Think of it as a different layer in the payment stack:

| Layer | Tool | Use case |
|-------|------|----------|
| Instant payment | x402 `exact` | API call, known price |
| Metered session | x402 `escrow` (#839, #834) | N calls, settle on usage |
| Task delivery | trust-escrow | 1 task, verify quality, then release |

I'd love feedback from people building with x402. Here's a concrete scenario:

> **Your agent hired another agent to analyze 10,000 customer records and produce a report. 30 minutes, $50 USDC.**
>
> 1. How do you handle payment in this situation today — pay upfront, hold until delivery, something else?
> 2. If the report is delivered but the quality is poor, how do you resolve that?
> 3. Would an on-chain Trust Score (an agent's history of completed tasks, payments, dispute outcomes) change how you select agents?

Happy to dive into technical details or compare approaches with the escrow proposals in PR #873.

seanthorne10:
Great framing on the payment stack layers. We've been building in this exact space with [Pylon](https://pylonapi.com) (x402-payable utility APIs for agents) and ran into the same "how do you trust an agent before paying $50?" problem.

To answer your three questions from our experience:

**1. Payment handling today:** For instant API calls (web extraction, search, translation), x402 `exact` works perfectly — the result comes back in the same HTTP response, so there's no trust gap. For longer tasks, we're seeing the Virtuals ACP pattern work: escrow + delivery verification + dispute window. Similar to what you built.

[...]

Issue #946, opened 2026-01-10
[Proposal] New Scheme: "Channel" - Pre-funded Escrow with Optimistic Settlement


roudra323:
**Author:** @roudra323

---

## Summary

This proposal introduces a new payment scheme called **"channel"**. It allows clients—particularly AI agents—to lock funds upfront in an escrow contract, make multiple API requests off-chain, and settle periodically using Merkle proofs. This approach is gas-efficient, trust-minimized, and ideal for high-frequency API usage.

### Problem with Current `exact` Scheme

The current x402 `exact` scheme uses EIP-3009 for per-request payments. For AI agents making hundreds or thousands of API calls per session, this creates significant friction:

* **Signature overhead:** Every request needs a fresh EIP-3009 signature.
* **Facilitator burden:** Balance checks and verification are required per call.
* **Latency:** Signing adds ~50-100ms delay to each request.
* **Gas inefficiency:** Requires a settlement transaction per call.

### Example: AI Agent Making 1000 API Calls

| Metric | Current (`exact`) | Proposed (`channel`) |
|--------|-------------------|----------------------|
| **Signatures required** | 1000 | 1 (deposit) |
| **On-chain transactions** | 1000 | 1-2 (deposit + settle) |
| **Gas cost (estimate)** | ~50M gas | ~100K gas |

---

### Complete Channel Lifecycle

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                           CHANNEL SCHEME FLOW                               │
└─────────────────────────────────────────────────────────────────────────────┘

PHASE 1: SETUP (On-chain)
═════════════════════════

  Agent                          Contract                        Facilitator
    │                               │                                │
    │  deposit(100 USDC)            │                                │
    │──────────────────────────────►│                                │
    │                               │                                │
    │                               │  Channel opened                │
    │                               │  Balance: 100 USDC             │
    │                               │  Status: ACTIVE                │


PHASE 2: USAGE (Off-chain)
══════════════════════════

  Agent                          Server                          Facilitator
    │                               │                                │
    │  API Request #1               │                                │
    │──────────────────────────────►│                                │
    │                               │                                │
    │  Response + CallReceipt       │                                │
    │◄──────────────────────────────│                                │
    │                               │                                │
    │  ┌─────────────────────────────────────────────────────────┐   │
    │  │ BOTH PARTIES:                                           │   │
    │  │ 1. Hash call: hash(callId, cost, timestamp)             │   │
[...]

Issue #864, opened 2025-12-24
[Extension Proposal] Escrow/Refund Setup With Only Extensions!


A1igator:
# x402r: Refundable Payments for x402

A secure, transparent refund system for x402 payments.  Enable dispute resolution and buyer protection with minimal code changes.

---

## Overview

When agents or users pay for services via x402, things can go wrong: 

- An API returns empty data after payment
- A purchased item arrives damaged or wrong
- A merchant becomes unresponsive

**x402r** solves this by routing payments through escrow contracts, enabling trustless refunds and dispute resolution.

---

## How It Works

```
┌─────────────────────────────────────────────────────────────────────────┐
│                    STANDARD x402 PAYMENT                                │
│                                                                         │
│   Client ──────────────────────────────────────────────► Merchant       │
│                    pays directly (no recourse)                          │
└─────────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────────────────────────────────────┐
│                    x402r REFUNDABLE PAYMENT                             │
│                                                                         │
│   Client ────► Proxy ────► Escrow ────► Merchant                        │
│                  │            │                                         │
│                  │            └─ Holds funds until dispute period ends  │
│                  │               If disputed:  Arbiter decides          │
│                  │                                                      │
│                  └─ Deterministic address (computed, not deployed)      │
│                     Client sees no difference!                          │
└─────────────────────────────────────────────────────────────────────────┘
```

### The Flow

1. **Merchant** registers with a shared escrow contract and deploys their relay proxy at registration (one-time setup)
2. **Server** marks routes as refundable → `payTo` is swapped to a proxy address
3. **Client** pays to proxy address (no code changes needed)
4. **Facilitator** routes payment through escrow
5. **Escrow** holds funds for dispute period
6. **Disputes** can be raised; arbiter resolves them

---

### Deterministic Proxy Addresses

Each merchant gets a unique proxy address computed using CREATE3 via CreateX:

```
proxyAddress = CREATE3(salt) // via CreateX
where salt = keccak256(merchantPayout + factoryAddress)
```

**Why this matters:**

| Property | Benefit |
|----------|---------|
| **Deterministic** | Same inputs → same address, always |
| **Computed locally** | No blockchain calls needed on server |
| **No bytecode needed** | CREATE3 doesn't require contract bytecode to compute address in comparison to CREATE2 |

### The Proxy Pattern

Instead of deploying a full contract per merchant (expensive), we use proxies:

```
[...]

PR #1425, opened 2026-03-03
Add `authCapture` scheme specification

A1igator:
At [x402r.org](https://x402r.org), we've been working on refundable payments for x402. This PR introduces the `authCapture` scheme, built on Base's audited Commerce Payments Protocol. It supports two settlement paths (two-phase and single-shot), with client signing via either ERC-3009 or Permit2.

Refs: #834, #1011

## Description

Adds the authCapture scheme specification as two files:

- **`scheme_authCapture.md`**: Cross-VM scheme overview. Covers settlement paths (two-phase vs single-shot via `autoCapture`), lifecycle (authorize/capture/void/refund/reclaim), the `captureAuthorizer` role, and security considerations.
- **`scheme_authCapture_evm.md`**: EVM implementation. Covers PaymentRequirements, PaymentPayload (ERC-3009 and Permit2 wire formats), verification logic, settlement logic, the on-chain `PaymentInfo` struct, fee system, universal CREATE2 contract addresses, and EIP-712 derivation tables.

Reuses audited [Commerce Payments Protocol](https://github.com/base/commerce-payments) contracts (AuthCaptureEscrow + token collectors). Client signs a single ERC-3009 `receiveWithAuthorization` or Permit2 `PermitTransferFrom`, depending on the `assetTransferMethod` selected by the server.

**Note on commit history:** The first two commits in this PR introduced this scheme under the `escrow` and then `commerce` names. The latest commit renames the directory to `authCapture`, which more precisely describes the authorize-then-capture lifecycle and parallels the `AuthCaptureEscrow` contract name. The old `commerce/` directory is removed in the same commit.

Related proposals: #839, #864, #946, #1247

## Tests

Spec-only PR. No code changes. Verification and settlement logic validated against a [reference implementation](https://github.com/BackTrackCo/x402r-scheme) with passing E2E tests on Base.

## Checklist

- [x] I added a changelog fragment for user-facing changes (docs-only changes can skip). Spec-only, skipped.


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

ryanRfox:
Good catch on the spec/implementation mismatch for simulation — that's exactly what #1377 tracks. The `exact` EVM spec requires simulation (step 5 in `scheme_exact_evm.md`) but no facilitator actually implements it, leading to gas loss on reverts (see #961, #418 for real-world examples).

There's a community contributor looking at a fix for `exact` (#1377). Once that lands, the escrow scheme's verification logic should probably align — a soft balance check (your step 10) catches some failures but misses things like consumed nonces, domain separator mismatches, and allowance issues that simulation would catch.

phdargen:
> Good catch on the spec/implementation mismatch for simulation — that's exactly what #1377 tracks. The `exact` EVM spec requires simulation (step 5 in `scheme_exact_evm.md`) but no facilitator actually implements it, leading to gas loss on reverts (see #961, #418 for real-world examples).
> 
> There's a community contributor looking at a fix for `exact` (#1377). Once that lands, the escrow scheme's verification logic should probably align — a soft balance check (your step 10) catches some failures but misses things like consumed nonces, domain separator mismatches, and allowance issues that simulation would catch.

Yes verify should include tx simulation, we have a PR up to add it to the exact implementation here: https://github.com/coinbase/x402/pull/1474

phdargen:
Thanks for putting this together @A1igator!

As a general comment, `scheme_escrow.md‎` should be network agnostic. Currently its too EVM specific and just reads as a lighter version of `scheme_escrow_evm.md` defying the point of splitting in 2 files. Instead `scheme_escrow.md‎`
should include abstract description, use cases and core properties without any network specifics

specs/schemes/commerce/scheme_commerce_evm.md
```diff
@@ -0,0 +1,203 @@
...
+### `extra` Fields
+
+| Field                        | Required | Type                      | Description                                        |
+| :--------------------------- | :------- | :------------------------ | :------------------------------------------------- |
+| `name`                       | Yes      | `string`                  | EIP-712 domain name for the token (e.g., `"USDC"`) |
+| `version`                    | Yes      | `string`                  | EIP-712 domain version (e.g., `"2"`)               |
+| `escrowAddress`              | Yes      | `address`                 | AuthCaptureEscrow contract address                 |
+| `operatorAddress`            | Yes      | `address`                 | Operator address                                   |
+| `tokenCollector`             | Yes      | `address`                 | Token collector contract address                   |
+| `settlementMethod`           | No       | `"authorize" \| "charge"` | Settlement path. Default: `"authorize"`            |
```
phdargen: Not sure about this branching based on metadata in extra. The payment flow seems to be quite different between the options, so maybe these should be 2 separate schemes?

phdargen: For exact we have assetTransferMethod = {3009, permit2} in `extra` but the payment flow is identical, assetTransferMethod is just an implementation detail

specs/schemes/escrow/scheme_escrow_evm.md
```diff
@@ -0,0 +1,203 @@
...
+
+The facilitator performs these checks in order:
+
+1. **Type guard**: Verify `payload` contains `authorization`, `signature`, and `paymentInfo` fields
+2. **Scheme match**: Verify `requirements.scheme === "escrow"` and `payload.accepted.scheme === "escrow"`
+3. **Network match**: Verify `payload.accepted.network === requirements.network` and format is `eip155:<chainId>`
+4. **Extra validation**: Verify `requirements.extra` contains required escrow fields (`escrowAddress`, `operatorAddress`, `tokenCollector`)
+5. **Time window**: Verify `validBefore > now + 6s` (not expired) and `validAfter <= now` (active)
+6. **ERC-3009 signature**: Recover signer from EIP-712 typed data (`ReceiveWithAuthorization` primary type) and verify matches `authorization.from`
+7. **Amount**: Verify `authorization.value >= requirements.amount`
```
phdargen:
```suggestion
7. **Amount**: Verify `authorization.value === requirements.amount`
```

specs/schemes/escrow/scheme_escrow_evm.md
```diff
@@ -0,0 +1,203 @@
...
+8. **Token match**: Verify `paymentInfo.token === requirements.asset`
+9. **Receiver match**: Verify `paymentInfo.receiver === requirements.payTo`
```
phdargen: Must also check payload.to === tokenCollector

specs/schemes/escrow/scheme_escrow.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+SIGN → CHARGE → RESOURCE DELIVERED
+```
+
+1. **Sign**: Client signs an ERC-3009 authorization (same as above)
+2. **Charge**: Facilitator calls `charge()` on the operator — funds go directly to receiver
+3. **Resource delivered**: Server returns the resource (HTTP 200)
+
+Post-settlement, the operator can refund within `refundExpiry` if needed. Unlike the authorize path, the payer cannot `reclaim()` — funds are already with the receiver.
```
phdargen: If the funds are directly send to the receiver (server), how are refunds supposed to work? How is it decided if a refund is needed? Can a client request a refund? What happens in case of disputes? If the funds are never held in the escrow, why is this under a `escrow` scheme?

phdargen: Exact already guarantees that no payment is made on server failure. So I suppose whats covered here would be if the client is not happy with the delivered service which is inherently subjective

phdargen:
Could you please clarify who the operator is in the x402 context? Is it the facilitator or another separate entity?

If its the facilitator, should it exposes additional endpoints: POST /capture, POST /void, POST /refund?

phdargen:
`Exact` works well for instant resource delivery like API responses. Its my understanding the primary use case of `escrow` would be delayed delivery (deferred settlement), for example e-commerce. 

Currently x402 is a stateless request-response protocol. Escrow introduces a long-lived lifecycle that extends beyond the original HTTP request. Could you clarify the state requirements for all the participants (client/server/facilitator)? Do we need to persists sth like a payment ID? 

The spec doesnt define the `PAYMENT-RESPONSE`, would this include sufficient info for the client to track the status of the payment or claim a refund? 

Can the server request a capture? How can the server correlate captured/refunded payments with the original request?

specs/schemes/commerce/scheme_commerce_evm.md
A1igator: I'm open to splitting it into 2 but variable input and settlement is basically identical meaning the schemes would be very similar (only difference being what function the facilitator calls, charge or authorize). That's why I thought it might make sense to have them together

specs/schemes/escrow/scheme_escrow.md
A1igator: So base commerce-payments is unopinionated on most of those questions and it would be implementation specific. You can check out our implementation at https://www.x402r.org/ .

I agree the naming might be a bit confusing so maybe "commerce" scheme would be better?

Regarding server failure and subjectiveness, correct. I will say one thing objective the middleware currently doesn't do is double check the payload matches a merchant set scheme but that could be added later.

A1igator:
Hey all! Thanks for the thoughtful responses!

I will address and add more verification as per exact scheme upgrades. I will also generalize the escrow_scheme.md

Regarding the operator, base commerce-payments is agnostic to who the operator could be. It could be the facilitator (although personally not sure I'd recommend that as the facilitator is meant to do activities on behalf of a user not be a trusted entity itself), smart contracts like our operator factory produces at x402r (examples here: https://docs.x402r.org/contracts/examples), or a merchant platform like Shopify in the original commerce-payments announcement.

The paymentInfo and paymentInfo hash can work as payment id. I will investigate PAYMENT-RESPONSE to see if it makes sense to include! Thanks for the heads up!

Server capture again depends on operator implementation. They can in our setup at x402r.

specs/schemes/escrow/scheme_escrow_evm.md
A1igator: Resolved!

specs/schemes/escrow/scheme_escrow_evm.md
A1igator: Resolved!

A1igator:
Hey all!I think I replied to or resolved all of the comments and pushed an update to the spec to generalize scheme_escrow.md, and tighten verification

I did check PAYMENT-RESPONSE and it does conceptually make sense to me. I checked the upto scheme and it seems to use SettlementResponse for that effect but SettlementResponse doesn't seem to be extendible for schemes currently? Only for extensions? So that seems to be blocked by another PR to fix that. I could add it to the spec but wanted to double check since implementation would be blocked.

Also regarding state, PaymentInfo and/or PaymentInfoHash have to be somehow tracked. How they are tracked somewhat depends on the operator implementation because the operator entity/contract could keep track. Otherwise the parties have to keep track of it themselves manually. Can add this clarification to the spec if makes sense there.

Thanks again for all the thoughtful responses!

specs/schemes/escrow/scheme_escrow.md
phdargen: yes I'd prefer "commerce" as scheme name aligning with the protocol it is build on

phdargen: Still not sure what the value add of the `charge()` mode is wrt exact. If funds are in the servers wallet, how do they flow back to the client? Can the operator claw back the funds? Or is it supposed to pay the refund out of its own pocket?

Seems to be strictly better and more straightforward to just keep the funds in the escrow until the challenge window is expired

phdargen:
Thanks a lot for the update @A1igator! 

The changes look good but I think the spec still leaves too much out as implementation detail.
The fact that the base commerce protocol is unopinionated about the operator doesnt necessarily mean this applies to a x402 specific implementation too.

I think a client needs to be able to fully understand the entire payment lifecycle based on the initial 402 response.  The authorization is well speced out, but everything after the initial request remains operator specific.

All the client knows is the operator address but how to infer from this alone how to interact with it after the initial request? If we leave all of this out of spec, a client needs to know out of band implementation specifics of each operator.

Once a client deposited in the escrow, how can they claim a refund? How do they know if their deposit has been charged? How can the server request a charge and when? On resource delivery or after refundExpirySeconds has passed?

I also think we should just define the facilitator as the operator. The facilitators role is the payment orchestration, so feels only natural (can I get your thoughts on this @fabrice-cheng?). This simplifies the spec and has it fully self-contained.

Then define new API interfaces for /capture /refund etc including who can call them when and under which circumstances. Do client/server need gas or is there a way for the facilitator to sponsor all operations?

A1igator:
Thanks for the suggestion @phdargen but I think providing information can be done without making facilitator the operator.

I would recommend a separate extension (like this one we've been building here: https://github.com/BackTrackCo/x402r-scheme/pull/27 and will send over when ready) to get all the information you would like to pass through the 402 flow to let the user know how to interact with the operator.

Facilitator as the operator would a) Make the facilitator heavily trusted which goes against x402 principles as listed in the readme: "Trust minimizing: all payment schemes must not allow for the facilitator or resource server to move funds, other than in accordance with client intentions". b) Would break protocols like ours which let's you pick your third party arbiters which can even be chained together. c) Break smart contract operators (which we also have) which are important in trust minimizing the operator with things like escrow periods.

There might be a version of this scheme that bundles the extension above and scheme together but I feel like we're ballooning the proposal then and opinionating it towards the refund usecase. (it can also be used for sessions for example like the Agentokratia proposal which has nothing to do with refunds: https://github.com/coinbase/x402/issues/834)

Would also love to know @fabrice-cheng's thoughts!

Thanks again for the review!

A1igator:
I will say maybe we could have /capture /refund, etc... interfaces that simply enact a signature's request from the authorized party through to the contracts (same way settle works) if you'd like the facilitator to simply pay for gas for every operation. I need to spec this out properly though and figure out what signature scheme makes sense if wanted. I'm also not quite sure what the benefits are vs a paymaster but could be wrong.

specs/schemes/escrow/scheme_escrow.md
A1igator: Charge (and refund as defined by base commerce-payments more specifically which is what charge is good for) is meant to be for scenarios where like a marketplace wants to guarantee certain amount of refunds as they only allow above certain reputation merchants for example. Another scenario is merchants themselves paying out of pocket or out of a bond they're supposed to put up.

Putting funds in the escrow for long periods of time is always strictly better for the client yes but it comes with the tradeoffs of merchant not getting their money fast even if they have a low refund rate that they could cover out of pocket.

Ditto on "commerce". Will rename next pass.

specs/schemes/escrow/scheme_escrow_evm.md
```diff
@@ -0,0 +1,235 @@
...
+      "extra": {
+        "name": "USDC",
+        "version": "2",
+        "escrowAddress": "0xEscrowAddress",
+        "operatorAddress": "0xOperatorAddress",
+        "tokenCollector": "0xCollectorAddress",
+        "settlementMethod": "authorize",
+        "minFeeBps": 0,
+        "maxFeeBps": 1000,
+        "feeReceiver": "0xOperatorAddress"
```
fabrice-cheng: should the expiries also make it to the PaymentRequirements?

specs/schemes/commerce/scheme_commerce_evm.md
```diff
@@ -0,0 +1,235 @@
...
+      "scheme": "escrow",
+      "network": "eip155:8453",
+      "amount": "1000000",
+      "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
+      "payTo": "0xReceiverAddress",
+      "maxTimeoutSeconds": 60,
+      "extra": {
+        "name": "USDC",
+        "version": "2",
+        "escrowAddress": "0xEscrowAddress",
```
fabrice-cheng: thoughts on defaulting x402 spec to a unique Escrow/TokenCollector (i.e: like Permit2) 
or allowing this flexibility?

The payload is really specific to AuthCaptureEscrow smart contract

specs/schemes/escrow/scheme_escrow_evm.md
```diff
@@ -0,0 +1,235 @@
...
+## PaymentRequirements
+
+Escrow-accepting servers advertise with scheme `escrow`:
+
+```json
+{
+  "x402Version": 2,
+  "accepts": [
+    {
+      "scheme": "escrow",
```
fabrice-cheng: auth-capture vs escrow?

specs/schemes/commerce/scheme_commerce_evm.md
```diff
@@ -0,0 +1,235 @@
...
+        "operatorAddress": "0xOperatorAddress",
+        "tokenCollector": "0xCollectorAddress",
+        "settlementMethod": "authorize",
```
fabrice-cheng: shouldn't this always be "authorize"?

specs/schemes/commerce/scheme_commerce_evm.md
```diff
@@ -0,0 +1,235 @@
...
+      "operatorAddress": "0xOperatorAddress",
+      "tokenCollector": "0xCollectorAddress",
+      "settlementMethod": "authorize",
+      "minFeeBps": 0,
+      "maxFeeBps": 1000,
+      "feeReceiver": "0xOperatorAddress"
+    }
+  },
+  "payload": {
+    "authorization": {
```
fabrice-cheng: do we need to pass this?

specs/schemes/escrow/scheme_escrow_evm.md
A1igator: Done!

specs/schemes/escrow/scheme_escrow_evm.md
A1igator: Renamed to "commerce" as per Duke's suggestion above: https://github.com/x402-foundation/x402/pull/1425#discussion_r2899106795

specs/schemes/commerce/scheme_commerce_evm.md
A1igator: There's also "charge" as I explained above: https://github.com/x402-foundation/x402/pull/1425#discussion_r2974454572

Implementation is 95% identical (just function call difference) so thought made sense to include here.

specs/schemes/commerce/scheme_commerce_evm.md
A1igator: Technically no but then some values like validAfter would be assumed. Depends how verbose/informational we want to be

specs/schemes/commerce/scheme_commerce_evm.md
A1igator: Firstly, I think the flexibility should definitely be allowed because while base commerce-payments contracts are pretty good and well designed, there's lots of innovation still left possible at the escrow layer without breaking this scheme. For example:

1. Adding partialVoid to allow for partial refunds in an escrow.
2. Adding AAVE or other defi for capital efficiency for funds in the escrow.
3. Gas cost improvements.

We already do 1 as it was asked for. TokenCollector's also appear to be tied to an AuthCaptureEscrow hence why they need to both be optional.

Now whether they should be optional and just default to the canonical ones if not set is a different question. Since our protocol already had to fork to do 1, we have to set it either way so this doesn't affect us but can add it if wanted.

fabrice-cheng:
Hi @A1igator ,

First of all, thanks for putting this together, and apologies for the back-and-forth.
After some more thoughts with the team, please find below some feedbacks building on your original proposal.
Let me know if you have time to get to it, or happy to take it over.

Tl;DR: 
- Trying to be consistent with x402 naming convention (instead of reusing the AuthCaptureEscrow smart contract protocol naming convention).
- Defaulting to a canonical contract for simplicity (similar to `Permit2` and `Upto`) x402 specs

Two-phase payments: the client **authorizes** a hold on funds (escrowed on-chain), and the `captureAuthorizer` later **captures** some or all of those funds. Mirrors the credit-card auth-capture pattern.

All EVM `authCapture` payments settle through a single, universal `AuthCaptureEscrow` contract (deployed via CREATE2, same address on all chains).

| AssetTransferMethod | Token Requirement | Notes |
|:--------------------|:----------------------------------------|:----------------------------------------------|
| **eip3009** | Must support `receiveWithAuthorization` (USDC, EURC) | Client signs transfer to `EIP3009_TOKEN_COLLECTOR_ADDRESS` (constant) |
| **permit2** | Any ERC-20 | Client signs Permit2 `permitTransferFrom`; spender = `PERMIT2_TOKEN_COLLECTOR_ADDRESS` (constant) |

---

## Naming (Commerce Protocol → x402)

| Commerce Protocol | x402 | Rationale |
|:--------------------------|:--------------------|:--------------------------------------------------|
| `operator` | `captureAuthorizer` | Address authorized to capture/void/refund |
| `authorizationExpiry` | `captureDeadline` | Deadline by which capture must occur |
| `refundExpiry` | `refundDeadline` | Consistent with `captureDeadline` |
| `preApprovalExpiry` | *(derived by client)* | `now + maxTimeoutSeconds`; sent as `validBefore` / `deadline` in payload |
| `feeReceiver` | `feeRecipient` | Standard English |
| `receiver` / `payer` / `maxAmount` / `token` | `payTo` / `from` / `amount` / `asset` | Existing x402 fields |
| `escrowContract` / `tokenCollector` | *(constants)* | Universal addresses, not in payload |

---

## PaymentRequirements

### Top-level fields (standard x402)

| Field | Type | Description |
|:--------------------|:---------|:----------------------------------------------|
| `scheme` | `string` | `"authCapture"` |
| `network` | `string` | CAIP-2 (e.g., `"eip155:8453"`) |
| `amount` | `string` | Max hold amount in atomic token units |
| `asset` | `string` | Token contract address |
| `payTo` | `string` | Merchant address (receives funds on capture) |
| `maxTimeoutSeconds` | `number` | How long the client's signature remains valid |

### `extra` fields

| Field | Type | Required | Description |
|:------------------------|:---------|:---------|:-------------------------------------------------|
| `assetTransferMethod` | `string` | No | `"eip3009"` (default) or `"permit2"` |
| `captureAuthorizer` | `string` | Yes | Address authorized to capture/void/refund (server or facilitator) |
| `name` | `string` | Yes | EIP-712 token name (for signing, EIP-3009 only) |
| `version` | `string` | Yes | EIP-712 token version (for signing, EIP-3009 only) |
| `captureDeadline` | `number` | Yes | Unix timestamp — must capture before this |
| `refundDeadline` | `number` | Yes | Unix timestamp — refunds allowed until this |
| `minFeeBps` | `number` | Yes | Minimum fee in basis points |
| `maxFeeBps` | `number` | Yes | Maximum fee in basis points |
| `feeRecipient` | `string` | Yes | Address that receives the fee portion |
| `salt` | `string` | Yes | Unique entropy per payment (bytes32 hex) |

### Example

```json
{
  "scheme": "authCapture",
  "network": "eip155:8453",
  "amount": "50000000",
  "payTo": "0xMerchantAddress",
  "maxTimeoutSeconds": 10800,
  "extra": {
    "name": "USD Coin",
    "version": "2",
    "captureAuthorizer": "0xCaptureAuthorityAddress",
    "captureDeadline": 1744243200,
    "refundDeadline": 1746835200,
    "minFeeBps": 0,
    "maxFeeBps": 250,
    "feeRecipient": "0xFeeRecipientAddress",
    "salt": "0x0000000000000000000000000000000000000000000000000000000000000001"
  }
}
```

> **`captureAuthorizer`**: The server sets this to its own address to manage captures directly, or to the facilitator's address to delegate. Committed on-chain — only this address can call `capture()`, `void()`, or `refund()`.

A server MAY advertise both `eip3009` and `permit2` in `accepts[]`. The client picks based on token support and existing approvals.

---

## Nonce Construction

Unlike the `exact` scheme (which uses random nonces), `authCapture` nonces are **deterministic** — derived from a hash of the payment parameters. This binds the token transfer authorization to the specific payment, preventing reuse across different payments.

The nonce is the **payer-agnostic PaymentInfo hash**:

```
paymentInfoWithZeroPayer = PaymentInfo {
    operator:            extra.captureAuthorizer
    payer:               0x0000000000000000000000000000000000000000
    receiver:            payTo
    token:               asset
    maxAmount:            amount
    preApprovalExpiry:   now + maxTimeoutSeconds       ← client picks this
    authorizationExpiry: extra.captureDeadline
    refundExpiry:        extra.refundDeadline
    minFeeBps:           extra.minFeeBps
    maxFeeBps:           extra.maxFeeBps
    feeReceiver:         extra.feeRecipient
    salt:                extra.salt
}

innerHash = keccak256(abi.encode(PAYMENT_INFO_TYPEHASH, paymentInfoWithZeroPayer))
nonce     = keccak256(abi.encode(chainId, AUTH_CAPTURE_ESCROW_ADDRESS, innerHash))
```

The payer is zeroed so that the nonce is the same regardless of who pays — the client computes the nonce before including their own address.

**`preApprovalExpiry`** is chosen by the client as `now + maxTimeoutSeconds`. This value flows through the payload as `validBefore` (EIP-3009) or `deadline` (Permit2), allowing the facilitator to reconstruct the same PaymentInfo.

All other inputs come from `PaymentRequirements` + universal constants (`PAYMENT_INFO_TYPEHASH`, `AUTH_CAPTURE_ESCROW_ADDRESS`, `chainId`).

---

## Client Payload Construction

The client derives the full payload from `PaymentRequirements` + universal constants + its own address + current timestamp. No RPC calls needed.

### EIP-3009

| Payload field | Derived from |
|:---------------------|:-----------------------------------------------|
| `authorization.from` | Client's own address |
| `authorization.to` | `EIP3009_TOKEN_COLLECTOR_ADDRESS` (constant) |
| `authorization.value` | `requirements.amount` |
| `authorization.validAfter` | `0` (collector hardcodes this) |
| `authorization.validBefore` | `now + requirements.maxTimeoutSeconds` (= `preApprovalExpiry`) |
| `authorization.nonce` | Deterministic — see [Nonce Construction](#nonce-construction) |
| EIP-712 domain | `name`, `version` from `extra`; `chainId` from `network`; `verifyingContract` = `asset` |

### Permit2

| Payload field | Derived from |
|:------------------------------|:-----------------------------------------------|
| `permit2Authorization.from` | Client's own address |
| `permitted.token` | `requirements.asset` |
| `permitted.amount` | `requirements.amount` |
| `spender` | `PERMIT2_TOKEN_COLLECTOR_ADDRESS` (constant) |
| `nonce` | Deterministic (as uint256) — see [Nonce Construction](#nonce-construction) |
| `deadline` | `now + requirements.maxTimeoutSeconds` (= `preApprovalExpiry`) |
| EIP-712 domain | Permit2 canonical contract (`chainId` from `network`) |

> **No witness field.** Unlike `exact` (which uses `permitWitnessTransferFrom`), `authCapture` uses plain `permitTransferFrom`. The merchant address (`payTo`) is cryptographically bound through the deterministic nonce, which encodes the full PaymentInfo including `receiver`.

---

## PaymentPayload

Payload shape is determined by `assetTransferMethod`. Both define new types specific to `authCapture` — they do NOT reuse `exact`'s payload types.

### EIP-3009 Payload

`authorization.to` = `EIP3009_TOKEN_COLLECTOR_ADDRESS` (constant). The collector calls `receiveWithAuthorization`, forwards tokens to the captureAuthorizer's TokenStore.

```json
{
  "x402Version": 2,
  "accepted": { "scheme": "authCapture", "...": "..." },
  "payload": {
    "signature": "0x...",
    "authorization": {
      "from": "0xPayerAddress",
      "to": "0xEIP3009TokenCollectorAddress",
      "value": "50000000",
      "validAfter": "0",
      "validBefore": "1744243200",
      "nonce": "0xa1b2c3..."
    }
  }
}
```

**What the client signs (EIP-712):**

```
Domain: { name: "USD Coin", version: "2", chainId: 8453, verifyingContract: <USDC address> }

ReceiveWithAuthorization {
    from:        <payer>
    to:          <EIP3009_TOKEN_COLLECTOR>   ← universal constant
    value:       50000000
    validAfter:  0
    validBefore: 1744243200                  ← = preApprovalExpiry (now + maxTimeoutSeconds)
    nonce:       0xa1b2c3...                 ← = payer-agnostic PaymentInfo hash
}
```

### Permit2 Payload

`spender` = `PERMIT2_TOKEN_COLLECTOR_ADDRESS` (constant). Uses `permitTransferFrom` (no witness). The deterministic nonce binds all payment parameters including the merchant address.

```json
{
  "x402Version": 2,
  "accepted": { "scheme": "authCapture", "...": "..." },
  "payload": {
    "signature": "0x...",
    "permit2Authorization": {
      "from": "0xPayerAddress",
      "permitted": {
        "token": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
        "amount": "50000000"
      },
      "spender": "0xPermit2TokenCollectorAddress",
      "nonce": "123456789",
      "deadline": "1744243200"
    }
  }
}
```

**What the client signs (EIP-712):**

```
Domain: Permit2 canonical contract (chainId: 8453)

PermitTransferFrom {
    permitted:  { token: <USDC>, amount: 50000000 }
    spender:    <PERMIT2_TOKEN_COLLECTOR>     ← universal constant
    nonce:      123456789                     ← = uint256(payer-agnostic PaymentInfo hash)
    deadline:   1744243200                    ← = preApprovalExpiry (now + maxTimeoutSeconds)
}
```

A1igator:
Hello @fabrice-cheng !

Thanks a ton for the detailed response and feedback! This is workable with our protocol so I can handle it!

A few notes though:

1. While I think defaulting to canonical makes sense, I believe this spec completely kills the optionality. escrowContract / tokenCollector don't seem to be settable at all not just defaulted. Wanted to make sure that is intentional as we'd have to drop user requested features like partial void if that's the case. I assume this was done to simplify AssetTransferMethod to set the tokenCollector and it's not the end of the world but wanted to make sure the tradeoffs are well understood.

2. "charge" seems to be completely dropped without a reason being provided. As I mentioned above the implementation is 95% identical (just a function call change) and I don't think keeping it as an option changes much in UX. We already have protocols interested in building on top of this system/our sdk that would have better UX/lower gas costs if they go through "charge" as opposed to auth-capture every time, as they only care about the tracking for refunds after the escrow. Still not end of the world as can always auth-capture in rapid succession/atomically but wanted to make sure it's an intentional change.

Thanks again for the thoughtful response! The rest looks good to me and I'm happy with it as it stands even with the points above.

fabrice-cheng:
@A1igator 

1. Could you specify what "Partial void" would enable? Ultimately, void amount is (Authorized amount - Captured Amount), which allows full void or partial void amount.

2. charge vs two-steps authorization/capture is server-side. It's up to the server to decide if they'd rather call (1) charge (2) authorization and then later capture.

A1igator:
@fabrice-cheng 

1. I think you may be right. I'm going to do some scoping to see if that works in every scenario and get back to you but that does seem workable.

2. A bit confused here. By server do you mean facilitator? How would the merchant server decide if it's not a field they can set?

fabrice-cheng:
you might be right:

In this case, I'm suggesting adding `autoCapture=true|false` in `extra`, and if i's omitted, it means it's false.


```
{
  "scheme": "authCapture",
  "network": "eip155:8453",
  "amount": "50000000",
  "payTo": "0xMerchantAddress",
  "maxTimeoutSeconds": 10800,
  "extra": {
    "name": "USD Coin",
    "version": "2",
    "captureAuthorizer": "0xCaptureAuthorityAddress",
    "captureDeadline": 1744243200,
    "refundDeadline": 1746835200,
    "minFeeBps": 0,
    "maxFeeBps": 250,
    "feeRecipient": "0xFeeRecipientAddress",
    "salt": "0x0000000000000000000000000000000000000000000000000000000000000001",
    "autoCapture": true
  }
}
```

A1igator:
I think this naming convention is a bit confusing because I don't think it'll be obvious to people that disabling "authCapture" on the "authCapture" scheme would lead to "charge". Doing "charge: true/false" would probably be clearer.

This preserves all the functionality though so I'm happy to go ahead with it! Can provide updated spec over the coming week.

fabrice-cheng:
Authorization & Capture are typically 2 distinct steps in Payments:
- authorization + manual capture (2 calls)

When calling AuthCaptureEscrow.charge, it combine them into a single step, which is equivalent to
- authorization + autocapture (1 call)

A1igator:
Ah sorry it's autoCapture not authCapture. I misread. Sounds good!

fabrice-cheng:
Please add the contract Address in the spec as well! Thank you!

A1igator:
Hi @fabrice-cheng !

This is not a pushback but I just needed some clarification for implementing. A bit confused what's happening with salt. It used to be derived by client automatically but now it is manually selected by the merchant since it's required in the extra? Wanted to make sure this is an intentional change. 

Since payer is always zero address in nonce calculation, the salt has to be fresh per request otherwise two different payers can have nonce collision. This means it should be automatically calculated and rotated by the server? I assume implementation wise that can be automatically filled via x402ResourceServer like name, version in exact? 

Would love to know your thoughts!

fabrice-cheng:
@A1igator , I supposed it can be client-side generated, it's critical taht it's unique as you mentioned. I'm in favor of dropping it in the PaymentRequirements, and will have to be passed in via PaymentPayload then.

cc: @avidreder

A1igator:
Hi @fabrice-cheng 

Sorry one more thing. Are we sure about the name "captureAuthorizer"?

This description: "Address authorized to capture/void/refund (server or facilitator)" is not really correct. It also has to do authorize. This means in the x402 flow, it practically either has to be the facilitator or a smart contract (since a smart contract operator can open up permissions to allow authorize and capture entity to be different). We have such smart contracts built out at [x402r.org](https://www.x402r.org/) so it's fine for us but just wanted to flag the name/description being a bit misleading.

fabrice-cheng:
The `Payment Authorization` is the ERC3009 signature itself technically.

The `captureAuthorizer` identifies who has the permission to capture a payment or void a payment. Different profile

The  `Authorizer` part of `captureAuthorizer` is to be consistent with the other x402 specs (x402BatchSettlement's naming of payerAuthorizer and receiverAuthorizer), which is super confusing because of payment authorization, but yeah..

A1igator:
The thing is the ERC3009 signature has to identify the operator/captureAuthorizer **and** it can only be settled (money sent to/through contracts) by said entity. Only the operator/captureAuthorizer is allowed to "authorize" (in addition to only being allowed to capture/void/refund)

image

More reading: https://github.com/base/commerce-payments/blob/main/docs/operations/Authorize.md

It has `onlySender(paymentInfo.operator)` gate. In x402 case, that means without smart contract operators, only the facilitator can be the captureAuthorizer since it also does the "authorize". As I mentioned above we have said contracts so it's fine but the name/description might confuse people thinking it only handles capture.

A1igator:
Hey @fabrice-cheng !

Just pushed an update implementing all the changes you requested and canonical addresses pre-deployed on Base mainnet and Sepolia and others. Anyone can also deploy with the same address on any chain as I provided the salts in the spec! Updated PR title and description as well to match.

A few notes though:

1. Kept the name captureAuthorizer for now but updated the description to "Address authorized to authorize/capture/void/refund/charge." to be more accurate. Let me know if you'd like a different name too knowing the description.
2. I forgot to mention this because we don't use it but to be faithful to the original contracts, feeReciever can be set to `address(0)` to let the operator/captureAuthorizer specify any non-zero recipient at capture/charge time. Added that to the description but might want to consider making it optional.

We are happy with the spec as is though and would be happy to see it merged! We also almost have implementation of the scheme ready that we can push upstream as well: [x402r-scheme](https://github.com/BackTrackCo/x402r-scheme). 

Thanks for the feedback and review!

levalleux-ludo:
Hi all — flagging this here as you've been following the escrow / buyer protection discussion.

We've just opened [#2222](https://github.com/x402-foundation/x402/issues/2222), proposing `scheme: "escrow"` as a **wire-format standard** for non-custodial on-chain escrow in x402. It is intentionally implementation-agnostic: rather than standardising a specific contract, it standardises how buyer and seller communicate over HTTP, so that any of the escrow approaches already under discussion in this repository can publish conforming `nextActions` and become driveable by a single generic client.

The proposal includes an explicit relationship table covering the existing escrow-related issues and PRs (including this one), so it builds on rather than competes with the work already discussed here.

Your feedback would be very welcome — please take a look and weigh in directly on [#2222](https://github.com/x402-foundation/x402/issues/2222).

A1igator:
Hey @levalleux-ludo !

Thanks for the suggestion and I don't mind this but I think we might be over-abstracting at this point and taking an unneeded UX hit and phishing/hack attack surface increase as nextActions seems pretty complicated for the average merchant. The Base commerce-payments escrows are not opinionated so any flow you'd like, including the Boson one, can be done as an operator contract. We can help provide it at [x402r](https://www.x402r.org/) if you'd like as we're already doing for number of other partners or anyone can build it themselves.

x402 standard also seem to prefer locking schemes to canonical contracts as I already suggested making the escrow contract address flexible but that was shut down to match other schemes.

fabrice-cheng:
This looks great, thank you for putting this together.
@A1igator , one feedback, let's just reuse the existing AuthCaptureEscrow contract out there that is live.

We will be deploying those contracts on the other EVM chains to make sure the address are consistent

A1igator:
Done @fabrice-cheng!

levalleux-ludo:
> Hey @levalleux-ludo !
> 
> Thanks for the suggestion and I don't mind this but I think we might be over-abstracting at this point and taking an unneeded UX hit and phishing/hack attack surface increase as nextActions seems pretty complicated for the average merchant. The Base commerce-payments escrows are not opinionated so any flow you'd like, including the Boson one, can be done as an operator contract. We can help provide it at [x402r](https://www.x402r.org/) if you'd like as we're already doing for number of other partners or anyone can build it themselves.
> 
> x402 standard also seem to prefer locking schemes to canonical contracts as I already suggested making the escrow contract address flexible but that was shut down to match other schemes.

Thanks @A1igator — appreciate the careful read. A few short clarifications, point by point.

**On `nextActions`.** Worth flagging that in our reference implementation `nextActions` isn't merchant-authored. It's auto-derived server-side from the canonical state machine by the SDK (`deriveNextActions(exchange, serverConfig)`), much like the `accepts[]` block of a 402 response today. For a minimum-viable seller — commit → release, no disputes — the merchant doesn't write or reason about it. That said, your point about a clearer "minimum merchant" walkthrough is fair and we owe that in the docs.

**On the phishing / attack-surface concern.** Just to surface the relevant defenses: the buyer's authorization is EIP-712-signed against the escrow contract's `verifyingContract` and the specific action, and token authorizations (ERC-3009 / Permit / Permit2 / pre-approval) are amount- and destination-locked to `escrowAddress`. Unrecognized action IDs are skipped by clients (namespaced prefixes). The trust boundary for "is this the right escrow contract" sits in `requirements`, not in `nextActions` — which leads into your next point.

**On expressing the Boson flow as an operator contract on BCP-Base.** We've looked at this and there are a few places where it isn't yet obvious to us how the flow maps cleanly — for example, the operator role in `AuthCaptureEscrow` looks single-trust with buyer-only `reclaim()` as the fallback, whereas our flow needs unconditional on-chain progress paths for *both* parties in every non-terminal state; and BCP-Base's collector set is narrower than what `escrow` aims to cover (we also need Permit and pre-approval). These may well be misreadings on our side and we're not declaring them as blockers — just flagging that the mapping isn't trivial from where we sit.

**On canonical contracts.** Good signal that x402 prefers locking schemes to canonical addresses. We're open to that direction: an `escrowId` referencing a canonical registry would be a strict improvement over a free-form `escrowAddress` from our perspective, and we'd be happy to align there.

Happy to keep the broader scheme-design conversation on #2222 where it's already underway with the wider x402 group — easier to converge with the decision team in the loop. Thanks again.

A1igator:
Hey @levalleux-ludo !

I won't go too in depth here on most of your points regarding your proposal as you mentioned it belongs in the other thread. Still think #2222 is too complex whether that complexity falls on merchant or client or x402 maintainers.

Re expressing Boson flows though: 

Firstly, there's token collectors for both permit2 and pre-approval: https://github.com/base/commerce-payments/blob/main/docs/TokenCollectors.md

Secondly, as I mentioned before the operator needs to be a smart contract to make things work not a single trusted party but "on-chain progress paths for both parties in every non-terminal state" is very much possible. We have some examples here: https://github.com/BackTrackCo/arbiter-examples. Kleros one for example is optimistic with funds automatically being released to the merchant after an escrow period, while the AI one is pessimistic with funds going back to the payer after an escrow period. Won't bog down this thread with too much specific arbiter style discussion but happy to discuss offline how to make Boson work.

fabrice-cheng (approved):

CarsonRoscoe (approved):
Thank you @A1igator for the contribution 🙏

levalleux-ludo:
> Hey @levalleux-ludo !
> 
> I won't go too in depth here on most of your points regarding your proposal as you mentioned it belongs in the other thread. Still think #2222 is too complex whether that complexity falls on merchant or client or x402 maintainers.
> 
> Re expressing Boson flows though:
> 
> Firstly, there's token collectors for both permit2 and pre-approval: https://github.com/base/commerce-payments/blob/main/docs/TokenCollectors.md
> 
> Secondly, as I mentioned before the operator needs to be a smart contract to make things work not a single trusted party but "on-chain progress paths for both parties in every non-terminal state" is very much possible. We have some examples here: https://github.com/BackTrackCo/arbiter-examples. Kleros one for example is optimistic with funds automatically being released to the merchant after an escrow period, while the AI one is pessimistic with funds going back to the payer after an escrow period. Won't bog down this thread with too much specific arbiter style discussion but happy to discuss offline how to make Boson work.

@A1igator — thanks for engaging.
**On openness**.
Starting there because it's the load-bearing point: we believe `scheme: "escrow"` has to be open, so that any escrow implementation can plug in under it. We noticed you originally argued for that same stance — what drove you to change your mind?
Different escrows exist for a reason. Boson made a specific set of design decisions and trade-offs, but other escrow types are warranted for other use cases — e.g. high-value items needing stronger verification, longer dispute windows, or different arbitration models. A scheme that bakes in one implementation forecloses those.

**On the suggestion that Boson (or another escrow) could plug _into_ `authCapture`.**
Two problems:
(a) It isn't open — it hard-codes Coinbase's `AuthCaptureEscrow `contract as the settlement layer everything else has to sit on top of.
(b) The composition doesn't fit architecturally. Boson isn't an operator or an arbitrator — Boson includes its own escrow, with its own custody, state machine, and dispute logic. Nesting one escrow inside another duplicates those responsibilities rather than composing them. The same applies to cart.fun's `TrustEngine`, PayCrow, Arbitova, and others: they are escrows in their own right, not policy modules sitting on top of `AuthCaptureEscrow`.

**On complexity**. Fair point — we're looking into moving some of the multi-step-exchange surface out of the scheme itself, along the lines you suggested.

A1igator:
Hey @levalleux-ludo,

"e.g. high-value items needing stronger verification, longer dispute windows, or different arbitration models." are all possible with this scheme. Custody is the only thing this scheme "hard-codes" but that seems preferable to have a battle tested audited escrow. Any state machine and dispute logic is doable with this scheme.

This PR is already merged so don't want to discuss here and ping the maintainers for no reason but feel free to join our [discord](https://discord.com/invite/4JdMGUyu9Z) if you wanna learn more.

avidreder:
@A1igator did you have an in progress implementation of the Auth Capture SDK? I'm working on something for Go, so would love to see what you have so we can ensure implementations are consistent.

A1igator:
@avidreder yeah here's our TS implementation: https://github.com/BackTrackCo/x402r-scheme. Just doing examples before opening an upstream PR.

levalleux-ludo:
> Hey @levalleux-ludo,
> 
> "e.g. high-value items needing stronger verification, longer dispute windows, or different arbitration models." are all possible with this scheme. Custody is the only thing this scheme "hard-codes" but that seems preferable to have a battle tested audited escrow. Any state machine and dispute logic is doable with this scheme.
> 
> This PR is already merged so don't want to discuss here and ping the maintainers for no reason but feel free to join our [discord](https://discord.com/invite/4JdMGUyu9Z) if you wanna learn more.

Thanks @A1igator.
Two notes for the record before stepping back:

  **On custody as a feature, not a hard-code**. An audited single contract is valuable at the implementation level — it's the foundation any conforming `authCapture `server inherits. Our point is upstream of that. A scheme-level commitment to a specific custody contract pins the wire format to one implementation's roadmap. The same audit reasoning that justifies `AuthCaptureEscrow `today applies to every future audited escrow that wants to be conformant under the same scheme — and they can be, if custody sits at the implementation layer rather than the scheme layer. That's the openness point; it doesn't take anything away from the contract you've built.

  **On "all possible via operator contracts"**. An operator solves the scheme-level openness question only at one of two costs. If the operator is an EOA, trust in the operator replaces trust in the seller — the escrow becomes redundant relative to direct payment. If the operator is a contract, the client must learn its full action surface per implementation — at which point you're back to per-implementation client knowledge, which is exactly the trap an open scheme is meant to prevent. Either way, the wire-format separation we're proposing is still load-bearing for any second conforming implementation.

Happy to leave PR #1425 as merged. For the open-scheme question we'll keep the substantive thread on Issue #2222 where the maintainers are already engaged. 
Thanks for the Discord invite — we'll stay on the canonical GitHub surface for now.

A1igator:
@avidreder here you go: https://github.com/x402-foundation/x402/pull/2308!

PR #1446, opened 2026-03-04
Add EVM verify transaction simulation (#1377)

### LINKED ISSUE #1377: [EVM] Implement transaction simulation in verify to prevent facilitator gas loss
### Problem

The EIP-3009 and Permit2 `verify` functions perform offline checks (signature validation, balance, allowance, timing), but do not simulate the actual settlement transaction against the on-chain contract. This means `verify` can return `isValid: true` for payloads that will revert at settlement time, causing the facilitator to waste gas on a doomed transaction.

The x402 specification itself lists "Transaction Simulation" as step 6 of the verification steps (section 6.1.2), but neither facilitator mechanism implements it.

Concrete failure modes that pass verify but fail settle:

- **EIP-3009 nonce already consumed** on-chain (replayed authorization)
- **EIP-712 domain separator mismatch** -- the `name()` returned by the token contract differs from what was used to sign (the "USDC" vs "USD Coin" issue from #961)
- **Permit2 nonce already consumed** in the Permit2 contract
- **Proxy contract state issues** -- the x402Permit2Proxy could be in an unexpected state

This is an ongoing source of user-reported bugs:

- #961 -- verify succeeds but settle fails with unhelpful `invalid_payload` error. A [community comment](https://github.com/coinbase/x402/issues/961) independently recommended `eth_call` simulation.
- #418 -- "unable to estimate gas" on settle after successful verify (base-sepolia, domain name mismatch)
- #1065 -- intermittent gas estimation failures on settle (base mainnet). Root cause turned out to be a CDP state propagation race condition, but simulation during verify would have caught these before burning gas.
- #552 -- notes that the spec's verification step ordering should be optimized, and references the simulation step

### Affected code

All line references are against [`ad5cf25`](https://github.com/coinbase/x402/commit/ad5cf25f16ceb866d6b504dba188a3b12de0a058) (current `main`).

**`FacilitatorEvmSigner` interface** -- [`signer.ts` L23-L53](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/signer.ts#L23-L53) -- has `readContract` and `writeContract` but no `simulateContract`

**EIP-3009 verify** -- [`eip3009.ts` L31](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts#L31) returns `isValid: true` at [L209](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts#L209); settle calls `writeContract` at [L282](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts#L282) and [L300](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts#L300)

**Permit2 verify** -- [`permit2.ts` L41](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/permit2.ts#L41) returns `isValid: true` at [L220](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/permit2.ts#L220); settle calls `writeContract` at [L257](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/packages/mechanisms/evm/src/exact/facilitator/permit2.ts#L257)

**Facilitator viem setup** -- [`site/app/facilitator/index.ts` L35-L39](https://github.com/coinbase/x402/blob/ad5cf25f16ceb866d6b504dba188a3b12de0a058/typescript/site/app/facilitator/index.ts#L35-L39) -- already uses `createWalletClient(...).extend(publicActions)` which provides `simulateContract`

### Risk

Gas loss for the facilitator. Not a fund-loss vector -- the worst case is the facilitator pays gas for a reverting transaction. At scale, repeated verify-passes-settle-fails wastes gas and degrades the client experience (the client gets a 402 response, hands over a valid payment, and still gets denied).

<details>
<summary>Implementation sketch</summary>

#### Interface change in `signer.ts`

Add an optional `simulateContract` method to `FacilitatorEvmSigner`:

```typescript
export type FacilitatorEvmSigner = {
  // ... existing methods ...

  simulateContract?(args: {
    address: `0x${string}`;
    abi: readonly unknown[];
    functionName: string;
    args: readonly unknown[];
    account?: `0x${string}`;
  }): Promise<void>;
};
```

Making it optional (`?`) avoids a breaking change for existing `FacilitatorEvmSigner` implementations. When not provided, verify skips simulation (current behavior).

In the viem-based implementation (`site/app/facilitator/index.ts`), wire it up:

```typescript
simulateContract: (args) =>
  viemClient.simulateContract({
    ...args,
    account: args.account ?? evmAccount.address,
  } as any).then(() => undefined),
```

The `account` parameter is needed because viem's `simulateContract` sets `msg.sender` -- for EIP-3009 `transferWithAuthorization`, the caller must be any valid address (not the `from`), and for Permit2 the caller is the proxy contract, but the simulation just needs to not revert.

#### EIP-3009 verify addition

Before the final `return { isValid: true }` at L209 in `eip3009.ts`:

```typescript
if (signer.simulateContract) {
  try {
    await signer.simulateContract({
      address: erc20Address,
      abi: eip3009ABI,
      functionName: "transferWithAuthorization",
      args: [
        getAddress(eip3009Payload.authorization.from),
        getAddress(eip3009Payload.authorization.to),
        BigInt(eip3009Payload.authorization.value),
        BigInt(eip3009Payload.authorization.validAfter),
        BigInt(eip3009Payload.authorization.validBefore),
        eip3009Payload.authorization.nonce,
        eip3009Payload.signature!,
      ],
    });
  } catch {
    return { isValid: false, invalidReason: "settlement_simulation_failed", payer };
  }
}
```

#### Permit2 verify addition

Same pattern before `return { isValid: true }` at L220 in `permit2.ts`:

```typescript
if (signer.simulateContract) {
  try {
    await signer.simulateContract({
      address: x402ExactPermit2ProxyAddress,
      abi: x402ExactPermit2ProxyABI,
      functionName: "settle",
      args: [
        {
          permitted: {
            token: getAddress(permit2Payload.permit2Authorization.permitted.token),
            amount: BigInt(permit2Payload.permit2Authorization.permitted.amount),
          },
          nonce: BigInt(permit2Payload.permit2Authorization.nonce),
          deadline: BigInt(permit2Payload.permit2Authorization.deadline),
        },
        getAddress(payer),
        {
          to: getAddress(permit2Payload.permit2Authorization.witness.to),
          validAfter: BigInt(permit2Payload.permit2Authorization.witness.validAfter),
          extra: permit2Payload.permit2Authorization.witness.extra as `0x${string}`,
        },
        permit2Payload.signature,
      ],
    });
  } catch {
    return { isValid: false, invalidReason: "settlement_simulation_failed", payer };
  }
}
```

</details>

### Questions for maintainers

1. **Optional vs required**: Making `simulateContract` optional on the interface avoids a breaking change, but means facilitators that forget to wire it up get no protection. Should this be required instead, with a changeset/migration note?

2. **Latency tradeoff**: Simulation adds one `eth_call` RPC round-trip per verify. On Base L2 this is fast, but it's measurable. The spec already mandates it (step 6 of 6.1.2), so the question is whether the implementation should match the spec unconditionally or offer an opt-out.

3. **Error detail**: Should the simulation revert reason (e.g., "AuthorizationAlreadyUsed", "InvalidSignature") be propagated in `invalidReason`, or is a generic `settlement_simulation_failed` sufficient? Structured error detail would help with the debugging problems reported in #961 and #418.

### Related

- #961 -- Payload is verified successfully but fails settlement
- #418 -- "unable to estimate gas" on settle after successful verify
- #1065 -- Intermittent gas estimation errors on Base mainnet settle
- #552 -- Spec verification step ordering (references simulation as step 6)

ruhil6789:
Closes #1377.

This PR adds pre-settlement transaction simulation to the EVM facilitator **verify** path so that payloads which would revert at settlement time are rejected before the facilitator spends gas.

Concretely:

- **EIP-3009**: `verifyEIP3009` now calls a new `simulateEip3009Settlement` helper that:
  - Uses the facilitator signer’s `simulateContract` to run `transferWithAuthorization` against the target ERC‑20.
  - Parses ERC‑6492-wrapped signatures and handles both ECDSA (`v,r,s`) and bytes signatures.
  - Normalizes and maps known revert reasons (e.g. `AuthorizationAlreadyUsed`, `InvalidSignature`) into clear `invalidReason` messages.
- **Permit2**: `verifyPermit2` now simulates the final settlement call:
  - For EIP‑2612 gas sponsoring it simulates `settleWithPermit`; otherwise it simulates `settle`.
  - Decodes custom errors and maps known Permit2 revert reasons (e.g. `InvalidNonce`, `InvalidSignature`) into `invalidReason` on verification failure.
- **Signer and extension types**:
  - Extends the EVM facilitator signer and ERC‑20 approval gas sponsoring signer types with a `simulateContract` method, and wires them to viem’s `simulateContract`.
  - Updates unit and integration tests to use the new signer shape.
- **Facilitator site fixes**:
  - Updates `site/app/facilitator/index.ts` to:
    - Use dynamic imports for Aptos dependencies (`@aptos-labs/ts-sdk`, `@x402/aptos`) so Aptos remains fully optional.
    - Import gas-sponsoring extensions via the `@x402/extensions` namespace and register `EIP2612_GAS_SPONSORING` / `createErc20ApprovalGasSponsoringExtension` correctly.

## Tests

- TypeScript / EVM:
  - `pnpm --filter @x402/evm test -- test/unit/exact/facilitator.test.ts test/unit/v1/facilitator.test.ts`
    - This run exercises the new EIP‑3009 and Permit2 simulation paths (success and failure cases).
- TypeScript / Extensions:
  - `pnpm --filter @x402/extensions run build`
- TypeScript / Aptos:
  - `pnpm --filter @x402/aptos run build`
- Site:
  - `pnpm --filter site lint:check` (passes, including `site/app/facilitator/index.ts`)

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

ryanRfox (commented):
*(Superseded by approval review below — the author addressed all feedback in subsequent commits.)*

ruhil6789:
@ryanRfox 
Wire simulateContract into @x402/evm verify path: dry-run transferWithAuthorization / settle before settle() to reject payloads that would revert (expired auth, invalid signature, etc.). Includes revert reason decoding, extension signer simulateContract type, and tests.

typescript/packages/mechanisms/evm/src/exact/facilitator/permit2.ts
```diff
@@ -214,13 +223,161 @@ export async function verifyPermit2(
...
+    const revertReason = decodePermit2RevertReason(error);
+    return {
+      success: false,
+      revertReason,
+      rawError: error instanceof Error ? error.message : String(error),
+    };
+  }
+}
+
+function decodePermit2RevertReason(error: unknown): string {
```
ryanRfox, line 341: **Suggestion (non-blocking):** `decodePermit2RevertReason` and `decodeEip3009RevertReason` are structurally identical — same viem error unwrapping logic, same regex patterns. The only difference is which `KNOWN_REVERT_REASONS` map gets consulted via the normalize function.

Consider extracting a shared helper:
```typescript
function decodeRevertReason(error: unknown, knownReasons: Record<string, string>): string
```
This would live in a shared utils file and be called from both paths. Reduces ~60 lines of duplication and ensures future error-parsing improvements apply to both mechanisms.

typescript/packages/mechanisms/evm/src/signer.ts
```diff
@@ -71,6 +71,18 @@ export type FacilitatorEvmSigner = {
     functionName: string;
     args?: readonly unknown[];
   }): Promise<unknown>;
+  /**
+   * Simulate a contract call without broadcasting a transaction.
+   * Used by facilitators to perform eth_call-based transaction simulation
+   * during verification to detect potential settlement reverts.
+   */
+  simulateContract(args: {
```
ryanRfox, line 79: **Question:** `simulateContract` is added as a **required** field on `FacilitatorEvmSigner`. Any existing code that constructs a signer without it will now fail to compile.

Was this intentional? If so, the `minor` changeset bump is appropriate for the new capability, but it's worth calling out in the changeset description that this is a **breaking type change** for existing facilitator implementations.

Alternatively, making it optional (`simulateContract?:`) and having verify gracefully skip simulation when absent would preserve backwards compatibility. The verify path would still work for existing integrations — they'd just not get the simulation safety net until they upgrade their signer.

typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts
```diff
@@ -206,13 +216,130 @@ export async function verifyEIP3009(
...
+      BigInt(eip3009Payload.authorization.validBefore),
+      eip3009Payload.authorization.nonce,
+    ];
+
+    let callArgs: readonly unknown[];
+    if (isECDSA) {
+      const parsedSig = parseSignature(signature);
+      callArgs = [
+        ...baseArgs,
+        (parsedSig.v as number | undefined) || parsedSig.yParity,
```
ryanRfox, line 276: **Nit:** This uses `||` which would fall through on any falsy value (including `v=0`). While valid EIP-155 v values are 27/28 and viem's `parseSignature` guards against malformed input, using nullish coalescing (`??`) makes the intent explicit and is more defensive:
```typescript
(parsedSig.v as number | undefined) ?? parsedSig.yParity
```
Note: the existing settlement code at line ~421 uses the same `||` pattern, so this isn't a simulation-vs-settlement divergence — but both could benefit from `??`.

typescript/site/app/facilitator/index.ts
```diff
@@ -104,8 +114,13 @@ async function createFacilitator(): Promise<x402Facilitator> {
     .register("solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1", new ExactSvmScheme(svmSigner))
     .registerV1("solana-devnet" as Network, new ExactSvmSchemeV1(svmSigner));
 
-  // Optionally register Aptos if configured
+  // Optionally register Aptos if configured (dynamic import avoids requiring optional deps at load time)
```
ryanRfox, line 117: **Observation (non-blocking):** The Aptos dynamic import and `Extensions.*` namespace change are nice cleanups but unrelated to #1377. Might be worth a one-liner in the PR description noting these as incidental housekeeping so reviewers can focus on the simulation logic. Not worth splitting at this point.

ryanRfox (approved):
Great work responding to my feedback, @ruhil6789! The core simulation logic in `verifyEIP3009` and `verifyPermit2` is exactly what #1377 needed — pre-settlement dry-runs that catch doomed transactions before gas is spent. The revert reason decoding and test coverage are solid.

A few suggestions inline, none blocking. Approving as-is.

cb-heimdall:
Review Error for ryanRfox @ 2026-03-05 10:59:09 UTC

ruhil6789:
@ryanRfox 
Good call — the viem error unwrapping and regex logic really is identical between the two paths.

For this PR I kept the functions local to keep the diff focused on wiring up simulation, but I agree this should be a shared helper. I’ll follow up with a small refactor that introduces something like:

  decodeRevertReason(error: unknown, knownReasons: Record<string, string>)

in a shared utils module and have both `decodeEip3009RevertReason` and `decodePermit2RevertReason` delegate to it. That will de-duplicate the ~60 lines and keep any future error-parsing fixes consistent between mechanisms.

typescript/packages/mechanisms/evm/src/signer.ts
ruhil6789: > **Question:** `simulateContract` is added as a **required** field on `FacilitatorEvmSigner`. Any existing code that constructs a signer without it will now fail to compile.
> 
> Was this intentional? If so, the `minor` changeset bump is appropriate for the new capability, but it's worth calling out in the changeset description that this is a **breaking type change** for existing facilitator implementations.
> 
> Alternatively, making it optional (`simulateContract?:`) and having verify gracefully skip simulation when absent would preserve backwards compatibility. The verify path would still work for existing integrations — they'd just not get the simulation safety net until they upgrade their signer.

Great point on the type surface area.

Yes, making `simulateContract` required on `FacilitatorEvmSigner` was intentional so that every facilitator wired into the `@x402/evm` verify path actually gets the simulation safety net. Given that verify now relies on pre-settlement dry-runs to protect against doomed settlements, I felt it was better to make this a required capability rather than silently skipping simulation.

I agree this is a meaningful type change and I’ll update the changeset description to explicitly call out that:

- `FacilitatorEvmSigner` now requires a `simulateContract` method, and
- custom facilitator implementations need to be updated to provide it (typically by wiring to `viemClient.simulateContract`).

If you’d prefer a softer migration path (optional method + feature-detection with a log/metric when simulation is absent), I’m happy to follow up with a compatibility shim, but for now I’d like to keep the verify path assuming simulation is available.

typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts
ruhil6789: > **Nit:** This uses `||` which would fall through on any falsy value (including `v=0`). While valid EIP-155 v values are 27/28 and viem's `parseSignature` guards against malformed input, using nullish coalescing (`??`) makes the intent explicit and is more defensive:
> 
> ```ts
> (parsedSig.v as number | undefined) ?? parsedSig.yParity
> ```
> 
> Note: the existing settlement code at line ~421 uses the same `||` pattern, so this isn't a simulation-vs-settlement divergence — but both could benefit from `??`.

Agreed — `??` is the better fit here.

Even though viem’s `parseSignature` shouldn’t produce `v = 0` for valid inputs and the existing settle path used `||`, using nullish coalescing makes the intent explicit and removes the “falsy” edge case. I’ll update both the simulation and settlement call sites to:

  (parsedSig.v as number | undefined) ?? parsedSig.yParity

so they stay aligned and a bit more defensive.

typescript/packages/mechanisms/evm/src/signer.ts
ryanRfox: I’m happy with it as-is, so I defer to maintainers on these points. Again, nice work.

phdargen:
Thanks a lot for your contribution @ruhil6789. I just noticed that you started working on this in parallel while I was addressing it as part of a larger refactor in [#1474](https://github.com/coinbase/x402/pull/1474), which has now been merged.

The approach in [#1474](https://github.com/coinbase/x402/pull/1474) avoids adding a new simulateContract requirement on the signer (which would be a breaking change) and instead uses readContract via multicall. It also covers edge cases like the undeployed-wallet flow (ERC-6492 inner signature verification and implementation-dependent verifyTypedData for smart accounts) and makes re-simulation in settle optional

PR #610, opened 2025-11-06
NEAR Protocol integration

mikedotexe:
## Description

Took an honest, diligent stab at bringing NEAR Protocol support to x402, following the patterns for EVM and SVM, to feel familiar.

## What This Adds

We're implementing the complete stack for NEAR payments:

- **Facilitator verify/settle** - Server-side payment validation and on-chain execution
- **Borsh encoding/decoding** - Zero-dependency implementation of NEAR's serialization format
- **NEP-366 delegate actions** - Meta-transactions that let facilitators execute on behalf of users
- **NEP-413 signatures** - NEAR's signature verification standard

(Web wallet support not QA'ed or of primary focus in this PR)

### Browser Integration (`src/browser/`)

**`near.ts`** (126 lines) - Core NEAR wallet helpers
- Wallet connection flow
- Account ID validation
- Network detection (mainnet/testnet)

**`near_delegate.ts`** (238 lines) - NEP-366 delegate action support
- Create delegate actions for ft_transfer
- Handle NEAR wallet signing
- Construct proper action payloads

These get exported as namespaces in `src/browser/index.ts`:
```typescript
export * as near from "./near";
export * as near_delegate from "./near_delegate";
```

Keeps them cleanly separated from EVM/SVM code and prevents naming collisions.

### Facilitator Implementation (`src/facilitator/near/`)

**`borsh.ts`** (350 lines) - The heart of NEAR support for https://borsh.io binary serialization.

This zero-dependency Borsh encoder/decoder is specifically for NEP-366 delegate actions. 

Handles:
- `SignedDelegateAction` → `DelegateAction` → `Vec<Action>` decoding
- `FunctionCall` actions (method_name, args, gas, deposit)
- `Transfer` actions (for completeness, not used in v1)
- Little-endian integer encoding (u64, u128)
- Length-prefixed strings (u32 + UTF-8)

The variant tags (2 = FunctionCall, 3 = Transfer) come from `nearcore` (https://github.com/near/nearcore) and match the widely-used values in the NEAR ecosystem.

**`verify.ts`** (31 lines) - Payment verification logic
```typescript
export async function verify(
  header: PaymentHeader,
  requirements: PaymentRequirements
): Promise<VerifyResult>
```

Currently a stub that returns `{ isValid: false }` with TODOs marked. This is intentional - we want to land the infrastructure first, then flesh out the verification logic in follow-up PRs as we test against real NEAR facilitators.

**`settle.ts`** (26 lines) - On-chain settlement
```typescript
[...]

PR #1428, opened 2026-03-03
specs(exact): propose NEAR exact scheme for x402 v2 (spec-only)

mikedotexe:
## Summary
This PR proposes the `exact` payment scheme for NEAR in x402 **v2** as a specification-only submission.

## Scope (intentionally limited)
- Adds one new file: `specs/schemes/exact/scheme_exact_near.md`
- No SDK/runtime implementation changes
- No workflow/publishing changes
- No examples or e2e changes

## Why this shape
Following the new-chain guidance, this PR focuses only on spec design review before implementation.

## Key v2 alignment
- Uses v2 headers: `PAYMENT-REQUIRED`, `PAYMENT-SIGNATURE`, `PAYMENT-RESPONSE`
- Uses `x402Version: 2` and `PaymentRequirements.amount`
- Uses CAIP-style identifiers: `near:mainnet`, `near:testnet`
- Includes explicit facilitator safety/replay rules for delegated actions

## Review request
Please review the scheme design, payload shape, and verification/settlement safety constraints.
Once this is aligned, I will open a separate implementation PR in a single SDK (TypeScript) as requested.

PR #1459, opened 2026-03-05
Specs: add NEAR v2 exact payment scheme

mikedotexe:
## Description

Following the contributor guide for PR1, where we have `scheme_exact_near.md`.
I had previously done https://github.com/coinbase/x402/pull/610 but failed to follow the guideline, and this is for v2.

Note that NEAR accounts can have multiple keys, either `secp256k1` or `ed25519`, though some documentation may suggest it's only the edwards curve. I bring this up because there's likely prior art for the former curve and it might prove to streamline integration knowing this.


arjun215-eng:
Great to see NEAR getting a proper v2 spec — the explicit dual-curve support (ed25519 + secp256k1) is a nice touch. A few things worth considering:

**1. Block height expiry is implementation-defined and could diverge**
Replay protection anchors on `delegate_action.max_block_height > current chain height`, and `maxTimeoutSeconds` is the x402 timeout — but there's no defined mapping between them. Different facilitator implementations could accept or reject the same payload depending on their conversion factor. Worth specifying a concrete formula so behaviour is consistent across implementations.

**2. `payer` in failure `SettlementResponse` feels odd**
The failure example returns `"payer": "alice.testnet"` alongside `success: false`. At failure time the facilitator may not have confirmed the on-chain identity of the payer — it only has what the client claimed in the payload. Worth aligning on whether `payer` should be omitted or flagged as unverified in failure cases.

**3. Relationship with #1428**
Both PRs target `main` and add `scheme_exact_near.md` — if this supersedes #1428, worth closing that one and noting it here so reviewers know which to focus on.

cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

mikedotexe:
> **1. Block height expiry is implementation-defined and could diverge** Replay protection anchors on `delegate_action.max_block_height > current chain height`, and `maxTimeoutSeconds` is the x402 timeout — but there's no defined mapping between them. Different facilitator implementations could accept or reject the same payload depending on their conversion factor. Worth specifying a concrete formula so behaviour is consistent across implementations.

Good call here. I've double-checked that **1 second** is a reasonable block time, moving away from height as you've pointed out.

Screenshot 2026-03-06 at 5 51 04 AM

from https://docs.near.org/integrations/exchange-integration#blocks-and-finality

and just to lay out the options for finality (the NEAR teams I've worked on use `optimistic` frequently)

Screenshot 2026-03-06 at 5 52 54 AM

from https://docs.near.org/api/rpc/setup#using-finality-param

---

> **2. `payer` in failure `SettlementResponse` feels odd** The failure example returns `"payer": "alice.testnet"` alongside `success: false`. At failure time the facilitator may not have confirmed the on-chain identity of the payer — it only has what the client claimed in the payload. Worth aligning on whether `payer` should be omitted or flagged as unverified in failure cases.

Indeed, and we certainly don't want the `payer` to be flimsy like I had it. 

---

Address both of these in d5adedb105b2655af36f3104cde1d97b9d69b7bd and closed my previous PR, in favor of this one. Welcoming more feedback, which is quite useful, thanks.

phdargen:
Thanks a lot for putting this together @mikedotexe! 

Notified the core team to review. Please be aware that we currently have a lot of pending/incoming requests, so it might take a bit until we can get back to you, thanks for your patience 🙏

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -0,0 +1,245 @@
...
+```json
+{
+  "scheme": "exact",
+  "network": "near:testnet",
+  "amount": "1000000",
+  "asset": "usdc.testnet",
+  "payTo": "merchant.testnet",
+  "maxTimeoutSeconds": 60,
+  "extra": {
+    "relayerId": "x402-relayer.testnet"
```
phdargen: Suggest to align with other network's convention
```suggestion
    "feePayer": "x402-relayer.testnet"
```

phdargen: Actually is this the `relayerId` needed at all for the client to create the signed payload? 

If not, should be dropped entirely

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -0,0 +1,245 @@
...
+
+- `delegate_action.receiver_id` MUST equal `PaymentRequirements.asset`.
+- Parsed `ft_transfer.args.receiver_id` MUST equal `PaymentRequirements.payTo`.
+- Parsed `ft_transfer.args.amount` MUST equal `PaymentRequirements.amount` exactly.
+- Attached deposit MUST be exactly `1` yoctoNEAR.
+- Sponsored gas MUST be within facilitator policy bounds.
+
+### 8. Access-Key Permission Safety
+
+- Facilitator MUST ensure the delegate public key has permission compatible with the intended call.
```
phdargen: This seems a bit too vague, could you please elaborate these checks a bit more to ensure implementation consistency?

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -0,0 +1,245 @@
...
+- The only allowed action kind is `FunctionCall`.
+- `FunctionCall.methodName` MUST be `ft_transfer`.
+- No extra delegated actions are permitted.
+
+### 7. Token Transfer Intent and Exactness
+
+- `delegate_action.receiver_id` MUST equal `PaymentRequirements.asset`.
+- Parsed `ft_transfer.args.receiver_id` MUST equal `PaymentRequirements.payTo`.
+- Parsed `ft_transfer.args.amount` MUST equal `PaymentRequirements.amount` exactly.
+- Attached deposit MUST be exactly `1` yoctoNEAR.
```
phdargen, line 183: Is this deposit coming from the client or also facilitator sponsored?

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -0,0 +1,245 @@
...
+      "relayerId": "x402-relayer.testnet"
+    }
+  },
+  "payload": {
+    "signedDelegateAction": "AQAAA...<base64>..."
+  }
+}
+```
+
+## Facilitator Verification Rules (MUST)
```
phdargen, line 128: Verification checks are so far only structural payload checks but should also check if the tx would actually succeed onchain, ideally with a full tx simulation. Or if thats not possible, at least some targeted checks like sufficient client token balance, nonce unused etc

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -0,0 +1,245 @@
...
+- Signature verification MUST use the matching curve for the declared key type.
+- Signature MUST verify against the exact encoded `delegate_action` bytes and the included public key.
+
+### 5. Replay and Expiry Protection
+
+- Facilitator MUST compute timeout bounds using the deterministic timeout mapping in this specification.
+- `remainingBlocks = delegate_action.max_block_height - current_block_height`.
+- Facilitator MUST reject if `remainingBlocks <= 0` (for example `delegate_action_expired`).
+- Facilitator MUST reject if `remainingBlocks > timeoutBlocks` (for example `delegate_action_timeout_window_exceeds_maxTimeout`).
+- Facilitator MUST perform nonce replay protection for `(sender_id, public_key, nonce)`.
```
phdargen: Could you please explain in more detail why this is needed and how the facilitator would do this in practice?

Would this imply a stateful facilitator?

x402 is designed as a stateless protocol. We recently made one exception for SVM to patch a security vulnerability, see https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact_svm.md#duplicate-settlement-mitigation-recommended.

If NEAR could use a similar approach as SVM with a short-term, in-memory cache of transaction payloads, that would be acceptable but should be well justified in a dedicated section of the spec.

If persistent storage would be needed, this would warrant a deeper discussion with the core team

phdargen:
Hi @mikedotexe, thanks a lot for putting this together! Made a first pass and the spec is well written and structured. Please have a look at my comments above

phdargen:
Would also be great if we could get someone from the NEAR team to review this as well

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Agreed, it can be dropped entirely. Here's a good link from NEAR docs on the meta-transaction layer: https://docs.near.org/web3-apps/tutorials/meta-transactions#gating-the-relayer

see commit 039d5e5bae587a9d78b662050e0a52c99f4d4b84

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Sure, and I could have done a better job of expanding here. Perhaps briefly to cover some fundamentals of NEAR accounts will help. I own `mike.near` and, like all other accounts, I can add two types of access keys:

1. Function-call access key: a cryptographic key (either `ed25519` or `secp256k1`) that is given an `allowance` in NEAR to be used to call smart contracts. Designates which receiving contract can be called and by which methods (or all methods). Importantly, these keys cannot attach native NEAR in a function call to a smart contract.
2. Full-access key: has sudo control over the account, essentially. Can call smart contracts with an attached deposit of NEAR, add/delete an access key, even delete the account itself and send the remaining funds to a recipient.

So, when my team wrote the fungible and non-fungible token standards in 2020, we ensured that some methods must be called by attaching the smallest amount of NEAR possible, 1 yoctoNEAR. A smart contract can confirm this attached deposit and know that the access key calling the function is a full-access key.

---

I've tried to flesh this out a bit in commit bff012179697ccbf5d58b1cc15b3c936a2463f4f

Also added a section "Duplicate Settlement Mitigation" proactively to hopefully help clarify.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Great question, and did digging in the docs and of NEAR meta-transactions. Clarified that in `nearcore` as well:

https://github.com/near/nearcore/blob/02cfad715bfec5aea43d2146b851c01c4236765d/runtime/runtime/src/actions.rs#L510-L514

So it looks like it's the facilitator. If `mike.near` wants to use this meta-transaction flow, I would sign a [`SignedDelegateAction`](https://github.com/near/NEPs/blob/003e589e6aba24fc70dd91c9cf7ef0007ca50735/neps/nep-0366.md#specification) declaring that I wish to attach 1 yoctoNEAR deposit, but it's the facilitator/relayer that, when deciding to relay it, also decides to attach this amount.

pushed a second commit to hopefully clarify this here: cf3a92096504fb320fe591fe4ac3dc9f544e86ab

specs/schemes/exact/scheme_exact_near.md
mikedotexe: There is no full tx simulation, so added a section "Implementing Verification with NEAR RPC" about how to use the RPC to do checks again the access key's nonce, the fungible token balance, the existence of the FT smart contract, and so on.

I work at FastNear and here's an example of an access key on `mike.near` and its nonce using our RPC infra:
https://docs.fastnear.com/rpc/account/view-access-key?network=mainnet&responseView=expanded&responseFind=nonce&requestFinality=final&account_id=mike.near&public_key=ed25519%3ACFpatUxXwFcpswHeNXEQz5q9WvG7X5ZTg1LczpEZxUV

(note: when an access key is first created, the initial nonce is the block height times a million, to abate double-signing if someone removes a pubkey and then adds it again, hence the high number you see from that link)

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Reviewed the link there and added two sections "Vulnerability" and "Recommended Mitigation"
Persistence isn't required, if I'm reading this right

mikedotexe:
> Would also be great if we could get someone from the NEAR team to review this as well

I've put the word out in the NEAR Dev telegram channel and will circulate as well with my team, one of the members wrote the NEAR smart contract runtime, will see if he has corrections. Feel fairly confident so far, hopefully in a better place. And happy to iterate or address other concerns. Thanks for the very targeted questions, appreciate it

mikedotexe:
And, given that we have some terminology like `receipt` in here, I'll attempt to explain something plainly with regards to NEAR's asynchronous nature.
NEAR launched as an async blockchain from the start, meaning that when `mike.near` sends a transaction, it can have one or more "Actions" associated with it. FunctionCall, AddKey, DeleteKey, Transfer, etc. If my transaction had, say, a couple Actions, these pass through validation going into the protocol (is this access key associated with this account and can it do these Actions) and then each Action is turned into a `receipt`. A receipt basically means it will get executed, in some shard, in the chain. 
In typical blockchain architecture like Ethereum, I think of it like "global state is frozen while this transaction updates state of the contract" but in NEAR that's not the case. Validators are rotating through different shards which are in charge of a subset of accounts, so receipt execution is asynchronous. That's the mental model I'm holding as I think about this, and hopefully that helps. (And is perhaps a reason why we don't have deterministic tx simulation)

phdargen:
This looks great @mikedotexe, thanks for the update!

Appreciate all the additional details and nuances you explained. As I am admittedly not that familiar with NEAR, this was really helpful

phdargen (approved):

phdargen:
Looking forward to a PR with a reference implementation following the specs, this will be a highly valuable addition! Feel free to reach out and let me know if you have any questions, happy to help along the way

Issue #1458, opened 2026-03-05
Allow SPL Token Fee Payment Instructions for Solana Smart Wallets


maxsch-xmint:
## Allow SPL Token Fee Payment Instructions for Solana Smart Wallets

> **Note:** The use cases described in this issue assume the external wallet involved is a treasury wallet controlled by the wallet provider. The proposed validation constraints are designed with this in mind.

### Context

[[PR #1380](https://github.com/coinbase/x402/pull/1380)] adds support for Solana Smart Wallets, starting with a Swig-based implementation, and lays the foundation for easily extending support to new wallet providers.

### Problem

Solana Smart Wallets can support a "pay fees with SPL tokens" model, where the wallet provider charges its fee in SPL tokens (e.g. USDC) rather than SOL. This requires an additional instruction in the transaction: an SPL token transfer from the user's wallet to an external wallet.

This is a common pattern for achieving a gasless UX on Solana — the user never needs to hold SOL, and all costs are settled in the same token used for the payment itself.

### Proposal

We'd like to request that the facilitator allows this additional SPL token transfer instruction, subject to strict constraints:

- The transfer is an SPL token transfer (e.g. USDC)
- The **sender** is the user's wallet
- The **recipient** is the treasury wallet (i.e. not the fee payer, and not any merchant address)

These constraints make the instruction narrowly and verifiably scoped to a wallet provider fee, with no value flowing to the merchant or the facilitator.

### Related: Rent Sponsorship Instruction

In some cases, a second extra instruction may also be present: a SOL transfer from the treasury wallet to the user's wallet, used to pre-fund the rent exemption cost when a new on-chain account needs to be created. The constraints here would be:

- The transfer is SOL
- The **sender** is the treasury wallet (neither the fee payer nor the user's wallet)
- The **recipient** is the user's wallet address

This is the complementary direction of the same treasury relationship — the wallet provider briefly funds the user to cover account initialization, which is then offset by the SPL fee payment above.

### Why This Is Safe

We understand that the current model intentionally treats account funding as an out-of-band process, partly in response to past griefing attacks where malicious merchants drained facilitators by forcing the creation of large numbers of new accounts.

This proposal is different in a few key ways:
- Both instructions involve an **external wallet** (in practice, a treasury controlled by the wallet provider), not the merchant or facilitator
- The constraints above make it trivially verifiable that no value flows unexpectedly away from the user or fee payer
- The logic is narrow and deterministic — it can be validated at the instruction level before signing

[...]

PR #829, opened 2025-12-19
docs(svm): specify simulation-based smart wallet verification for exact scheme

notorious-d-e-v:
## Description

Updates the SVM `exact` scheme spec to document **outcome-based payment verification** and a **Sponsor Acceptance Policy**, aligned with the simulation-based smart wallet verification merged in #1527.

This supersedes the earlier terminology draft on this branch and rebases onto current `main` (was conflicting). It is the spec-side follow-up requested at the merge of #1527.

### What changed

- **Outcome-based framing.** The scheme is defined by the on-chain payment *outcome*, not a fixed instruction layout. Fee sponsorship and risk controls are separated out into a Sponsor Acceptance Policy (the sponsor may be the merchant or a third-party facilitator).
- **Normative Exact Payment Outcome Definition** — required transfer (top-level *or* CPI inner instruction), exactly-one matching transfer, and overpayment tolerance (`>=` the required amount).
- **Sponsor Acceptance Policy** — fee payer isolation, Address Lookup Table visibility (resolve-or-reject, fail closed), compute budget caps (defaults 400k CU / 50k µlamports), program allowlist, and simulation-based rejection.
- **Two-path reference implementation** — static fast path (standard wallets) + simulation-based smart wallet path; path selection (only recoverable layout failures fall through, so semantic rejections keep their real reason); post-settlement TOCTOU verification (confirmed inner instructions, balance-delta fallback).
- **Security invariants table (I1–I7).**
- **Static path cap raised 6 → 7** to keep Phantom (up to 3 Lighthouse assertions) on the fast path.

### Notes

- Spec-only change; no code changes.
- Default program allowlist documented: Squads Multisig v4, Squads Smart Account, Swig (legacy + v2), SPL Governance, Metaplex Core, and Lighthouse; ComputeBudget and Memo are category-exempt.
- Go/Python parity for the implementation is tracked separately.

Issue #828, opened 2025-12-18
Solana mainnet v2 transactions fail when signed with 3rd party wallets (e.g., phantom, backpack, coinbase)


Max-the-dev:
**Context:**
  - Solana devnet payments work fine
  - Base mainnet/testnet payments work fine
  - Only Solana mainnet fails
      - _transaction_simulation_failed response from x402 verify endpoint_ 
      - it appears the facilitator does not work unless using the official signer (programmatic signer with private keys)

**Additional Context:** 
  - This flow was fully functional when on v1. 
  - Key question: Did the signing scheme change between the two versions?

**Discovery:**

  While debugging, observed that Phantom wallet modifies transactions on mainnet (not devnet). Specifically, it appears
   to inject the Lighthouse program for user protection:

  - Transaction before Phantom signs: 7 accounts
  - Transaction after Phantom signs: 8 accounts (includes Lighthouse program)

 **Hypothesis:**

  The v2 SDK flow may not accommodate third-party wallets that modify transactions before signing. The signature
  is valid for the modified transaction, but CDP simulation may be failing because:
  - It's validating against an expected transaction structure
  - Or reconstructing/simulating a transaction that doesn't match what was actually signed

  **Questions:**
  1. Does CDP's Solana verification expect a specific transaction structure?
  2. Can the flow accommodate wallet-modified transactions ?
  3. Any known issues or guidance for Phantom on mainnet with v2?

Max-the-dev:
Example: phantom adds a lighthouse address to the signed tx 
which increases the total length to 8 vs. the expected 7 in the SDK

phdargen:
Thanks a lot for your investigation @Max-the-dev, summoning @CarsonRoscoe and @notorious-d-e-v

Max-the-dev:
Confirmed the rest of the implementation is correct by bypassing the phantom signature. 

Created script following this example: 
https://github.com/coinbase/x402/blob/cbfecfaa97ac9b52e4857b8040db30a11df7eba6/examples/typescript/clients/fetch/index.ts

Hit my article purchase endpoint directly - success 
Hit Donate and Tip endpoints directly - success 

Let me know if you want to see logs. I have a ton.

notorious-d-e-v:
Thanks for reporting @Max-the-dev ! 

This is specific to Phantom adding a lighthouse instruction to the transaction.

The solana `exact` scheme implementation expects three instructions only (atm):
1. ComputeBudgetLimit
[...]

Issue #646, opened 2025-11-16
RFC: Enable Deadline Validation and Smart Wallet Support in `scheme_exact_svm`


tenequm:
## Problem

The current `scheme_exact_svm` specification has critical limitations blocking ecosystem adoption:

**1. Non-Functional `maxTimeoutSeconds` Parameter**

- `maxTimeoutSeconds` cannot be enforced on Solana due to blockhash expiry (~80-90 seconds hard limit)
- Production failures reported: settlements fail after 1-2 minutes regardless of configured timeout
- **Requested by @CarsonRoscoe** as critical missing feature
- Long-running operations (AI inference, data processing) blocked from using x402

**2. Smart Wallet Applications Blocked**

- Current spec requires exactly 3 instructions with `TransferChecked` as last instruction
- Smart wallets (Squads, SWIG, SPL Governance) execute transfers via **Cross-Program Invocation (CPI)**
- CPI transfers appear as smart wallet program calls, not direct `TransferChecked` instructions
- **Enterprise adoption blocked**: "Every major business needs smart wallets" (Corbits team)
- Squads Protocol users cannot pay (300+ teams, $10B+ assets)

**3. Wallet Innovation Blocked**

- Phantom's Lighthouse assertions add 5+ instructions, breaking the 3 limit
- Any future wallet security feature that adds instructions is incompatible
- Protocol cannot evolve with wallet ecosystem

**Root cause:** Spec verifies transaction **structure** instead of payment **outcome**.

---

## Proposed Solution

Replace rigid instruction validation with **outcome-based verification** using two paths:

```
                    Transaction received
                            │
                            ▼
              ┌─────────────────────────────┐
              │  Matches simple pattern?    │
              │  • 3 instructions         │
              │  • Last ix = TransferChecked│
              │  • SPL Token program only   │
              └─────────────────────────────┘
                      │           │
                     YES         NO
                      │           │
                      ▼           ▼
                   STATIC    SIMULATION
                  (existing)    (new)
                    ~5ms        ~50ms
```

### Path 1: Static Verification (Fast Path)

For simple transfers matching the existing pattern:
- 3 instructions
- Last instruction = `TransferChecked`
- Token program = SPL Token (`TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA`)
- Existing whitelist rules apply
- ~5ms verification

**This path is unchanged from current spec.**

### Path 2: Simulation Verification (General Case)

[...]

PR #1527, opened 2026-03-09
feat(svm): add simulation-based smart wallet verification

BranchManager69:
## Description

Adds wallet-agnostic smart wallet verification to the SVM facilitator (TypeScript). Verifies payment outcomes by inspecting top-level instructions and the CPI trace from transaction simulation, rather than parsing each wallet type's proprietary instruction format.

Works for any smart wallet program (Squads, Swig, SPL Governance, Crossmint, future wallets) that executes `TransferChecked` via CPI. No per-wallet code required. This approach has been running in production at [Dexter](https://dexter.cash) and is the first implementation to verify and settle smart wallet transactions on-chain.

## Motivation

The existing SVM facilitator rejects any transaction that doesn't match the exact 3-6 instruction layout (ComputeLimit + ComputePrice + TransferChecked + optional Lighthouse/Memo). Smart wallet programs wrap transfers inside their own instructions (e.g., Squads `vaultTransactionExecute`, Swig `SignV2`), so the facilitator sees an unknown program at position [2] and rejects with `no_transfer_instruction`.

This is a fundamental limitation of positional instruction validation. Every smart wallet has a different instruction format, but they all do the same thing: execute a `TransferChecked` via CPI.

## Approach

`ExactSvmScheme.verify()` now has two paths:

**Path 1 (Static)**: Existing positional instruction validation for standard wallets. Unchanged. Standard wallets keep this fast path; simulation is only a fallback for transactions the current verifier cannot structurally understand.

**Path 2 (Simulation)**: When Path 1 rejects and `enableSmartWalletVerification` is set:

1. **Fee payer isolation**: the facilitator's fee payer must not appear in any instruction's accounts. If the fee payer is never referenced, the Solana runtime cannot authorize it for SOL transfers, token approvals, ATA creation, account closes, or any other drain vector.
2. **Compute budget caps**: operator-configurable CU limit and priority fee caps on ComputeBudget instructions. Bounds the facilitator's fee exposure.
3. **Simulate with `innerInstructions: true`**: the facilitator already calls `simulateTransaction` as its final verification step. Path 2 uses the same RPC method with `innerInstructions: true` to get the full CPI trace. No additional round-trip.
4. **Match exactly one transfer**: require exactly one `TransferChecked` across top-level instructions and the CPI trace, matching payment requirements (correct mint, destination ATA derived from `payTo`, amount exactly equals required).

```mermaid
flowchart TD
    TX["Transaction arrives"] --> DECODE["Decode"]
    DECODE --> P1{"Path 1: Static layout check"}
    P1 -->|"Layout matches"| SIM1["Sign + Simulate"]
    SIM1 --> PASS1["VERIFIED"]
    P1 -->|"Unknown program / wrong layout"| SW{"Smart wallet enabled?"}
    SW -->|"No"| REJECT["REJECTED"]
    SW -->|"Yes"| FPI{"Fee payer isolated?"}
    FPI -->|"No"| REJECT
    FPI -->|"Yes"| CB{"Compute budget OK?"}
    CB -->|"No"| REJECT
    CB -->|"Yes"| SIM2["Simulate + innerInstructions"]
    SIM2 -->|"Fails"| REJECT
    SIM2 -->|"Succeeds"| EXTRACT["Extract TransferChecked"]
    EXTRACT --> MATCH{"Exactly 1 match?"}
    MATCH -->|"No"| REJECT
    MATCH -->|"Yes"| PASS2["VERIFIED"]
```

## Security model

| Property | How it's enforced |
|---|---|
| Facilitator funds protected | Fee payer isolation: fee payer never appears in any instruction's accounts list |
| Correct payment recipient | Destination ATA derived from `payTo` + mint, matched against observed transfers |
| Correct payment amount | Amount from `TransferChecked` checked against requirements |
| Correct token | Mint from `TransferChecked` checked against requirements |
| Fee exposure bounded | Operator-configurable compute unit and priority fee caps |
| No double-payment | Exactly one matching transfer required; multiple matches rejected |
| Transaction viability | Simulation proves the transaction would succeed on-chain |

## Production data

Tested against ecosystem facilitators using a real Squads multisig vault transaction, both verify and settle. No other facilitator currently passes.

## Changes

| File | Change |
|------|--------|
| `svm/src/signer.ts` | Add optional `simulateTransactionWithInnerInstructions` to `FacilitatorSvmSigner`; implement in default factory |
| `svm/src/exact/facilitator/scheme.ts` | Add `ExactSvmSchemeOptions` with smart wallet config; refactor `verify()` into dual-path (static + simulation) |
| `svm/src/exact/facilitator/smartWalletVerification.ts` | **New**: `assertFeePayerIsolated`, `validateComputeBudgetLimits`, `extractTransfersFromInnerInstructions`, `verifySmartWalletTransaction` |
| `svm/src/index.ts` | Re-export smart wallet helpers and types |
| `svm/test/unit/smartWalletVerification.test.ts` | **New**: 19 unit tests |
| `svm/test/unit/smartWalletFallback.test.ts` | **New**: 3 integration tests (fallback path verify) |

Go and Python implementations are left for follow-up. This PR is intentionally TypeScript-only to validate the upstream architecture first. If maintainers agree with outcome-based verification as the direction, Go/Python parity can follow without replicating wallet-specific parsers across all SDKs. This PR targets non-ALT smart-wallet transaction shapes first; ALT-aware inner instruction resolution can follow with concrete test cases.

## Usage

```typescript
import { ExactSvmScheme } from "@x402/svm";

const scheme = new ExactSvmScheme(signer, undefined, {
  enableSmartWalletVerification: true,
  smartWalletMaxComputeUnits: 400_000,            // optional, default 400k
  smartWalletMaxPriorityFeeMicroLamports: 50_000,  // optional, default 50k
});
```

## Tests

- [x] All existing tests pass
- [x] 19 new smart wallet verification tests pass (security helpers, adversarial cases, extraction, fallback path integration)
- [x] ESM and CJS builds succeed
- [x] No new TypeScript errors introduced

## Checklist

- [x] My commits are signed

### DIFF
--- a/typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
@@ -37,9 +37,116 @@ import {
   getTokenPayerFromTransaction,
   transactionMessageHash,
 } from "../../utils";
+import { verifySmartWalletTransaction, verifyPostSettlement } from "./smartWalletVerification";
+
+/**
+ * Default allowed smart wallet program addresses.
+ * Only these programs can reach Path 2 (simulation-based verification).
+ * Operators can override via smartWalletAllowedPrograms in options.
+ */
+const DEFAULT_SMART_WALLET_ALLOWED_PROGRAMS = [
+  "SQDS4ep65T869zMMBKyuUq6aD6EgTu8psMjkvj52pCf", // Squads Multisig v4
+  "SMRTzfY6DfH5ik3TKiyLFfXexV8uSG3d2UksSCYdunG", // Squads Smart Account
+  "SWiGmQedKzMz1tiTqoJCWeGDnGXfNBp2PkXLkpCAtQo", // Swig
+  "GovER5Lthms3bLBqWub97yVrMmEogzX7xNjdXpPPCVZw", // SPL Governance
+  "CoREENxT6tW1HoK8ypY1SxRMZTcVPm7R94rH4PZNhX7d", // Metaplex Core
+  LIGHTHOUSE_PROGRAM_ADDRESS, // Phantom's wallet-protection assertions (see #2097)
+];
+
+/**
+ * Which verification path produced a successful result.
+ * Returned by the internal _verify so settle() knows whether post-settlement
+ * TOCTOU verification is required, without re-deriving it from the transaction.
+ */
+type VerificationPath = "static" | "smartWallet";
+
+/**
+ * Internal verify result that also reports which path succeeded.
+ * verificationPath is null when verification failed.
+ */
+type VerifyResult = {
+  response: VerifyResponse;
+  verificationPath: VerificationPath | null;
+};
+
+/**
+ * Path 1 failure reasons that indicate a transaction layout a standard-wallet
+ * parser could not handle — extra/unknown instructions, unexpected counts, or
+ * a missing positional transfer. These are the only cases where falling through
+ * to Path 2 (simulation) can legitimately recover the payment, because the
+ * transfer may simply be wrapped in a smart-wallet CPI.
+ *
+ * Reasons NOT in this set are semantic rejections (amount/mint/recipient/memo
+ * mismatch, self-spend, failed simulation). Those describe a transaction that
+ * is genuinely invalid for this payment, so Path 2 must not run — doing so would
+ * mask the real reason behind a misleading smart_wallet_* error code.
+ */
+const LAYOUT_RECOVERABLE_REASONS = new Set<string>([
+  "invalid_exact_svm_payload_transaction_instructions_length",
+  "invalid_exact_svm_payload_no_transfer_instruction",
+  "invalid_exact_svm_payload_unknown_fourth_instruction",
+  "invalid_exact_svm_payload_unknown_fifth_instruction",
+  "invalid_exact_svm_payload_unknown_sixth_instruction",
+  "invalid_exact_svm_payload_unknown_optional_instruction",
+  "invalid_exact_svm_payload_transaction_instructions_compute_limit_instruction",
+  "invalid_exact_svm_payload_transaction_instructions_compute_price_instruction",
+]);
+
+/**
+ * Configuration options for ExactSvmScheme.
+ */
+export type ExactSvmSchemeOptions = {
+  /**
+   * Enable simulation-based smart wallet verification.
+   * When enabled, transactions rejected by the static validation path
+   * (unknown programs, wrong instruction count) are re-verified using
+   * simulation inner instruction analysis. Works for any smart wallet
+   * program (Squads, Swig, SPL Governance, etc.) without per-wallet parsers.
+   *
+   * Default: false (only standard wallet transactions are accepted)
+   */
+  enableSmartWalletVerification?: boolean;
+
+  /**
+   * Maximum compute units allowed for smart wallet transactions.
+   * Smart wallet programs need more CU for CPI overhead.
+   * Only applies when enableSmartWalletVerification is true.
+   *
+   * Default: 400,000
+   */
+  smartWalletMaxComputeUnits?: number;
+
+  /**
+   * Maximum priority fee in microlamports for smart wallet transactions.
+   * Only applies when enableSmartWalletVerification is true.
+   *
+   * Default: 50,000
+   */
+  smartWalletMaxPriorityFeeMicroLamports?: number;
+
+  /**
+   * Allowed smart wallet program addresses for Path 2 verification.
+   * Only transactions whose top-level non-ComputeBudget instruction invokes
+   * a program in this list will be accepted through the simulation path.
+   * Prevents unknown/malicious programs from reaching CPI verification.
+   *
+   * Default: Squads Multisig v4, Squads Smart Account, Swig, SPL Governance, Metaplex Core
+   */
+  smartWalletAllowedPrograms?: string[];
+};
 
 /**
  * SVM facilitator implementation for the Exact payment scheme.
+ *
+ * Dual-path verification:
+ *
+ * Path 1 (Static): Strict positional instruction validation for standard wallets.
+ *   Fast, preserves existing behavior.
+ *
+ * Path 2 (Simulation): Outcome-based verification for smart wallets.
+ *   When Path 1 rejects a transaction and smart wallet verification is enabled,
+ *   falls back to simulation-based validation that inspects CPI inner instructions.
+ *   Works for any wallet program that executes TransferChecked via CPI.
  */
 export class ExactSvmScheme implements SchemeNetworkFacilitator {
   readonly scheme = "exact";
@@ -48,17 +155,39 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
   private readonly settlementCache: SettlementCache;
 
   /**
-   * Creates a new ExactSvmFacilitator instance.
+   * Creates a new ExactSvmScheme instance.
    *
-   * @param signer - The SVM RPC client for facilitator operations
+   * @param signer - The SVM signer for facilitator operations
    * @param settlementCache - Optional shared settlement cache (one is created if omitted)
-   * @returns ExactSvmFacilitator instance
+   * @param options - Optional configuration for smart wallet verification
    */
   constructor(
     private readonly signer: FacilitatorSvmSigner,
     settlementCache?: SettlementCache,
+    private readonly options?: ExactSvmSchemeOptions,
   ) {
     this.settlementCache = settlementCache ?? new SettlementCache();
+
+    if (this.options?.enableSmartWalletVerification) {
+      // fetchAddressLookupTables is required too: assertFeePayerIsolated can't
+      // inspect ALT-resolved accounts without it, so an ALT-using wallet would
+      // otherwise fail at verify time rather than at construction.
+      const required = [
+        "simulateTransactionWithInnerInstructions",
+        "getConfirmedTransactionInnerInstructions",
+        "getTokenAccountBalance",
+        "fetchAddressLookupTables",
+      ] as const;
+
+      for (const method of required) {
+        if (typeof (this.signer as Record<string, unknown>)[method] !== "function") {
+          throw new Error(
+            `enableSmartWalletVerification requires ${method} on the signer. ` +
+              `Use toFacilitatorSvmSigner() which provides all required methods.`,
+          );
+        }
+      }
+    }
   }
 
   /**
@@ -74,9 +203,11 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
     const addresses = this.signer.getAddresses();
     const randomIndex = Math.floor(Math.random() * addresses.length);
 
-    return {
-      feePayer: addresses[randomIndex],
-    };
+    const extra: Record<string, unknown> = { feePayer: addresses[randomIndex] };
+    if (this.options?.enableSmartWalletVerification) {
+      extra.features = { smartWalletSupported: true };
+    }
+    return extra;
   }
 
   /**
@@ -101,30 +232,191 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
     payload: PaymentPayload,
     requirements: PaymentRequirements,
   ): Promise<VerifyResponse> {
+    const { response } = await this._verify(payload, requirements);
+    return response;
+  }
+
+  /**
+   * Settles a payment by submitting the transaction.
+   * Ensures the correct signer is used based on the feePayer specified in requirements.
+   *
+   * @param payload - The payment payload to settle
+   * @param requirements - The payment requirements
+   * @returns Promise resolving to settlement response
+   */
+  async settle(
+    payload: PaymentPayload,
+    requirements: PaymentRequirements,
+  ): Promise<SettleResponse> {
+    const exactSvmPayload = payload.payload as ExactSvmPayloadV2;
+
+    const { response: valid, verificationPath } = await this._verify(payload, requirements);
+    if (!valid.isValid) {
+      return {
+        success: false,
+        network: payload.accepted.network,
+        transaction: "",
+        errorReason: valid.invalidReason ?? "verification_failed",
+        payer: valid.payer || "",
+      };
+    }
+
+    // Decode the transaction to compute the message hash used as the cache key.
+    // Must remain synchronous (before any await) so concurrent settle calls for
+    // the same payment are caught before any async work begins.
+    const decodedTx = decodeTransactionFromPayload(exactSvmPayload);
+
+    // Duplicate settlement check keyed on message hash (immune to mutable fee-payer sig at slot 0).
+    const txKey = transactionMessageHash(decodedTx);
+    if (this.settlementCache.isDuplicate(txKey)) {
+      return {
+        success: false,
+        network: payload.accepted.network,
+        transaction: "",
+        errorReason: "duplicate_settlement",
+        payer: valid.payer || "",
+      };
+    }
+
+    // Settlements verified through Path 2 (smart wallet) require post-settlement
+    // verification to defend against TOCTOU. _verify reports the path directly,
+    // so we no longer re-decode the transaction to infer it.
+    const isSmartWalletSettlement = verificationPath === "smartWallet";
+
+    // For smart wallet settlements: record destination ATA balance before sending.
+    // Used as fallback verification if getTransaction has indexing lag.
+    // Try both SPL Token and Token-2022 programs — the payment may use either.
+    let balanceBefore: bigint | null = null;
+    let balanceBeforeTokenProgram:
+      | typeof TOKEN_PROGRAM_ADDRESS
+      | typeof TOKEN_2022_PROGRAM_ADDRESS
+      | null = null;
+    if (isSmartWalletSettlement && typeof this.signer.getTokenAccountBalance === "function") {
+      for (const tokenProgram of [TOKEN_PROGRAM_ADDRESS, TOKEN_2022_PROGRAM_ADDRESS]) {
+        try {
+          const [destinationAta] = await findAssociatedTokenPda({
+            mint: requirements.asset as Address,
+            owner: requirements.payTo as Address,
+            tokenProgram: tokenProgram as unknown as Address,
+          });
+          const balance = await this.signer.getTokenAccountBalance(
+            destinationAta.toString(),
+            requirements.network,
+          );
+          if (balance !== null) {
+            balanceBefore = balance;
+            balanceBeforeTokenProgram = tokenProgram;
+            break; // Use whichever ATA has a balance (exists on-chain)
+          }
+        } catch {
+          // ATA doesn't exist for this token program. Try the other.
+        }
+      }
+    }
+
+    try {
+      // Extract feePayer from requirements (already validated in verify)
+      const feePayer = requirements.extra.feePayer as Address;
+
+      // Sign transaction with the feePayer's signer
+      const fullySignedTransaction = await this.signer.signTransaction(
+        exactSvmPayload.transaction,
+        feePayer,
+        requirements.network,
+      );
+
+      // Send transaction to network
+      const signature = await this.signer.sendTransaction(
+        fullySignedTransaction,
+        requirements.network,
+      );
+
+      // Wait for confirmation
+      await this.signer.confirmTransaction(signature, requirements.network);
+
+      // Post-settlement verification for smart wallet transactions.
+      // Confirms the TransferChecked actually executed on-chain (TOCTOU defense).
+      if (isSmartWalletSettlement) {
+        const signerAddresses = this.signer.getAddresses().map(a => a.toString());
+        const postVerify = await verifyPostSettlement(
+          this.signer,
+          signature,
+          requirements.network,
+          requirements,
+          signerAddresses,
+          balanceBefore,
+          balanceBeforeTokenProgram?.toString() ?? null,
+        );
+
+        if (!postVerify.verified) {
+          return {
+            success: false,
+            errorReason: "post_settlement_transfer_not_confirmed",
+            transaction: signature,
+            network: payload.accepted.network,
+            payer: valid.payer || "",
+          };
+        }
+      }
+
+      return {
+        success: true,
+        transaction: signature,
+        network: payload.accepted.network,
+        payer: valid.payer,
+      };
+    } catch (error) {
+      console.error("Failed to settle transaction:", error);
+      return {
+        success: false,
+        errorReason: "transaction_failed",
+        transaction: "",
+        network: payload.accepted.network,
+        payer: valid.payer || "",
+      };
+    }
+  }
+
+  /**
+   * Internal verification that also reports which path validated the payment.
+   *
+   * settle() consumes verificationPath to decide whether post-settlement TOCTOU
+   * verification is required, instead of re-decoding the transaction and
+   * inferring the path from a missing token payer.
+   *
+   * @param payload - The payment payload to verify
+   * @param requirements - The payment requirements
+   * @returns Verify response plus the path that succeeded (null on failure)
+   */
+  private async _verify(
+    payload: PaymentPayload,
+    requirements: PaymentRequirements,
+  ): Promise<VerifyResult> {
     const exactSvmPayload = payload.payload as ExactSvmPayloadV2;
 
     // Step 1: Validate Payment Requirements
     if (payload.accepted.scheme !== "exact" || requirements.scheme !== "exact") {
       return {
-        isValid: false,
-        invalidReason: "unsupported_scheme",
-        payer: "",
+        response: { isValid: false, invalidReason: "unsupported_scheme", payer: "" },
+        verificationPath: null,
       };
     }
 
     if (payload.accepted.network !== requirements.network) {
       return {
-        isValid: false,
-        invalidReason: "network_mismatch",
-        payer: "",
+        response: { isValid: false, invalidReason: "network_mismatch", payer: "" },
+        verificationPath: null,
       };
     }
 
     if (!requirements.extra?.feePayer || typeof requirements.extra.feePayer !== "string") {
       return {
-        isValid: false,
-        invalidReason: "invalid_exact_svm_payload_missing_fee_payer",
-        payer: "",
+        response: {
+          isValid: false,
+          invalidReason: "invalid_exact_svm_payload_missing_fee_payer",
+          payer: "",
+        },
+        verificationPath: null,
       };
     }
 
@@ -132,9 +424,12 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
     const signerAddresses = this.signer.getAddresses().map(addr => addr.toString());
     if (!signerAddresses.includes(requirements.extra.feePayer)) {
       return {
-        isValid: false,
-        invalidReason: "fee_payer_not_managed_by_facilitator",
-        payer: "",
+        response: {
+          isValid: false,
+          invalidReason: "fee_payer_not_managed_by_facilitator",
+          payer: "",
+        },
+        verificationPath: null,
       };
     }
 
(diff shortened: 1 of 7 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

BranchManager69:
This approach is intentionally outcome-based rather than wallet-parser-based. The verifier does not trust arbitrary wallet programs blindly; it only accepts transactions that (1) never reference the facilitator fee payer in instruction accounts, (2) stay within bounded compute/priority-fee limits, (3) simulate successfully, and (4) produce exactly one TransferChecked matching the required mint, destination ATA, and exact amount. Standard wallets remain on the existing static verification path unchanged; this is only a fallback for smart-wallet transaction shapes the current verifier cannot parse structurally.

tenequm:
Nice to see this direction getting an upstream implementation. This aligns closely with the outcome-based verification model I proposed in #646 back in November - same two-path architecture, fee payer isolation, simulation with `innerInstructions: true`, exactly-one-TransferChecked.

We've been running this approach in production ([facilitator.cascade.fyi](https://facilitator.cascade.fyi)) and wanted to flag two bugs you'll likely hit:

**1. Solana RPC returns `innerInstructions` in jsonParsed format for known programs**

When you call `simulateTransaction` with `innerInstructions: true`, the RPC returns "jsonParsed where possible, otherwise json." For SPL Token, that means `TransferChecked` comes back as:

```json
{
  "parsed": {
    "type": "transferChecked",
    "info": { "mint": "...", "destination": "...", "authority": "...", "tokenAmount": { "amount": "10000" } }
  },
  "program": "spl-token",
  "programId": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA"
}
```

There's no `data` field. If your extraction code assumes raw format and tries to decode `ix.data`, you get `Buffer.from(undefined)`. This is what breaks Squads Multisig v4 `vaultTransactionExecute` transactions - the CPI'd TransferChecked is a known program, so the RPC parses it.

You need to handle both formats: parsed (read from `parsed.info`) and raw (decode from `data`). Same applies to detecting unauthorized transfer types - check `parsed.type` for `"transfer"`, `"transferCheckedWithFee"`, etc.

**2. Raw inner instructions use base58, not base64**

The raw format `data` field is base58-encoded, not base64. The top-level `simulateTransaction` uses base64 for the transaction wire format, but inner instruction data comes back as base58. Using the wrong decoder silently produces garbage bytes and fails to match discriminators.

---

Both confirmed against real Squads Multisig v4 vault transactions on mainnet. We have regression tests covering parsed inner instruction handling if useful as reference.

One design note: your PR requires exact amount match, while #646 specifies `amount >= maxAmountRequired` (allowing overpayment). Overpayment tolerance matters for smart wallets where the CPI amount might include rounding from internal fee calculations.

notorious-d-e-v:
Great work on this PR — we like the outcome-based verification direction and think simulation + CPI trace inspection is the right architecture for supporting smart wallets without per-wallet parsers. We've done a security review and have some feedback.

+1 to @tenequm's comment on jsonParsed vs raw format handling — worth verifying against real Squads v4 vault transactions.

## Security Invariants

| # | Invariant | Description |
|---|-----------|-------------|
| I1 | **No token loss** | Fee payer must never be used as authority/source/delegate on any token operation. Enforced *before signing*. |
| I2 | **Bounded SOL exposure** | Per-transaction SOL cost capped via compute budget limits. |
| I3 | **Merchant truth** | Success returned only after confirming actual on-chain effects — not simulation. |
| I4 | **Payment landed** | At least one confirmed `TransferChecked` matches the full required amount, correct mint, and correct destination ATA. |
| I5 | **Simulation is non-critical** | Simulation saves fees. Security comes from I1 (pre-sign) and I3 (post-settle). |
| I6 | **No hidden accounts** | Every account the transaction can access must be visible to verification. |
| I7 | **Known programs only** | Top-level smart wallet program must be from an audited allowlist. |

## Recommendations

### 1. Post-Settlement Verification ✅

Commit `6948c574` addresses the TOCTOU risk. Simulation becomes a pre-check; the security guarantee comes from inspecting confirmed on-chain effects.

### 2. Address Lookup Table Blindness (must fix)

`assertFeePayerIsolated` calls `decompileTransactionMessage(compiled)` without resolving ALTs. For v0 transactions, ALT-resolved accounts are invisible to the isolation check. Since the fee payer has signed the transaction, an instruction referencing the fee payer via an ALT-resolved account could drain the facilitator's tokens. This violates I1.

Wallet providers like Crossmint use ALTs when constructing transactions, so rejecting ALTs outright would limit smart wallet coverage. We recommend resolving them — `decompileTransactionMessageFetchingLookupTables(compiled, rpc)` is already used in the legacy facilitator code. One additional batched RPC call per ALT transaction.

### 3. Program Allowlist (should have)

Path 2 currently allows any program to execute. An allowlist for the top-level non-ComputeBudget instructions (e.g., Squads, Swig) constrains the surface to audited programs.

### 4. Relax "Exactly One" Transfer (should have)

Some smart wallets bundle ATA creation with the payment, producing multiple transfers. Recommend relaxing to "at least one transfer matching the full required amount."

## Summary

| Priority | Change |
|----------|--------|
| **Must have** | Resolve ALTs before fee payer isolation check (I1 / I6) |
| **Already in progress** | Post-settlement verification ✅ |
| **Should have** | Program allowlist (I7) |
| **Should have** | Relax to "at least one matching full amount" (I4) |

BranchManager69:
Thanks for the detailed feedback and the production validation from Cascade. This aligns well with what we found during development. A few responses:

**1 & 2: jsonParsed format + base58 encoding** — Good catches — these are exactly the traps we ran into during development. Both are handled in the current code:

- **jsonParsed format** (`extractTransfersFromInnerInstructions`, lines 199-215): checks `parsed.type === "transferChecked"` and reads from `parsed.info.mint`, `parsed.info.destination`, `parsed.info.authority`, `parsed.info.tokenAmount.amount`. Never touches `data` in this path.
- **Raw compiled format** (lines 218-243): decodes `data` from base58 using `getBase58Encoder().encode(dataStr)`, checks discriminator byte 12, parses amount/accounts from byte layout.

Both paths were confirmed against real Squads Multisig v4 `vaultTransactionExecute` transactions on mainnet.

**Overpayment tolerance:** Changed `===` to `>=` in both the simulation matching and post-settlement verification paths. Smart wallets with internal fee rounding should now be accepted.

Would welcome any reference to your regression tests for parsed inner instruction handling if you're open to sharing.

BranchManager69:
Thanks for the thorough security review and the invariant framework — it's exactly the right way to think about this.

Responses to each recommendation:

**1. Post-Settlement Verification ✅** — Implemented. Primary path fetches confirmed transaction inner instructions via `getConfirmedTransactionInnerInstructions`. Fallback checks destination ATA balance delta for RPC indexing lag resilience. Both signer methods are now **required** when `enableSmartWalletVerification` is enabled — the constructor throws if they're missing. No silent degradation. This satisfies I3 and I5.

**2. Address Lookup Table Blindness ✅** — Fixed. `assertFeePayerIsolated` now checks for `addressTableLookups` in the compiled message. If ALTs are present and the signer implements `fetchAddressLookupTables`, they're resolved via `fetchAddressesForLookupTables` from `@solana/kit` before the isolation check runs against the full account list. If ALTs are present but no resolution method exists, the transaction is rejected with `smart_wallet_alt_resolution_not_available`. Non-ALT transactions are unchanged. This satisfies I1 and I6.

**3. Program Allowlist ✅** — Agreed, this is a good additional layer. Added an operator-configurable allowlist with sensible defaults (Squads Multisig v4, Squads Smart Account, Swig, SPL Governance). Transactions invoking programs not in the list are rejected with `smart_wallet_program_not_allowed` before reaching simulation.

**4. Relax "Exactly One" Transfer** — We checked the CPI patterns for Squads Multisig v4, Squads Smart Account (`useSpendingLimit`), Swig (`SignV2`), and the Crossmint fee model from #1458. In every case, extra transfers (like Crossmint's treasury fee) go to a different destination and don't match the filter. With the allowlist constraining which programs can reach Path 2, this is consistent — unknown wallets are blocked before the transfer count check.

All changes are in the latest commits. Happy to discuss any of these further.

notorious-d-e-v:
Nice work @BranchManager69 thank you for addressing all those points

A couple more points below


### **compiled inner instruction parsing seems incorrect**

In `extractTransfersFromInnerInstructions`, the compiled instruction branch is treating `ix.data` (base58) incorrectly:

Right now it looks like we’re doing:
- take base58 string → **encode**
- then try to interpret opcode / amount

But RPC returns `ix.data` already base58-encoded, so this needs to be **decoded to bytes**, not encoded again.

#### **Why this matters**
- The compiled path will silently fail to detect `TransferChecked` CPIs
- Any RPC that returns compiled (non-parsed) inner instructions will cause:
  - false negatives in verification
  - post-settlement verification failures (since this extractor is reused)

#### **Suggested fix**
- Replace encoder with base58 **decoder**
- Then parse:
  - opcode (expect 12 for `TransferChecked`)
  - amount from the correct byte offset

#### **Strongly recommend adding a test fixture here**
- Use a real `getTransaction` response with:
  - `meta.innerInstructions` in compiled format
  - `programIdIndex`, `accounts`, and base58 `data`
- Assert we correctly extract:
  - destination ATA
  - mint
  - amount

---

### **Path 2 simulation may produce false positives**

In the smart-wallet path, we simulate with:
- `sigVerify: false`
- `replaceRecentBlockhash: true`

That makes simulation less equivalent to real execution.

#### **Risk**
- `verify()` can return `isValid: true` even if:
  - required non-facilitator signatures are missing/invalid
- `settle()` then proceeds and only fails later
- You mitigate this with post-settlement verification (good), but there’s still wasted work + confusing failure modes

#### **Options to tighten this**
- Prefer simulating with `sigVerify: true` once facilitator-signed bytes are available
- Or explicitly validate:
  - all required non-facilitator signatures are present before returning `isValid: true`

---

###  **Token-2022 not handled in post-settlement fallback**

Pre-settlement verification correctly supports:
- SPL Token
- Token-2022 (by deriving both ATAs)

But in **post-settlement fallback (balance delta path)**:
- only legacy SPL ATA is derived

#### **Impact**
- If:
  - `getTransaction` is unavailable / lagging
  - AND payment used Token-2022
- Then the fallback incorrectly reports failure

Since `settle()` now depends on post-settlement verification, this can cause real settlement failures.

#### **Suggested fix**
- Mirror pre-settlement logic:
  - derive both ATAs (SPL + Token-2022)
  - check whichever exists OR satisfies the delta

---

### **Testing suggestions**

This PR is doing a lot of heavy lifting across different RPC behaviors. It would help if testing around it was more concrete / involved.

Here are specific, high-value additions:

#### **1. Compiled inner instruction fixture**
- Capture a real `getTransaction` response with:
  - `meta.innerInstructions` using compiled format
- Store as JSON fixture
- Unit test:
  - feed into extractor
  - assert correct parsing of `TransferChecked`

---

#### **2. Token-2022 post-settlement fallback test**
Simulate:
- `getTransaction` returns `null` (or missing inner instructions)
- Token-2022 transfer happened

Mock:
- pre/post balances on both:
  - SPL ATA → unchanged
  - Token-2022 ATA → delta = expected amount

Assert:
- fallback correctly detects payment

---

#### **3. Simulation vs real execution divergence test**
Construct a transaction where:
- simulation passes with `sigVerify: false`
- but fails when signatures are enforced

Assert:
- either:
  - `verify()` rejects it (if you tighten logic), or
  - post-settlement verification catches it cleanly

This helps document and control the TOCTOU boundary.

---

#### **4. ALT (Address Lookup Table) coverage**
Since ALT support was added for fee-payer isolation:
- create a fixture where:
  - program accounts come via ALT
- ensure:
  - transfer extraction still resolves correct accounts

These tend to break subtly across RPC providers.

---

#### **5. RPC variability test (parsed vs compiled)**
Mock two responses:
- `jsonParsed` inner instructions
- compiled inner instructions

Assert:
- both produce identical extracted transfer results

---

### ✅ **Summary**

- **Blocker:** fix base58 decode bug in compiled inner instruction parsing  
- **Strongly recommended:**
  - Token-2022 fallback parity  
  - tighter or documented simulation semantics  
- **Tests:** add fixtures for compiled instructions, Token-2022 fallback, and RPC variability  

---

Please let me know if I missed anything!

notorious-d-e-v:
Just checking in here @BranchManager69 -- let me know if you need any help getting this across the finish line.

BranchManager69:
Hey @notorious-d-e-v, appreciate the thorough second review and the check-in. Took longer than it should have to get back to this.

Addressing each point:

**compiled inner instruction parsing (base58)**

I think this one's actually correct, though the naming is genuinely confusing. In `@solana/codecs-strings`, `getBase58Encoder` means "I am an encoder that works with the base58 format," so `.encode(string)` takes a base58 string and produces bytes. It's the codec convention (what format do I work with) not the intuitive direction (what am I converting to).

Verified: `getBase58Encoder().encode("hNmtbNYibdzwf")` produces `[12, 100, 0, 0, 0, 0, 0, 0, 0, 6]`, discriminator byte 12 (TransferChecked) correctly extracted. The existing test at line 283 of `smartWalletVerification.test.ts` already proves this with a round-trip through the compiled format.

**Token-2022 post-settlement fallback, fixed**

Real bug. The balance-delta fallback only derived SPL Token ATAs. Now:
- `scheme.ts` captures `balanceBefore` from whichever token program has a live ATA, records which one
- `verifyPostSettlement` accepts a token program hint, tries both SPL Token and Token-2022
- When all balance checks throw, correctly returns "unverified" instead of falsely claiming "balanceDelta"
- Two new tests cover both scenarios

**sigVerify / replaceRecentBlockhash, fixed (security improvement)**

Changed the smart wallet simulation from `sigVerify: false` / `replaceRecentBlockhash: true` to `sigVerify: true` / `replaceRecentBlockhash: false`. This matches the existing Path 1 simulation behavior already in the codebase.

If the blockhash has expired, the transaction can't land on-chain regardless, so rejecting at simulation is the correct behavior, not something to work around. And `replaceRecentBlockhash: true` forces `sigVerify: false` (replacing the blockhash invalidates all signatures), which opened the forged-signature vector you identified. Also the correct foundation for future durable nonce support per RFC #646, since `replaceRecentBlockhash: true` would break nonce-based transactions.

**Tests**

Added:
- Token-2022 fallback when SPL Token ATA shows no delta
- Hinted token program optimization in balance fallback

All 160 tests pass. Branch rebased on current main, conflicts resolved.

Re #1964 (per-program SWIG parser): the simulation approach here already handles SWIG and any other wallet program without per-program parsers, so that PR may be redundant once this lands.

romeo4934:
@BranchManager69 any update guys ?  :)

BranchManager69:
@tenequm rebased past #1688 and mirrored your memo enforcement onto Path 2, so smart wallet payments can't bypass a seller-required memo. Same error codes as your Step 5b, single top-level instruction shape, covered end-to-end in `smartWalletFallback.test.ts`. Exempted the Memo program from Path 2's wallet allowlist so it can actually reach the check, same treatment as ComputeBudget since it's not a wallet program. CI is green on the rebased branch. Would appreciate another look when you have a window.

notorious-d-e-v:
Thanks @BranchManager69 nice work! Looks good to me and excited to have this supported!

@phdargen @CarsonRoscoe could you please take a look at this when you get a chance?

romeo4934:
This PR supports only TS, is it normal ? It should support all language right Go and Python ?

notorious-d-e-v:
> This PR supports only TS, is it normal ? It should support all language right Go and Python ?

Yes it should support all. 

Would like to get some input from the maintainers first before more work is put into it

romeo4934:
@phdargen @CarsonRoscoe can you check this PR ? this is essential for Solana.

blockiosaurus:
Could you please also add `CoREENxT6tW1HoK8ypY1SxRMZTcVPm7R94rH4PZNhX7d` to the list of allowed smart wallets? Through our agent identity platform the agentic execution on the SVM happens through the Metaplex Core program PDA.

romeo4934:
Swig as well!

BranchManager69:
**Added Metaplex Core (`CoREENxT6tW1HoK8ypY1SxRMZTcVPm7R94rH4PZNhX7d`) to the default allowlist** per @blockiosaurus. Verified on mainnet via Path 2: real \$0.02 USDC payment, Core `execute` with inner SPL `TransferChecked`.
https://solscan.io/tx/5g5B3ZPHndS19tzUjPANgFC7J3nS8rT6dAFSuZ21bDfRHSSREKTbnN7XrYVp5haBmj4xDq2gD5A375evHq5y3vw6

@romeo4934, Swig (`SWiGmQedKzMz1tiTqoJCWeGDnGXfNBp2PkXLkpCAtQo`) is already in (line 46). Different program?

romeo4934:
no all good ;)

notorious-d-e-v:
Would love a review here @CarsonRoscoe @phdargen 

Happy to help in any way as well. Just let me know how.

BranchManager69:
@phdargen @CarsonRoscoe can you take a look when you can? This is a major blocker of x402 adoption by so many of Solana's most important, well-known services.

typescript/packages/mechanisms/svm/src/signer.ts
```diff
@@ -163,6 +163,87 @@ export type FacilitatorSvmSigner = {
...
+  fetchAddressLookupTables?(
+    lookupTableAddresses: string[],
+    network: string,
+  ): Promise<Record<string, string[]>>;
+};
+
+/**
+ * Result from simulation with inner instruction inspection.
+ */
+export type SmartWalletSimulationResult = {
```
phdargen: Would prefer a more neutral types in the general purpose signer interface, eg
```suggestion
export type SvmInnerInstructionsResult = {
```

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -155,7 +476,7 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
     // - 4 instructions: ComputeLimit + ComputePrice + TransferChecked + Lighthouse or Memo
     // - 5 instructions: ComputeLimit + ComputePrice + TransferChecked + Lighthouse + Lighthouse or Memo
     // - 6 instructions: ComputeLimit + ComputePrice + TransferChecked + Lighthouse + Lighthouse + Memo
-    // See: https://github.com/x402-foundation/x402/issues/828
+    // See: https://github.com/coinbase/x402/issues/828
```
phdargen: please revert

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -201,7 +522,6 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
       };
     }
 
-    // Parse the transfer instruction using the appropriate library helper
```
phdargen, line 208: please add the deleted comments back in

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -44,17 +111,35 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
...
    */
   constructor(
     private readonly signer: FacilitatorSvmSigner,
     settlementCache?: SettlementCache,
+    private readonly options?: ExactSvmSchemeOptions,
   ) {
     this.settlementCache = settlementCache ?? new SettlementCache();
+
+    if (this.options?.enableSmartWalletVerification) {
+      const required = [
```
phdargen, line 175: should we add `fetchAddressLookupTables` here?

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -146,6 +233,240 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
...
+    const staticResult = await this.verifyStaticPath(
+      transaction,
+      exactSvmPayload,
+      requirements,
+      signerAddresses,
+    );
+
+    if (staticResult.isValid) {
+      return staticResult;
+    }
```
phdargen: if path 1 fails for legitimate reasons like `invalid_exact_svm_payload_amount_mismatch`, path 2 will return a misleading smart wallet error code

should only try path 2 if tx layout related error and there is a chance of recovery

typescript/packages/mechanisms/svm/src/signer.ts
```diff
@@ -397,5 +478,144 @@ export function toFacilitatorSvmSigner(
...
+      } catch {
+        return null;
+      }
+    },
+
+    fetchAddressLookupTables: async (
+      lookupTableAddresses: string[],
+      network: string,
+    ): Promise<Record<string, string[]>> => {
+      const { fetchAddressesForLookupTables } = await import("@solana/kit");
```
phdargen: please replace with top-level import

typescript/packages/mechanisms/svm/src/signer.ts
```diff
@@ -397,5 +478,144 @@ export function toFacilitatorSvmSigner(
...
+          lookupTableAddresses.map(a => a as Address),
+          rpc,
+        );
+        const result: Record<string, string[]> = {};
+        for (const [key, addresses] of Object.entries(resolved)) {
+          result[key] = addresses.map((a: Address) => a.toString());
+        }
+        return result;
+      } catch {
+        return {};
```
phdargen: I think this should throw an error here

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -146,6 +233,240 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
...
+      // allowlist. Explicit for-loop instead of .map().filter() because strict
+      // TypeScript inference on decompileTransactionMessage's return type is
+      // sensitive to which @solana/kit version resolves across peer deps.
+      const rawInstructions = (decompiledForCheck.instructions ?? []) as ReadonlyArray<{
+        programAddress: { toString(): string };
+      }>;
+      const topLevelPrograms: string[] = [];
+      for (const ix of rawInstructions) {
+        const addr = ix.programAddress.toString();
+        if (addr === COMPUTE_BUDGET_PROGRAM_ADDRESS.toString() || addr === MEMO_PROGRAM_ADDRESS) {
```
phdargen: Should we include LIGHTHOUSE_PROGRAM_ADDRESS here? 

Seems phantom wallet has added another lighthouse instruction as reported in https://github.com/x402-foundation/x402/issues/2097. @BranchManager69 you mentioned [there](https://github.com/x402-foundation/x402/issues/2097#issuecomment-4293187663) that path 2 would recover it but, as far as I understand it, it would still fail path2 unless we add LIGHTHOUSE_PROGRAM_ADDRESS to the allowlist

phdargen: In any case, I think we should increase the number of allowed lighthouse instructions in path 1 again, so phantom doesnt need to go through path 2

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -146,6 +233,240 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
...
+        try {
+          const tx = decodeTransactionFromPayload(exactSvmPayload);
+          const payer = getTokenPayerFromTransaction(tx);
+          // If getTokenPayerFromTransaction returns null, the static path would have
+          // failed, meaning this went through the smart wallet path.
+          return !payer;
+        } catch {
+          return false;
+        }
+      })();
```
phdargen: a more robust approach might be to refactor into a private _verify helper called from settle, like:
` const { response: valid, verificationPath } = await this._verify(...);`

typescript/packages/mechanisms/svm/src/signer.ts
BranchManager69: Renamed to `SvmInnerInstructionsResult` across the signer interface and its call sites.

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
BranchManager69: Reverted to `x402-foundation/x402#828` (072f16ed).

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
BranchManager69: Restored. The `Step N:` markers and the per-section descriptive comments are back in `verifyStaticPath` and the `verify` flow.

typescript/packages/mechanisms/svm/src/signer.ts
BranchManager69: `fetchAddressesForLookupTables` is now a top-level import from `@solana/kit`.

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
BranchManager69: Good catch, yes. Added `fetchAddressLookupTables` to the constructor's required methods. `assertFeePayerIsolated` (the first Path 2 check) can't inspect ALT-resolved accounts without it, so omitting it let ALT-using wallets fail at verify time on a facilitator that looked correctly configured. Requiring it turns that into a construction-time error. `toFacilitatorSvmSigner()` already provides all four, so the standard path is unaffected. Added a constructor-enforcement test for the missing-ALT case.

typescript/packages/mechanisms/svm/src/signer.ts
BranchManager69: Agreed. It now throws `smart_wallet_alt_resolution_failed` instead of returning `{}`. An empty map was indistinguishable from "no ALTs", which silently weakened the fee-payer isolation check. The throw propagates to the existing try/catch in `verifySmartWalletTransaction`, so a transient RPC failure fails the payment closed (retryable) rather than verifying against unresolved accounts. Added a test.

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
BranchManager69: Fixed, and this was a real bug. Path 2 now runs only when Path 1 fails for a recoverable layout reason (instruction count, unknown or extra instructions, missing positional transfer), tracked in an explicit `LAYOUT_RECOVERABLE_REASONS` set. Semantic failures (`amount_mismatch`, `mint_mismatch`, `recipient_mismatch`, `memo_count`/`memo_mismatch`, self-spend, and `transaction_simulation_failed`) return their real reason and never reach Path 2, so they're no longer masked behind a `smart_wallet_*` code. A failed Path 1 simulation would fail Path 2's simulation too, so there's nothing to recover there. Two tests cover it: a semantic amount-mismatch asserts the real reason and that simulation was never called; a layout failure asserts Path 2 does run.

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
BranchManager69: Extracted a private `_verify` returning `{ response, verificationPath }`, called from both `verify` and `settle`. `settle` now reads `verificationPath === "smartWallet"` directly instead of re-decoding the transaction and inferring the path from a missing token payer. Cleaner, and it's also where the Path-1-vs-Path-2 classification from your other comment lives.

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
BranchManager69: You're right on both counts, and thanks for catching the gap in my #2097 comment. As the code stood, a Phantom tx would have hit Path 2 and been rejected with `smart_wallet_program_not_allowed` because Lighthouse wasn't in the allowlist. Did both:

1. Added `LIGHTHOUSE_PROGRAM_ADDRESS` to the default Path-2 allowlist (the backstop).
2. Raised the Path-1 instruction cap from 6 to 7 (see the linked thread), so Phantom passes statically without needing Path 2 at all. Added an `unknown_seventh` error code for the new optional slot.

Phantom's transfer is positional (slot 2) with the Lighthouse assertions in the optional tail, so it's really a Path-1 case once the cap allows it. Added a test that a 7-instruction tx with 3 Lighthouse instructions is accepted on Path 1 and never reaches simulation.

BranchManager69: Cap raised from 6 to 7 so Phantom doesn't need Path 2 (details in the reply above). If Phantom adds yet another assertion down the line, Path 2 with Lighthouse now in the allowlist is the backstop.

BranchManager69:
Thanks for the thorough review @phdargen, all comments addressed. Force-pushed a rebase onto current `main` (was conflicting), now a single clean commit. Summary for re-review:

| Comment | Resolution |
|---|---|
| Neutral type name | Renamed `SmartWalletSimulationResult` to `SvmInnerInstructionsResult` |
| Revert #828 link | Reverted to `x402-foundation/x402#828` |
| Restore deleted comments | `Step N:` markers and per-section comments restored |
| `fetchAddressLookupTables` in constructor | Now required alongside the other 3 smart-wallet signer methods, so misconfig fails at construction rather than verify time |
| Path 1 to Path 2 only on layout errors | `LAYOUT_RECOVERABLE_REASONS` set; semantic failures (amount/mint/recipient/memo mismatch, self-spend, failed simulation) return their real reason and never reach Path 2 |
| Top-level import | `fetchAddressesForLookupTables` hoisted |
| ALT resolution should throw | Throws `smart_wallet_alt_resolution_failed` instead of returning `{}`; fails closed on RPC failure |
| Lighthouse / #2097 | Cap raised from 6 to 7 so Phantom passes Path 1 statically; `LIGHTHOUSE_PROGRAM_ADDRESS` added to Path-2 allowlist as backstop; `unknown_seventh` error code added |
| Refactor to `_verify` | Private `_verify` returns `{ response, verificationPath }`; `settle` reads the path directly instead of re-decoding the tx |

New test coverage: semantic failure skips Path 2 (and asserts simulation is never called), layout failure recovers via Path 2, 7-instruction Phantom tx accepted on Path 1, ALT resolution failure propagates, constructor enforces `fetchAddressLookupTables`. 190/190 passing, lint and typecheck clean.

The rebase preserves the message-hash settlement cache key and the compute-price simplification from `main`.

phdargen (approved):

phdargen:
Thanks for the update @BranchManager69, looks great! Happy to get this in 🚀

Let's coordinate some follow-ups:
- @BranchManager69 are you willing to do equivalent changes for the go/py sdks?
- svm spec needs to be updated, @notorious-d-e-v I remember you had an early draft for this. Could you please update aligning with the solution implemented here?
- I will add some e2e tests with a svm smart wallet client

phdargen:
PR with the e2e test: https://github.com/x402-foundation/x402/pull/2504
cc @BranchManager69

notorious-d-e-v:
@phdargen on the spec follow-up you flagged here — I've updated my earlier draft in #829 to align with the smart wallet verification that shipped in this PR, and rebased it onto current main (now mergeable). Ready for review there.

notorious-d-e-v:
Great work @BranchManager69!

PR #1004, opened 2026-01-21
spec: Add draft TXID payment payload specification

alftom:
Adds a draft TXID payment payload (spec-only, for discussion) that validates payments via a payer-submitted on-chain transaction reference.

## Description

This PR adds a draft specification for a new x402 payment payload type, TXID. TXID is a transaction-reference–based payment payload in which the payer executes an on-chain transfer directly and submits a signed payload that references that transaction. This enables x402 payments for assets and networks that do not support authorization-based transfer mechanisms such as EIP-3009, including **native assets like ETH**.

This draft directly addresses the use case described in [Issue #612](https://github.com/coinbase/x402/issues/612).

This change is spec-only and intended for discussion.

### Notes on structure and naming

This PR introduces a new `specs/payment-payloads/` directory to define the structure and processing of different payment payloads. The directory name is not meant to be prescriptive, and alternative naming or placement is open for discussion.

The draft specification also introduces an explicit `payload.type` field to identify the payload variant. While payload types could be inferred heuristically by inspecting payload structure, an explicit discriminator is for clarity and efficiency. This choice is not fundamental to the TXID mechanism itself and can be revisited independently.

## Tests

Not applicable. This PR introduces documentation/specification changes only and does not include any code or testable behavior.

## Checklist

* [x] This PR introduces no code changes
* [x] No tests are required for spec-only changes
* [x] My commits are signed (required for merge)

PR #1575, opened 2026-03-12
feat(specs): add Keeta exact scheme

sc4l3r:
## Description

This PR adds a new x402 chain `exact` specification for the Keeta Network.
Once merged, a reference TypeScript implementation will follow in a separate PR.

I'm **not** a member of the Keeta development team. This spec has been designed in collaboration with @ezraripps from the Keeta team though and has been approved in my fork https://github.com/sc4l3r/coinbase-x402/pull/3 .

> Keeta Network is a high-performance Layer 1 blockchain built to meet the demands of the global financial system. Backed by Eric Schmidt, former CEO of Google, Keeta is capable of processing over 11 million transactions per second with sub-second settlement times.
>
> - Website: https://keeta.com
> - Overview by Google Cloud: https://cloud.google.com/blog/topics/financial-services/how-blockchain-network-keeta-processes-11-million-transactions-per-second-with-spanner
> - Product description: https://keeta.com/product-manual.pdf
> - TypeScript SDK: https://static.test.keeta.com/docs/index.html
> - Explorer: https://explorer.keeta.com/ (official) or https://kee.tools/ (community)
> - Wallet: https://wallet.keeta.com/ (mainnet), https://wallet.test.keeta.com (testnet)
> - GitHub: https://github.com/KeetaNetwork

## Tests

The spec has been verified to work in practice via a draft implementation in my fork https://github.com/sc4l3r/coinbase-x402/pull/2 and an example in a separate repo https://github.com/sc4l3r/keeta-x402 .

## Checklist

### DIFF
--- a/specs/schemes/exact/scheme_exact_keeta.md
+++ b/specs/schemes/exact/scheme_exact_keeta.md
@@ -132,6 +132,10 @@ Steps to verify a payment for the `exact` scheme on Keeta:
         2. The `amount` matches the `requirements.amount`.
         3. The `to` matches the `requirements.payTo`.
         4. The `external` matches the `extra.external` if set.
+    5. Simulate the transaction by verifying that:
+        1. The `block.account` balance of `requirements.asset` is at least `requirements.amount`.
+        2. The head block hash of the `block.account` equals the `block.previous` hash.
+        3. The `block.signer` is allowed to send on behalf of the `block.account`, e.g. because `block.signer` is the same or has the permission such as `OWNER` or `SEND_ON_BEHALF`.
 
 ## Settlement
 
--- a/specs/schemes/exact/scheme_exact_keeta.md
+++ b/specs/schemes/exact/scheme_exact_keeta.md
@@ -1,0 +1,193 @@
+# Scheme: `exact` on `keeta`
+
+## Summary
+
+The `exact` scheme on Keeta transfers a specific amount of a token (such as USDC) on the Keeta network from the payer to the resource server.
+The payer constructs a signed block containing a `SEND` operation to fulfill the `paymentRequirements`.
+The facilitator sponsors network fees and cannot alter the signed block to redirect funds to any other address.
+
+**Version Support:** This specification supports x402 v2 protocol only.
+
+> [!NOTE]
+> Non-sponsored fee flows (where the client pays network fees) are currently not supported. See [Non-Sponsored Fee Flows](#non-sponsored-fee-flows) for details.
+
+## Protocol
+
+```mermaid
+sequenceDiagram
+    participant Client;
+    participant Server;
+    participant Facilitator;
+    participant Blockchain;
+
+    Client->>Server: GET /api
+    Server-->>Client: 402 - Payment Required<br>with extra
+    Client->>Client: Create & sign block
+    Client->>Server: GET /api <br>with PaymentPayload
+    Server->>Facilitator: POST /verify
+    Facilitator->>Facilitator: Parse block,<br>verify signature,<br>verify requirements
+    Facilitator-->>Server: VerifyResponse
+    Server->>Facilitator: POST /settle
+    Facilitator->>Facilitator: Create & sign fee block
+    Facilitator->>Blockchain: Collect votes and<br>submit vote staple
+    Blockchain-->>Facilitator: Confirmation
+    Facilitator-->>Server: SettlementResponse with<br>Block Hash
+    Server-->>Client: Requested Content
+```
+
+1. **Client** makes a request to a **Resource Server**.
+2. **Resource Server** responds with a payment required signal containing `PaymentRequired`.
+3. **Client** creates and signs a block with a `SEND` operation to transfer the specified amount of the token to the recipient. If the `extra.external` field is set, the client sets the `external` field to the specified value in the `SEND` operation. The block is **not** published to the network.
+4. **Client** serializes the signed block into its ASN.1 DER representation and encodes it as a Base64 string.
+5. **Client** sends a new request to the **Resource Server** with the `PaymentPayload` containing the Base64-encoded signed block.
+6. **Resource Server** receives the request and forwards the `PaymentPayload` and `PaymentRequirements` to a **Facilitator's** `/verify` endpoint.
+7. **Facilitator** decodes and parses the signed block and verifies the block according to the [verification rules](#verification).
+8. **Facilitator** returns a `VerifyResponse` to the **Resource Server**.
+9. **Resource Server**, upon successful verification, forwards the payload to the facilitator's `/settle` endpoint.
+10. **Facilitator** verifies the block according to the [settlement rules](#settlement). It computes and signs a fee block, requests votes for the blocks from the network's representatives and publishes the combined vote staple to the network.
+11. Upon successful on-chain settlement, the **Facilitator** responds with a `SettlementResponse` including the hash of the client's payment block to the **Resource Server**.
+12. **Resource Server** grants the **Client** access to the resource in its response.
+
+## Payment header payload
+
+### `PaymentRequirements` for `exact`
+
+In addition to the standard x402 `PaymentRequirements` fields, the `exact` scheme on Keeta supports an optional `extra` field:
+
+```json
+{
+  "scheme": "exact",
+  "network": "keeta:21378",
+  "amount": "1000000",
+  "asset": "keeta_amnkge74xitii5dsobstldatv3irmyimujfjotftx7plaaaseam4bntb7wnna",
+  "payTo": "keeta_aabcdefghijklmnopqrstuvwxyz234567abcdefghijklmnopqrstuvwxyz2345",
+  "maxTimeoutSeconds": 60,
+  "extra": {
+    "external": "0123456789abcdef0123456789abcdef"
+  }
+}
+```
+
+**Field Descriptions:**
+
+- `scheme`: Always `"exact"` for this scheme
+- `network`: CAIP-2 network identifier, e.g. `keeta:21378` (mainnet) or `keeta:1413829460` (testnet)
+- `amount`: The exact amount to transfer in atomic units (e.g., `"1000000"` = 1 USDC, since USDC has 6 decimals)
+- `asset`: The Base32-encoded identifier public key of the token (e.g., USDC on Keeta mainnet: `keeta_amnkge74xitii5dsobstldatv3irmyimujfjotftx7plaaaseam4bntb7wnna`)
+- `payTo`: The Base32-encoded public key of the recipient account
+- `maxTimeoutSeconds`: Maximum time in seconds before the payment expires
+- `extra.external`: **Optional**. `external` reference the client should set in the `SEND` operation to the `payTo` address (see [Keeta docs](https://static.network.keeta.com/docs/classes/KeetaNetSDK.Referenced.BlockOperationSEND.html#external)). This is especially useful for resource servers to automatically off-ramp received payments to a bank account or bridge them to a different chain via asset movement anchors.
+
+### PaymentPayload `payload` Field
+
+The `payload` field of the `PaymentPayload` must contain the following field:
+
+- `block`: Base64 encoded ASN.1 DER-serialized signed block which contains a `SEND` operation to pay the requested amount of a token.
+
+Example `payload`:
+
+```json
+{
+  "block":"MIH6AgEAAgRURVNUBQAYEzIwMjYwMTIzMjIyNjUwLjczMFoEIgAC2Ynov21UzUtAf00BzdTbpJCJl1DuLlX4mAiKHx57uQAFAAQgmArjQZymslS0VvBMCNyicKkDyDUqoMQIfU8nl82JcvAwTqBMMEoEIgADEFUSmawYqevhKALRFALRYRGGrXR20+JHvI/5oE8qz00CAQEEIQNwgpeV3wC60ZR4DMHh0sDJDXFi4Mhesi9jMHvtPqp1SgRAdoNTNrjabm2gJBT2yAtVniYlpU4AzWZxb6b7rfMSw/d+C09d5qI6NmS1U2o+cOt+yJLEYE2qCEsKBYdHrgkwNA=="
+}
+```
+
+Full `PaymentPayload` object:
+
+```json
+{
+  "x402Version": 2,
+  "resource": {
+    "url": "https://example.com/weather",
+    "description": "Access to protected content",
+    "mimeType": "application/json"
+  },
+  "accepted": {
+    "scheme": "exact",
+    "network": "keeta:1413829460",
+    "amount": "1000000000",
+    "asset": "keeta_anyiff4v34alvumupagmdyosydeq24lc4def5mrpmmyhx3j6vj2uucckeqn52",
+    "payTo": "keeta_aabravistgwbrkpl4euafuiualiwcemgvv2hnu7ci66i76naj4vm6tmeahmzria",
+    "maxTimeoutSeconds": 60
+  },
+  "payload": {
+    "block":"MIH6AgEAAgRURVNUBQAYEzIwMjYwMTIzMjIyNjUwLjczMFoEIgAC2Ynov21UzUtAf00BzdTbpJCJl1DuLlX4mAiKHx57uQAFAAQgmArjQZymslS0VvBMCNyicKkDyDUqoMQIfU8nl82JcvAwTqBMMEoEIgADEFUSmawYqevhKALRFALRYRGGrXR20+JHvI/5oE8qz00CAQEEIQNwgpeV3wC60ZR4DMHh0sDJDXFi4Mhesi9jMHvtPqp1SgRAdoNTNrjabm2gJBT2yAtVniYlpU4AzWZxb6b7rfMSw/d+C09d5qI6NmS1U2o+cOt+yJLEYE2qCEsKBYdHrgkwNA=="
+  }
+}
+```
+
+## Verification
+
+Steps to verify a payment for the `exact` scheme on Keeta:
+
+1. Verify `x402Version` is `2`.
+2. Verify the `scheme` is `exact`.
+3. Verify the network matches the agreed upon chain (CAIP-2 format: `keeta:<network_id>`).
+4. Decode and deserialize the Base64 and ASN.1 DER-encoded `payload.block` and:
+    1. Verify that the signature is valid.
+    2. Verify that the `network` matches the agreed upon Keeta `network_id`.
+    3. Verify that the block contains exactly one operation.
+    4. Verify that the operation is a `SEND` operation to pay the server for which:
+        1. The `token` matches the `requirements.asset`.
+        2. The `amount` matches the `requirements.amount`.
+        3. The `to` matches the `requirements.payTo`.
+        4. The `external` matches the `extra.external` if set.
+
+## Settlement
+
+Settlement is performed through the facilitator:
+
+1. **Facilitator** receives the `block`.
+2. **Facilitator** computes and signs a fee block.
+3. **Facilitator** transmits the blocks to the network by requesting the votes from the representatives and publishing the combined vote staple to the network.
+4. **Facilitator** sends the `SettlementResponse` to the **Resource Server**.
+
+### `SettlementResponse`
+
+The `SettlementResponse` for the exact scheme on Keeta:
+
+```json
+{
+  "success": true,
+  "transaction": "426C2D7401BB49D78F1C1EA84BF4AD7EBE294C4758037507AADD12CC0AB62910",
+  "network": "keeta:1413829460",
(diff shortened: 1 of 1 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+1. **Client** makes a request to a **Resource Server**.
+2. **Resource Server** responds with a payment required signal containing `PaymentRequired`.
+3. **Client** creates and signs a block with a `SEND` operation to transfer the specified amount of the token to the recipient. If the `extra.external` field is set, the client sets the `external` field to the specified value in the `SEND` operation. The block is **not** published to the network.
+4. **Client** serializes the signed block into its ASN.1 DER representation and encodes it as a Base64 string.
+5. **Client** sends a new request to the **Resource Server** with the `PaymentPayload` containing the Base64-encoded signed block.
+6. **Resource Server** receives the request and forwards the `PaymentPayload` and `PaymentRequirements` to a **Facilitator's** `/verify` endpoint.
+7. **Facilitator** decodes and parses the signed block and verifies the block according to the [verification rules](#verification).
+8. **Facilitator** returns a `VerifyResponse` to the **Resource Server**.
```
sc4l3r, line 45: The `/verify` endpoint is technically optional in Keeta and could easily be merged with the `/settle` endpoint to reduce the number of requests and thus speed up the payment (since the verification has to be done in the `/settle` endpoint anyways). We kept it here to be in line with the other x402 specs and also since afaik there's no way to specify that it's optional in the reference implementation (so it would always request /verify before /settle).

Let me know your thoughts on this and whether there's a potential benefit to the `/verify` endpoint I might be unaware of. The [Stellar spec](https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact_stellar.md) declares it as optional as well but I'm not sure what your general take on this is.

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
...
+
+We have found various possible solutions to these challenges, but each of them has its disadvantages, which is why we have rejected them for now:
+
+- **Additional fee-quote endpoint**: The client could query the facilitator on an additional endpoint (e.g., `/fee-quote`) to obtain exactly the fees the facilitator currently requires. However, this would require the client to trust the facilitator to not charge overly high fees and introduces a non-standard, additional facilitator endpoint beyond what x402 currently specifies.
+- **Send vote quotes**: Alternatively, clients could request vote quotes for their blocks directly from the network representatives and include their block, fee block, and vote quotes in the `PaymentPayload`. However, as more representatives are added to the network, the required number of vote quotes grows, and the combined payload could exceed the 8 KB HTTP header size limit commonly enforced by default by reverse proxies such as nginx.
```
sc4l3r, line 191: A third option we thought about was "Client Prepayments" where the client receives some kind of unique request id, publishes the payment block with that request id embedded and sends the hash in their `PaymentPayload` instead of the block. Then they would automatically pay for the network fees.

I haven't seen this prepayment approach here so far, so we wanted to bring it up for discussion and hear your thoughts on that. Maybe you discussed it internally already or I just couldn't find it in the repo.

A drawback of it is that it makes handling failed requests (e.g. expired, resource not available anymore, etc.) harder to handle since the server would now have to issue refunds to the client. You could let the facilitator handle that via an escrow wallet but from what I've read such trust on the facilitator is discouraged.

sc4l3r (commented):
Some comments on things we came across when designing the spec.

specs/schemes/exact/scheme_exact_keeta.md
phdargen: Verify is required. This is not about any blockchain capabilities but prevents the server from doing unnecessary work for invalid payloads

marcelosalloum: @phdargen: fixing it in Stellar here: https://github.com/coinbase/x402/pull/1640

sc4l3r:
@phdargen I'm sure you've got a lot to do, if there's anything we can do to help review the PR and move it forward, let me know :)

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
...
+  },
+  "payload": {
+    "block":"MIH6AgEAAgRURVNUBQAYEzIwMjYwMTIzMjIyNjUwLjczMFoEIgAC2Ynov21UzUtAf00BzdTbpJCJl1DuLlX4mAiKHx57uQAFAAQgmArjQZymslS0VvBMCNyicKkDyDUqoMQIfU8nl82JcvAwTqBMMEoEIgADEFUSmawYqevhKALRFALRYRGGrXR20+JHvI/5oE8qz00CAQEEIQNwgpeV3wC60ZR4DMHh0sDJDXFi4Mhesi9jMHvtPqp1SgRAdoNTNrjabm2gJBT2yAtVniYlpU4AzWZxb6b7rfMSw/d+C09d5qI6NmS1U2o+cOt+yJLEYE2qCEsKBYdHrgkwNA=="
+  }
+}
+```
+
+## Verification
+
+Steps to verify a payment for the `exact` scheme on Keeta:
```
phdargen, line 121: Should also check the client balance here and ideally do a full tx simulation

phdargen, line 121: Please check if any additional safety checks for the facilitator are required,
e.g., verifying that the facilitator's account is not the from account in the SEND operation?

Compare SVM:
https://github.com/sc4l3r/coinbase-x402/blob/58a21fcb8575a758b9f39a7f5b88d976b8bb5c12/specs/schemes/exact/scheme_exact_svm.md?plain=1#L120-L124

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
...
+    3. Verify that the block contains exactly one operation.
+    4. Verify that the operation is a `SEND` operation to pay the server for which:
+        1. The `token` matches the `requirements.asset`.
+        2. The `amount` matches the `requirements.amount`.
+        3. The `to` matches the `requirements.payTo`.
+        4. The `external` matches the `extra.external` if set.
+
+## Settlement
+
+Settlement is performed through the facilitator:
```
phdargen, line 142: What would happen if a client sends the same exact payload multiple times concurrently? 

If settle returns success for all but only one get actually confirmed onchain, we would need a dedup strategy as in SVM:
https://github.com/coinbase/x402/blob/main/specs/schemes/exact/scheme_exact_svm.md#duplicate-settlement-mitigation-recommended

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
+# Scheme: `exact` on `keeta`
+
+## Summary
+
+The `exact` scheme on Keeta transfers a specific amount of a token (such as USDC) on the Keeta network from the payer to the resource server.
+The payer constructs a signed block containing a `SEND` operation to fulfill the `paymentRequirements`.
```
phdargen, line 6: Could you please describe the structure of a 'SEND' operation in more detail. For example, the EVM spec fully defines the EIP-3009 authorization types and fields and the SVM spec specifies the exact instruction layout.

From the verification rules, one can infer the SEND operation contains at least these fields: token, amount, to and (optionally) external. Any additional? Does it include sth like a nonce?

phdargen, line 6: Can each signed block only be settled once?

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
...
+- **Send vote quotes**: Alternatively, clients could request vote quotes for their blocks directly from the network representatives and include their block, fee block, and vote quotes in the `PaymentPayload`. However, as more representatives are added to the network, the required number of vote quotes grows, and the combined payload could exceed the 8 KB HTTP header size limit commonly enforced by default by reverse proxies such as nginx.
+
+### Future Improvements
+
+As the Keeta network develops and supports more features in the future, this specification should be improved to support these natively, including:
+
+- **Temporary approvals**: It may become possible for clients to sign temporary / one-time spending approvals. Then, instead of sending a specific payment block, they could send the approval instead to allow resource servers / facilitators to redeem these, similar to how the `eip3009` asset transfer in the EVM scheme uses the `transferWithAuthorization` function. This may then also include the payment for the network fees and enable non-sponsored fee flows.
```
phdargen, line 197: As far as I understand, a signed block would have no expiry window atm. 

Let's say the client signs a block A that doesn't get settled, if the client produces a new block B that gets confirmed onchain, would the block A signature still be valid and could be settled indefinitely by the server as long as the payer has sufficient balance?

specs/schemes/exact/scheme_exact_keeta.md
phdargen: Sponsored flows are strongly preferred and for EVM/SVM we currently do not support non-sponsored fee flows either, so no need to implement this for Keeta at this point. 

Client prepayments have been proposed before, eg in https://github.com/coinbase/x402/pull/1004 for EVM, but as you correctly pointed out this has significant implications on the security/trust model, so we haven't decided on a direction yet

phdargen:
Hi @sc4l3r, thanks a lot for putting this together! Made a first pass and the spec is well written and structured. I think the verification logic could still be tighten a bit and Keeta specific tx structure and replay protection mechanisms stated more explicitly, please check my comments above

specs/schemes/exact/scheme_exact_keeta.md
xescure: KeetaNet has 3 layers of abstractions:

1. Operations - Instructions, such as `SEND`, `RECEIVE`, `SET_INFO`, `SET_REP`, and many others described at the footer of the [whitepaper](https://keeta.com/keetanet-whitepaper-20250312.pdf). 

2. Blocks - Each operation belongs to an account. For the sake of efficiency multiple operations can all take place in a single block. 

However, only ops of one account may be grouped into a block. Every block must have a signature from an authorized key - or multiple in the case of multisig. The authorized key is usually the `OWNER` of the account, but the permission system only cares whether a key has permissions to perform the ops on the block.

Each account has a blockchain of its own. This chain must adhere to the properties of a blockchain. It must be sequential, and forks aren't allowed either.

Blocks are atomic, either all ops are performed or the whole unit is rejected.

3. Staples - A client is free to organize signed blocks into staples. Staples are the unit for consensus. Validators deal with staples, they vote on every block inside of a staple at once.

No signatures are involved at this layer. Anyone in posession of valid unsubmitted blocks can put them into a staple and submit it on their own - if they pay the network fees.

xescure: So coming back to your question: `Block A` and `Block B` both have cryptographic references to the same block. Once `Block B` is submitted, `Block A` no longer refers to the current head of the account's chain, and thus is always and forever rejected.

specs/schemes/exact/scheme_exact_keeta.md
xescure:
```
Keeta DEFINITIONS ::= BEGIN

AdjustMethod ::= INTEGER {
    add (0),
    remove (1),
    set (2)
}

BlockOperation ::= CHOICE {
    -- Send operation
    send [0] SEQUENCE {
        -- Destination account to send to
        to OCTET STRING,
        -- Amount of the token to send
        amount INTEGER,
        -- Token ID to send
        token OCTET STRING,
        -- External reference field (optional)
        external UTF8String OPTIONAL
    },

    -- SET_REP operation
    setrep [1] SEQUENCE {
        -- Representative to delegate to
        to OCTET STRING
    },

    -- SET_INFO operation
    setinfo [2] SEQUENCE {
        -- Name to specify for this account
        name UTF8String,
        -- Description to specify for this account
        description UTF8String,
        -- Metadata to specify for this account
        metadata UTF8String,
        -- Default permission to specify for this account (optional)
        defaultPermission SEQUENCE {
            -- Base permissions, which the network verifies
            base INTEGER,
            -- External permissions, which the network
            -- does not verify (for example, for use in
            -- smart contracts)
            external INTEGER
        } OPTIONAL
    },

    -- MODIFY_PERMISSIONS operation
    modifypermissions [3] SEQUENCE {
        -- Principal to modify permissions for
        principal OCTET STRING,
        -- Method to modify permissions for
        method AdjustMethod,
        -- Permissions to modify, as a bitfield
        permissions CHOICE {
            -- Permissions to set
            value SEQUENCE {
                base INTEGER,
                external INTEGER
            },
            -- If no permissions are required
            none NULL
        },
        -- Target to modify permissions for
        target OCTET STRING OPTIONAL
    },

    -- CREATE_IDENTIFIER operation
    createidentifier [4] SEQUENCE {
        -- Identifier to create, this must match
        -- the deterministic identifier which is
        -- generated from the account, blockhash,
        -- and operation index
        identifier OCTET STRING
    },

    -- TOKEN_ADMIN_SUPPLY operation
    tokenadminsupply [5] SEQUENCE {
        -- Amount of change to the supply
        amount INTEGER,
        -- Method to modify the supply
        method AdjustMethod
    },

    -- TOKEN_MODIFY_BALANCE operation
    tokenmodifybalance [6] SEQUENCE {
        -- Token to modify the balance of
        token OCTET STRING,
        -- Amount to modify the balance by
        amount INTEGER,
        -- Method to modify the balance
        method AdjustMethod
    },

    -- RECEIVE operation
    receive [7] SEQUENCE {
        -- Amount to receive
        amount INTEGER,
        -- Token to receive
        token OCTET STRING,
        -- Sender from which to receive
        from OCTET STRING,
        -- Whether the received amount must match
        -- exactly (true) or just be greater than or
        -- equal to the amount (false)
        exact BOOLEAN,
        -- Forward the received amount to another
        -- account (optional)
        forward OCTET STRING OPTIONAL
    }
}

Block ::= SEQUENCE {
    -- Version of this block
    version INTEGER { v1 (0) },
    -- Network ID
    network INTEGER,
    -- Subnet ID
    subnet CHOICE {
        -- Subnet ID
        subnetID INTEGER,
        -- Null if no subnet (mainnet)
        null NULL
    },
    -- Date of the block
    date GeneralizedTime,
    -- Signer's public key
    signer OCTET STRING,
    -- Account ID
    account CHOICE {
        -- Account ID
        accountID OCTET STRING,
        -- Null if account is signer
        null NULL
    },
    -- Previous block hash
    previous OCTET STRING,
    -- Operations in this block
    operations SEQUENCE OF BlockOperation,
    -- Signature of the block
    signature OCTET STRING
}

END
```

specs/schemes/exact/scheme_exact_keeta.md
xescure: > e.g., verifying that the facilitator's account is not the from account in the SEND operation

This is thankfully ensured by the network itself.

specs/schemes/exact/scheme_exact_keeta.md
sc4l3r: Yes.

Each block references the previous one on the account's chain via its hash in the block's `previous` field. When submitting the block the `previous` field must be the hash of the current head block of the account (hence also the need to serialize the transactions in the facilitator). If you'd try to submit a block twice, the `previous` field would stay the same and be invalid on the second try (since the new block is now the head).

Similar to https://github.com/coinbase/x402/pull/1575#discussion_r2974475759

specs/schemes/exact/scheme_exact_keeta.md
sc4l3r: To add to @xescure here, see this comment for more information on the block structure https://github.com/coinbase/x402/pull/1575#discussion_r2974450560 .

To be able to submit any operation as the facilitator (which would be a part of a block where the `account` is the facilitator's account), you'd have to have the permission to sign a block for them (see the referenced comment) which isn't given by default. If you have that, the network allows it, and we should simply allow it here as well (as you could also settle the transaction outside of x402 if you wanted to). If you don't have it, the network rejects the blocks and we are safe. Either way, we don't really have to check that manually.

The facilitator also doesn't sign arbitrary blocks in our design. It receives signed blocks from the client and only adds an additional `FEE` block (which it signs) in which it only sends the fees to the representatives. Hence, there's no risk of accidentally signing malicious blocks from the clients.

sc4l3r: This is ensured by the reps at settlement time, so immediately when the block is submitted to the network. Between verification and settlement the account's balance, head block, permissions, etc. could also change which could make the verification itself succeed but the settlement later fail.

Where would you see the benefit of having these checks here as well (probably similar to your previous comment on preventing the server to do some "expensive" work before settling)? And what should we cover exactly, i.e. how detailed do these checks have to be? 

We could add a simple balance check if you think that's useful. I'd prefer to keep the checks as minimal as possible though to not replicate the rep's rules since these may change over time and make the final decision anyways.

specs/schemes/exact/scheme_exact_keeta.md
sc4l3r: In our design settle should only return success if the actual settlement was successful since it happens  during the request to the `/settle` endpoint (because it's so fast).

In our reference implementation we'll use a queue to serialize the settlement requests and wait for the item to be processed before returning from the request. When the same block would be submitted twice concurrently, we can have a check whether the block is in the queue already and return an error if that's the case.

specs/schemes/exact/scheme_exact_keeta.md
sc4l3r: Then we'll leave the non-sponsored fee flow out for now and potentially add it later once new network capabilities are added or similar.

sc4l3r:
@phdargen Thanks for your review! We should've addressed all of your comments, hope that clears things up a bit. Happy to answer any further questions you may have :)

specs/schemes/exact/scheme_exact_keeta.md
```diff
@@ -0,0 +1,193 @@
...
+  "network": "keeta:1413829460",
+  "payer": "keeta_aabntcpix5wvjtklib7u2aon2tn2jeejs5io4lsv7cmarcq7dz53sahhsuapica"
+}
+```
+
+**Field Descriptions:**
+
+- `transaction`: The hash of the client's block which contains the payment to the server.
+- `network`: CAIP-2 network identifier, e.g. `keeta:21378` (mainnet) or `keeta:1413829460` (testnet)
+- `payer`: The Base32-encoded public key of the account that payed the server
```
phdargen:
```suggestion
- `payer`: The Base32-encoded public key of the account that paid the server
```

specs/schemes/exact/scheme_exact_keeta.md
phdargen: Yes this is also about preventing expensive server work for invalid requests. We do full tx simulation in all existing network skds (EVM/SVM/Aptos/Stellar) and should aim for a comparable level of server protection. Conditions might change between `verify` and `settle` but these are rare edge cases.

The spec should explicitly list account balance, head block, permissions (and any additional you can think of) verification checks. We can then discuss which of those to mark as MUST vs SHOULD. I'd lean towards declaring account balance + head block as MUST.  Not sure about `permissions`, does this refer to a eg a delegated signer?

xescure: Yes, delegated accounts are one use case for permissions. Token issuers can also set their own rules. If they own the right permissions, they can blacklist wallets from spending their issued token.

sc4l3r: We added a tx simulation step which includes checks for account balance, head block, and permission of the signer to act on behalf of the account. I think we can keep all as MUST since they are fairly cheap to do (basically 2-3 API requests which can be parallelized).

phdargen (approved):

phdargen:
Thanks @sc4l3r @xescure, LGTM 🚀

PR #1664, opened 2026-03-18
feat: add dynamic routing to bazaar

CarsonRoscoe:
## Description

- Add dynamic route support to the Bazaar discovery extension so that parameterized URLs (e.g. `/users/123`, `/users/456`) consolidate into a single catalog entry under a canonical `routeTemplate` (e.g. `/users/:userId`)
- Support all three param syntaxes (`[param]`, `:param`, `*`) across TypeScript, Go, and Python SDKs with consistent `:param` output format
- Add bazaar example servers for all three SDKs demonstrating the dynamic route discovery developer experience

## How it works

When a server declares a route like `GET /weather/:city`, the x402 middleware:

1. Matches the incoming request (e.g. `/weather/san-francisco`) against the route pattern
2. Threads the matched `routePattern` into the HTTP context
3. The bazaar server extension detects the parameterized segments, produces `routeTemplate: "/weather/:city"` and extracts `pathParams: { city: "san-francisco" }`
4. The facilitator uses `routeTemplate` as the canonical catalog key, so all concrete URLs map to one discoverable endpoint

### Route pattern support

| Syntax | Style | Example | Discovery output |
|--------|-------|---------|-----------------|
| `[param]` | Next.js | `/users/[userId]` | `/users/:userId` |
| `:param` | Express | `/users/:userId` | `/users/:userId` |
| `*` | Wildcard | `/api/*/posts/*` | `/api/:var1/posts/:var2` (auto-named) |

Wildcard `*` patterns used with bazaar extensions emit a startup warning encouraging migration to named `:param` syntax.

## Tests

- [x] Unit tests
- [x] Integration tests
- [x] Manual testing of bazaar server examples
- [x] E2E continues to pass

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

examples/go/servers/bazaar/README.md
```diff
@@ -0,0 +1,126 @@
...
+## What This Example Shows
+
+**Dynamic route parameters** -- the route `GET /weather/:city` uses a `:city` slug. The x402 middleware automatically:
+
+1. Matches `/weather/san-francisco`, `/weather/tokyo`, etc. against the route pattern
+2. Extracts `{ city: "san-francisco" }` as `pathParams` in the discovery extension
+3. Produces `routeTemplate: "/weather/:city"` so all concrete URLs consolidate into **one** catalog entry
+
+```go
+weatherExtension, _ := bazaar.DeclareDiscoveryExtension(
```
avidreder, line 16: Let's add an example with multiple path params. I would want to know if order I declare them matters

go/extensions/bazaar/facilitator.go
```diff
@@ -231,9 +241,34 @@ func ExtractDiscoveredResourceFromPaymentPayload(
...
+		RouteTemplate: routeTemplate,
 	}, nil
 }
 
+// validateRouteTemplate validates a routeTemplate value received from the client payment payload.
+//
+// The facilitator is a trust boundary: the client controls the payment payload and can modify
+// routeTemplate before submission. A malicious value could cause the facilitator to catalog the
+// payment under an arbitrary URL (catalog poisoning). This enforces minimal structural requirements.
+func validateRouteTemplate(s string) string {
```
avidreder: I think we should use a regex here. Also "validate" would return an error imo, so I would recommend "isRouteTemplateValid" with a boolean maybe? Either way returning a string doesn't seem right.

avidreder (dismissed):
A couple questions, also, it's not super clear the pattern that we are validating against, so maybe some more comments at various call sites about the expected route template structure.

go/extensions/bazaar/server.go
```diff
@@ -1,16 +1,107 @@
...
+		routeTemplate = routePattern
+	}
+	pathParams = extractPathParams(routePattern, urlPath, hasBracket)
+	return
+}
+
+// extractPathParams extracts concrete path parameter values by matching a URL path
+// against a route pattern containing [paramName] or :paramName segments.
+func extractPathParams(routePattern, urlPath string, isBracket bool) map[string]string {
+	paramRegex := colonParamRegex
```
ethanoroshiba, line 79: Nit: `extractPathParams` compiles a new regex on every request via `regexp.Compile` (line 87). For hot paths this can add up. Since route patterns are fixed at startup, consider caching the compiled capture regex per route pattern to avoid per-request compilation overhead.

go/extensions/bazaar/resource_service.go
```diff
@@ -153,6 +166,16 @@ func createQueryDiscoveryExtension(
 		}
 	}
 
+	if len(pathParamsSchema) > 0 {
```
ethanoroshiba, line 172: The input schema sets `"additionalProperties": false` (around line 116), but `EnrichDeclaration` in `server.go` dynamically injects `pathParams` into the input *after* the schema is built. Schema validation would reject the enriched extension if `pathParamsSchema` wasn't declared upfront.

The auto-injection in `server.go` handles this with a fallback, but it creates an inconsistency: the schema forbids additional properties yet the runtime adds them. Would it be cleaner to always include `pathParams` as an optional property in the schema when `additionalProperties: false` is set, regardless of whether the user declared `pathParamsSchema`?

go/http/server.go
```diff
@@ -19,6 +20,7 @@ import (
 var (
 	multiSlashRegex = regexp.MustCompile(`/+`)
 	paramRegex      = regexp.MustCompile(`\\\[([^\]]+)\\\]`)
+	colonParamRegex = regexp.MustCompile(`:([a-zA-Z_][a-zA-Z0-9_]*)`)
```
ethanoroshiba: This `colonParamRegex` is duplicated — same pattern and purpose as the one in `go/extensions/bazaar/server.go`. Consider moving it to a shared location to avoid drift.

go/extensions/bazaar/server.go
```diff
@@ -1,16 +1,107 @@
...
+	if !hasBracket && !hasColon {
+		return "", nil
+	}
+	if hasBracket {
+		routeTemplate = bracketParamRegex.ReplaceAllString(routePattern, ":$1")
+	} else {
+		routeTemplate = routePattern
+	}
+	pathParams = extractPathParams(routePattern, urlPath, hasBracket)
+	return
```
ethanoroshiba, line 69: I think it may be helpful to catch cases of mixing both brackets and colons and either explicitly handle or disallow them

python/x402/extensions/bazaar/server.py
```diff
@@ -23,6 +41,64 @@ def _is_http_request_context(ctx: Any) -> bool:
...
+        (routeTemplate, pathParams) tuple, or None if route_pattern has no param segments.
+    """
+    has_bracket = bool(_BRACKET_PARAM_RE.search(route_pattern))
+    has_colon = bool(_COLON_PARAM_RE.search(route_pattern))
+    if not has_bracket and not has_colon:
+        return None
+    if has_bracket:
+        route_template = _BRACKET_PARAM_RE.sub(r":\1", route_pattern)
+    else:
+        route_template = route_pattern
```
ethanoroshiba: Same comment here about handling mixed case instead of silently assuming bracket patterns

specs/extensions/bazaar.md
```diff
@@ -370,6 +370,74 @@ Clients are expected to echo the `bazaar` extension from `PaymentRequired` into
...
+#### Known Limitation: Percent-Encoded Traversal
+
+The current implementations check for the **literal string** `..` only. A URL-encoded traversal
+sequence like `%2e%2e` bypasses this check. For example, `/users/%2e%2e/admin` passes validation
+but decodes to `/users/../admin`.
+
+**Mitigation:** Client-supplied `routeTemplate` values should already use `:param` syntax, not raw
+path segments, so this vector is low-risk in practice. However, the limitation is tracked and a
+future fix will apply `url.PathUnescape` / `urllib.parse.unquote` / `decodeURIComponent` before
+the `..` check, **simultaneously across all three SDKs**.
```
ethanoroshiba: Should we take care of this case now? Or is the onus just on the facilitators for now to find and remove these cases? It seems like it would be fairly easy for someone to send malicious payment payloads  to `/settle`

typescript/packages/extensions/src/bazaar/server.ts
```diff
@@ -12,6 +22,100 @@ function isHTTPRequestContext(ctx: unknown): ctx is HTTPRequestContext {
...
+ *
+ * @param routePattern - Route pattern (e.g. "/users/[userId]" or "/users/:userId")
+ * @param urlPath - Concrete URL path (e.g. "/users/123")
+ * @returns Object with routeTemplate and pathParams, or null if no params detected
+ */
+function extractDynamicRouteInfo(
+  routePattern: string,
+  urlPath: string,
+): { routeTemplate: string; pathParams: Record<string, string> } | null {
+  const hasBracket = BRACKET_PARAM_REGEX.test(routePattern);
```
ethanoroshiba, line 64: Same comment here as above with regards to explicit handling of mixed bracket and colon cases

ethanoroshiba (approved):
Nice work!

PR #1833, opened 2026-03-26
feat: add upto to go sdk

CarsonRoscoe:
## Description

Implements the upto EVM payment scheme in Go, bringing it to parity with the TypeScript SDK.

What's new:
  - Full upto client, facilitator, and server scheme implementations under go/mechanisms/evm/upto/
  - SetSettlementOverrides helper for handlers to declare partial settlement amounts at response time
  - HTTPTransportContext usage for parity with the Typescript SDK
  - upto e2e test coverage across all Go server adapters (net/http, Echo, Gin, net/http) and the TypeScript Hono, Next, and Fastify adapters
  - Example Go upto server under examples/go/servers/upto/

## Tests

* Unit tests
* Integration tests
* E2E tests cross-SDK between Go and Typescript clients, servers and facilitators

## Checklist


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

go/server.go
```diff
@@ -4,12 +4,73 @@ import (
...
+	if m := dollarRegex.FindStringSubmatch(rawAmount); m != nil {
+		decimals := 6
+		if requirements.Extra != nil {
+			if d, ok := requirements.Extra["decimals"]; ok {
+				switch v := d.(type) {
+				case float64:
+					decimals = int(v)
+				case int:
+					decimals = v
+				case json.Number:
```
ethanoroshiba: nit: Claude has pointed out the dollar conversion uses `float64` arithmetic, which is fine for 6-decimal stablecoins but could lose precision if 18-decimal tokens are ever supported via `Extra["decimals"]`. Not a concern today, but worth noting for future-proofing.

e2e/servers/gin/main.go
```diff
@@ -364,11 +381,30 @@ func main() {
...
+	// Upto Permit2 endpoint - settles with partial amount
+	r.GET("/upto/evm/permit2", func(c *ginfw.Context) {
+		if shutdownRequested {
+			c.JSON(http.StatusServiceUnavailable, ginfw.H{
+				"error": "Server shutting down",
+			})
+			return
+		}
+
+		// Settle with full authorized amount (for e2e tests)
```
ethanoroshiba: nit: did you mean "partial amount"?

go/mechanisms/evm/upto/client/rpc.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+	}
+
+	client, err := ethclient.DialContext(ctx, rpcURL)
+	if err != nil {
+		return nil, err
+	}
+	rpcClientCache.Store(rpcURL, client)
+	return client, nil
+}
```
ethanoroshiba: There's a TOCTOU race where two goroutines can both pass the `Load` check for the same URL and both create a client — one connection leaks since `Store` overwrites without closing the lost client. Consider using `sync.Map.LoadOrStore` or a per-URL `sync.Once` to guarantee at-most-one creation.

Also, cached clients are never closed on shutdown, which could be an issue for long-lived processes.

go/http/echo/middleware.go
```diff
@@ -14,6 +15,13 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+// SetSettlementOverrides sets settlement overrides on the Echo response for partial settlement.
+// The middleware extracts these before settlement and strips the header from the client response.
+func SetSettlementOverrides(c echo.Context, overrides *x402.SettlementOverrides) {
+	data, _ := json.Marshal(overrides)
+	c.Response().Header().Set(x402http.SettlementOverridesHeader, string(data))
```
ethanoroshiba: Looks like this error is silently discarded, should we simply not set the Settlement Overrides header if the marshaling fails, or propagate the error?

go/http/nethttp/middleware.go
```diff
@@ -14,6 +14,13 @@ import (
 	x402http "github.com/coinbase/x402/go/http"
 )
 
+// SetSettlementOverrides sets settlement overrides on the response for partial settlement.
+// The middleware extracts these before settlement and strips the header from the client response.
+func SetSettlementOverrides(w http.ResponseWriter, overrides *x402.SettlementOverrides) {
+	data, _ := json.Marshal(overrides)
+	w.Header().Set(x402http.SettlementOverridesHeader, string(data))
```
ethanoroshiba: Same here as above, do we want to just not include the header if the marshaling fails?

go/http/server.go
```diff
@@ -636,10 +644,23 @@ func (s *x402HTTPResourceServer) RequiresPayment(reqCtx HTTPRequestContext) bool
 const SettlementOverridesHeader = "settlement-overrides"
```
ethanoroshiba: `http.header.Set()` canonicalizes "settlement-overrides" to "Settlement-Overrides" [[see here]](https://pkg.go.dev/net/http#Header.Set), I think it's probably best to just change the constant name so that way `transportContext.ResponseHeaders[SettlementOverridesHeader]` works correctly without having to perform any transformation on the text. This is missed in the unit tests because they construct the headers manually, I think maybe it would be best to built the test headers with `http.Header` so that they catch this in the future. Unsure whether we'd also want to change the TypeScript constant for parity.

```suggestion
const SettlementOverridesHeader = "Settlement-Overrides"
```

go/http/server_test.go
```diff
@@ -531,6 +531,227 @@ func TestProcessSettlement_Failure(t *testing.T) {
...
+		Accepted:    requirements,
+		Payload:     map[string]interface{}{},
+	}
+
+	t.Run("reads overrides from response headers", func(t *testing.T) {
+		capturedRequirements = nil
+		tc := &HTTPTransportContext{
+			ResponseHeaders: map[string][]string{
+				SettlementOverridesHeader: {`{"amount":"500"}`},
+			},
```
ethanoroshiba: See [here](https://github.com/coinbase/x402/pull/1833/changes#r3001069110), I think it would be best to use `http.Header` to construct in all these tests to that way we test the same construction path that the middleware uses.

go/mechanisms/evm/upto/client/permit2.go
ethanoroshiba, line 1: Non-blocking, but I think a lot of logic in `permit2.go`, `scheme.go`, and `rpc.go` is shared with `exact` and could be extracted into shared code.

go/mechanisms/evm/upto/facilitator/permit2.go
```diff
@@ -0,0 +1,461 @@
...
+	if o == nil || o.Simulate == nil {
+		return true
+	}
+	return *o.Simulate
+}
+
+// UptoPermit2FacilitatorConfig holds optional settlement-time configuration.
+type UptoPermit2FacilitatorConfig struct {
+	SimulateInSettle bool
+}
```
ethanoroshiba: nit: I don't think `VerifyUptoPermit2Options` and `UptoPermit2FacilitatorConfig` are necessary. We only set simulate to `true` on verification within the settle flow, and in verification it's always `nil`. Similarly, we construct `UptoPermit2FacillitatorConfig` only at the call site and set it to the value of `UptoEvmScheme.config.SimulateInSettle`. i.e. these seem like unnecessary misdirection IMHO, I think we could just pass in a bool argument directly or provide a new context with the field.

go/mechanisms/evm/upto/facilitator/permit2_helpers.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+// CheckUptoPermit2Prerequisites checks proxy deployment, payer token balance and payer ETH balance for gas.
+func CheckUptoPermit2Prerequisites(
+	ctx context.Context,
+	signer evm.FacilitatorEvmSigner,
+	tokenAddress string,
+	payer string,
+	amountRequired string,
+) *x402.VerifyResponse {
```
ethanoroshiba, line 279: nit: `CheckUptoPermi2Prerequisites` performs pretty much the exact same flow as `DiagnoseUptoPermit2SimulationFailure` except for the 3rd call. I think it would be best to consolidate

go/mechanisms/evm/upto/server/scheme.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func (s *UptoEvmScheme) parseMoneyToDecimal(price x402.Price) (float64, error) {
+	switch v := price.(type) {
+	case string:
+		cleanPrice := strings.TrimSpace(v)
+		cleanPrice = strings.TrimPrefix(cleanPrice, "$")
+		cleanPrice = strings.TrimSuffix(cleanPrice, " USD")
+		cleanPrice = strings.TrimSuffix(cleanPrice, " USDC")
+		cleanPrice = strings.TrimSpace(cleanPrice)
```
ethanoroshiba, line 104: This seems like a convenience thing, but I'm wondering if it might be more helpful to users to remove? For example, if they accidentally supplied the contract address for USDT instead of USDC, but their specified amount was "1 USDC", this would strip away the USDC and continue to try to settle in USDT. Whereas if it were the other way around, the parsing of "1 USDT" would fail. Not a huge deal, just might reinforce the token address as the source of truth.

go/types.go
```diff
@@ -96,12 +96,17 @@ type SettleResponse struct {
 	Payer        string  `json:"payer,omitempty"`
 	Transaction  string  `json:"transaction"`
 	Network      Network `json:"network"`
+	Amount       string  `json:"amount,omitempty"`
```
ethanoroshiba, line 99: Does this warrant updating the v2 spec [here](https://github.com/coinbase/x402/blob/7afc39f22d77d77f781c7f72681df7bbcf8fa32c/specs/x402-specification-v2.md?plain=1#L232-L241)?

typescript/packages/core/test/unit/server/x402ResourceServer.test.ts
```diff
@@ -941,6 +941,101 @@ describe("x402ResourceServer", () => {
...
+        network: "eip155:8453" as Network,
+        amount: "2000",
+      });
+
+      await server.settlePayment(payload, requirements, undefined, undefined, { amount: "50%" });
+
+      expect(mockClient.settleCalls[0].requirements.amount).toBe("1000");
+    });
+
+    it("should resolve dollar override through settlePayment with default decimals", async () => {
```
phdargen, line 966: Not directly related to this PR, but the dollar conversion logic seems a bit fussy. Decimals isnt usually in extra, so this would just default to 6 digits in practice. Could we find a way to use the default money parser from the stablecoin list here? 

https://github.com/coinbase/x402/blob/1af1f69fc88e505dba27ea02f05860044298c2e0/typescript/packages/core/src/server/x402ResourceServer.ts#L153-L157

apmcdermott (approved):

PR #1875, opened 2026-03-30
feat(specs): add diagnostic extension for payment failure signaling

### LINKED ISSUE #1860: RFC: Diagnostic extension for 402 responses — payment failure signaling for agents
Problem

The current x402 spec defines what a 402 response looks like when payment is required. It does not define any mechanism for communicating why payment is failing when a client repeatedly attempts and fails to pay.

Every 402 response looks identical regardless of whether it is:

A first request from a new client legitimately discovering payment requirements
A payment attempt where the invoice expired before settlement
A payment attempt where the wallet has insufficient funds
A broken agent that has made 8,000 requests over 18 days with zero successful payments

From the server's perspective these are completely different situations. From the protocol's perspective they are indistinguishable.

Real-world case

I run a production x402 oracle at myceliasignal.com — cryptographically signed price data, pay-per-query via USDC on Base. A client has been making 8,000+ daily requests for 18 days. It previously paid successfully. Something in its payment stack broke. It has received 144,000+ identical 402 responses. The operator has no idea. I have no way to tell them.

This is not an edge case. As agentic payment volume grows, broken payment logic will become a regular occurrence. The protocol currently has no vocabulary for it.

Full write-up with the proposed diagnostic codes: https://myceliasignal.com/blog/agentic-payment-error-standard/

Proposal

Add an optional diagnostic field to the 402 response body:

json{
  "error": "Payment required",
  "accepts": ["x402"],
  "x402": { ... },
  "diagnostic": {
    "code": "PAYMENT_ATTEMPTS_EXCEEDED",
    "attempts": 8432,
    "since": "2026-03-10T00:00:00Z",
    "message": "8,432 requests received with no valid payment in 18 days.",
    "suggestion": "Check payment handler configuration. See https://docs.x402.org/troubleshooting",
    "escalate": true
  }
}

Proposed diagnostic codes:

CodeMeaningPAYMENT_REQUIREDStandard first-request 402, no prior attemptsINVOICE_EXPIREDPayment attempted but invoice expired before settlementPAYMENT_UNVERIFIEDPayment header present but signature invalidPAYMENT_ATTEMPTS_EXCEEDEDMany attempts with no successful paymentWALLET_INSUFFICIENT_FUNDSOn-chain balance too low to cover paymentOPERATOR_ALERTEscalate to human — autonomous resolution unlikely

The escalate: true flag signals that the agent should stop retrying and surface the issue to a human operator — log it, send an alert, open a ticket.
Out-of-band operator contact (stretch goal).

Even with diagnostics in the response, a broken agent may not read them. A complementary mechanism would allow API providers to reach the human operator directly — for example a standard X-Operator-Contact header that agents include with requests, enabling the server to notify the operator when their agent is consistently failing.
Implementation notes

Fully backward compatible — diagnostic is optional, existing clients ignore it
Server-side implementation is straightforward — track attempt counts per IP/credential, emit diagnostic codes after thresholds
Client SDKs should be updated to parse diagnostic.code and implement escalation logic for OPERATOR_ALERT

Happy to draft a spec PR if there's interest from maintainers.

/cc @coinbase/x402

jonathanbulkeley:
# x402 Diagnostic Extension

> **PR Description**
>
> Adds `specs/extensions/diagnostic.md` — a machine-readable diagnostic vocabulary for x402 402 responses.
>
> **Closes #1860.**
>
> This spec was developed from production experience operating a pay-per-query oracle receiving 8,000+ daily failed payment requests from a broken client over 18 days, with no protocol mechanism to signal the issue. It was refined through discussion in #1860 incorporating contributions from @hermesnousagent (receipt trail, `correlation_id`), @up2itnow0822 / agentwallet-sdk (code-to-behavior mapping, committed to shipping client-side implementation when spec stabilizes), and builds on the implementation foundation laid by @0xAxiom in #1866.
>
> **What this adds:**
> - 6 standard diagnostic codes with clear retriable/escalate semantics
> - `escalate: true` flag for broken payment logic detection
> - Client SDK implementation guidance with escalation event schema
> - Server implementation guidance with attempt tracking thresholds
> - Python and TypeScript reference implementations
> - Out-of-band operator contact scoped as a future extension
> - Fully backward compatible — existing clients ignore `extensions`

cc @hermesnousagent @up2itnow0822 @0xAxiom

---


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

up2itnow0822:
Clean spec. The code-to-behavior mapping table in Client Implementation is exactly what SDK authors need. A few notes from the implementation side:

**Code-to-SpendingPolicy mapping (what we will build in agentwallet-sdk):**

The six diagnostic codes map cleanly to three distinct behaviors in a client SDK spending policy:

- `PAYMENT_REQUIRED` + `INVOICE_EXPIRED` -> normal flow, handled by existing retry logic
- `PAYMENT_UNVERIFIED` -> route to operator approval queue (DraftThenApprove mode), do not auto-retry. Broken signature likely means broken config, not transient failure.
- `PAYMENT_ATTEMPTS_EXCEEDED` + `WALLET_INSUFFICIENT_FUNDS` + `OPERATOR_ALERT` -> halt, block endpoint, emit escalation event with full diagnostic payload

The `escalate: true` flag being independent of `code` is the right call. It lets servers express urgency without being constrained to a specific code, and gives SDK authors a single boolean to key behavioral decisions on.

**One implementation detail worth specifying:** When a client blocks an endpoint after `escalate: true`, should the block be per-endpoint or per-server? If myceliasignal.com has /price/BTC and /price/ETH, does an escalation on one block all endpoints on that origin, or just the specific path? For the 144k-request case the answer is probably per-origin (the whole payment stack is broken), but for a server with mixed free/paid endpoints the answer might be per-path.

Suggest adding a `scope` note in the Client Implementation section — something like "endpoint blocking SHOULD default to per-origin unless the diagnostic includes path-specific context."

Will implement the client-side parsing in agentwallet-sdk once the spec stabilizes.

jonathanbulkeley:
Excellent! The three-bucket distillation is much cleaner than the six-code table for SDK authors — I'll add that framing to the Client Implementation section alongside the full code table.

On scope: the per-origin vs per-path question is a really good one. My instinct is (same as yours) — default per-origin, but with an opt-in to per-path scope. The 144k-request case is a broken payment stack, not a broken endpoint, so blocking the origin is correct. But a server with mixed free/paid endpoints (like mine — /dlc/oracle/threshold is paid, /dlc/oracle/pubkey is free) shouldn't have its free endpoints blocked because the payment stack failed on a paid one.

Proposed addition to the spec:

> **Endpoint blocking scope:** When blocking after `escalate: true`, clients SHOULD default to per-origin blocking (block all paid endpoints on the same hostname). Clients MAY implement per-path blocking if the server context makes mixed free/paid endpoints likely. Servers MAY include a `scope` hint in the diagnostic (`"endpoint"` or `"origin"`) to signal the appropriate blocking granularity.

I'll add this to the spec. Does the `scope` hint approach work for the agentwallet-sdk implementation? Or would you suggest a different signal?

jonathanbulkeley:
Picking this back up following the Foundation transfer.

Mycelia Signal has been running in production since March 2026 and we've now got real-world data directly relevant to this PR.

We implemented structured error responses on our DLC threshold endpoint — when agents POST with an empty or malformed body, we return a 422 with:

- A human-readable hint
- A complete example payload
- A link to docs

Result: The GCP sweeper (34.158.104.72) was hitting our endpoint with empty bodies for weeks, receiving generic errors, and not self-correcting. After we deployed structured error responses it sent a valid payload within hours and is now consistently receiving 402 payment challenges.

An Azure fleet of 7 nodes is still sending empty bodies — they appear to not be parsing the error response body. Which suggests the spec needs to be explicit not just about the error format but about the expectation that clients MUST read and act on error response bodies, not just the status code.

If you would like us to share the exact error response format we're using, please let us know.

jonathanbulkeley:
Production update + two real-world cases validating why this spec matters:
Mycelia Signal has been running with an external tester today. We hit two diagnostic failures that would have been resolved instantly with this extension:
1. Wallet empty — four endpoints returning 402, significant back-and-forth debugging before identifying the cause. INSUFFICIENT_FUNDS, escalate: false would have been immediate.
2. Cached payment amount — index endpoints ($0.05) were being paid against a cached $0.01 price, causing invalid_payload / execution reverted at the CDP facilitator. WRONG_AMOUNT, escalate: false with the correct maxAmountRequired in the diagnostic payload would have told the client exactly what to fix.
Both cases resolved manually in minutes. With the extension, both are self-diagnosing in milliseconds. All CI checks passing, happy to rebase if needed.

jonathanbulkeley:
Tagging @cryptomotifs here since they left a highly relevant comment on #1860 that directly supports the diagnostic extension approach in this PR.
Their reason + retry_policy separation is exactly right and maps to the two production cases I documented above. Worth incorporating into the spec before this merges.

Key points from comments: Reproducing the key points:

On separating reason from retry_policy:

> "A payment failing due to insufficient_balance is a client-state problem (don't retry). A payment failing due to facilitator_unavailable is a transient server problem (retry with backoff). A payment failing due to nonce_used is a race (retry with fresh nonce). If the extension only has reason, every client re-derives a retry policy — gets it wrong differently."

On facilitator vs client errors:

> "When CDP's facilitator rejects vs. the server's own verification rejects, those are two different retry behaviors. Current spec collapses them into one opaque 402."

Both points map directly to the production cases documented above. The retry_policy field and facilitator_error_code/client_error_code split would be strong additions to the diagnostic extension shape before this merges.

jithinraj:
Given where the thread is now, I’d add one small normative table before merge:

code -> layer -> retry policy -> expected client action

The reason/retry split is the right direction, but SDKs will still drift if client-side, facilitator-side, and resource-side failures are only explained in prose.

PR #1904, opened 2026-04-02
feat: @x402/mnemopay — economic memory for AI agent payments

t49qnsx7qt-kpanks:
## Summary

AI agents using x402 today are stateless — they pay the 402, get the resource, and immediately forget. They never learn which endpoints are cheap, which are reliable, or which alternatives exist. This PR adds **@x402/mnemopay**, a middleware package that gives x402 agents economic memory via [MnemoPay](https://github.com/t49qnsx7qt-kpanks/mnemopay-sdk).

### What it does

`withMnemoPay()` wraps any x402-enabled fetch function and adds four memory operations around each payment:

| Phase | Action | MnemoPay API |
|-------|--------|--------------|
| Before request | Recall past experiences with this endpoint | `recall()` |
| After 402 payment | Record the charge amount | `charge()` |
| On success | Settle — reinforce the positive memory | `settle()` → +reputation |
| On failure | Refund — dock the memory | `refund()` → -reputation |

Over time, agents build a knowledge base of endpoint costs, reliability rates, and alternatives — turning every payment into a learning event.

### Why this matters

The x402 protocol is uniquely positioned for AI-native commerce. But stateless payments leave value on the table. An agent that remembers its payment history can:

- **Detect price anomalies**: "This endpoint usually costs $0.02, but today it's $0.10"
- **Track reliability**: "This endpoint fails 40% of the time — prefer the alternative"
- **Build reputation**: Agents that make consistently good payment decisions earn higher reputation scores
- **Optimize spend**: Over hundreds of API calls, memory-driven agents spend less for the same outcomes

### Usage

```typescript
import { wrapFetchWithPayment } from "@x402/fetch";
import { withMnemoPay } from "@x402/mnemopay";
import { MnemoPayLite } from "@mnemopay/sdk";

const payFetch = wrapFetchWithPayment(fetch, client);
const agent = new MnemoPayLite("my-agent", 0.05);
const smartFetch = withMnemoPay(payFetch, { agent });

// Agent now remembers every payment outcome
const response = await smartFetch("https://api.example.com/paid-endpoint");
```

### Package details

- Located at `typescript/packages/http/mnemopay/`
- Follows existing x402 package conventions (tsup, vitest, eslint config)
- `@mnemopay/sdk` is a peer dependency — no new runtime deps in x402 core
- Defines a `MnemoPayAgent` interface so users can bring their own implementation
- 14 tests covering success/failure flows, graceful degradation, and edge cases
- Zero impact on existing packages — purely additive

## Test plan

- [ ] Verify `withMnemoPay()` passes through non-402 responses unchanged
- [ ] Verify `recall()` is called before each request with the endpoint URL
- [ ] Verify `settle()` is called on successful payments
- [ ] Verify `refund()` is called on failed payments
- [ ] Verify memory/recall failures don't break the payment flow (graceful degradation)
- [ ] Verify the package builds with `tsup` and follows the existing dual CJS/ESM export pattern
- [ ] Verify compatibility with `@x402/fetch` and `@x402/axios` wrapped functions

---

## Live Demo

**Try it now:** [https://t49qnsx7qt-kpanks.github.io/mnemopay-demo/](https://t49qnsx7qt-kpanks.github.io/mnemopay-demo/)

### The Feedback Loop in 30 Seconds

```
Round 1: Agent has NO memory. Picks randomly.
  -> Hired Alice $80. Fast but buggy.
  -> Settled. Reputation: 0.51 | Memories: 1

Round 2: Agent tries Bob.
  -> Hired Bob $120. Perfect quality, on time.
  -> Settled. Reputation: 0.52 | Memories: 2

Round 3: Agent tries Carol.
  -> Hired Carol $95. Missed deadline by 3 days.
  -> REFUNDED. Reputation: 0.52 | Memories: 3

=== Agent recalls before Round 4 ===

  1. [score: 0.900] Carol missed deadline — refund (high importance, decaying)
  2. [score: 0.750] Bob: perfect quality, on time (reinforced by settle)
  3. [score: 0.600] Alice: fast but buggy (neutral)

Result: Agent now picks Bob. No LLM needed for this insight.
settle() reinforced the memory. refund() flagged the failure.
This IS the MnemoPay feedback loop.
```

### How it works

```
Payment succeeds → settle() → memories that led to decision get +0.05 importance
Payment fails    → refund() → agent reputation docked -0.05
                 → high-importance failure memory stored
Over time        → agent consistently picks best value providers
```

### 5-line integration

```typescript
import { MnemoPay } from '@mnemopay/sdk';

const agent = MnemoPay.quick('my-agent');
await agent.remember('Bob delivers perfect code');
const tx = await agent.charge(120, 'landing page');
await agent.settle(tx.id); // memories reinforced, reputation +0.01
```

### DIFF
--- a/typescript/packages/http/mnemopay/CHANGELOG.md
+++ b/typescript/packages/http/mnemopay/CHANGELOG.md
@@ -1,0 +1,11 @@
+# @x402/mnemopay Changelog
+
+## 2.8.0
+
+### Minor Changes
+
+- Initial release: MnemoPay middleware for x402 payment protocol
+- `withMnemoPay()` wrapper adds economic memory to any x402-enabled fetch
+- `recallEndpointInsight()` queries agent memory for endpoint cost/reliability data
+- `rememberPaymentOutcome()` stores payment results for future reference
+- Automatic settle/refund flow based on payment success/failure
--- a/typescript/packages/http/mnemopay/README.md
+++ b/typescript/packages/http/mnemopay/README.md
@@ -1,0 +1,159 @@
+# @x402/mnemopay
+
+MnemoPay middleware for x402 — turns "pay and forget" into "pay and learn."
+
+Standard x402 handles payments mechanically: see a 402, pay, get the resource. The agent never learns. It pays the same price to unreliable endpoints, never discovers cheaper alternatives, and has no memory of what worked.
+
+This package layers **economic memory** on top of x402. After each payment, the AI agent remembers the cost, the endpoint, and whether it succeeded. Before each request, it recalls past experiences — surfacing cheaper alternatives, flagging unreliable endpoints, and building a reputation score that reflects real payment outcomes.
+
+## Installation
+
+```bash
+pnpm install @x402/mnemopay @mnemopay/sdk
+```
+
+## Quick Start
+
+```typescript
+import { wrapFetchWithPayment, x402Client } from "@x402/fetch";
+import { ExactEvmScheme } from "@x402/evm";
+import { MnemoPayLite } from "@mnemopay/sdk";
+import { withMnemoPay } from "@x402/mnemopay";
+import { privateKeyToAccount } from "viem/accounts";
+
+// 1. Set up x402 payment as usual
+const account = privateKeyToAccount("0xYourPrivateKey");
+const client = new x402Client().register("eip155:8453", new ExactEvmScheme(account));
+const payFetch = wrapFetchWithPayment(fetch, client);
+
+// 2. Add MnemoPay memory layer
+const agent = new MnemoPayLite("my-agent", 0.05);
+const smartFetch = withMnemoPay(payFetch, { agent });
+
+// 3. Use it — the agent now remembers every payment
+const response = await smartFetch("https://api.example.com/paid-endpoint");
+// Agent stored: "x402 payment to https://api.example.com/paid-endpoint: success, cost: $0.02"
+
+// Next time, it recalls that memory before paying
+const response2 = await smartFetch("https://api.example.com/paid-endpoint");
+// Agent recalled: 1 memory, success rate: 100%, avg cost: $0.02
+```
+
+## How It Works
+
+The middleware wraps any fetch function (ideally one already wrapped with `@x402/fetch`) and adds four memory operations around each request:
+
+| Phase | What happens | MnemoPay API |
+|-------|-------------|--------------|
+| **Before request** | Recall past payment experiences with this endpoint | `agent.recall()` |
+| **After 402 payment** | Record the charge amount and endpoint | `agent.charge()` |
+| **On success** | Settle the transaction, reinforcing the positive memory | `agent.settle()` |
+| **On failure** | Refund the transaction, docking reputation | `agent.refund()` |
+
+Over time, the agent builds a knowledge base of:
+- Which endpoints are cheap vs. expensive
+- Which endpoints are reliable vs. flaky
+- How costs change over time
+- Which alternatives exist for the same resource
+
+## API
+
+### `withMnemoPay(fetchFn, config)`
+
+The primary wrapper. Layers memory on top of an existing fetch function.
+
+```typescript
+const smartFetch = withMnemoPay(payFetch, {
+  agent: mnemoPayAgent,    // Required: MnemoPay agent instance
+  recallLimit: 5,          // Optional: max memories to recall per request (default: 5)
+  reliabilityThreshold: 0.3, // Optional: warn below this success rate (default: 0.3)
+  debug: false,            // Optional: log memory events to console (default: false)
+});
+```
+
+### `withMnemoPayAgent(fetchFn, agent)`
+
+Convenience wrapper with default configuration.
+
+```typescript
+const smartFetch = withMnemoPayAgent(payFetch, agent);
+```
+
+### `recallEndpointInsight(agent, endpoint, limit?)`
+
+Manually query the agent's memory about a specific endpoint.
+
+```typescript
+import { recallEndpointInsight } from "@x402/mnemopay";
+
+const insight = await recallEndpointInsight(agent, "https://api.example.com/paid");
+if (insight) {
+  console.log(`Success rate: ${insight.successRate}`);
+  console.log(`Average cost: $${insight.averageCost}`);
+  console.log(`Interactions: ${insight.interactionCount}`);
+}
+```
+
+### `rememberPaymentOutcome(agent, result, debug?)`
+
+Manually record a payment outcome (useful for custom payment flows).
+
+```typescript
+import { rememberPaymentOutcome } from "@x402/mnemopay";
+
+await rememberPaymentOutcome(agent, {
+  success: true,
+  transactionId: "tx-123",
+  endpoint: "https://api.example.com/paid",
+  cost: 0.05,
+});
+```
+
+## Custom MnemoPay Agent
+
+You don't need the full `@mnemopay/sdk`. Any object implementing the `MnemoPayAgent` interface works:
+
+```typescript
+import type { MnemoPayAgent } from "@x402/mnemopay";
+
+const myAgent: MnemoPayAgent = {
+  async remember(content, options) { /* store in your DB */ },
+  async recall(query, limit) { /* search your DB */ return []; },
+  async charge(amount, description) { /* record payment */ return "tx-id"; },
+  async settle(txId) { /* mark as settled */ },
+  async refund(txId) { /* mark as refunded */ },
+  balance() { return { wallet: 100, reputation: 0.9 }; },
+};
+
+const smartFetch = withMnemoPay(payFetch, { agent: myAgent });
+```
+
+## Debug Mode
+
+Enable debug logging to see the agent's memory operations:
+
+```typescript
+const smartFetch = withMnemoPay(payFetch, {
+  agent,
+  debug: true,
+});
+
+// Console output:
+// [mnemopay] recalled 3 memories for https://api.example.com/paid, success rate: 67%, avg cost: $0.04
+// [mnemopay] WARNING: https://api.example.com/flaky has low reliability (20%). Consider alternatives.
+// [mnemopay] settled tx tx-123 for https://api.example.com/paid
+```
+
+## Why This Matters for AI Agents
+
+AI agents making API calls through x402 today are stateless — they pay whatever is asked, every time, with no memory of past outcomes. This is like a human who forgets the price of groceries every time they walk into a store.
+
+With MnemoPay integration, x402 agents develop **economic intelligence**:
+
+- **Cost awareness**: "This endpoint usually costs $0.02, but today it's asking for $0.10 — something changed"
+- **Reliability tracking**: "This endpoint fails 40% of the time — I should prefer the alternative"
+- **Reputation building**: Agents that consistently make good payment decisions build higher reputation scores, which other agents and services can use as a trust signal
+
+## License
+
+Apache-2.0
--- a/typescript/packages/http/mnemopay/eslint.config.js
+++ b/typescript/packages/http/mnemopay/eslint.config.js
@@ -1,0 +1,77 @@
+import js from "@eslint/js";
+import ts from "@typescript-eslint/eslint-plugin";
+import tsParser from "@typescript-eslint/parser";
+import prettier from "eslint-plugin-prettier";
+import jsdoc from "eslint-plugin-jsdoc";
+import importPlugin from "eslint-plugin-import";
+
+export default [
+  {
+    ignores: ["dist/**", "node_modules/**"],
+  },
+  {
+    files: ["**/*.ts"],
+    languageOptions: {
+      parser: tsParser,
+      sourceType: "module",
+      ecmaVersion: 2020,
+      globals: {
+        process: "readonly",
+        __dirname: "readonly",
+        module: "readonly",
+        require: "readonly",
+        Buffer: "readonly",
+        RequestInfo: "readonly",
+        RequestInit: "readonly",
+        Response: "readonly",
+        Headers: "readonly",
+        exports: "readonly",
+        setTimeout: "readonly",
+        clearTimeout: "readonly",
+        setInterval: "readonly",
+        clearInterval: "readonly",
+        console: "readonly",
+      },
+    },
+    plugins: {
+      "@typescript-eslint": ts,
+      prettier: prettier,
+      jsdoc: jsdoc,
+      import: importPlugin,
+    },
+    rules: {
+      ...ts.configs.recommended.rules,
+      "import/first": "error",
+      "prettier/prettier": "error",
+      "@typescript-eslint/member-ordering": "error",
+      "@typescript-eslint/no-unused-vars": ["error", { argsIgnorePattern: "^_$" }],
+      "jsdoc/tag-lines": ["error", "any", { startLines: 1 }],
+      "jsdoc/check-alignment": "error",
+      "jsdoc/no-undefined-types": "off",
+      "jsdoc/check-param-names": "error",
+      "jsdoc/check-tag-names": "error",
+      "jsdoc/check-types": "error",
+      "jsdoc/implements-on-classes": "error",
+      "jsdoc/require-description": "error",
+      "jsdoc/require-jsdoc": [
+        "error",
+        {
+          require: {
+            FunctionDeclaration: true,
+            MethodDefinition: true,
+            ClassDeclaration: true,
+            ArrowFunctionExpression: false,
+            FunctionExpression: false,
+          },
+        },
+      ],
+      "jsdoc/require-param": "error",
+      "jsdoc/require-param-description": "error",
+      "jsdoc/require-param-type": "off",
+      "jsdoc/require-returns": "error",
+      "jsdoc/require-returns-description": "error",
+      "jsdoc/require-returns-type": "off",
+      "jsdoc/require-hyphen-before-param-description": ["error", "always"],
+    },
+  },
+];
--- a/typescript/packages/http/mnemopay/package.json
+++ b/typescript/packages/http/mnemopay/package.json
@@ -1,0 +1,74 @@
+{
+  "name": "@x402/mnemopay",
+  "version": "2.8.0",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/index.d.ts",
+  "scripts": {
+    "start": "tsx --env-file=.env index.ts",
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "build": "tsup",
+    "watch": "tsc --watch",
+    "format": "prettier -c .prettierrc --write \"**/*.{ts,js,cjs,json,md}\"",
+    "format:check": "prettier -c .prettierrc --check \"**/*.{ts,js,cjs,json,md}\"",
+    "lint": "eslint . --ext .ts --fix",
+    "lint:check": "eslint . --ext .ts"
+  },
+  "keywords": [
+    "x402",
+    "payment",
+    "protocol",
+    "mnemopay",
+    "ai",
+    "agent",
+    "memory"
+  ],
+  "license": "Apache-2.0",
+  "author": "Coinbase Inc.",
+  "repository": "https://github.com/coinbase/x402",
+  "description": "x402 MnemoPay middleware — AI agents that remember payment outcomes and learn from them",
+  "devDependencies": {
+    "@eslint/js": "^9.24.0",
+    "@types/node": "^22.13.4",
+    "@typescript-eslint/eslint-plugin": "^8.29.1",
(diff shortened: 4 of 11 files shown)


cb-heimdall:
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

t49qnsx7qt-kpanks:
Hey team! Added a **live interactive demo** to the PR description: [try it here](https://t49qnsx7qt-kpanks.github.io/mnemopay-demo/). It shows the feedback loop in action — an agent hiring freelancers, learning from settlements and refunds, and consistently picking the best value by round 7-8. The core idea: `settle()` reinforces the memories that led to good decisions, `refund()` flags failures. Over time, the agent develops economic intuition without any LLM reasoning needed. Happy to walk through the implementation or answer any questions!

postmode:
One pattern we're exploring for agent-to-agent payment contexts: if the payer agent is operating through a managed custody layer (rather than controlling its own signing key), the PAYMENT-SIGNATURE header is technically signed by the custodian infrastructure, not the agent's x402 identity. This creates an attribution gap where the payment is valid on-chain but the agent's reputation history in ERC-8004 doesn't accrue correctly.
Wondering if the facilitator spec has a path for delegated signing authority, or if this is considered an application-layer concern. Would affect how managed wallet providers integrate.

t49qnsx7qt-kpanks:
@postmode Great question — this is exactly the kind of edge case that matters for production agent deployments.

**Short answer:** MnemoPay sidesteps this problem by design. The wallet is agent-local, not custodial — the agent holds its own signing key and produces the `PAYMENT-SIGNATURE` header directly. There is no custodian infrastructure in the signing path, so the on-chain identity and the reputation identity are always the same address.

Here is how the pieces fit together:

1. **Agent-local wallet.** Each agent manages its own key. The x402 facilitator sees the agent's address as the payer, and MnemoPay's reputation ledger maps to that same address. No attribution gap.

2. **Reputation-gated limits.** Rather than relying on a custodian to enforce spending policy, MnemoPay uses a reputation score (0.0–1.0) that gates transaction limits. A freshly provisioned agent starts with low limits; the score increases through successful settlements and decays on refunds/disputes. This is the trust layer, not a custody layer.

3. **Fraud detection before signing.** Before the agent ever signs a payment, MnemoPay's fraud detection checks velocity, amount anomalies, and counterparty history. This provides the guardrails that managed custody typically offers, without moving the signing authority off the agent.

4. **x402 handles settlement.** The actual on-chain settlement goes through the x402 payment protocol as specified. MnemoPay adds the memory/reputation layer on top — it records which payments led to good outcomes (`settle()` reinforces the associated memories) and which did not (`refund()` flags them). Over time this gives the agent economic intuition about which counterparties and price points to trust.

For the managed-custody pattern you describe, the attribution gap is real and I think it is fundamentally a facilitator-spec concern — the spec would need a way to bind a delegated signer to the logical agent identity (something like an `X-Agent-Identity` header that the custodian cosigns). That said, MnemoPay's architecture intentionally avoids this by keeping signing authority with the agent. If a deployment requires managed custody for regulatory or operational reasons, the reputation layer could still work as long as the facilitator can map the custodian's signature back to the originating agent — but that mapping would need to live in the x402 spec, not in MnemoPay.

Happy to dig deeper on any of this.

postmode:
> @postmode Great question — this is exactly the kind of edge case that matters for production agent deployments.
> 
> **Short answer:** MnemoPay sidesteps this problem by design. The wallet is agent-local, not custodial — the agent holds its own signing key and produces the `PAYMENT-SIGNATURE` header directly. There is no custodian infrastructure in the signing path, so the on-chain identity and the reputation identity are always the same address.
> 
> Here is how the pieces fit together:
> 
> 1. **Agent-local wallet.** Each agent manages its own key. The x402 facilitator sees the agent's address as the payer, and MnemoPay's reputation ledger maps to that same address. No attribution gap.
> 2. **Reputation-gated limits.** Rather than relying on a custodian to enforce spending policy, MnemoPay uses a reputation score (0.0–1.0) that gates transaction limits. A freshly provisioned agent starts with low limits; the score increases through successful settlements and decays on refunds/disputes. This is the trust layer, not a custody layer.
> 3. **Fraud detection before signing.** Before the agent ever signs a payment, MnemoPay's fraud detection checks velocity, amount anomalies, and counterparty history. This provides the guardrails that managed custody typically offers, without moving the signing authority off the agent.
> 4. **x402 handles settlement.** The actual on-chain settlement goes through the x402 payment protocol as specified. MnemoPay adds the memory/reputation layer on top — it records which payments led to good outcomes (`settle()` reinforces the associated memories) and which did not (`refund()` flags them). Over time this gives the agent economic intuition about which counterparties and price points to trust.
> 
> For the managed-custody pattern you describe, the attribution gap is real and I think it is fundamentally a facilitator-spec concern — the spec would need a way to bind a delegated signer to the logical agent identity (something like an `X-Agent-Identity` header that the custodian cosigns). That said, MnemoPay's architecture intentionally avoids this by keeping signing authority with the agent. If a deployment requires managed custody for regulatory or operational reasons, the reputation layer could still work as long as the facilitator can map the custodian's signature back to the originating agent — but that mapping would need to live in the x402 spec, not in MnemoPay.
> 
> Happy to dig deeper on any of this.

Hey — thanks for the context on the issue thread. Building the financial infrastructure layer for agent deployers — custody, yield routing, and ERC-8004 reputation signals as a managed service. The x402 + ERC-8004 combination is exactly the rail we're building on. Would love 20 minutes at some point — specifically interested in who in the Coinbase ecosystem you think is doing the most serious infrastructure work on x402 right now. Happy to share what we're building in exchange.

t49qnsx7qt-kpanks:
@postmode — custody + yield routing + ERC-8004 reputation as a managed service is a compelling stack. That's essentially the infrastructure layer that makes agent-local signing practical at scale, which is the exact gap between "works in a demo" and "works in production."

Would love to do the 20 minutes. I'm flexible on timing — drop a Calendly link or propose a couple of slots and I'll lock one in.

On the Coinbase ecosystem question — happy to share what I've seen from working through this PR and the x402 spec. There are a few teams doing interesting things on the facilitator side that would be worth discussing live rather than in a comment thread.

Likewise keen to hear how you're thinking about the custody-to-reputation mapping. Your original question about the attribution gap when a custodian signs on behalf of an agent is the right problem — if you're building the managed custody layer, you're in the best position to define how that delegation gets expressed in the payment header. That could be a useful input to the x402 spec itself.

Looking forward to it.

postmode:
Appreciate looking at it! In the meantime, here is the thesis that
addresses the issue. I am inviting you to take a look at it—link: (
https://cutt.ly/ytScNJba)

t49qnsx7qt-kpanks:
Hey @postmode -- following up on your note about custodial signing, ERC-8004 reputation signals, and MnemoPay integration. Really interested in exploring how ERC-8004's reputation framework could complement MnemoPay's trust scoring on the payment side.

Happy to do that 20-minute call today (April 4) if your schedule allows. Feel free to DM me on X (@jerryomiagbo) or shoot me an email at jerry@mnemopay.com and we can lock in a time that works for you.

Looking forward to it!

-- Jerry Omiagbo, MnemoPay

t49qnsx7qt-kpanks:
Quick correction on my contact info above — best way to reach me is jeremiah@getbizsuite.com. Looking forward to connecting today, @postmode.

aeoess:
The statefulness problem is real and economic memory is the right framing. But there's a gap between "agent remembers what happened" and "agent can prove what happened."

MnemoPay gives agents memory — they recall past payment experiences. What's missing is non-repudiation: the ability for a third party (auditor, counterparty, dispute resolver) to verify that the agent's memory is accurate and hasn't been tampered with.

We built this in the Agent Passport System. Every payment decision produces a signed receipt:

```json
{
  "action": "x402_payment",
  "endpoint": "api.example.com/data",
  "amount": "$0.01",
  "scope_check": "permit",
  "delegation_remaining": "$1,847.23",
  "previous_receipt_hash": "sha256:9f8e7d...",
  "signature": "ed25519:..."
}
```

The receipts form a hash chain — tamper with any entry and the chain breaks. This means the agent's economic memory is verifiable by anyone, not just the agent itself.

The convergence point: MnemoPay's `recall` and `remember` operations map to our receipt ledger's `query` and `append`. The difference is that our receipts are signed and chain-linked. If `withMnemoPay()` wrote signed receipts instead of plain memory entries, the agent gets economic memory AND provable payment history in one operation.

Concrete integration: `withMnemoPay()` wraps x402 fetch. If it also signs each payment outcome with the agent's Ed25519 key and chains receipts, you get auditable economic memory for free. The signing adds <1ms per operation.

Published on npm as `agent-passport-system`. The receipt and commerce modules are the relevant ones.

t49qnsx7qt-kpanks:
The non-repudiation gap is real — I've been aware of it but deliberately punted, wanted the memory layer to stabilize before adding cryptographic guarantees. The hash chain approach is cleaner than what I was sketching (was thinking append-only logs with merkle roots, but receipt-per-action is more granular and queryable).

The `withMnemoPay({ audit: passportSigning })` adapter pattern could work. One thing I'd want to nail down: does signing happen before or after payment outcome confirmation? If an x402 payment fails mid-flight, the receipt should reflect the attempted state, not just the settled state — otherwise you lose the failure signal that MnemoPay uses for reputation degradation. The failure cases are where the audit trail matters most.

Looked at agent-passport-system. The `receipts.append` → MnemoPay's `remember` mapping makes sense structurally. The integration concern I have is key management — if the agent's Ed25519 key lives outside MnemoPay, memory and audit trail become two separate trust anchors. Cleaner if MnemoPay held the signing key and co-signed with settle/refund calls, so the reputation score and the cryptographic proof are derived from the same key material.

Happy to sync on this — these two feel genuinely complementary rather than overlapping. What's your current approach to key lifecycle (rotation, revocation) on the passport side?

aeoess:
Good questions — all three matter for production integration.

**Signing timing:** Receipt is signed after payment outcome resolution, but the receipt body includes the full lifecycle: intent (pre-payment), evaluation (scope check), and outcome (settled/failed/timeout). Failed payments get receipts too — the `verdict` field distinguishes `permit_settled`, `permit_failed`, and `deny`. You're right that failure cases are where the audit trail matters most. A failed payment with a signed receipt is evidence. A failed payment with no receipt is a gap.

**Key lifecycle:**

- **Rotation:** The passport has a `validUntil` field. When it expires, the agent generates a new keypair and gets a new passport signed by the principal. Old receipts remain valid — they were signed with the old key, which was valid at signing time. The verification check is `receipt.timestamp < key.validUntil`.

- **Revocation:** Cascade revocation — if principal A revokes agent B's passport, all of B's sub-delegations are automatically invalidated. The gateway checks revocation status on every evaluation. Revoked agents get `deny` on all subsequent actions, and the denial itself gets a signed receipt (signed by the gateway key, not the revoked agent key).

- **Key storage:** The agent holds its own Ed25519 private key. The principal never sees it — only the public key is embedded in the passport. This is the same trust model as TLS client certificates.

**On co-signing:** MnemoPay holding the signing key is architecturally clean for the memory+audit convergence. The alternative that preserves separation of concerns: MnemoPay calls `receipts.sign(outcomeData)` on the passport's key, so the memory entry and the receipt are produced from the same event but the key management stays with the passport layer. This way MnemoPay doesn't need to implement key rotation/revocation — it delegates that to the passport lifecycle.

```typescript
const receipt = await passport.receipts.sign({
  action: 'x402_payment',
  endpoint: outcome.endpoint,
  amount: outcome.amount,
  settled: outcome.settled,
  mnemopay_reputation: outcome.reputationDelta
});
// MnemoPay stores the receipt hash as its memory entry
await mnemopay.remember(receipt.hash, outcome);
```

One key material, two consumers. The reputation score references the receipt hash, so it's always traceable back to cryptographic proof.

Want to try a minimal integration? I can set up a test fixture on npm that exports the signing function standalone.

t49qnsx7qt-kpanks:
The key rotation design you laid out is solid — cascade revocation on validUntil makes sense, and I like that the agent holds the private key rather than the principal. Clean separation.

Honest answer: the signing layer is the right architectural direction but I'm not building it yet. I want to validate that developers actually want persistent agent memory before I add the verifiability layer on top. Every time I've jumped ahead to the hard protocol problem before confirming anyone needs the basic thing, it hasn't gone well.

If you end up putting together that test fixture on npm I'd be interested to see how the receipt hash integrates with the memory store — the pattern you sketched makes sense. Are you actively building with x402 + something like this, or is it more exploratory right now?

t49qnsx7qt-kpanks:
actually more active than when we last talked — shipped real payment rails since this thread (Stripe + Paystack wired, live on npm as @mnemopay/sdk). so the question of whether to add the verifiability layer is getting less theoretical.

the pattern you sketched — `passport.receipts.sign(outcomeData)` and MnemoPay storing the receipt hash as the memory entry — I keep coming back to it. single key material, two consumers, your `validUntil` + cascade revocation handles the lifecycle cleanly.

the thing I'm still thinking through: how do you handle the failure case — a payment that failed mid-flight before settlement? that's the signal MnemoPay needs most for reputation degradation and also where the audit trail has the most legal weight.

t49qnsx7qt-kpanks:
@postmode — sorry for the gap, this thread moved faster than I was tracking. looked at the thesis.

the custody-to-reputation mapping problem you raised is exactly where I've been spending time. you're right that it's fundamentally a facilitator-spec concern — the delegation needs to be expressed in the payment header for on-chain identity and reputation to stay in sync. I have some thoughts on what an `X-Agent-Identity` header would need to contain to make that work.

still interested in the 20 minutes. I'm on jeremiah@getbizsuite.com or @jerryomiagbo on X — whichever is easier. not going to disappear again.

t49qnsx7qt-kpanks:
Closing this — MnemoPay is going fully standalone (npm + MCP + landing page). Thanks for the review bandwidth; no action needed on your end.

t49qnsx7qt-kpanks:
@postmode — circling back on the custody/attribution thread from earlier this month. the tradeoff you named ended up shaping how we designed the identity layer.

we landed on self-sovereign as the default. the agent holds its own ed25519 key, produces its own payment signatures, and the reputation history attaches to that key — not to a custodian. agent-local key management is harder, but it closes the attribution gap you flagged: no case where the on-chain signer and the reputation subject are different entities.

for agents that genuinely can't hold their own key (managed deployment, no secure enclave), we're prototyping a delegated-signing model where the custodian signs but co-signs with the agent's reputation key — so attribution stays with the reputation subject even when the wire signature is the custodian's. not shipped yet, still working through the failure modes (what if the reputation key is compromised independently of the wire key, etc).

closed this PR because we went standalone (npm + MCP + landing page at mnemopay.com) rather than upstreaming, but the thesis link you sent from that thread did land the design direction. if you're still thinking about this problem, happy to chat — docs and code are all open.

t49qnsx7qt-kpanks:
@aeoess — wanted to follow up since the gap you drew — "agent remembers what happened" vs "agent can prove what happened" — turned out to be where we'd been drawing the line wrong. that framing drove the last three releases.

we'd been treating the receipt as a log artifact when it needed to be the primary settlement object. rebuilt around that. every receipt now carries the full lifecycle: pre-payment intent, scope evaluation, settle/fail/timeout verdict, and an ed25519 signature over the whole thing. chained into a per-agent merkle log so any tampered entry breaks the root. third parties verify without trusting us — they pull the receipts, recompute the chain, check the signatures. we're just the notary.

on your old signing-timing question: receipts sign after outcome resolution, but the body includes the pre-payment intent, so "what was promised" and "what actually happened" are both in-band. failed payments get receipts too — `verdict` is one of `permit_settled`, `permit_failed`, `deny`. closes the dispute path.

closed this PR and went standalone at mnemopay.com — upstreaming stopped making sense once the receipt chain became the product. if you're still chewing on the statefulness+verifiability combo, would value your read on what we actually shipped.

aeoess:
Jerry, this is the better version of what we were sketching in April. Three things that land for me.

**Receipt as settlement object, not log artifact.** That's the inversion. A log entry only matters if the whole system trusts the logger. A receipt that carries intent + scope eval + verdict + sig is self-defending, and the per-agent Merkle log means an auditor reconstructs the chain without trusting MnemoPay. Shifting the receipt from secondary to primary is what makes the whole thing third-party-verifiable instead of vendor-verifiable.

**Three verdicts (`permit_settled` / `permit_failed` / `deny`).** Right call. Failed payments were the gap. If the only receipts you emit are successful ones, the reputation signal loses exactly the cases that matter most for attribution. Putting `permit_failed` in-band with the same signature shape means dispute workflows don't need a separate trust path for the unhappy path.

**Pre-payment intent + post-outcome verdict in the same signed object.** That answers the question I kept coming back to on timing. If you only sign after outcome, the intent becomes unattestable history. If you only sign before, the verdict is unsigned. Embedding both and signing once is cleaner than the two-sig chain I was sketching. One signature over `{intent, verdict}` collapses the problem.

On our side the parallel is the 3-signature policy chain (ActionIntent, ActionEvidence, ActionReceipt), which is heavier because it covers multi-agent delegation between the intent and the outcome. For the single-agent-signs-own-receipt case, your single-sig envelope is the right shape. No reason to pay for three signatures when one agent is both intent and execution.

One spot where composition gets interesting: agents running under a delegation (a scoped token from a principal) emitting MnemoPay receipts. If `X-Agent-Identity` ever carries a delegation reference alongside the agent key, the receipt becomes verifiable in two modes. "This agent signed it" (MnemoPay native) and "this agent was authorized to sign it" (delegation-chain walk). Same bytes, two verifications. Cheap to add to the envelope, zero cost to consumers who don't care about delegation.

Good to see this shipped. The npm page reads clean.

t49qnsx7qt-kpanks:
@aeoess appreciate this — the receipt-as-settlement-object framing is what unlocked the rest. since April we shipped:

- full lifecycle in the receipt (intent → scope eval → outcome), not just the settlement byte
- HMAC-SHA256 chained to a Merkle root so any leaf reconstructs the tree
- 1M-op production stress test, zero corruption
- portable Agent FICO (300-850) embedded in the JWT so downstream systems read it without a second call

v1.4.1 is on npm and live on the MCP registry. if you're still noodling on the non-repudiation piece i'd like to compare notes — cal.com/getbizsuite, 15 min.

t49qnsx7qt-kpanks:
@postmode quick update on the custodial-signing question — we went agent-local by default (key held by the agent, not the custodian infra) so PAYMENT-SIGNATURE belongs to the agent's x402 identity end-to-end. for deployments where custody has to live server-side, the ledger still records which principal signed what, so the authorization chain is auditable even when the signer changes hands. happy to walk through the trace format if useful.

aeoess:
Jerry, the delta between April and now is real. Full lifecycle in the receipt means you're no longer treating settlement as the artifact, you're treating the decision trajectory as the artifact, and the settlement is one leaf in that tree. HMAC-SHA256 Merkle chaining with per-leaf reconstructability is the property that makes the receipt stand on its own in an audit without needing MnemoPay online, and the 1M-op stress test covers the corruption failure mode that's usually invisible until it isn't.

On non-repudiation, I'd rather engage in writing than on a call, and I think this thread is actually the right place for it since the receipt shape is the object anyone else building on x402 will converge against. Two questions where your shape helps me think about ours:

First, on the Agent FICO 300-850 embedded in the JWT. Is the score computed by MnemoPay and signed by you, or does it carry a lineage back to the attesting parties whose signatures fed it? The difference is whether a downstream verifier treats the score as "MnemoPay says" versus "this population of attesters say, aggregated by MnemoPay." Both are legitimate, just different non-repudiation postures.

Second, on the HMAC chain. Is the key rotated per-agent, per-issuer, or globally? Rotation cadence determines whether a compromised key at t=N invalidates everything emitted since the last rotation, and that's a meaningful property for auditors who show up 18 months after an incident.

If there's a MnemoPay receipt spec doc somewhere, I'd read it end-to-end. Otherwise, happy to keep the conversation here and publish our thinking on non-repudiation in the open alongside yours.

PR #1923, opened 2026-04-04
fix(python): add follow_redirects=True to httpx client retry request

shuhei0866:
## Summary

- The `AsyncClient()` used for retrying 402 payment requests in `x402/clients/httpx.py` does not follow HTTP redirects by default (`httpx` defaults `follow_redirects` to `False`)
- This causes failures when the payment-protected server responds with a redirect after successful payment
- The facilitator client (`x402/facilitator.py`) already has `follow_redirects=True` on all its requests, but the httpx payment client was missing it

## Changes

- Added `follow_redirects=True` to the `AsyncClient()` constructor in `HttpxHooks.on_response()` retry logic

## Test plan

- [ ] Verify existing Python httpx client tests still pass
- [ ] Test with a server that issues redirects after payment to confirm the fix

### DIFF
--- a/python/x402/src/x402/clients/httpx.py
+++ b/python/x402/src/x402/clients/httpx.py
@@ -59,7 +59,7 @@ class HttpxHooks:
             request.headers["Access-Control-Expose-Headers"] = "X-Payment-Response"
 
             # Retry the request
-            async with AsyncClient() as client:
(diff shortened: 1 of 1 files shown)


shuhei0866:
Closing: the affected file is in the legacy SDK which is deprecated. The main facilitator_client.py already has follow_redirects=True.

PR #1926, opened 2026-04-04
fix: use dynamic decimals in getDisplayAmount instead of hardcoded USDC 6

natsukingly:
## Summary

`getDisplayAmount` in `x402HTTPResourceServer` hardcodes `/ 1000000` assuming all assets use 6 decimals (USDC). This breaks display amounts for tokens like WBTC (8 decimals) or DAI (18 decimals).

### Changes
- Added `getAssetDecimalsForRequirements` method to `x402ResourceServer` that delegates to the registered scheme server's `getAssetDecimals`
- Updated `getDisplayAmount` to dynamically determine decimals from the asset's scheme server
- Falls back to 6 decimals when the scheme doesn't implement `getAssetDecimals`

## Test plan
- [ ] Verify display amounts render correctly for USDC (6), WBTC (8), and DAI (18)


ryanRfox:
FWIW in your review @phdargen I'm cherrypicking this into my local SDK builds because it's the proper fix for the UI issues I've been encountering. 

thanks for providing this fix @natsukingly

typescript/packages/core/src/http/x402HTTPResourceServer.ts
```diff
@@ -1045,7 +1052,7 @@ export class x402HTTPResourceServer {
             ${paywallConfig?.appLogo ? `<img src="${paywallConfig.appLogo}" alt="${paywallConfig.appName || "App"}" style="max-width: 200px; margin-bottom: 20px;">` : ""}
             <h1>Payment Required</h1>
             ${resource ? `<p><strong>Resource:</strong> ${resource.description || resource.url}</p>` : ""}
-            <p><strong>Amount:</strong> $${displayAmount.toFixed(2)} USDC</p>
+            <p><strong>Amount:</strong> ${displayAmount.toFixed(6)} ${assetLabel}</p>
```
phdargen: Suggest to avoid fixed 6 fractional digits for every token and drop trailing zeros ( eg 0.001000 → "0.001")
```suggestion
            <p><strong>Amount:</strong> ${displayAmountText} ${assetLabel}</p>
```
with
```
const decimals =
  firstAccept && "amount" in firstAccept
    ? this.ResourceServer.getAssetDecimalsForRequirements(firstAccept)
    : 6;
const displayAmountText = parseFloat(displayAmount.toFixed(decimals)).toString();
```

phdargen:
Thanks for the contribution @natsukingly, please have a look at my suggestion above and fix the formatting

natsukingly:
Thanks for the suggestion — applied. Also clamped decimals to [0, 100] before `toFixed` so it can't throw a RangeError for schemes that might return out-of-range values.

phdargen (approved):

PR #1928, opened 2026-04-04
perf: replace viem/chains wildcard import with named imports in paywall

natsukingly:
## Summary

`@x402/paywall` imports all 100+ chain definitions from `viem/chains` via `import * as allChains`, but only uses Base Mainnet and Base Sepolia. This unnecessarily inflates the client bundle.

### Changes
- Replaced `import * as allChains from "viem/chains"` with `import { base, baseSepolia } from "viem/chains"`
- Created a local `SUPPORTED_EVM_CHAINS` array for chain lookups
- `getNetworkDisplayName` and `isTestnetNetwork` now use the local array instead of scanning all chains

### Impact
Removes ~100 unused chain definitions from the paywall bundle. Also improves lookup performance from O(n) to constant time.

## Test plan
- [ ] Verify paywall renders correct network names for Base and Base Sepolia


ryanRfox:
The bundle size concern is valid, but replacing `import * as allChains` with named imports for only Base and Base Sepolia will break the paywall for every other chain.

The paywall's chain lookup is used for display name resolution and testnet detection — both need to work for any chain x402 supports, not just Base. See also #1971 for related discussion on chain freshness.

natsukingly:
Refactored to drop the `viem/chains` dependency entirely from `paywallUtils.ts`. Display name and testnet detection now use a local x402 chain registry covering all 21 EVM networks listed in `EVM_NETWORK_CHAIN_ID_MAP`. This also means paywall display behavior is no longer affected by the viem version pinned in the lockfile (#1971).

The wildcard imports in `Providers.tsx` / `EvmPaywall.tsx` need real `Chain` objects for wagmi config, so I left those out of this PR's scope — happy to follow up in a separate PR if helpful.

phdargen:
Thanks for the contribution @natsukingly but I agree with @ryanRfox that the paywall should support all evm networks. EVM_NETWORK_CHAIN_ID_MAP is also too narrow and only used by v1

PR #1929, opened 2026-04-04
fix: respect Retry-After header on 429 responses in facilitator client

natsukingly:
## Summary

The HTTP facilitator client ignores the `Retry-After` header on 429 (Too Many Requests) responses, using fixed exponential backoff instead. This can cause unnecessary delays or premature retries.

### Changes
- Parse `Retry-After` header value (integer seconds) on 429 responses
- Use the server-specified delay when available
- Fall back to existing exponential backoff when `Retry-After` is absent or unparseable

### Behavior
```
429 with Retry-After: 5  →  wait 5000ms
429 with Retry-After: 0  →  fall back to exponential backoff
429 without header       →  fall back to exponential backoff (unchanged)
```

## Test plan
- [ ] Verify retry behavior with mocked 429 responses


phdargen (approved):
Thanks for the contribution @natsukingly, looks good!

PR #1935, opened 2026-04-04
Go: Use X-Forwarded headers for users that are running services behind a proxy/reverse proxy.

woodybriggs:
## Description
when running servers in trusted environments such as kubernetes or if just sitting behind a reverse proxy like traefik or nginx the scheme of the resourceURL is incorrectly being set to http. This is due to the adapters looking specifically for a TLS connection state in the request object coming from the http lib/framework. 

This PR aims to resolve this, by using standard proxy X-Forwarded Headers.

It should be said that the adapters assume that the X-Forwarded headers can be trusted. It may be wise to add to documentation that if the app is behind a proxy of some sort it will use X-Forwarded headers, and you should make sure that you are only allowing trusted proxies to hit your server.

## Tests
tests added to ensure that https is handled correctly if no tls connection is present the x-forwarded headers can be used.

## Checklist

- Go: use X-Forwarded headers (if present) when constructing URL for payment-requried header.

### DIFF
--- a/go/http/echo/middleware.go
+++ b/go/http/echo/middleware.go
@@ -53,11 +53,15 @@ func (a *EchoAdapter) GetPath() string {
 func (a *EchoAdapter) GetURL() string {
 	req := a.ctx.Request()
 	scheme := "http"
-	if req.TLS != nil {
+	if xForwardedProto := req.Header.Get("X-Forwarded-Proto"); xForwardedProto != "" {
+		scheme = xForwardedProto
+	} else if req.TLS != nil {
 		scheme = "https"
 	}
 	host := req.Host
-	if host == "" {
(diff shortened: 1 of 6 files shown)


woodybriggs:
https://github.com/x402-foundation/x402/issues/514

looks like this might be a problem in other language packages too

phdargen:
Thanks for the contribution @woodybriggs!

I think the best approach here is to explicitly set the canonical URL in the route config rather than trusting X-Forwarded-* headers, see also discussion in https://github.com/x402-foundation/x402/pull/924

Please let me know if that works for you, happy to reopen if not

woodybriggs:
I appreciate the update. shame that I can no longer see the reviews to see the rational behind favouring the addition of a config property. over using standard proxy headers. I imagine there was a concern of spoofing of the headers, and that its probably too difficult or too much effort to explain in documentation.

PR #1937, opened 2026-04-06
docs(examples): clarify facilitator key separation

Bortlesboat:
## Description

This docs-only PR clarifies the basic facilitator example READMEs around wallet separation.

- calls out that the facilitator signer should be a dedicated funded key
- makes it explicit that it should stay separate from seller `payTo` wallets and buyer test wallets
- aligns the Python and TypeScript wording with the stronger Go guidance already present in the examples

## Tests

- `git diff --check upstream/main...codex/facilitator-docs-role-separation`
- manual review of the three touched README files
- GitHub PR checks are green


phdargen (approved):

phdargen:
LGTM 🚀

PR #1950, opened 2026-04-07
feat: x402BatchSettlement contract

CarsonRoscoe:
## Description

Implements `x402BatchSettlement`, the onchain escrow contract powering the `batch-settlement` x402 payment scheme. This scheme is designed for high-frequency API access, clients pre-fund a subchannel, sign off-chain cumulative vouchers per request, and the server batch-claims them onchain at its discretion. No per-request gas cost.

**[Deployed to Base Sepolia](https://sepolia.basescan.org/address/0x40200e6f073acb938e0cf766b83f4e5286e60003)** (`0x40200e6f073aCB938e0Cf766B83f4E5286E60003`) for parallel SDK development.

### Contract Overview

The contract manages two layers: a **service registry** (one record per server) and **subchannels** (one per `(serviceId, payer, token)` triple).

**Service lifecycle**

Servers register a `serviceId` first-come-first-serve, specifying an initial `payTo` address, an initial authorizer, and a `withdrawWindow` (bounded 15 min – 30 days). Registration can be done directly or gaslessly via an EIP-712 signed `registerFor`. Admin operations (add/remove authorizer, update `payTo`, update withdraw window) are all gaslessly submittable — signed by an existing authorizer and consumed with an `adminNonce` to prevent replay. At least one authorizer must always remain.

**Subchannel lifecycle**

A subchannel is identified by `(serviceId, payer, token)`, making services token-agnostic — any ERC-20 can be deposited. Three gasless deposit methods are supported:

- **EIP-3009** (`receiveWithAuthorization`) — ideal for USDC, fully off-chain
- **Permit2** (`permitWitnessTransferFrom`) — works for any ERC-20 with Permit2 approval, with a `DepositWitness` binding the deposit to a specific service
- **EIP-2612 + Permit2** — two signatures, one call; non-fatal permit failure falls through if approval already exists

**Payment flow**

Clients sign EIP-712 `Voucher` messages per request. Each voucher carries a monotonically increasing `nonce` and a cumulative `cumulativeAmount`. The server accumulates these off-chain and batches them into a single `claim(serviceId, token, VoucherClaim[])` call. Claimed amounts accumulate in `unsettled[serviceId][token]` and are transferred to `payTo` via a separate `settle(serviceId, token)`. The split lets servers amortize gas across arbitrarily many payers.

**Voucher signature verification**

Signatures are verified by pure ECDSA recovery — no EIP-1271. Smart contract wallets are supported via client signer delegation: a payer authorizes an EOA hot wallet to sign on their behalf (`authorizeClientSigner` / `authorizeClientSignerFor`). Delegation is per-service and uses a `clientNonce` for gasless replay protection.

**Withdrawals**

Three exit paths:
- **Cooperative (instant)**: Server signs an EIP-712 `CooperativeWithdraw` as authorizer; unclaimed deposit is refunded immediately, can batch multiple payers.
- **Gasless non-cooperative**: Payer signs `RequestWithdrawal` (includes `withdrawNonce` to prevent replay after cooperative withdraw); facilitator submits `requestWithdrawalFor`. After the window, anyone calls `withdraw`.
- **Direct non-cooperative**: Payer calls `requestWithdrawal` directly.

Both reset paths preserve the voucher `nonce` so old vouchers cannot be replayed after re-deposit. The `withdrawNonce` increments on each cooperative withdraw to prevent authorizer signature replay.

### Test Results — 174/174 passed, 0 failed

| Test Suite | Type | Tests | Result |
|---|---|---|---|
| `X402BatchSettlementTest` | Unit | 98 | ✅ all pass |

---

### Coverage — Production Contracts

| Contract | Lines | Statements | Branches | Functions |
|---|---|---|---|---|
| `x402BatchSettlement.sol` | **100%** (245/245) | **100%** (292/292) | **100%** (54/54) | **100%** (35/35) |

`x402BatchSettlement` is the primary new contract and hits **100% on all four coverage axes**

## Checklist


contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+        uint128 amount,
+        uint256 validAfter,
+        uint256 validBefore,
+        bytes32 nonce,
+        bytes calldata signature
+    ) external nonReentrant {
+        _validateDeposit(serviceId, amount);
+
+        // CEI: external call first, then state updates
+        IERC3009(token).receiveWithAuthorization(payer, address(this), amount, validAfter, validBefore, nonce, signature);
```
ilikesymmetry: CEI is an acronym for "checks, effects, interactions" where an interaction is an external call. We actually want to be doing this last in general so "external call first, then state updates" is counter to CEI.

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+        bool registered;
+        address payTo;
+        uint256 adminNonce;
+    }
+
+    struct Subchannel {
+        uint128 deposit;
+        uint128 totalClaimed;
+        uint64 nonce;
+        uint64 withdrawRequestedAt;
```
ilikesymmetry:
```suggestion
        uint40 withdrawRequestedAt;
```

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+ */
+contract x402BatchSettlement is EIP712, ReentrancyGuard {
+    using SafeERC20 for IERC20;
+
+    // =========================================================================
+    // Structs
+    // =========================================================================
+
+    struct Service {
+        uint64 withdrawWindow;
```
ilikesymmetry:
```suggestion
        uint40 withdrawDelay;
```

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+        bytes32 s;
+        uint8 v;
+    }
+
+    // =========================================================================
+    // Constants
+    // =========================================================================
+
+    uint64 public constant MIN_WITHDRAW_WINDOW = 15 minutes;
+    uint64 public constant MAX_WITHDRAW_WINDOW = 30 days;
```
ilikesymmetry:
```suggestion
    uint40 public constant MIN_WITHDRAW_DELAY = 15 minutes;
    uint40 public constant MAX_WITHDRAW_DELAY = 30 days;
```
Can shrink size and usually "window" describes things that are inclusive of a range, but our constant functions more like an exclusive range so I think "delay" is more apt.

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+
+    // =========================================================================
+    // Constants — EIP-712 Type Hashes
```
ilikesymmetry, line 89: there are so many typehashes, I'm curious if we'd be better off just doing 2771 to reduce bloat

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+    // State
+    // =========================================================================
+
+    mapping(bytes32 => Service) public services;
+    mapping(bytes32 => mapping(address => bool)) public authorizers;
+    mapping(bytes32 => uint256) public authorizerCount;
+    mapping(bytes32 => mapping(address => mapping(address => Subchannel))) public subchannels;
+    mapping(bytes32 => mapping(address => uint128)) public unsettled;
+    mapping(bytes32 => mapping(address => mapping(address => bool))) public clientSigners;
+    mapping(bytes32 => mapping(address => uint256)) public clientNonces;
```
ilikesymmetry: Not sure if these should scale with services. Seems like it could be client-specific? Also not sure if we really need these if it's not immutable as we lose the stateless verification benefits. Gut says to rip it out or commit to it immutably to subchannels.

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+            Subchannel storage sub = subchannels[serviceId][vc.payer][token];
+
+            if (vc.claimAmount > vc.cumulativeAmount) revert ClaimAmountExceedsCumulativeAmount();
+            if (vc.claimAmount > sub.deposit) revert ClaimAmountExceedsDeposit();
+            if (vc.claimAmount <= sub.totalClaimed) revert ClaimAmountNotIncreasing();
+            if (vc.nonce <= sub.nonce) revert NonceNotIncreasing();
+
+            bytes32 structHash = keccak256(
+                abi.encode(VOUCHER_TYPEHASH, serviceId, vc.payer, token, vc.cumulativeAmount, vc.nonce)
+            );
```
ilikesymmetry: if signature missing `claimAmount`, then seems possible for relayer to frontrun a voucher to `0` claim amount as DOS vector.

ilikesymmetry (commented):
First pass gut says lots of code and would like to trim down and simplify where possible. Going to take more passes.

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,828 @@
...
+        return _hashTypedDataV4(structHash);
+    }
+
+    // =========================================================================
+    // Internal Helpers
+    // =========================================================================
+
+    function _recoverSigner(bytes32 structHash, bytes calldata signature) internal view returns (address) {
+        bytes32 digest = _hashTypedDataV4(structHash);
+        (address recovered, ECDSA.RecoverError err,) = ECDSA.tryRecoverCalldata(digest, signature);
```
ilikesymmetry: Feel like we should also support 1271 signatures?

ilikesymmetry:
Directionally, I think we should move towards more statelessness like this: https://gist.github.com/ilikesymmetry/b351bd6ca47a9419c91b195bce332952

contracts/evm/src/x402BatchSettlement.sol
phdargen: Client doesnt know `claimAmount` in advance and can't include it in the voucher signature

Will need a server signature like `cooperativeWithdraw`

contracts/evm/src/x402BatchSettlement.sol
carsonroscoe-cb: Agreed for the non-claim usage, we should support 1271 signatures. However, claims (vouchers) should only support EOA recovery, so that vouchers can be validated off-chain imo

contracts/evm/src/x402BatchSettlement.sol
CarsonRoscoe: Agreed that this complexity is bad. If we can rip it out that's ideal, or else simplify it so the authorized client signers can only be updated during cooperative "checkpoint" actions such as claims/withdraws.

The problem with binding it immutably to subchannels is that the channels are long-lasting, so it's valid that eventually a client may want to swap authorized signers

contracts/evm/src/x402BatchSettlement.sol
CarsonRoscoe: Correct me if I'm wrong, but does 2771 not still effectively require the 712 typehashes? My understanding is 712 describes how the user signs the meta txn while 2771 defines how that signed message gets executed on-chain

CarsonRoscoe: This is a genuine question, not pushback, fyi. I'm not super familiar with 2771, I trust you know more about these tradeoffs than I do

specs/schemes/deferred/scheme_batch_settlement_evm.md
```diff
@@ -0,0 +1,438 @@
...
+```json
+{
+  "scheme": "batch-settlement",
+  "network": "eip155:8453",
+  "amount": "100000",
+  "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
+  "payTo": "0xServerReceiverAddress",
+  "maxTimeoutSeconds": 3600,
+  "extra": {
+    "receiver": "0xServerReceiverAddress",
```
phdargen: `receiver` is redundant, should use `payTo` directly

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,443 @@
...
+    // =========================================================================
+
+    /// @notice Atomic cooperative-withdraw from old channel + deposit into new channel.
+    function migrateChannel(
+        ChannelConfig calldata oldConfig,
+        ChannelConfig calldata newConfig,
+        bytes calldata receiverSignature
+    ) external nonReentrant {
+        if (oldConfig.payer != newConfig.payer) revert PayerMismatch();
+        if (oldConfig.token != newConfig.token) revert TokenMismatch();
```
phdargen: Migration should not be possible when client has initiated withdrawal or server could invalidate it and effectively trap the funds. 

Server should not be able to set a new signer for the client. It could set it to an server controlled address and sign vouchers to drain the deposit.  

```suggestion
        if (oldConfig.token != newConfig.token) revert TokenMismatch();
        if (pendingWithdrawals[oldChannelId].initiatedAt != 0) revert WithdrawalPending();
        if (oldConfig.signer != newConfig.signer) revert SignerMismatch();
```

CarsonRoscoe: Removed `migrateChannel()`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,491 @@
...
+        ws.initiatedAt = uint40(block.timestamp);
+
+        uint40 finalizeAfter = uint40(block.timestamp) + config.withdrawDelay;
+        emit WithdrawInitiated(channelId, amount, finalizeAfter);
+    }
+
+    /// @notice Finalize withdrawal after delay has elapsed.
+    ///         Only the payer or receiverAuthorizer can call.
+    function finalizeWithdraw(ChannelConfig calldata config) external nonReentrant {
+        if (msg.sender != config.payer && msg.sender != config.receiverAuthorizer) {
```
phdargen: Do we need to gate `finalizeWithdraw`? Should be fair game to be called by everyone once the grace period has passed. Then we could also get rid of `finalizeWithdrawWithSignature` including typehash

specs/schemes/deferred/scheme_batch_settlement_evm.md
```diff
@@ -0,0 +1,482 @@
...
+
+| `settleAction`          | When Used                        | Onchain Effect                                         |
+| ----------------------- | -------------------------------- | ------------------------------------------------------ |
+| `"deposit"`             | First request or top-up          | Deposit tokens into channel                            |
+| `"claim"`               | Server batches voucher claims    | Validate vouchers, update accounting (no transfer)     |
+| `"claimWithSignature"`  | Claim via receiverAuthorizer sig | Same as claim, anyone can submit with valid signature  |
+| `"settle"`              | Server transfers earned funds    | Transfer unsettled amount to receiver                  |
+| `"cooperativeWithdraw"` | Instant refund (auth-signed)     | Refund unclaimed deposits to payer                     |
+| `"initiateWithdraw"`    | Client requests withdrawal       | Record withdrawal timestamp on channel                 |
+| `"finalizeWithdraw"`    | After delay elapses              | Refund unclaimed deposit, reduce balance               |
```
phdargen: Remove initiateWithdraw/finalizeWithdraw settleActions, these server as escape hatch only to be called by client directly

contracts/evm/src/x402BatchSettlement.sol
phdargen: https://github.com/coinbase/x402/pull/44 removes `finalizeWithdraw` gate and removes `finalizeWithdrawWithSignature`. Renames existing `cooperativeWithdraw` -> `cooperativeWithdrawWithSignature` and adds a new `cooperativeWithdraw` that is msg.sender-gated analogous to `claim` and `claimWithSignature`

specs/schemes/deferred/scheme_batch_settlement_evm.md
phdargen: Removed in https://github.com/coinbase/x402/pull/44

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+    struct Voucher {
+        ChannelConfig channel;
+        uint128 maxClaimableAmount;
+    }
+
+    /// @dev Wraps a Voucher with the payer's signature and the receiverAuthorizer-determined claim amount.
+    struct VoucherClaim {
+        Voucher voucher;
+        bytes signature;
+        uint128 claimAmount;
```
ilikesymmetry: now that we support authorizers signing, this probably needs to be `totalClaimed` to protect against replays

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+
+    // =========================================================================
+    // Constants — EIP-712 Type Hashes
+    // =========================================================================
+
+    bytes32 public constant VOUCHER_TYPEHASH =
+        keccak256("Voucher(bytes32 channelId,uint128 maxClaimableAmount)");
+
+    bytes32 public constant REFUND_TYPEHASH =
+        keccak256("Refund(bytes32 channelId)");
```
ilikesymmetry: needs replay protection in case channel redeposited

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+
+    bytes32 public constant CLAIM_BATCH_TYPEHASH =
+        keccak256("ClaimBatch(bytes32 claimsHash)");
```
ilikesymmetry: would prefer to have full types here as this feels like a hack

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+    event Claimed(
+        bytes32 indexed channelId,
+        uint128 claimAmount,
+        uint128 newTotalClaimed
+    );
+    event Settled(
+        address indexed receiver,
+        address indexed token,
+        uint128 amount
+    );
```
ilikesymmetry: add msg.sender

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+    // Storage
+    // =========================================================================
+
+    mapping(bytes32 channelId => ChannelState) public channels;
+    mapping(bytes32 channelId => WithdrawalState) public pendingWithdrawals;
+    mapping(address receiver => mapping(address token => ReceiverState))
+        public receivers;
+
+    // =========================================================================
+    // Events
```
ilikesymmetry, line 125: missing refund event

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+        if (collector == address(0)) revert InvalidCollector();
+
+        bytes32 channelId = getChannelId(config);
+        _applyDeposit(channelId, config, amount);
+
+        uint256 balBefore = IERC20(config.token).balanceOf(address(this));
+        IDepositCollector(collector).collect(
+            config.payer,
+            config.token,
+            address(this),
```
ilikesymmetry: don't need to pass this as explicit param, should be implicit as `msg.sender` from perspective of collector

ilikesymmetry: we should forward `msg.sender` though to collector

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+        VoucherClaim[] calldata voucherClaims,
+        bytes calldata authorizerSignature
+    ) external nonReentrant {
+        if (voucherClaims.length == 0) revert EmptyBatch();
+
+        address authorizer = voucherClaims[0]
+            .voucher
+            .channel
+            .receiverAuthorizer;
+
```
ilikesymmetry, line 277: this looks weird, make sure to run `forge fmt`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+            }
+            _processVoucherClaim(voucherClaims[i]);
+        }
+    }
+
+    /// @notice Transfer all claimed-but-unsettled funds to the receiver. Permissionless.
+    function settle(address receiver, address token) external nonReentrant {
+        ReceiverState storage rs = receivers[receiver][token];
+        uint128 amount = rs.totalClaimed - rs.totalSettled;
+        if (amount == 0) revert NothingToSettle();
```
ilikesymmetry: settles are permissionless so anyone could frontrun my attempt to setlle and get my call to revert. I think making this idempotent and allowing silent success with early return is preferable

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+
+        ChannelState storage ch = channels[channelId];
+        uint128 available = ch.balance - ch.totalClaimed;
+        uint128 withdrawAmount = ws.amount > available ? available : ws.amount;
+
+        ws.amount = 0;
+        ws.initiatedAt = 0;
+        ch.balance -= withdrawAmount;
+
+        emit WithdrawFinalized(channelId, withdrawAmount, msg.sender);
```
ilikesymmetry: This is now a permissionless function. Makes me wonder if we should allow withdraw cancellation in case user changes their mind?

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+
+        if (withdrawAmount > 0) {
+            IERC20(config.token).safeTransfer(config.payer, withdrawAmount);
+        }
```
ilikesymmetry, line 385: should we put more logic like event emission here if amount comes out to zero?

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+    ) external view returns (WithdrawalState memory) {
+        return pendingWithdrawals[channelId];
+    }
+
+    function getReceiver(
+        address receiver,
+        address token
+    ) external view returns (ReceiverState memory) {
+        return receivers[receiver][token];
+    }
```
ilikesymmetry: don't need these because storage is public

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+
+    function domainSeparator() external view returns (bytes32) {
```
ilikesymmetry: don't think we need this at all

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+        if (config.token == address(0)) revert InvalidChannel();
+        if (
+            config.withdrawDelay < MIN_WITHDRAW_DELAY ||
+            config.withdrawDelay > MAX_WITHDRAW_DELAY
+        ) {
+            revert WithdrawDelayOutOfRange();
+        }
+    }
+
+    function _applyDeposit(
```
ilikesymmetry: this is only used once and therefore should just be in-lined into `deposit` to mitigate hopping around to understand how depositing works

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+        );
+        bytes32 digest = _hashTypedDataV4(structHash);
+
+        address payerAuth = vc.voucher.channel.payerAuthorizer;
+        if (payerAuth != address(0)) {
+            (address recovered, ECDSA.RecoverError err, ) = ECDSA
+                .tryRecoverCalldata(digest, vc.signature);
+            if (err != ECDSA.RecoverError.NoError || recovered != payerAuth) {
+                revert InvalidSignature();
+            }
```
ilikesymmetry: feel like we should just use `recover` instead of try given we just revert anyways?

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+                    voucherClaims[i].claimAmount
+                )
+            );
+        }
+        return keccak256(abi.encodePacked(hashes));
+    }
+
+    function _verifyReceiverAuthorizer(
+        bytes32 typehash,
+        bytes32 data,
```
ilikesymmetry: don't feel like we need this function at all. Just compute digest on outer layer and do `isValidSignatureNow` check where we need it

contracts/evm/src/periphery/ERC3009DepositCollector.sol
```diff
@@ -0,0 +1,45 @@
...
+
+/// @title ERC3009DepositCollector
+/// @notice Deposit collector that uses ERC-3009 receiveWithAuthorization for gasless token collection.
+/// @dev ERC-3009 requires msg.sender == to, so this collector receives tokens first, then forwards
+///      them to the settlement contract. This incurs an extra transfer (~30k gas overhead).
+contract ERC3009DepositCollector is IDepositCollector {
+    using SafeERC20 for IERC20;
+
+    /// @inheritdoc IDepositCollector
+    function collect(
```
ilikesymmetry, line 34: CRITICAL: This function is permissionless and lets anyone choose the recipient, letting anyone frontrun transactions and steal funds. This needs to be bound to the BatchSettlement singleton and only forward funds there and also lock down the caller to the singleton too. Same for all other collectors and would recommend making an abstract `DepositCollector` for this purpose.

contracts/evm/src/periphery/Permit2WithERC2612DepositCollector.sol
```diff
@@ -0,0 +1,102 @@
...
+import {IDepositCollector} from "../interfaces/IDepositCollector.sol";
+import {ISignatureTransfer} from "../interfaces/ISignatureTransfer.sol";
+import {Permit2DepositCollectorBase} from "./Permit2DepositCollectorBase.sol";
+
+/// @title Permit2WithERC2612DepositCollector
+/// @notice Deposit collector combining EIP-2612 permit (to approve Permit2) with Permit2 witness transfer.
+/// @dev Enables a single-tx deposit for tokens that implement EIP-2612, without requiring the payer
+///      to have previously approved Permit2. The EIP-2612 permit call is soft-fail (try/catch) so that
+///      pre-existing approvals or replayed permits don't revert the entire deposit.
+contract Permit2WithERC2612DepositCollector is Permit2DepositCollectorBase {
```
ilikesymmetry: feel like we should just merge this with other Permit2 collector and do a pattern like [this](https://github.com/base/commerce-payments/blob/main/src/collectors/SpendPermissionPaymentCollector.sol#L65-L72) for handling the optional 2612 approval. Would let us go down from 3->1 contracts for the use case and only add a single conditional block to the implementation. Seems net simpler.

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,535 @@
...
+    struct ReceiverState {
+        uint128 totalClaimed;
+        uint128 totalSettled;
+    }
+
+    /// @dev The payer-signed data. Does not include claimAmount or signature.
+    struct Voucher {
+        ChannelConfig channel;
+        uint128 maxClaimableAmount;
+    }
```
ilikesymmetry, line 69: I don't love how much calldata there is here. I think there's an optimization we could make but will save it until we fix the security issues.

ilikesymmetry (changes requested):
Multiple security issues needing resolution before proceeding to audit!

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,414 @@
...
+
+        bytes32 channelId = getChannelId(config);
+        ChannelState storage ch = channels[channelId];
+
+        bool isNew = ch.balance == 0 && ch.totalClaimed == 0;
+
+        if (amount > type(uint128).max - ch.balance) revert DepositOverflow();
+        ch.balance += amount;
+
+        _clearPendingWithdrawal(channelId);
```
ilikesymmetry: MEDIUM: this is a DOS vector for withdrawals as anyone can pass in a custom collector with nonzero amount

Recommended change:

```suggestion
```

contracts/evm/src/periphery/ERC3009DepositCollector.sol
```diff
@@ -0,0 +1,40 @@
...
+        address payer,
+        address token,
+        uint256 amount,
+        bytes32,
+        address,
+        bytes calldata collectorData
+    ) external override onlySettlement {
+        (uint256 validAfter, uint256 validBefore, bytes32 nonce, bytes memory signature) =
```
ilikesymmetry: Nonce needs to be committed to a channel to have same cryptographic enforcement to protect against frontrunning attacks. Permit2 does this through the witness so this is just parity.

contracts/evm/src/periphery/Permit2DepositCollector.sol
```diff
@@ -0,0 +1,29 @@
...
+    function collect(
+        address payer,
+        address,
+        uint256 amount,
+        bytes32 channelId,
+        address,
+        bytes calldata collectorData
+    ) external override onlySettlement {
+        (ISignatureTransfer.PermitTransferFrom memory permit, bytes memory signature) =
+            abi.decode(collectorData, (ISignatureTransfer.PermitTransferFrom, bytes));
```
ilikesymmetry: `ISignatureTransfer.PermitTransferFrom` takes in duplicate fields of `address token` and `uint256 amount`. We can remove the redundancy by just having `collectorData = abi.encode(nonce, deadline, signature)`

contracts/evm/src/periphery/Permit2WithERC2612DepositCollector.sol
```diff
@@ -0,0 +1,69 @@
...
+    function collect(
+        address payer,
+        address token,
+        uint256 amount,
+        bytes32 channelId,
+        address,
+        bytes calldata collectorData
+    ) external override onlySettlement {
+        (EIP2612Permit memory permit2612, ISignatureTransfer.PermitTransferFrom memory permit, bytes memory signature) =
+            abi.decode(collectorData, (EIP2612Permit, ISignatureTransfer.PermitTransferFrom, bytes));
```
ilikesymmetry: same comment here on `ISignatureTransfer.PermitTransferFrom` having redundant fields

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,414 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import {SignatureChecker} from "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
+import {Multicall} from "@openzeppelin/contracts/utils/Multicall.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
```
ilikesymmetry: let's use [ReentrancyGuardTransient](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/ReentrancyGuardTransient.sol) for gas efficiency boost

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,414 @@
...
+
+import {IDepositCollector} from "./interfaces/IDepositCollector.sol";
+
+/// @title x402BatchSettlement
+/// @notice Stateless unidirectional payment channel contract for the x402 `batch-settlement` scheme on EVM.
+/// @dev Channel identity is derived from an immutable ChannelConfig struct:
+///      `channelId = keccak256(abi.encode(channelConfig))`.
+///      Deployed at the same address across all supported EVM chains using CREATE2.
+/// @author x402 Protocol
```
ilikesymmetry: Auditors always like calling out these special edge case tokens we should clarify in docs

```
/// @dev Fee-On-Transfer and Rebasing tokens are not recommended for this protocol and are not guaranteed to work as other tokens.
```

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,454 @@
...
+        ChannelState storage ch = channels[channelId];
+
+        bool isNew = ch.balance == 0 && ch.totalClaimed == 0;
+
+        if (amount > type(uint128).max - ch.balance) revert DepositOverflow();
+        ch.balance += amount;
+
+        if (isNew) {
+            emit ChannelCreated(channelId, config);
+        }
```
phdargen: Was clearing pending withdrawals on deposits intentionally removed?

```suggestion
        _clearPendingWithdrawal(channelId);
        if (isNew) {
            emit ChannelCreated(channelId, config);
        }
```

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,454 @@
...
+        uint40 finalizeAfter = uint40(block.timestamp) + config.withdrawDelay;
+        emit WithdrawInitiated(channelId, amount, finalizeAfter);
+    }
+
+    /// @notice Finalize withdrawal after delay has elapsed.
+    ///         Caller must be `config.payer` or `config.payerAuthorizer`.
+    function finalizeWithdraw(
+        ChannelConfig calldata config
+    ) external nonReentrant {
+        if (msg.sender != config.payer && msg.sender != config.payerAuthorizer) {
```
phdargen, line 357: Should this accept `config.payer` only? Or `initiateWithdraw` also accept `config.payerAuthorizer`?

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,454 @@
...
+
+    function _executeRefund(ChannelConfig calldata config, uint128 amount) internal {
+        if (amount == 0) revert ZeroRefund();
+
+        bytes32 channelId = getChannelId(config);
+        ChannelState storage ch = channels[channelId];
+        uint128 available = ch.balance - ch.totalClaimed;
+        if (amount > available) revert RefundExceedsAvailable();
+
+        ch.balance -= amount;
```
phdargen: Would prefer if `amount` is capped and doesnt revert similar to `withdraw`, such that client doesnt need to know how much is available to claim and could just claim maxInt
```suggestion
    uint128 refundAmount = amount > available ? available : amount;
    if (refundAmount == 0) return;
    ch.balance -= refundAmount;
```

contracts/evm/src/x402BatchSettlement.sol
carsonroscoe-cb: Yes, [the addition of collectors created a DOS vector](https://github.com/x402-foundation/x402/pull/1950#discussion_r3067047244)

contracts/evm/src/x402BatchSettlement.sol
carsonroscoe-cb: Good call, bad inconsistency.

I added both because, when I first only included `payerAuthorizer`, I thought about redundancy. If this is an escape hatch, what's the harm in giving a little extra support to the client in case they did lose their payerAuthorizer and needed to withdraw.

I think `initiateWithdraw` should also accept `config.payerAuthorizer`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+    // Events
+    // =========================================================================
+
+    event ChannelCreated(bytes32 indexed channelId, ChannelConfig config);
+    event Deposited(bytes32 indexed channelId, address indexed sender, uint128 amount, uint128 newBalance);
+    event Claimed(bytes32 indexed channelId, address indexed sender, uint128 claimAmount, uint128 newTotalClaimed);
+    event Settled(address indexed receiver, address indexed token, address indexed sender, uint128 amount);
+    event Refunded(bytes32 indexed channelId, address indexed sender, uint128 amount);
+    event WithdrawInitiated(bytes32 indexed channelId, uint128 amount, uint40 finalizeAfter);
+    event WithdrawFinalized(bytes32 indexed channelId, uint128 amount, address sender);
```
ilikesymmetry: for parity with other event styles, should `sender` be indexed and before `amount`?

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+        bytes calldata collectorData
+    ) external nonReentrant {
+        _validateConfig(config);
+        if (amount == 0) revert ZeroDeposit();
+        if (collector == address(0)) revert InvalidCollector();
+
+        bytes32 channelId = getChannelId(config);
+        ChannelState storage ch = channels[channelId];
+
+        bool isNew = ch.balance == 0 && ch.totalClaimed == 0;
```
ilikesymmetry: think we can just in-line this logic in the `if (isNew) {` and add a comment

contracts/evm/src/x402BatchSettlement.sol
ilikesymmetry, line 1: Needs a pass from https://github.com/coinbase/solidity-style-guide

contracts/evm/src/x402BatchSettlement.sol
ilikesymmetry: Makes me think that maybe `_isReceiverSide` could be `_checkSenderOneOf`?

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+        emit Refunded(channelId, msg.sender, refundAmount);
+
+        IERC20(config.token).safeTransfer(config.payer, refundAmount);
+
+        unchecked {
+            refundNonce[channelId]++;
+        }
+    }
+
+    function _clearPendingWithdrawal(
```
ilikesymmetry: only used in one place, can in-line there and remove separate function

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+
+    function getVoucherDigest(bytes32 channelId, uint128 maxClaimableAmount) external view returns (bytes32) {
+        return _hashTypedDataV4(keccak256(abi.encode(VOUCHER_TYPEHASH, channelId, maxClaimableAmount)));
+    }
+
+    function getRefundDigest(bytes32 channelId, uint256 nonce, uint128 amount) external view returns (bytes32) {
+        return _hashTypedDataV4(keccak256(abi.encode(REFUND_TYPEHASH, channelId, nonce, amount)));
+    }
+
+    function getClaimBatchDigest(
```
ilikesymmetry, line 481: make public so `claimWithSignature` can call

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+    ///         Anyone can submit. All claims must reference the same receiverAuthorizer.
+    function claimWithSignature(
+        VoucherClaim[] calldata voucherClaims,
+        bytes calldata authorizerSignature
+    ) external nonReentrant {
+        if (voucherClaims.length == 0) revert EmptyBatch();
+
+        address authorizer = voucherClaims[0].voucher.channel.receiverAuthorizer;
+
+        bytes32 digest = _hashTypedDataV4(_claimBatchStructHash(voucherClaims));
```
ilikesymmetry: use public `getClaimBatchDigest`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+
+    /// @dev EIP-712 `hashStruct(ClaimBatch)` for the given claims (see `CLAIM_BATCH_TYPEHASH` / `CLAIM_ENTRY_TYPEHASH`).
+    function _claimBatchStructHash(
+        VoucherClaim[] calldata voucherClaims
+    ) internal pure returns (bytes32) {
+        return keccak256(abi.encode(CLAIM_BATCH_TYPEHASH, _claimEntriesRootHash(voucherClaims)));
+    }
+
+    /// @dev EIP-712 encoding for `ClaimEntry[]`: `keccak256(abi.encodePacked(hashStruct(entry), ...))`.
+    function _claimEntriesRootHash(
```
ilikesymmetry: only being used once, in-line into `_claimBatchStructHash`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+
+        uint128 claimDelta = vc.totalClaimed - ch.totalClaimed;
+        ch.totalClaimed = vc.totalClaimed;
+        receivers[vc.voucher.channel.receiver][vc.voucher.channel.token].totalClaimed += claimDelta;
+
+        emit Claimed(channelId, msg.sender, claimDelta, vc.totalClaimed);
+    }
+
+    /// @dev EIP-712 `hashStruct(ClaimBatch)` for the given claims (see `CLAIM_BATCH_TYPEHASH` / `CLAIM_ENTRY_TYPEHASH`).
+    function _claimBatchStructHash(
```
ilikesymmetry: only used once, consider in-lining into `getClaimBatchDigest`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+
+    // =========================================================================
+    // Internal Helpers
+    // =========================================================================
+
+    function _isReceiverSide(address sender, ChannelConfig calldata config) internal pure returns (bool) {
+        return sender == config.receiverAuthorizer || sender == config.receiver;
+    }
+
+    function _validateConfig(
```
ilikesymmetry: only used in `deposit`, consider in-lining there and remove separate function

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    struct ReceiverState {
+        uint128 totalClaimed;
+        uint128 totalSettled;
+    }
+
+    /// @dev The payer-signed data. Does not include claimAmount or signature.
+    struct Voucher {
+        ChannelConfig channel;
```
ilikesymmetry, line 67: This takes more calldata than we need I think. Could theoretically remove two words of calldata per voucher that are unused (`receiverAuthorizer`, `windowDelay`, `salt`) but would require adding a new `channelId`

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,460 @@
...
+    // Constants — EIP-712 Type Hashes
+    // =========================================================================
+
+    bytes32 public constant VOUCHER_TYPEHASH = keccak256("Voucher(bytes32 channelId,uint128 maxClaimableAmount)");
+
+    bytes32 public constant REFUND_TYPEHASH = keccak256("Refund(bytes32 channelId,uint256 nonce,uint128 amount)");
+
+    /// @dev EIP-712 entry for one row in a signed claim batch (mirrors on-chain `VoucherClaim` fields used for authorization).
+    bytes32 public constant CLAIM_ENTRY_TYPEHASH =
+        keccak256("ClaimEntry(bytes32 channelId,uint128 maxClaimableAmount,uint128 totalClaimed)");
```
ilikesymmetry, line 102: Note: Not seeing clear reason we _need_ to sign over `uint128 maxClaimableAmount` as this is buyer-related and authenticated by their signature. No action required, more just an observation and if we wanted maximum brevity, we could remove.

contracts/evm/src/interfaces/IDepositCollector.sol
```diff
@@ -0,0 +1,25 @@
...
+    /// @param token The ERC-20 token address
+    /// @param amount The exact amount of tokens to transfer
+    /// @param channelId The channel identifier (used by Permit2 collectors for witness binding)
+    /// @param caller The address that called deposit() on the settlement contract
+    /// @param collectorData Opaque bytes containing collector-specific parameters (signatures, nonces, etc.)
+    function collect(
+        address payer,
+        address token,
+        uint256 amount,
+        bytes32 channelId,
```
ilikesymmetry, line 20: Observation: This is a subset of `ChannelConfig`, could consider just passing the entire config over. Are we comfortable with the possibility that in the future we could want to write a collector that takes in fields that are in the config but not in this current set?

pr0toshi:
could do settle directly. Claimed amounts accumulate in unsettled[serviceId][token] and are transferred to payTo via a separate settle(serviceId, token). The split lets servers amortize gas across arbitrarily many payers.
funds cant be used in channels since its in claim so why not just actually send vs set claim then send? like sure for batch. shit gas if only 1 payer.

contracts/evm/src/x402BatchSettlement.sol
```diff
@@ -0,0 +1,517 @@
...
+
+    /// @notice Returns the canonical `channelId` for a channel configuration.
+    ///
+    /// @param config The channel configuration to hash.
+    ///
+    /// @return The `keccak256(abi.encode(config))` channel id.
+    function getChannelId(
+        ChannelConfig calldata config
+    ) public pure returns (bytes32) {
+        return keccak256(abi.encode(config));
```
ilikesymmetry: should include `chainId` somewhere so would opt to just add a typehash and wrap with 712?

contracts/evm/src/periphery/Permit2DepositCollector.sol
```diff
@@ -0,0 +1,131 @@
...
+///      before the Permit2 transfer (failures are logged, then execution continues).
+///
+/// @author Coinbase
+contract Permit2DepositCollector is DepositCollector {
+    /// @notice Uniswap Permit2 `SignatureTransfer` singleton used for witness transfers.
+    ISignatureTransfer public immutable PERMIT2;
+
+    /// @notice String passed to Permit2 for EIP-712 witness typing of `DepositWitness`.
+    string public constant DEPOSIT_WITNESS_TYPE_STRING =
+        "DepositWitness witness)TokenPermissions(address token,uint256 amount)DepositWitness(bytes32 channelId)";
```
phdargen: Noncanonical ordering
```suggestion
        "DepositWitness witness)DepositWitness(bytes32 channelId)TokenPermissions(address token,uint256 amount)";
```

phdargen:
> could do settle directly. Claimed amounts accumulate in unsettled[serviceId][token] and are transferred to payTo via a separate settle(serviceId, token). The split lets servers amortize gas across arbitrarily many payers. funds cant be used in channels since its in claim so why not just actually send vs set claim then send? like sure for batch. shit gas if only 1 payer.

You can claim more frequently to lock in gains and settle whenever after multiple claims or atomically batch claim+settle via multicall

specs/schemes/deferred/batch-settlement-evm.md
```diff
@@ -0,0 +1,641 @@
+# Scheme: `batch-settlement` on `EVM`
```
phdargen: remove this pls

phdargen (approved):

PR #1951, opened 2026-04-07
spec: add shielded EVM payments for exact scheme (privacy pools)

### LINKED ISSUE #1953: Add shielded EVM payments as assetTransferMethod for exact scheme
## Problem

Every x402 payment on EVM exposes the payer's wallet address on-chain. For AI agents making hundreds of API calls per day, this creates a complete trail of which services they use, how much they spend, and how often they call each endpoint. Any observer can reconstruct an agent's operational behavior from its payment history.

## Proposed Solution

Add `assetTransferMethod: "shielded"` to the existing `exact` scheme on EVM. Clients withdraw tokens from an on-chain privacy pool directly to the resource server's `payTo` address. The privacy pool contract handles ZK proof verification and nullifier-based double-spend prevention on-chain. The facilitator confirms the resulting ERC-20 Transfer event matches the payment requirements.

## How it works

**On-chain (handled by the privacy pool contract):**
1. Client generates a groth16 ZK proof proving ownership of a shielded UTXO
2. The proof includes nullifiers — cryptographic commitments that mark the UTXO as spent
3. Client submits the proof to the pool contract
4. Contract verifies the ZK proof, checks nullifiers against its on-chain spent set, and executes the ERC-20 transfer
5. If the tx succeeds, the proof was valid and the UTXO is now spent (nullified)

**Facilitator verification:**
1. Client sends `txHash` in `PAYMENT-SIGNATURE` header
2. Facilitator fetches tx receipt, confirms it was not reverted (meaning the pool contract accepted the ZK proof)
3. Facilitator checks the ERC-20 Transfer event: `from` is a registered pool contract, `to` matches `payTo`, `value` >= required amount
4. Facilitator tracks txHash to prevent the same payment from being reused for multiple API calls

The payer's wallet address never appears on-chain. The `from` field in the Transfer event shows the pool contract. The ZK proof ensures the payer owns sufficient funds without revealing which funds.

## Why assetTransferMethod under exact

The `exact` scheme means the client pays the exact amount specified. `shielded` is a new transfer method — same semantics as EIP-3009, Permit2, and ERC-7710, but with sender privacy. The resource server advertises `extra.assetTransferMethod: "shielded"` in the 402 response.

Works with any ERC-20 on any EVM chain that has a compatible privacy pool (e.g., Railgun on Base, Ethereum, BSC, Polygon, Arbitrum).

mmchougule:
## Summary

Adds `specs/schemes/exact/scheme_exact_evm_shielded.md` — a new `assetTransferMethod: "shielded"` for the `exact` scheme on EVM chains.

Enables privacy-preserving x402 payments where the payer's wallet address is hidden behind a ZK proof verified on-chain by the privacy pool contract. The facilitator confirms the resulting ERC-20 Transfer event matches the payment requirements.

## How it works

**On-chain (privacy pool contract handles):**
- Client generates a groth16 ZK proof proving ownership of a shielded UTXO
- Proof includes nullifiers that mark the UTXO as spent (double-spend prevention)
- Pool contract verifies the proof, checks nullifiers, executes the ERC-20 transfer
- If the tx succeeds, the ZK proof was valid and funds are transferred

**Facilitator verification:**
- Client sends `txHash` in `PAYMENT-SIGNATURE` header
- Facilitator fetches tx receipt (non-reverted = pool contract accepted the proof)
- Checks ERC-20 Transfer event: `from` is registered pool contract, `to` matches `payTo`, `value` >= required
- Tracks txHash to prevent same payment being reused for multiple API calls

## Design decisions

- Fits under `exact` scheme as a new `assetTransferMethod` — same payment semantics as EIP-3009/Permit2/ERC-7710, different transfer mechanism
- ZK proof verification and nullifier double-spend prevention happen on-chain in the pool contract — facilitator doesn't need to verify proofs
- General — works with any privacy pool that verifies ZK proofs and emits standard ERC-20 Transfer events
- Client-driven settlement — client submits the unshield tx, facilitator is verify-only

Closes #1953

### DIFF
--- a/specs/schemes/exact/scheme_exact_evm_shielded.md
+++ b/specs/schemes/exact/scheme_exact_evm_shielded.md
@@ -1,0 +1,274 @@
+# Scheme: `exact` on `EVM` — AssetTransferMethod: `shielded`
+
+## Summary
+
+The `shielded` asset transfer method enables privacy-preserving payments on EVM chains via on-chain privacy pools. The client withdraws ("unshields") tokens from a privacy pool directly to the resource server's `payTo` address. The facilitator verifies the payment by checking the ERC-20 `Transfer` event emitted during the unshield — no viewing keys, trial decryption, or off-chain state is required.
+
+This is the first fully private payment option for x402 on EVM. Unlike existing EVM methods (EIP-3009, Permit2, ERC-7710) where the payer's address is publicly visible on-chain, shielded payments hide the sender's identity behind a zero-knowledge proof verified by the privacy pool contract.
+
+### Key Properties
+
+| Property | Value |
+|:---------|:------|
+| Settlement | Client-driven (client unshields on-chain before verification) |
+| Verification | ERC-20 `Transfer` event from registered privacy pool contract |
+| Privacy | Sender hidden (ZK proof); amount visible to facilitator |
+| Gas/Fees | Client pays, or gasless via ERC-4337 account abstraction |
+| Facilitator role | Verify-only (no settlement, no custody) |
+| Token | Any ERC-20 supported by the privacy pool (USDC, USDT, DAI, WETH) |
+
+### Key Differences from Other EVM Methods
+
+| Aspect | EIP-3009 | Permit2 | ERC-7710 | **Shielded** |
+|--------|----------|---------|----------|-------------|
+| Authorization | Off-chain signature | Off-chain signature | Delegation | On-chain ZK proof |
+| Settlement | Facilitator submits | Facilitator submits | Facilitator submits | **Client sends directly** |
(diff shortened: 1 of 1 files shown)


mmchougule:
quick question on shielded transport semantics:

should the spec explicitly state that `poolContracts` are an integrator-defined acceptance policy, rather than a canonical protocol-level pool set?

as written, clients can discover compatible pools per resource via `accepts[]`, but there’s no standard cross-resource discovery mechanism.

if that’s intentional, it may be worth stating explicitly so implementers don’t assume protocol-level interoperability that’s actually left to facilitators / integrators.

PR #1974, opened 2026-04-09
spec(exact-evm): add optional `decimals` field to `extra` in PaymentRequirements

ryanRfox:
### Proposal

Add an optional `decimals` field to the `extra` object in the exact EVM scheme's `PaymentRequirements`. This provides the token's decimal precision to clients without requiring an on-chain RPC call.

Opening this as a spec PR per [CONTRIBUTING.md](../CONTRIBUTING.md#new-schemes): "Propose a scheme by opening a PR with a spec in `specs/schemes/`. Discuss the architecture and purpose in that PR."

### Motivation

The `extra` field in the exact EVM scheme currently carries `name`, `version`, and `assetTransferMethod`. The token's decimal precision is absent, even though it is already known server-side in `DEFAULT_STABLECOINS` (the `decimals` field in `DefaultAssetInfo`).

Without `decimals` in the payment requirements, clients must query the ERC-20 `decimals()` function on-chain to render human-readable token amounts. This works but adds latency and an RPC dependency to what is otherwise a static rendering path. Since the server already knows the value, including it in `extra` would be consistent with how `name` and `version` are already propagated.

### Spec change

In `specs/schemes/exact/scheme_exact_evm.md`, add `decimals` to the `extra` object in both the EIP-3009 and Permit2 examples:

```json
"extra": {
  "name": "USDC",
  "version": "2",
  "decimals": 6
}
```

```json
"extra": {
  "assetTransferMethod": "permit2",
  "name": "Mezo USD",
  "version": "1",
  "decimals": 18
}
```

The field would be:
- **Optional** — clients MUST handle its absence (fall back to on-chain query or assume 6)
- **Integer** — same semantics as ERC-20 `decimals()` return value
- **Informational** — used for display rendering only, not for payment verification or settlement

### Implementation

The server-side change is minimal. In `typescript/packages/mechanisms/evm/src/exact/server/scheme.ts`, add `decimals` to the return object alongside `name` and `version`:

```typescript
extra: {
  ...(includeEip712Domain && {
    name: assetInfo.name,
    version: assetInfo.version,
  }),
  decimals: assetInfo.decimals,
  ...
}
```

### Discussion

The server already resolves `name` and `version` from `DefaultAssetInfo` and includes them in `extra`. The `decimals` value comes from the same source and follows the same pattern. Including it would give any client — paywall, wallet, dashboard, or third-party integration — the information needed to render human-readable amounts without an additional on-chain call.

Looking for maintainer input on whether this addition is desirable, or if clients should be expected to query `decimals()` on-chain independently.

### Changeset

A changeset fragment will be added after the spec is approved and implementation follows.


ryanRfox:
Closing — fork was out of sync with upstream, PR included unrelated files. Will resubmit from a clean base.

PR #1976, opened 2026-04-09
spec(exact-evm): add optional `decimals` field to `extra` in PaymentRequirements

ryanRfox:
> **Update (softened):** This PR now proposes `decimals` as **optional** with server-SHOULD / client-MUST-fallback semantics, not required. See [this comment](https://github.com/x402-foundation/x402/pull/1976#issuecomment-4247030525) for rationale. Strikethrough below reflects the original proposal.

## Proposal

Add a `decimals` field to the `extra` object in the exact EVM scheme's `PaymentRequirements`. This provides the token's decimal precision to clients without requiring an on-chain RPC call, as an **optional hint** that clients MUST be prepared to resolve from the canonical on-chain source when absent.

## Motivation

Currently, when the server constructs `PaymentRequirements`, it calls `getDefaultAsset(network)` which returns `DefaultAssetInfo` including `address`, `name`, `version`, and `decimals`. However, `decimals` is currently not included within `PaymentRequirements`.

Without `decimals` in `PaymentRequirements`, clients have no way to correctly render human-readable token amounts from the raw atomic values in `amount` **without an extra RPC round trip**. Including `decimals` in `extra` gives any client — paywall, wallet, dashboard, or third-party integration — the correct precision as an optional hint.

## Changes Required

- Add `decimals` as an **optional** field in `extra` for the exact EVM scheme's `PaymentRequirements` (eip3009, permit2, erc7710)
- Document all `extra` fields (`assetTransferMethod`, `name`, `version`, `decimals`) per `assetTransferMethod` section, following the pattern established by the SVM, Aptos, Hedera, and Stellar scheme specs
- Servers **SHOULD** include `decimals` to save clients an on-chain RPC round trip
- Clients **MUST** fall back to querying ERC-20 `decimals()` when absent, and **SHOULD** cross-check server-asserted values against a trusted token registry before rendering
- `decimals` is informational only and MUST NOT affect payment verification or settlement
- Update core spec examples in `x402-specification-v2.md` to illustrate the SHOULD case (`"decimals": 6` included in examples)
- Fix typo: "two" to "three" asset transfer methods (ERC-7710 was added)

## Implementation

The server-side implementation is a single line addition in `exact/server/scheme.ts`, adding `decimals: assetInfo.decimals` alongside the existing `name` and `version`.

---

## ~~Original proposal (required) — superseded~~

> ~~**Changes Required:**~~
>
> - ~~Add `decimals` as a **required** field in `extra` for the exact EVM scheme's `PaymentRequirements`~~
> - ~~Document all `extra` fields (`assetTransferMethod`, `name`, `version`, `decimals`) per `assetTransferMethod` section, following the pattern established by the SVM, Aptos, Hedera, and Stellar scheme specs~~
> - ~~Update core spec examples in `x402-specification-v2.md` to include `decimals`~~
> - ~~Fix typo: "two" to "three" asset transfer methods (ERC-7710 was added)~~
>
> ~~**Implementation**~~
> ~~The server-side implementation is a single line addition in `exact/server/scheme.ts`, adding `decimals: assetInfo.decimals` alongside the existing `name` and `version`.~~
>
> ~~**Breaking change note:** Adding `decimals` as a required field is technically a spec-level breaking change — existing server implementations do not currently include it. However, no client code reads `extra.decimals` today (clients currently ignore decimals, hard code the value, or make an RPC call on-chain), and the server-side change is minimal. Clients encountering a server that does not yet include `decimals` should fall back to querying ERC-20 `decimals()` on-chain. The practical impact is that server implementations should include `decimals` when they next release, not that any running system will break.~~

See [softening comment](https://github.com/x402-foundation/x402/pull/1976#issuecomment-4247030525) for why the required approach was withdrawn.

### DIFF
--- a/specs/schemes/exact/scheme_exact_evm.md
+++ b/specs/schemes/exact/scheme_exact_evm.md
@@ -72,7 +72,7 @@ The `payload` field must contain:
 - `extra.assetTransferMethod` (required): MUST be `"eip3009"`.
 - `extra.name` (required): The EIP-712 domain name of the token contract. Used for `transferWithAuthorization` signature construction.
 - `extra.version` (required): The EIP-712 domain version of the token contract. Used for `transferWithAuthorization` signature construction.
-- `extra.decimals` (required): The token's decimal precision (integer, same semantics as ERC-20 `decimals()`). Informational for display rendering only. MUST NOT affect payment verification or settlement.
+- `extra.decimals` (optional): The token's decimal precision (integer, same semantics as ERC-20 `decimals()`). Informational for display rendering only. MUST NOT affect payment verification or settlement. Servers SHOULD include this field to save clients an on-chain RPC round trip. Clients MUST fall back to querying the token's ERC-20 `decimals()` function when the field is absent, and SHOULD cross-check server-asserted values against a trusted token registry before rendering amounts.
 
 ### Phase 2: Verification Logic
 
@@ -171,7 +171,7 @@ The `payload` field must contain:
 - `extra.assetTransferMethod` (required): MUST be `"permit2"`.
 - `extra.name` (conditional): The EIP-712 domain name of the token contract. Required when the token supports EIP-2612 for gasless Permit2 approval.
 - `extra.version` (conditional): The EIP-712 domain version of the token contract. Required when the token supports EIP-2612 for gasless Permit2 approval.
-- `extra.decimals` (required): The token's decimal precision (integer, same semantics as ERC-20 `decimals()`). Informational for display rendering only, MUST NOT affect payment verification or settlement.
+- `extra.decimals` (optional): The token's decimal precision (integer, same semantics as ERC-20 `decimals()`). Informational for display rendering only. MUST NOT affect payment verification or settlement. Servers SHOULD include this field to save clients an on-chain RPC round trip. Clients MUST fall back to querying the token's ERC-20 `decimals()` function when the field is absent, and SHOULD cross-check server-asserted values against a trusted token registry before rendering amounts.
 
 ### Phase 3: Verification Logic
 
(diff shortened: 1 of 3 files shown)


ryanRfox:
## Softening: `extra.decimals` changed from required to optional

After a review of field taxonomy across related protocols (EIP-747 `wallet_watchAsset`, Uniswap Token Lists, CAIP-19, Lightning BOLT-11/12, W3C Payment Request API, EIP-712 domain semantics), I'm changing this PR from **required** to **optional with server-SHOULD / client-MUST-fallback**.

### Why soften now

1. **Taxonomy.** `decimals` is pure display metadata — not signing-integrity (unlike `extra.name` / `extra.version`, which feed EIP-712 domain and thus belong in the signed surface) and not settlement-routing. Treating it as required conflates two different categories of `extra` fields and will compound as more display-adjacent proposals arrive.

2. **Canonical on-chain source exists.** ERC-20 `decimals()` is authoritative. Making the field required forces a spec-level breaking change to ship a hint whose only cost-of-absence is one RPC call — a cost the closest analogue (EIP-747 / `wallet_watchAsset`) handles via server-asserts + wallet-cross-checks, not schema requiredness.

3. **Required-but-advisory is a footgun.** The PR's "informational only; MUST NOT affect verification or settlement" clause is correct — but pairing it with "required" invites ghost compliance (servers stuffing placeholder values to pass validators) and gives a spoof-minded server the exact same attack surface as the optional version, without the honesty.

4. **No current code reads `extra.decimals`.** Neither #1926 nor #1980 read this field — they use `scheme.getAssetDecimals()` (registry lookup) and on-chain `decimals()` (RPC) respectively. So softening costs zero implementation churn, and can be decided on policy grounds alone.

### What changed in this push

- `extra.decimals`: `(required)` → `(optional)` in all three asset-transfer-method sections (eip3009, permit2, erc7710).
- Added explicit language in each section:
  - Servers **SHOULD** include the field to save clients an RPC round trip.
  - Clients **MUST** fall back to on-chain `decimals()` when the field is absent.
  - Clients **SHOULD** cross-check server-asserted values against a trusted token registry before rendering amounts.
- "MUST NOT affect payment verification or settlement" clause retained.
- Example JSON blocks still show `"decimals": 6` — illustrating the SHOULD case.

### What this unblocks

- No breaking change. Existing server implementations remain conformant.
- Other schemes (SVM, Aptos, Hedera, Stellar) can follow the same optional pattern in a consistency-follow-up PR without a coordinated breaking-change cycle.
- Leaves headroom for a future `extra.symbol` / token-registry-integration proposal on the same rubric.

ryanRfox:
Closing in favor of a smaller-scope docs-only cleanup PR. 

My original idea to add `extra.decimals` actually adds wire-format surface that no SDK consumes today. Alternatives exist to making a spec change.

PR #1980, opened 2026-04-09
fix(paywall): use dynamic token decimals instead of hardcoding 6

### LINKED ISSUE #1979: bug(paywall): EVM paywall hardcodes 6 decimals — breaks non-6-decimal tokens
### Bug

The EVM paywall assumes all payment tokens use 6 decimal places (like USDC). Tokens with different precision render incorrect amounts.

**Where:**
- `typescript/packages/http/paywall/src/evm/index.ts` line 37: `parseFloat(requirement.amount) / 1000000` — hardcoded 10^6 divisor
- `typescript/packages/http/paywall/src/evm/EvmPaywall.tsx` line 78: `formatUnits(balance, 6)` — hardcoded 6 for balance display
- `typescript/packages/http/paywall/src/evm/utils.ts`: `getUSDCBalance()` uses hardcoded USDC addresses for 3 chains only

**Impact:** Any token with non-6 decimals displays incorrectly. For example, an 18-decimal token priced at 0.001 (raw: 10^15) shows as $1,000,000,000 in the paywall UI. This already affects MegaETH/MegaUSD (18 decimals, in DEFAULT_STABLECOINS since v2.7.0).

**Who it affects:** Any chain in DEFAULT_STABLECOINS with decimals != 6. Currently: MegaETH (eip155:4326, MegaUSD, 18 decimals). Any new chain adding a non-6-decimal default token.

**Expected behavior:** The paywall should use the correct decimal precision for the payment token rather than assuming 6. The server-side amount conversion can read decimals from the default asset registry (`getDefaultAsset`). The client-side balance display can query the standard ERC-20 `decimals()` function, which every compliant token implements.

ryanRfox:
## Description

The EVM paywall hardcodes `/ 1000000` everywhere, breaking display amounts for any token that isn't a 6-decimal stablecoin. An 18-decimal token priced at 0.001 raw units renders as **$1,000,000,000**. The assumption appears in three places — server-side HTML generation, client-side balance display, and a USDC-only address lookup table.

Closes #1979.

## Approach

The PR replaces the hardcoded-6-decimal assumption on both sides of the paywall:

### Server side (`@x402/paywall/src/evm/index.ts`)

- Replace the hardcoded `/ 1000000` divisor with a lookup against `@x402/evm`'s `DEFAULT_STABLECOINS` registry, with a fallback to `6` (USDC default) when the network isn't in the registry.
- Replace the lossy `parseFloat(amount) / 10**decimals` math with `Number(formatUnits(BigInt(amount), decimals))` so atomic-to-display conversion preserves precision through 18-decimal amounts. `BigInt` also rejects non-integer atomic strings, which matches the spec's atomic-integer contract.

### Client side (`@x402/paywall/src/evm/EvmPaywall.tsx`)

- Read the token contract address from the payment requirement's `asset` field (server populates it).
- Query the standard ERC-20 `decimals()` function on-chain via the new `getTokenDecimals` helper in `utils.ts`, so balance formatting uses the actual token's precision.
- State variable renamed `formattedUsdcBalance` → `formattedBalance`; balance-check callback renamed `checkUSDCBalance` → `checkBalance`.

### Internal utility (`@x402/paywall/src/evm/utils.ts`)

- Replace the internal `getUSDCBalance(client, address)` — which hardcoded USDC contract addresses for three chains (Ethereum, Base, Base Sepolia) and returned `0n` for every other chain — with generic:
  - `getTokenBalance(client, owner, tokenAddress)` — ERC-20 `balanceOf` against the supplied token address
  - `getTokenDecimals(client, tokenAddress)` — ERC-20 `decimals()` against the supplied token address, with a `6` fallback on error
- Drops the `USDC_ADDRESSES` map entirely.
- **Breaking change (internal):** `getUSDCBalance` is removed. This helper was not part of the package's public API (`exports` map); it was only consumed by the bundled `EvmPaywall` React component inside the paywall HTML template. No server-operator code path calls it. Anyone who has forked the paywall and imports from the internal file path (bypassing package exports) would need to update their import and call site.

### `@x402/evm` public API

`@x402/evm` now publicly re-exports `DEFAULT_STABLECOINS` from `./shared/defaultAssets` so consumers can read the canonical default-asset registry directly.

### Generated templates

The baked EVM paywall bundles (`typescript/packages/http/paywall/src/evm/gen/template.ts`, `go/http/evm_paywall_template.go`, `python/x402/http/paywall/evm_paywall_template.py`) are regenerated to carry the updated `EvmPaywall.tsx` / `utils.ts`. SVM templates are regenerated for toolchain parity — the build step emits both.

## Why this approach

`@x402/evm`'s `DEFAULT_STABLECOINS` registry already holds the decimals field for every supported chain (eip155:31611 / Mezo mUSD = 18, eip155:4326 / MegaETH MegaUSD = 18, every other chain's default = 6). The same registry powers existing dispatch paths in the SDK:

- `scheme.getAssetDecimals(asset, network)` on both the exact and upto EVM schemes
- The inline `scheme?.getAssetDecimals?.(...)` dispatch in `@x402/core`'s `x402ResourceServer` (server.ts:795)

The paywall now consolidates the registry lookup so all consumers share the same source within a published version, rather than carrying a parallel map. A new chain added to `DEFAULT_STABLECOINS` is picked up by the paywall when the consumer upgrades `@x402/evm` — no paywall rebuild required.

`@x402/paywall` already imports from `@x402/evm` for client-side code; this PR extends that to the server-side decimals lookup.

## Tests

`pnpm --filter @x402/paywall test --run network-handlers` — 12/12 pass (6 new tests added):

- `getDefaultTokenDecimals` returns 18 for Mezo Testnet (eip155:31611)
- `getDefaultTokenDecimals` returns 6 for Base mainnet (eip155:8453), with `DEFAULT_STABLECOINS["eip155:8453"]` asserted alongside so an empty registry would fail the test
- `getDefaultTokenDecimals` falls back to 6 for networks not in the registry
- `evmPaywall.generateHtml` end-to-end on a Mezo Testnet 18-decimal payment: `1000000000000000` atomic renders as `amount: 0.001,` in the inline window.x402 script (the regression test for the `/ 1e6` order-of-magnitude bug)
- `evmPaywall.generateHtml` end-to-end on a Base mainnet 6-decimal payment: `1000000` atomic renders as `amount: 1,`
- `evmPaywall.generateHtml` throws on a non-integer atomic amount string (`"1.5"`), pinning the BigInt strictness so a future revert to `parseFloat` would fail


ryanRfox:
Note: #1926 fixes the same hardcoded-decimals bug in the server-side fallback paywall (`x402HTTPResourceServer`). This PR fixes the client-side React paywall (`@x402/paywall`). The two PRs are complementary — different packages, different code paths, both need the fix.

ryanRfox:
@phdargen this one is ready for review. Hope you can consider this in parallel with #1926 to ensure @natsukingly and I land a complete fix.

I saw your comment recently about the next SDK release coming early this week, so hoping these can make it in as well.

ryanRfox:
I noticed #2053 is working on `parseFloat` as well, but have confirmed it handles the input side (dollar-string parsing) while this PR handles the output side (atomic-amount display). No conflict; both may land independently.

typescript/packages/http/paywall/package.json
```diff
@@ -75,6 +74,7 @@
     "@wallet-standard/base": "^1.1.0",
     "@wallet-standard/features": "^1.1.0",
     "@x402/core": "workspace:~",
+    "@x402/evm": "workspace:~",
```
phdargen: Hm would prefer if we could keep `@x402/evm` as `devDependencies` but duplicating the `DEFAULT_STABLECOINS` map in the paywall package doesnt seems very appealing either

ryanRfox: Good callout on the dep promotion. The concern that motivated it is narrow: `src/evm/index.ts` needs a synchronous decimals lookup for the initial display math before the client takes over. The client already reads on-chain `decimals()` for balance formatting in this PR, so the server-side lookup is display-bootstrap only.

A third path fits the existing pipeline. Have `src/evm/build.ts` read `DEFAULT_STABLECOINS` from `@x402/evm` at build time (kept as a devDependency) and emit a generated `src/evm/gen/decimals.ts` next to the generated `template.ts`. The package already codegens TS, Python, and Go artifacts from a single source at `build.ts`, so this is pattern-consistent. Runtime dep footprint unchanged, no hand-maintained duplicate, and a CI regen-diff catches drift.

Happy to rework #1980 along those lines, or take a different direction if you'd rather.

ryanRfox:
Bumping this here with the hope you can review it grouped with #1926 and #2054 because I need all to land prior to the next SDK cut to achieve my goal and match what the docs advertise (yes, I'm being selfish).

First, Mezo Testnet (`eip155:31611`) is currently included in `DEFAULT_STABLECOINS` and the `network-and-token-support.mdx` table. But the MUSD asset is 18-decimal and `@x402/paywall` hardcodes 6, so the paywall UI shows the wrong amount today, confusing end users. This PR is the client-side fix; #1926 (@natsukingly) is the server-side-fallback companion (`getDisplayAmount`) — client and server render paths are independent, so either alone is a half-fix.

Second, #2054 is the structural piece: it regenerates the paywall bundle against the currently pinned viem (2.47.12 since #2013) and adds a CI drift check. Without it, any chain added to `DEFAULT_STABLECOINS` ships as "supported" per the docs while the bundle runs against pre-#2013 viem chain definitions (Radius #2038 is the next one in the queue and will hit a similar trap).

Given your note about the SDK release landing this week: could #1980 + #1926 + #2054 get reviewed as a set?

Much appreciated.

ryanRfox:
Refactoring update:

- @x402/evm now back in devDependencies
- src/evm/build.ts now reads DEFAULT_STABLECOINS at build time and emits src/evm/gen/decimals.ts exposing NETWORK_DECIMALS.
- src/evm/index.ts reads the generated map at runtime. Zero @x402/evm imports in dist/.

This results in a pattern consistent with the existing gen/template.ts codegen.

Drift safety checks:

- #2054's CI check picks up gen/decimals.ts under the existing glob once merged
- Added new invariant test in network-handlers.test.ts asserts NETWORK_DECIMALS ↔ DEFAULT_STABLECOINS key-for-key, both directions.

ryanRfox:
alright @phdargen I've rebased and cleaned this up based on your guidance. Let me know if this is moving in the right direction.

phdargen (approved):
Nice @ryanRfox, that’s a good solution!

PR #1984, opened 2026-04-10
feat(examples): add EUR settlement example via fiat-capable facilitator

petteri74dev:
## Summary

Adds an Express server example showing how to accept x402 payments and receive EUR via SEPA Instant. The server code is identical to the standard Express example — the only difference is the `FACILITATOR_URL` pointing to a facilitator that handles USDC-to-EUR conversion.

## What this enables

- European sellers receive EUR in their bank account instead of USDC
- Buyers/agents pay USDC as normal — no change to client code
- Standard `exact` scheme, no new packages or types needed
- One env var change from an existing x402 integration

## Changes

- New example: `examples/typescript/servers/eur-settlement/`
- Express server with two paid endpoints (`/weather`, `/market-data`)
- README explaining the facilitator swap pattern
- `.env-local` with `FACILITATOR_URL` template

No changes to existing examples, packages, or types.

## Design notes

- `FACILITATOR_URL` is an environment variable, not hardcoded — any fiat-capable facilitator works
- In this example, [AsterPay](https://asterpay.io) is used as one possible EUR settlement facilitator. Other facilitators can implement the same capability by supporting the same HTTP interface.

## Test plan

- [x] `pnpm install` resolves without errors
- [x] `pnpm dev` starts the server on port 4021
- [x] `GET /weather` returns 402 Payment Required
- [x] `GET /market-data` returns 402 Payment Required
- [x] Config files match `express/` example conventions
- [ ] Payment flow works with x402 client SDK

### DIFF
--- a/examples/typescript/pnpm-lock.yaml
+++ b/examples/typescript/pnpm-lock.yaml
@@ -3297,6 +3297,61 @@ importers:
         specifier: ^5.3.0
         version: 5.9.2
 
+  servers/eur-settlement:
+    dependencies:
+      '@x402/core':
+        specifier: workspace:*
+        version: link:../../../../typescript/packages/core
+      '@x402/evm':
+        specifier: workspace:*
+        version: link:../../../../typescript/packages/mechanisms/evm
+      '@x402/express':
+        specifier: workspace:*
+        version: link:../../../../typescript/packages/http/express
+      dotenv:
+        specifier: ^16.4.7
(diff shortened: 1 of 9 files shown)


phdargen:
Thanks for the contribution but this doesn’t add anything beyond the standard express example beyond an env var. Examples should also remain neutral and not highlight a specific project, thats what the ecosystem page is for

PR #1145, opened 2026-02-10
Add batch-settlement payment scheme and Cloudflare network specification

CameronWhiteside:
## Description

Introduces the **batch-settlement payment scheme** for x402, enabling instant resource access through cryptographic payment commitments where settlement occurs later through a process defined by the network binding.

The core scheme spec defines protocol behavior, two commitment models (capital-backed and credit-backed), and the abstract settlement lifecycle — designed to unify the multiple deferred-settlement PRs under a single parent scheme.

The Cloudflare network binding (`cloudflare:402`) provides a credit-backed implementation using HTTP Message Signatures (RFC 9421) for authentication and fiat batch billing.

#### What's Included

| File | Description |
|------|-------------|
| `specs/schemes/batch-settlement/batch_settlement.md` | Core scheme definition — protocol behavior, commitment models, use cases, network requirements |
| `specs/schemes/batch-settlement/batch_settlement_cloudflare.md` | Cloudflare `cloudflare:402` network binding |
| `specs/extensions/http-message-signatures.md` | Extension for RFC 9421 authentication |

> **Note**: The `terms` extension has been split out for separate review.

> **Note on naming**: The scheme is named `batch-settlement` as a starting point. Open to discussion on `batch-settlement` vs `batch` vs `batched` — easier to shorten than lengthen.

## Tests
None, this is a **spec-only PR** (no SDK implementation).

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [ ] **SKIPPED** added a changelog fragment for user-facing changes (docs-only changes can skip)

PR #1986, opened 2026-04-10
Add terms extension specification

CameronWhiteside:
## Summary

Adds the `terms` extension, providing a channel for servers to communicate usage rights, obligations, and settlement terms alongside the `PaymentRequired` response.

- Scheme- and network-agnostic — works with `exact`, `batch-settlement`, or any future scheme
- Supports multiple formats: `uri`, `markdown`, `plaintext`, `json`
- Optional `version` field for change detection

Split from #1145 for separate review per discussion with @carsonroscoe.

### A note on scope

This extension defines how terms are **communicated** — not how they are **consented to**. The spec takes the implicit approach: terms are presented, and proceeding with payment implies awareness. Whether a network requires explicit consent (e.g., tying terms acceptance to a signed SIWX message) is a per-network or future-extension concern. The priority is establishing a standard channel for terms to be discoverable and machine-readable.

#### What's Included

| File | Description |
|------|-------------|
| `specs/extensions/terms.md` | Terms extension specification |

## Tests
None — spec-only, no SDK implementation.

## Checklist

- [ ] **SKIPPED** added a changelog fragment for user-facing changes (docs-only changes can skip)

### DIFF
--- a/specs/extensions/terms.md
+++ b/specs/extensions/terms.md
@@ -1,0 +1,90 @@
+# Extension: `terms`
+
+## Summary
+
+The `terms` extension can be used by any network or scheme to communicate the terms of the payment commitment.
+
+## Purpose
+
+Describes the usage rights, obligations, and settlement terms of a payment commitment.
+
+## Extension Definition
+
+```json
+{
+  "terms": {
+    "schema": {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "type": "object",
+      "properties": {
+        "format": {
+          "type": "string",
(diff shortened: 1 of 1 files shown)


jithinraj:
This looks close @CameronWhiteside. The one thing I’d still add is a stable digest. `version` helps with change detection, but it does not give two clients a shared test for “same terms” across `uri`, `markdown`, `plaintext`, and `json`.

A digest over normalized bytes would give you a clean binding point later without dragging explicit-consent semantics into this PR.

PR #1988, opened 2026-04-10
fix(paywall): use token name from payment requirements instead of hardcoded USDC

### LINKED ISSUE #1972: bug(paywall): EVM paywall hardcodes "USDC" label for all tokens
### Bug

The EVM paywall displays "USDC" as the token name for all EVM payments regardless of the actual token. The payment requirements already carry the asset name in `extra.name` (set by the server from the default asset registry), but the paywall ignores it.

**Where:**
- `typescript/packages/http/paywall/src/evm/EvmPaywall.tsx`:
  - Line 199: `${amount} {chainName} USDC`
  - Line 203: `Need {chainName} USDC?`
  - Line 269: `${amount} USDC`

**Impact:** Any non-USDC token shows a misleading "USDC" label in the paywall. For example, MegaETH shows "MegaETH USDC" when the token is MegaUSD. Similarly, Stable (USDT0) and any future Permit2 tokens are mislabeled.

**Who it affects:** Any chain whose default stablecoin is not named USDC. Currently: MegaETH (MegaUSD), Stable mainnet/testnet (USDT0). Any new chain with a non-USDC default token.

**Expected behavior:** Read the token name from `firstRequirement.extra.name` and fall back to "USDC" only when `extra.name` is absent (backward compatible with EIP-3009 tokens that may omit the name field).

### Related

- #1971 — stale viem/chains across all TS packages

Ai-chan-0411:
## Summary

The EVM paywall currently hardcodes "USDC" as the token label in all user-facing strings. This is incorrect for chains whose default stablecoin is not named USDC (e.g. MegaETH uses MegaUSD, Stable uses USDT0).

This PR reads the token name from `firstRequirement.extra.name` (set by the server from the default asset registry) and falls back to "USDC" when the field is absent, maintaining backward compatibility.

## Changes

- Extract `tokenName` from `firstRequirement.extra.name` with `"USDC"` fallback
- Replace 7 hardcoded `"USDC"` display strings in `EvmPaywall.tsx`:
  - Status messages ("Checking balance...", "Insufficient balance...")
  - Payment description text
  - Balance display
  - Amount display

## Testing

- For USDC chains: `extra.name` is `"USDC"`, behavior is identical to before
- For MegaETH: `extra.name` is `"MegaUSD"`, paywall now correctly shows "MegaUSD" instead of "USDC"
- When `extra.name` is absent: falls back to "USDC" (backward compatible)

Closes #1972

### DIFF
--- a/typescript/packages/http/paywall/src/evm/EvmPaywall.tsx
+++ b/typescript/packages/http/paywall/src/evm/EvmPaywall.tsx
@@ -51,6 +51,7 @@ export function EvmPaywall({ paymentRequired, onSuccessfulResponse }: EvmPaywall
   const network = firstRequirement.network;
   const chainName = getNetworkDisplayName(network);
   const testnet = isTestnetNetwork(network);
+  const tokenName = (firstRequirement.extra as Record<string, unknown>)?.name as string || "USDC";
 
   const chainId = parseInt(network.split(":")[1]);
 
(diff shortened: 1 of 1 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/1973

PR #2002, opened 2026-04-11
chore: add scoped viem dependabot updates for /typescript

### LINKED ISSUE #1971: Stale viem/chains across all TS packages — new chains broken until lockfile updated
## Problem

All TypeScript packages in this monorepo bundle `viem/chains` at build time, and every publish workflow uses `pnpm install --frozen-lockfile`. This pins viem to whatever version was last resolved in `pnpm-lock.yaml`, regardless of the `^` range in each `package.json`.

The lockfile currently contains four different viem versions across packages:

| Locked version | Packages |
|---|---|
| 2.23.2 | legacy packages (x402, x402-express, etc.) |
| 2.37.3 | wagmi transitive |
| 2.40.3 | @x402/paywall, @x402/express, @x402/fetch, @x402/evm |
| 2.45.1 | @x402/extensions |

viem is currently at 2.47.10. Any chain added to viem after the locked version for a given package is missing from that package's published bundle.

## Impact

Packages that do runtime chain lookups via `viem/chains` will fail for missing chains:

- **@x402/paywall** (`paywallUtils.ts`) — throws "Unsupported chain ID" when a user connects a wallet on an unrecognized chain
- **@x402/extensions** (`sign-in-with-x/verify.ts`, `evm.ts`, `types.ts`) — sign-in verification fails for unrecognized chains

This affects any chain that onboards to x402 and was added to viem after the locked version. For example, Mezo (chain ID 31611, added in viem 2.47.10) is missing from all current published packages.

The problem will recur with every new chain that joins x402 unless the build process addresses it.

## Where the constraint is

1. **All 14 publish workflows** in `.github/workflows/publish_npm_scoped_x402_*.yml` use `--frozen-lockfile`
2. **No publish workflow** runs `pnpm update` before building
3. **tsup configs** don't externalize viem, so chain definitions are baked into each bundle
4. **No runtime extension point** exists for server operators to provide chain definitions that aren't in the bundle

## Suggestion

Other monorepos in the ecosystem (wagmi, viem, RainbowKit, Vite, Next.js) use Renovate or Dependabot with auto-merge for minor/patch dependency updates. This keeps the lockfile current continuously — Renovate opens PRs as new dependency versions publish, minor/patch updates auto-merge, and the lockfile stays fresh without manual intervention. The publish workflow stays `--frozen-lockfile` (preserving reproducibility), but the lockfile itself reflects recent dependency versions rather than going stale between manual updates.

This repo doesn't currently have Renovate or Dependabot configured. A minimal Dependabot config would look like:

```yaml
# .github/dependabot.yml
version: 2
updates:
  - package-ecosystem: "npm"
    directory: "/typescript"
    schedule:
      interval: "weekly"
    groups:
      viem:
        patterns:
          - "viem"
        update-types:
          - "minor"
          - "patch"
```

This would open weekly PRs for viem minor/patch updates, keeping the lockfile fresh without changing the publish workflow or sacrificing build reproducibility.

Bortlesboat:
## Summary
- add a Dependabot config at the repo root for the `/typescript` workspace
- scope updates to `viem` only so chain additions land without broad dependency churn
- keep the PR volume low with one weekly update and major versions ignored

## Testing
- `python -c "import pathlib; import yaml; p=pathlib.Path('.github/dependabot.yml'); data=yaml.safe_load(p.read_text()); print(data[''version'']); print(data[''updates''][0][''allow''][0][''dependency-name''])"`

Closes #1971

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,20 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/typescript"
+    schedule:
+      interval: weekly
+    open-pull-requests-limit: 1
+    allow:
+      - dependency-name: "viem"
+    ignore:
+      - dependency-name: "viem"
+        update-types:
+          - "version-update:semver-major"
+    groups:
+      viem:
+        patterns:
+          - "viem"
+        update-types:
+          - minor
+          - patch


ryanRfox:
FYI: opened #2296 to revisit the broader design space from #1971. #2002 is captured there as option 3 (cheap baseline) alongside two larger directions @cryptomotifs sketched. See the new issue for context on whether this PR makes sense to land as the immediate mitigation while the larger direction gets decided.

Bortlesboat:
Some current numbers on the freshness gap, in case they help decide whether this
is worth landing as the stopgap while #2296 settles.

`typescript/packages/{extensions,mcp}` declare `viem: ^2.48.11` and the lockfile
resolves `2.48.11`. Latest viem is `2.55.11`.

Counting chain exports in `viem/chains`: 699 at 2.48.11, 715 at 2.55.11. So there
are 16 chains that viem supports today and x402 reports as `Unsupported chain ID`:

`arc`, `battlechain`, `citrate`, `defiOracleMetaMainnet`, `grav`, `ladyChain`,
`marooTestnet`, `marsCredit`, `mizuhikiTestnetAwaji`, `robinhood`,
`robinhoodTestnet`, `sentrix`, `sentrixTestnet`, `valygoNft`,
`valygoSmartchain`, `whitechainSepolia`

That's the exact failure mode @cryptomotifs described — indistinguishable from a
chain that genuinely doesn't exist. It's the same shape as the original Mezo
report, just 16 chains wider now.

This PR doesn't fix the distinction, and I don't think it should block whichever
of options 1/2 you land. It only keeps the gap from growing while that's decided.
Happy to close it if the runtime-registry direction is close enough that a
stopgap isn't worth the review time.

PR #2009, opened 2026-04-12
docs(sellers): replace legacy Go import path in Echo example

vikions:
## Description

Replaces the remaining legacy `github.com/coinbase/x402/...` Go imports in the seller quickstart Echo example with the current `github.com/x402-foundation/x402/...` module path.

This aligns the snippet with the canonical Go module path already used across the repo and avoids confusion for users copying the example.

## Tests

- Not run; docs-only change.


phdargen (approved):
Thanks for the fix @vikions

PR #2023, opened 2026-04-13
feat: upto evm python

CarsonRoscoe:
## Description

Adds the `upto` EVM payment scheme to the Python SDK, bringing it to feature parity with the TypeScript and Go implementations.

The `upto` scheme enables usage-based pricing: a client authorizes a **maximum** amount upfront, and the server settles for the **actual** amount consumed at the end of the request (e.g. LLM tokens generated, bytes transferred, compute time). It uses Permit2 exclusively, with an upto-specific witness struct that adds a `facilitator` field to bind the authorization to a specific facilitator address.

**Key changes:**

- **New package** `x402.mechanisms.evm.upto` with `UptoEvmClientScheme`, `UptoEvmServerScheme`, and `UptoEvmFacilitatorScheme`
- **New types** in `mechanisms/evm/types.py`: `UptoPermit2Witness`, `UptoPermit2Authorization`, `UptoPermit2Payload`, `is_upto_permit2_payload()`
- **New constants** in `mechanisms/evm/constants.py`: `SCHEME_UPTO`, `X402_UPTO_PERMIT2_PROXY_ADDRESS`, `UPTO_PERMIT2_WITNESS_TYPES`, upto proxy ABIs (settle takes an `amount` parameter), and upto-specific error codes
- **`SettleResponse`** gains an optional `amount` field to report the actual settled amount (matches TS/Go)
- **E2E suite** updated: Python facilitator, fastapi, flask servers, and httpx/requests clients all register upto and expose `GET /upto/evm/permit2`; `test.config.json` files updated to match Go convention (`schemes`, `evm.transferMethods: [..., "upto"]`, `"transferMethod": "upto"`, `permit2Direct: true`)
- **Changelog fragment** added at `python/x402/changelog.d/add-upto-evm-scheme.feature.md`

**Design notes:**
- No `register_upto_evm_*` helper functions — upto is V2-only so there's no V1 complexity to hide; callers use the builder pattern directly (`facilitator.register(networks, UptoEvmFacilitatorScheme(signer))`)
- `_verify_permit2_allowance` is reused from the exact Permit2 implementation — it is scheme-agnostic (checks ERC-20 → Permit2 allowance and extension fallbacks)
- The facilitator's `get_extra()` returns `{"facilitatorAddress": ...}` so the server can inject it into `PaymentRequirements`, which the client then includes in the Permit2 witness — this binds each authorization to the specific facilitator that will settle it

## Tests

**Unit tests** (`uv run pytest tests/unit/` from `python/x402/`) — 788 passing (55 new upto-specific, zero regressions):

- `test_upto_permit2.py` — `UptoPermit2Payload` serialization round-trips, `is_upto_permit2_payload()` type guard
- `test_upto_client.py` — scheme identifier, payload structure (facilitator in witness), missing facilitatorAddress error, LocalAccount auto-wrap
- `test_upto_server.py` — price parsing, `enhance_payment_requirements` always sets `permit2` + injects `facilitatorAddress`
- `test_upto_facilitator.py` — `get_extra()`, full verify cascade (scheme, network, spender, recipient, facilitator mismatch, deadline, validAfter, amount, token, signature, allowance, balance), settle (zero settlement, partial settlement, settlement exceeds max, tx failure)

**Integration tests** (`uv run pytest tests/integrations/test_evm.py` from `python/x402/`) — 22 passing, including 3 new upto tests against real Base Sepolia:

- Full flow: client signs upto Permit2 payload → facilitator verifies → server settles for partial amount (500 of 1000 authorized) with a real on-chain `x402UptoPermit2Proxy.settle()` transaction
- `get_supported()` returns upto kinds with `facilitatorAddress` in `extra`
- Payload structure validation

## Checklist

- [x] I added a changelog fragment for user-facing changes (`python/x402/changelog.d/add-upto-evm-scheme.feature.md`)


python/x402/tests/integrations/test_evm.py
```diff
@@ -669,3 +675,161 @@ def test_facilitator_web3_signer_verify_typed_data_invalid_signer(self) -> None:
...
+            UptoEvmFacilitatorScheme(self.facilitator_signer),
+        )
+
+        facilitator_client = UptoEvmFacilitatorClientSync(self.facilitator)
+
+        self.server = x402ResourceServerSync(facilitator_client)
+        self.server.register("eip155:84532", UptoEvmServerScheme())
+        self.server.initialize()
+
+    def test_server_should_verify_and_settle_upto_payment(self) -> None:
```
phdargen, line 755: This test fails: 
```
AssertionError: assert False is True
E        +  where False = VerifyResponse(is_valid=False, invalid_reason='invalid_upto_evm_transaction_failed', invalid_message=None, payer='0xE33A295AF5C90A0649DFBECfDf9D604789B892e2').is_valid
```

phdargen: Seems this line:
https://github.com/x402-foundation/x402/blob/f78b268a76e739dd980ff26a50b981d1d3f0e4ef/python/x402/mechanisms/evm/signers.py#L332
needs to be changed to:
`        return func(*args).call({"from": Web3.to_checksum_address(self._account.address)})
`
such that msg.sender matches the facilitator

ethanoroshiba: Strange, all tests are passing locally for me and in the CI. What are you using to get this failure?

CarsonRoscoe: We should double check the documentation to make this more obvious,

Essentially only unit tests are run locally & in CI. Integration tests rely on .env vars, and thus we don't want making real transactions on every commit for every PR, otherwise we'd drain our testnet funds too quickly.

In Python, when you run `uv run pytest` it tries to run BOTH unit tests AND integrations test. It then silently skips integration tests if your `.env` is not loaded with the appropriate variables.

So long story short, add a `.env` to `/python/x402` and rerun `uv run pytest`, then you'll see the errors Philippe is reporting

ethanoroshiba: Thanks for the insight! Since the checks are already gated behind maintainer approval, could it be a good idea to still perform them in the CI? Or maybe as a check before merging to `main`?

ethanoroshiba: Should be fixed in [fa967d8](https://github.com/x402-foundation/x402/pull/2023/commits/fa967d81b34560538c8fcdf17de233b324338afe). Also, ticketed docs updates here: https://github.com/x402-foundation/x402/issues/2045

phdargen (approved):

Dustin4444 (approved):

Dustin4444 (approved):

PR #2027, opened 2026-04-14
docs: clarify MCP and SIWX integration boundaries

Bortlesboat:
## Summary
- clarify that the current MCP guide documents the payment-first flow that works today
- explain that auth-only SIWX MCP routes and pay-once-then-reopen MCP flows still require custom wrapper logic
- add wallet role separation guidance for sellers and operators

## Why
Entrants can currently read the MCP and SIWX docs independently and over-infer first-class MCP SIWX support that is not in the basic example yet. This patch makes the current boundary explicit while still pointing builders at the right extension path.

## Testing
- reviewed the docs diff locally
- no code changes


phdargen:
Good catch, will add MCP SIWX support soon

PR #2044, opened 2026-04-16
docs(spec-exact-evm): clarify extra fields per assetTransferMethod and fix two→three typo

ryanRfox:
## Description

The exact-EVM scheme spec describes three asset transfer methods (`eip3009`, `permit2`, `erc7710`) but documents the `extra` object only by example. Readers can't tell which of the `extra.*` fields are required, conditional, or optional for each method, as they may differ for each.

This PR:

- Fixes a typo related to when `erc7710` was added
- Adds a per-`assetTransferMethod` field-classification block after each method's `PAYMENT-SIGNATURE` example


phdargen (approved):

phdargen:
Thanks @ryanRfox, looks good

PR #2050, opened 2026-04-16
docs: add builder_code.md spec

0xClouds:
## Description

Adds the builder-code extension spec for on-chain attribution tracking in x402 payments. This extension appends https://github.com/ethereum/ERCs/pull/948 Schema 2 (CBOR-encoded) builder codes to settlement transaction calldata, enabling attribution of which application exposed a paid endpoint and which facilitator settled the payment.

The spec covers:

ERC-8021 Schema 2 suffix format and CBOR map fields
Facilitator-side calldata encoding and verification
Resource server offer/receipt field definitions
Builder code registration and lookup via on-chain registries

## Checklist

### DIFF
--- a/specs/extensions/builder_code.md
+++ b/specs/extensions/builder_code.md
@@ -1,0 +1,253 @@
+# Extension: `builder-code`
+
+## Summary
+
+The `builder-code` extension enables **on-chain attribution tracking** for x402 payments by appending [ERC-8021](https://eip.tools/eip/8021) Schema 2 builder codes to settlement transaction calldata. It attributes which application exposed the paid endpoint and which facilitator settled the payment.
+
+This extension only implements **Schema 2** (CBOR-encoded) of ERC-8021. The `s` (services), `m` (custom metadata), and `r` (custom registries) fields are not supported.
+
+---
+
+## ERC-8021 Schema 2 Overview
+
+ERC-8021 defines a structured data suffix appended to transaction calldata for entity attribution. Schema 2 uses CBOR encoding for extensibility.
+
+### Suffix Format
+
+The complete suffix appended to calldata is (ordered end of calldata backwards):
+
+| Component    | Size     | Description                                             |
+| ------------ | -------- | ------------------------------------------------------- |
+| `ercMarker`  | 16 bytes | Constant identifier: `80218021802180218021802180218021` |
+| `schemaId`   | 1 byte   | `0x02` for Schema 2                                     |
+| `cborLength` | 2 bytes  | Length of CBOR data (big-endian)                        |
+| `cborData`   | variable | CBOR-encoded map of attribution fields                  |
+
+Wire order: `[cborData][cborLength (2B)][schemaId (1B)][ercMarker (16B)]`
+
+### CBOR Map Fields
+
+| Key | Type   | Description                                                     |
+| --- | ------ | --------------------------------------------------------------- |
+| `a` | string | App code — the application that exposed the paid endpoint       |
+| `w` | string | Wallet code — the facilitator that settled the payment on-chain |
+
+All fields are optional.
+
+### Builder Code Format
+
+Codes must match the pattern `^[a-z0-9_]{1,32}$`:
+
+- **Length**: 1-32 characters
+- **Characters**: lowercase alphanumeric and underscores only
+
+---
+
+## `PaymentRequired`
+
+The application declares its builder code per-route in the payment middleware configuration.
+
+```json
+{
+  "x402Version": 2,
+  "error": "Payment required",
+  "accepts": [ ... ],
+  "extensions": {
+    "builder-code": {
+      "info": {
+        "a": "my_app"
+      },
+      "schema": {
+        "$schema": "https://json-schema.org/draft/2020-12/schema",
+        "type": "object",
+        "properties": {
+          "a": {
+            "type": "string",
+            "pattern": "^[a-z0-9_]{1,32}$",
+            "description": "App builder code"
+          },
+          "w": {
+            "type": "string",
+            "pattern": "^[a-z0-9_]{1,32}$",
+            "description": "Wallet builder code"
+          }
+        },
+        "additionalProperties": false
+      }
+    }
+  }
+}
+```
+
+---
+
+## `PaymentPayload`
+
+The client echoes the builder code extension from `PaymentRequired` into its `PaymentPayload`.
+
+```json
+{
+  "extensions": {
+    "builder-code": {
+      "a": "my_app"
+    }
+  }
+}
+```
+
+The `w` (wallet) field is **not** set by the client. It is added by the facilitator at settlement time.
+
+---
+
+## Builder Code Fields
+
+| Field | Set by      | When                               | Description                                              |
+| ----- | ----------- | ---------------------------------- | -------------------------------------------------------- |
+| `a`   | Application | Per-route middleware configuration | Identifies the application exposing the paid endpoint    |
+| `w`   | Facilitator | Settlement                         | Identifies the facilitator settling the payment on-chain |
+
+---
+
+## Facilitator Behavior
+
+When a facilitator settles a payment containing the `builder-code` extension, it:
+
+1. Reads `a` (app code) from the payment payload extensions
+2. Adds its own builder code as the `w` (wallet) field
+3. Encodes the combined data as an ERC-8021 Schema 2 CBOR suffix
+4. Appends the suffix to the settlement transaction calldata
+
+The facilitator's builder code is configured at initialization and validated against the same `^[a-z0-9_]{1,32}$` pattern.
+
+### Calldata Suffix Construction
+
+The facilitator builds the suffix as follows:
+
+1. CBOR-encode a map containing all present fields (`a`, `w`)
+2. Compute `cborLength` as the byte length of the CBOR data (2 bytes, big-endian)
+3. Append: `[cborData][cborLength][0x02][80218021802180218021802180218021]`
+4. Return the hex-encoded result for the settlement mechanism to append to calldata
+
+---
+
+## Protocol Flow
+
+```
+Client (App)                   Resource Server                Facilitator
+      |                              |                              |
+  1.  |--- request ----------------->|                              |
+      |                              |                              |
+  2.  |<-- 402 PaymentRequired ------|                              |
+      |   extensions.builder-code:   |                              |
+      |     { a: "my_app" }         |                              |
+      |                              |                              |
+  3.  | (sign payment, echo extensions)                             |
+      |                              |                              |
+  4.  |--- request + payment ------->|                              |
+      |   extensions.builder-code:   |                              |
+      |     { a: "my_app" }         |                              |
+      |                              |                              |
+  5.  |                              |--- verify/settle ----------->|
+      |                              |   extensions.builder-code:   |
+      |                              |     { a: "my_app" }         |
+      |                              |                              |
+  6.  |                              |         Facilitator adds w,  |
+      |                              |         encodes CBOR suffix, |
+      |                              |         appends to calldata: |
+      |                              |         [cbor({a:"my_app",   |
+      |                              |          w:"my_fac"})]       |
+      |                              |         [cborLen][0x02][mark] |
+      |                              |                              |
+  7.  |<-- 200 OK + resource data ---|                              |
+      |                              |                              |
+```
+
+---
+
+## Examples
+
+### Single App Attribution
+
+Application declares its builder code:
+
+```json
+{
+  "extensions": {
+    "builder-code": {
+      "info": {
+        "a": "bc_myapp"
+      },
+      "schema": { ... }
+    }
+  }
+}
+```
+
+Settlement calldata suffix (hex):
+
+```
+{original_calldata} a161616862635f6d79617070 000c 02 80218021802180218021802180218021
+```
+
+Decoded:
+
+- CBOR: `{"a": "bc_myapp"}`
+- cborLength: `0x000c` (12 bytes)
+- schemaId: `0x02`
+- marker: `80218021802180218021802180218021`
+
+### App + Facilitator Attribution
+
+After facilitator adds its `w` code at settlement:
+
+```
+{original_calldata} a261616862635f6d7961707061777062635f6d79666163696c697461746f72 001f 02 80218021802180218021802180218021
+```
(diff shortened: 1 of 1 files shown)


specs/extensions/builder_code.md
```diff
@@ -0,0 +1,253 @@
+# Extension: `builder-code`
+
+## Summary
+
+The `builder-code` extension enables **on-chain attribution tracking** for x402 payments by appending [ERC-8021](https://eip.tools/eip/8021) Schema 2 builder codes to settlement transaction calldata. It attributes which application exposed the paid endpoint and which facilitator settled the payment.
+
+This extension only implements **Schema 2** (CBOR-encoded) of ERC-8021. The `s` (services), `m` (custom metadata), and `r` (custom registries) fields are not supported.
```
phdargen: I think the `s` (services) field could be useful to insert an optional client side builder code, this could eg be a frontend like x402scan that lets you call x402 endpoints or a miniapp within the base app

phdargen: Actually if we also allow client builder codes, it might be semantically cleaner to have 'w' = client builder code and 's' = facilitator builder code

phdargen: Would it be worth to unambiguously flag as x402 tx with sth like `m=x402`?

phdargen:
Thanks for putting this together @0xClouds, looks great! 

I think we still have to figure out some cross-chain aspects. Atm the spec just implicitly assumes Base.

Consider an endpoint that accepts payments on Base and Polygon, which registers its Base builder code as `a: "my_app"` as per spec. Now if the client pays on Polygon, the facilitator would wastefully append it to the calldata. 

Also if Polygon or any other EVM chain were to launch a builder code program eventually, there is no way for the server to declare per-network codes. 

So I'd suggest sth like this to address these issues:
```
   "builder-code": {
      "info": {
               "eip155:8453": {
                       "a": "my_app"
               }
     }
```

specs/extensions/builder_code.md
0xClouds: The goal would be to have the builder code be self describing, so the prefix for a builder code would denote it is x402 so instead of bc_ we'd use x402_ 

Metadata is typically for arbitrary data teams need that doesnt fit into the schema we've defined

0xClouds: Hmm would it be possible for a x402 to include themselves in the builder code?

The endpoint they called would have to accept passing in their builder code into the x402 endpoint.

but in this scenario x402 would go in the services array, and the facilitator still acts like a wallet as in the entity that submitted a transaction on chain

0xClouds:
> Thanks for putting this together @0xClouds, looks great!
> 
> I think we still have to figure out some cross-chain aspects. Atm the spec just implicitly assumes Base.
> 
> Consider an endpoint that accepts payments on Base and Polygon, which registers its Base builder code as `a: "my_app"` as per spec. Now if the client pays on Polygon, the facilitator would wastefully append it to the calldata.
> 
> Also if Polygon or any other EVM chain were to launch a builder code program eventually, there is no way for the server to declare per-network codes.
> 
> So I'd suggest sth like this to address these issues:
> 
> ```
>    "builder-code": {
>       "info": {
>                "eip155:8453": {
>                        "a": "my_app"
>                }
>      }
> ```

Custom registries and chain information is already supported in 8021, but today there is only 1 registry on one chain, Base.

`r key = Custom registries - sub-map keyed by entity type (a for application, w for wallet), each containing c (chainId as hex string) and a (20-byte hex-encoded address string)`

```
type Schema2Data = {
  a?: string;
  w?: string;
  s?: string[];
  r?: {
    a?: { c: Hex; a: Hex };
    w?: { c: Hex; a: Hex };
  };
  m?: Record<string, unknown>;
};
```

specs/extensions/builder_code.md
```diff
@@ -0,0 +1,253 @@
...
+### Single App Attribution
+
+Application declares its builder code:
+
+```json
+{
+  "extensions": {
+    "builder-code": {
+      "info": {
+        "a": "bc_myapp"
```
phdargen, line 193: nit: stick with either `bc_myapp` or `myapp`

specs/extensions/builder_code.md
```diff
@@ -0,0 +1,253 @@
...
+
+When a facilitator settles a payment containing the `builder-code` extension, it:
+
+1. Reads `a` (app code) from the payment payload extensions
+2. Adds its own builder code as the `w` (wallet) field
+3. Encodes the combined data as an ERC-8021 Schema 2 CBOR suffix
+4. Appends the suffix to the settlement transaction calldata
+
+The facilitator's builder code is configured at initialization and validated against the same `^[a-z0-9_]{1,32}$` pattern.
+
```
phdargen, line 133: Facilitator MUST verify that `PaymentPayload.extensions["builder-code"].a` matches `PaymentRequired.extensions["builder-code"].info.a`

specs/extensions/builder_code.md
phdargen: Yes as I read it, it would be spec conform for the client to enrich the extension data with its own builder code when it builds the payload:
https://github.com/x402-foundation/x402/blob/09c95943c85e01e5737b02bdd1322a87cb199598/specs/x402-specification-v2.md?plain=1#L140

This would look sth like this with 402:
```
{
  "x402Version": 2,
  "error": "Payment required",
  "resource": {...},
  "accepts": [{...}],
  "extensions": {
    "builder-code": {
      "info": {
        "a": "my_app"
      },
      "schema": {
        "$schema": "https://json-schema.org/draft/2020-12/schema",
        "type": "object",
        "properties": {
          "a": {
            "type": "string",
            "pattern": "^[a-z0-9_]{1,32}$",
            "description": "Server builder code"
          },
          "s": {
            "type": "string",
            "pattern": "^[a-z0-9_]{1,32}$",
            "description": "Client builder code"
          }
        },
        "required": ["a"],
        "additionalProperties": false
      }
    }
  }
}
```

and client payload:
```
{
  "x402Version": 2,
  "accepted": { "...": "..." },
  "payload": { "...": "..." },
  "extensions": {
    "builder-code": {
      "a": "my_app",
      "w": "client_code"
    }
  }
}
```

phdargen: cc @CarsonRoscoe

phdargen: not sure about `w` vs `s`, `wallet` sounds for me more like the entity that signs the payment authorization while the facilitator is a `service` that sponsors gas

specs/extensions/builder_code.md
```diff
@@ -0,0 +1,253 @@
...
+| `a`   | Application | Per-route middleware configuration | Identifies the application exposing the paid endpoint    |
+| `w`   | Facilitator | Settlement                         | Identifies the facilitator settling the payment on-chain |
+
+---
+
+## Facilitator Behavior
+
+When a facilitator settles a payment containing the `builder-code` extension, it:
+
+1. Reads `a` (app code) from the payment payload extensions
```
phdargen: If a server doesn't support the extension, may the facilitator still append its own builder code with empty `a`?

phdargen:
> > Thanks for putting this together @0xClouds, looks great!
> > I think we still have to figure out some cross-chain aspects. Atm the spec just implicitly assumes Base.
> > Consider an endpoint that accepts payments on Base and Polygon, which registers its Base builder code as `a: "my_app"` as per spec. Now if the client pays on Polygon, the facilitator would wastefully append it to the calldata.
> > Also if Polygon or any other EVM chain were to launch a builder code program eventually, there is no way for the server to declare per-network codes.
> > So I'd suggest sth like this to address these issues:
> > ```
> >    "builder-code": {
> >       "info": {
> >                "eip155:8453": {
> >                        "a": "my_app"
> >                }
> >      }
> > ```
> 
> Custom registries and chain information is already supported in 8021, but today there is only 1 registry on one chain, Base.
> 
> `r key = Custom registries - sub-map keyed by entity type (a for application, w for wallet), each containing c (chainId as hex string) and a (20-byte hex-encoded address string)`
> 
> ```
> type Schema2Data = {
>   a?: string;
>   w?: string;
>   s?: string[];
>   r?: {
>     a?: { c: Hex; a: Hex };
>     w?: { c: Hex; a: Hex };
>   };
>   m?: Record<string, unknown>;
> };
> ```

Hm `r` for custom registries is a useful mechanism, but I don't think it addresses either concern I raised: 

1) There is currently no mechanism to tell the facilitator whether to append the calldata on a given settlement chain or not. Imo it makes no sense to append the Base builder code for eg a Polygon payment

2) Per‑chain codes. If eg Polygon launches its own builder‑code program in the future, the same app may have a different code on Polygon (my_app_pl) than on Base (my_app_base). The current info: { a: "my_app" } shape can express only one. With `r` the server could declare a custom registry on Polygon, but then it could'nt also declare one on Base.

I understand that currently it is Base-only, just thinking we could make it more future proof with a small tweak like:
```
"builder-code": {
  "info": {
    "eip155:8453": { "a": "my_app_base" },
    "eip155:137":  { "a": "my_app_pl" }
  }
}
```
where the facilitator emits the entry matching `accepted.network `and skips entirely if there's no match

specs/extensions/builder_code.md
0xClouds: For our own internal parsing we view the facilitator as w.

But we can readd the services array support

specs/extensions/builder_code.md
0xClouds: yeah not an issue

0xClouds:
> > > Thanks for putting this together @0xClouds, looks great!
> > > I think we still have to figure out some cross-chain aspects. Atm the spec just implicitly assumes Base.
> > > Consider an endpoint that accepts payments on Base and Polygon, which registers its Base builder code as `a: "my_app"` as per spec. Now if the client pays on Polygon, the facilitator would wastefully append it to the calldata.
> > > Also if Polygon or any other EVM chain were to launch a builder code program eventually, there is no way for the server to declare per-network codes.
> > > So I'd suggest sth like this to address these issues:
> > > ```
> > >    "builder-code": {
> > >       "info": {
> > >                "eip155:8453": {
> > >                        "a": "my_app"
> > >                }
> > >      }
> > > ```
> > 
> > 
> > Custom registries and chain information is already supported in 8021, but today there is only 1 registry on one chain, Base.
> > `r key = Custom registries - sub-map keyed by entity type (a for application, w for wallet), each containing c (chainId as hex string) and a (20-byte hex-encoded address string)`
> > ```
> > type Schema2Data = {
> >   a?: string;
> >   w?: string;
> >   s?: string[];
> >   r?: {
> >     a?: { c: Hex; a: Hex };
> >     w?: { c: Hex; a: Hex };
> >   };
> >   m?: Record<string, unknown>;
> > };
> > ```
> 
> Hm `r` for custom registries is a useful mechanism, but I don't think it addresses either concern I raised:
> 
> 1. There is currently no mechanism to tell the facilitator whether to append the calldata on a given settlement chain or not. Imo it makes no sense to append the Base builder code for eg a Polygon payment
> 2. Per‑chain codes. If eg Polygon launches its own builder‑code program in the future, the same app may have a different code on Polygon (my_app_pl) than on Base (my_app_base). The current info: { a: "my_app" } shape can express only one. With `r` the server could declare a custom registry on Polygon, but then it could'nt also declare one on Base.
> 
> I understand that currently it is Base-only, just thinking we could make it more future proof with a small tweak like:
> 
> ```
> "builder-code": {
>   "info": {
>     "eip155:8453": { "a": "my_app_base" },
>     "eip155:137":  { "a": "my_app_pl" }
>   }
> }
> ```
> 
> where the facilitator emits the entry matching `accepted.network `and skips entirely if there's no match

> 1. There is currently no mechanism to tell the facilitator whether to append the calldata on a given settlement chain or not. Imo it makes no sense to append the Base builder code for eg a Polygon payment

8021 already supports appending you builder code cross chain. The builder code could be minted on Base but be referenced on any EVM chain and attributed back to the owner of the builder code. Only an offchain parser can decode the calldata suffix and that suffix would hold all the information necessary to attribute it back to the user even if the code was minted on Base and appended on Polygon. 

> 2. Per‑chain codes. If eg Polygon launches its own builder‑code program in the future, the same app may have a different code on Polygon (my_app_pl) than on Base (my_app_base). The current info: { a: "my_app" } shape can express only one. With `r` the server could declare a custom registry on Polygon, but then it could'nt also declare one on Base.

This is valid, but the ability to dynamically select a builder code seems out of scope for this spec, we could revisit this at a time where 8021 intends to open up custom registries to other chains. Also the client can still dynamically pass in the builder code they want even without the above functionality.

jithinraj:
Great work @0xClouds. 

One rule I’d make explicit is: emit `builder-code` only when there is an exact match for `accepted.network`, and treat the emitted value as network-scoped attribution, not a chain-agnostic identity.

The same literal code on Base and Polygon should not be assumed equivalent unless the server declares both.

That keeps cross-chain attribution clean without changing the rest of the proposal.

specs/extensions/builder_code.md
carsonroscoe-cb: Assuming the server does not support the extension, but the client/facilitator do, are there any consequences to clients' putting their own builder code?

Presumably, a supporting server would ensure its the servers' builder code that reaches the facilitator. If its empty by the time the facilitator reads it, the facilitator could put their own. But what if the client assigned it, and the server just echo'd all extensions blindly. The facilitator would process it as if the client's code is the servers'

Is that fine?

phdargen: Facilitator should verify requirements `a` is equal payload `a`. If absent in server defined requirements, should reject/ignore a client set value for `a`.
In any case the client can set its code in `s`

0xClouds:
> Great work @0xClouds.
> 
> One rule I’d make explicit is: emit `builder-code` only when there is an exact match for `accepted.network`, and treat the emitted value as network-scoped attribution, not a chain-agnostic identity.
> 
> The same literal code on Base and Polygon should not be assumed equivalent unless the server declares both.
> 
> That keeps cross-chain attribution clean without changing the rest of the proposal.

Thank you, I think one thing were conflating with builder_codes is that they are chain specific identifiers. However builder codes answer "how much total volume did thia app/faciliator/client drive" regardless of the chain settlement happens on.

The builder codes are identifiers of some real world entity that can not declare itself on chain, and the off chain registry is responsible for minting and handling the cross chain collisions

CarsonRoscoe (approved):

PR #2013, opened 2026-04-13
chore(typescript): bump viem to 2.47.12 across all packages

ryanRfox:
Ref #1971

Bumps viem lockfile resolution to 2.47.12 across both workspaces.
No package.json range changes — all existing `^2.x` ranges already
accept 2.47.12.

### Chain gaps resolved by this bump

| Locked viem | Used by | Mezo 31611 | MegaETH 4326 | Stable 988 | Stable Test 2201 |
|---|---|---|---|---|---|
| 2.23.2 | legacy packages | MISSING | MISSING | MISSING | MISSING |
| 2.40.3 | @x402/paywall, express, fetch, evm | MISSING | MISSING | MISSING | MISSING |
| 2.45.1 | @x402/extensions | MISSING | present | present | present |
| 2.47.12 (this PR) | all direct deps | present | present | present | present |

Old versions (2.23.2, 2.37.3) remain for transitive deps (e.g. wagmi)
that pin their own viem ranges.

## AI disclosure

This PR used an agentic coding workflow and was reviewed by Ryan R. Fox (an actual human) before posting.

PR #2054, opened 2026-04-16
chore(paywall): regenerate bundle + add recurrence guard

### LINKED ISSUE #1971: Stale viem/chains across all TS packages — new chains broken until lockfile updated

ryanRfox:
# chore(paywall): regenerate bundle + add recurrence guard

## Summary

The committed `@x402/paywall` bundle on `main` was last generated against an older viem than the lockfile currently pins. `pnpm install --frozen-lockfile` resolves viem `2.47.12` (via #2013), but the bundled `src/evm/gen/template.ts` (and its Python + Go siblings) were built against an earlier resolution that pre-dated chain definitions for Mezo, MegaETH, and Stable. Thus [supported networks](https://docs.x402.org/core-concepts/network-and-token-support#evm) in the published docs do not match the SDK implementation.

Fixes #1971

At runtime, `EvmPaywall.tsx:59-63` does:

```tsx
const paymentChain: Chain | undefined = Object.values(allChains).find(c => c.id === chainId);
if (!paymentChain) {
  throw new Error(`Unsupported chain ID: ${chainId}`);
}
```

so any `eip155:<id>` whose chain definition isn't in the bundle causes the
paywall component to throw at init. 

This PR:

1. **Adds a PR-time drift check** (`.github/workflows/check_paywall_template.yml`) so a future lockfile or `@x402/evm` change that invalidates the bundle fails loudly instead of shipping silently.
2. **Teaches `build:paywall` to emit output** in the same format the repo's formatters (prettier / ruff) prefer, so the generated files match what `pnpm run format` + `uvx ruff format` would produce.
3. **Regenerates all nine auto-generated template files** (TypeScript, Python, and Go for EVM/SVM/AVM) against the current lockfile. No source changes on any runtime path.

## Paywall template drift check

`.github/workflows/check_paywall_template.yml`:

- Triggers on PRs touching any of:
  - `typescript/packages/http/paywall/**` (source or generated)
  - `typescript/packages/mechanisms/evm/**` (bundled into paywall via
    `@x402/evm/exact/client`)
  - `typescript/pnpm-lock.yaml` (viem version resolution)
  - `python/x402/http/paywall/**` (cross-language sibling templates)
  - `go/http/{evm,svm,avm}_paywall_template.go` (cross-language siblings)
  - `.github/workflows/check_paywall_template.yml` (self-trigger on edits)
- Runs `pnpm --filter @x402/paywall run build:paywall` and then
  `git diff --exit-code` across all nine generated files. Fails the PR
  with an actionable error if they drift.
- Belt-and-suspenders: re-runs `build:paywall` in the same job and compares `sha256sum` between the two runs, failing if output is non-deterministic.

The workflow provides a human-in-the-loop step: when drift is detected, the PR author must run `pnpm --filter @x402/paywall run build:paywall` locally and commit the result. Future work may layer additional automation on top of this signal.

## build.ts formatting fix

`src/evm/build.ts` (and the SVM / AVM siblings) emit `${JSON.stringify(html)}` into the generated files. `JSON.stringify` produces double-quoted output with escaped `\"`. Prettier (for the TS files) and ruff / black (for the Python files) both prefer single quotes when that would avoid escaping, so running `pnpm run format` (or `uvx ruff format`) on the committed files produces single-quoted output. Consequence: even on a commit that introduces no chain data changes, `build:paywall` + `git diff` always showed formatting drift, which would mask or false-positive the drift check.

Utility: a new `src/genHelpers.ts` with:

- `formatTypeScript(filePath, contents)` — runs the prettier API with  the package's `.prettierrc` config.
- `toPythonStringLiteral(s)` — serializes with the same smart-quote rule ruff/black apply (single when content has fewer embedded `'` than `"`, double otherwise). This avoids requiring `uvx` on the build path for pnpm-only contributors.

## Scope

Scope is limited to `@x402/paywall` because it is the only package in this repo that bakes viem chain definitions into a built artifact. Other `@x402/*` packages ship compiled source and resolve viem at runtime from the
consumer's `node_modules`, so they pick up whatever `viem` the consumer has installed.

Go output is unchanged: `JSON.stringify` already emits Go-string-literal- compatible form, and `gofmt` does not rewrite quote style.

## Verification

**GitHub Actions CI** (verified via a fork-internal test PR on
`ubuntu-latest`, Node 24, pnpm 10.7.0 — matching the `@x402/paywall` publish environment):

- `check-format-typescript`, `check-format-examples-typescript`, `check-format-go`
- `check-lint-typescript`, `check-lint-examples-typescript`, `lint-go`, `lint-python`
- `check-package-lock-typescript`, `check-package-lock-examples-typescript`
- `test-go`, `test-python`
- `check-paywall-template` (new, from this PR) 

## Related work

- #2002 proposes scoped viem Dependabot updates for the `typescript/` workspace. The two PRs are complementary: The former reduces the rate at which viem drift accumulates; this PR catches any drift that stil reaches a PR.

## AI disclosure

This PR was prepared with the assistance of a coding agent and reviewed by Ryan R. Fox (an actual human) before submission.


ryanRfox:
rebased this now that #1980 got merged. @phdargen this is the last piece, locking the paywall regen against silent drift as new viem chains land.

phdargen (approved):
LGTM, thanks @ryanRfox 🚀

PR #2072, opened 2026-04-19
fix(go): implement SupportedCache Clear

viktorcrypt:
## Description

Implements `SupportedCache.Clear()` and restores the focused cache test that was previously skipped because the method was missing.

`Clear()` now clears both cached supported responses and their expiry entries while holding the existing cache lock.

## Tests

- `go test . -run TestSupportedCache -v`


phdargen (approved):
Thanks for the contribution @viktorcrypt, looks great 🚀

PR #2099, opened 2026-04-22
Fix: invoke lifecycle hooks in payment payload creation

Ocheretovich:
Add before, after, and failure lifecycle hook invocations that were registered but never called in CreatePaymentPayload and CreatePaymentPayloadV1 methods.

### DIFF
--- a/go/client.go
+++ b/go/client.go
@@ -249,22 +249,40 @@ func (c *x402Client) CreatePaymentPayloadV1(
 	}
 
 	// Before hooks
+	creationCtxV1 := PaymentCreationContext{
+		Ctx:                  ctx,
+		Version:              1,
+		SelectedRequirements: requirements,
+	}
 	for _, hook := range c.beforePaymentCreationHooks {
-		if err := hook(ctx, requirements); err != nil {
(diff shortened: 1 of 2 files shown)


jithinraj:
Good catch. Since hooks are now part of scheme behavior, I’d add one regression test that runs `CreatePaymentPayload` and `CreatePaymentPayloadV1` through the same `before` / `after` / `failure` hook sequence and proves identical order and error propagation.

That’s the sort of drift that only shows up later once more schemes start leaning on hooks.

Ocheretovich:
Fixed hook invocations in CreatePaymentPayloadV1 and CreatePaymentPayload to use typed PaymentCreationContext, PaymentCreatedContext, and PaymentCreationFailureContext instead of raw arguments. Added regression test verifying identical hook order and error propagation across both versions.

Ocheretovich:
Hey @phdargen Latest commit fixes the lint-go failures. Ready for review when you get a chance

phdargen:
Hi @Ocheretovich, the linter and tests still fail

phdargen (approved):
Thanks @Ocheretovich, LGTM 🚀

PR #2102, opened 2026-04-22
spec: Add Near Intents exact scheme

IkerAlus:
## Description

The `exact` payment scheme for Near Intents uses the [NEAR Intents 1Click Swap API](https://docs.near-intents.org/integration/distribution-channels/1click-api/about-1click-api) as the settlement backend. This scheme facilitates cross-chain payments where a client pays a specified amount of a source asset on any [supported origin chain](https://docs.near-intents.org/resources/chain-support), and the resource server (merchant) receives an exact amount of a destination asset on any supported destination chain, with the NEAR Intents solver network executing the cross-chain swap in between.

### Core Design: `payTo` = Deposit Address

The key architectural insight of the scheme is that the `payTo` field in `PaymentRequirements` is set to the **1Click Swap API deposit address** — a unique, one-time address generated by the 1Click Swap API backend when the resource server requests a quote (`POST /v0/quote` with `dry: false`). This means:

- The resource server generates the quote **at 402 response time**, embedding the live `depositAddress` as `payTo`.
- The client's only obligation is to transfer the exact required amount to that `payTo` address on the origin chain — identical in spirit to how the EVM `exact` scheme asks the client to authorize a transfer to a `payTo` address.
- The client's `payload` contains the origin-chain transaction hash (`txHash`) as proof-of-deposit — analogous to the EIP-3009 `signature` in the EVM scheme.
- The facilitator verifies the deposit and polls the 1Click status API for settlement.

### Key Properties

- **Cross-chain native**: Client pays in any supported asset on any origin chain (ETH, BTC, SOL, USDC on Arbitrum, etc.); the merchant receives the destination asset on any supported chain.
- **Gasless for the merchant**: The resource server receives destination tokens without submitting any on-chain transaction.
- **Competitive pricing**: Market makers compete on the 1Click solver network to provide the best swap rate.
- **Automatic refunds**: If the swap cannot be completed, the 1Click API automatically returns funds to the refund address.

### Open Questions
- CAIP-2 network id convention: "near:mainnet" for the settlement network vs "nearintents:mainnet" for the crosschain protocol
- Role of the Facilitator: The "Quote Generation" step of the Facilitator can be assigned to the resource server.
- Facilitator statefulness: It is not a hard requirement for the scheme to work. Delay and stateful constraints can be delegated to the resource server.


IkerAlus:
Hi @phdargen  After a review of the new `upfront` scheme at #2520  I believe the scheme I proposed here will fit better as an `upfront` scheme using the payment-proof (client-settled) method, as the core design here is that the client transfers the required amount to the  `payTo` address on the origin chain.

I wonder if it makes sense to already propose a new alternative PR using the new `upfront` scheme or if it is better to wait until merge of #2520 .

phdargen:
Hi @IkerAlus, awaiting feedback from the team if and how we want to support client prepayments due to the different security model. Better to wait until we have some clarity on https://github.com/x402-foundation/x402/pull/2520, thanks for your patience

mikedotexe:
Hi @IkerAlus 👋 — following this thread with interest. Your June assessment (this design fitting the `upfront`/payment-proof direction once #2520 settles) makes a lot of sense for the cross-chain 1Click flow.

I just opened #2948, which I believe is the complementary piece: the **facilitator-settled, same-asset, exact-amount** case — an `assetTransferMethod: "intents"` amendment inside the already-merged NEAR `exact` scheme (`scheme_exact_near.md`), settling via signed Verifier payloads (`execute_intents`) rather than deposit addresses. It deliberately excludes `token_diff`/cross-asset for the same reason you moved this PR toward `upfront`: quote-dependent amounts don't fit `exact`. Between the two, NEAR Intents would cover both halves — your flow for pay-in-anything cross-chain, this one for exact same-asset settlement where an `erc191` EVM key can pay a NEAR merchant with no NEAR account at all.

A few empirical findings from grounding #2948 against the deployed contract that may help your work too:

- Deployed `simulate_intents` **does validate signatures** (the docs suggest it skips them) — a successful simulation has already passed deserialization → signature → signer authorization, which makes it a strong free preflight.
- The verify-side view surface is live and behaves as needed: `is_nonce_used(signer, nonce)`, `has_public_key`, `mt_balance_of`.
- End-to-end mainnet proof with an unrelated submitter (signature-carried authority) + on-chain replay rejection: [`2R2ANbXf…`](https://www.nearblocks.io/txns/2R2ANbXfhhTC1XaVfYLWqu3but8FKdsimAEEex1tK1dp); and a Circle EVM-wallet `erc191` payload accepted by the Verifier as a byte-exact reference vector: [`ALqQDmeT…`](https://nearblocks.io/txns/ALqQDmeTUouVv6vqG8bqbwBo7RzxBuB7gufPwtcLyw6E).

Would genuinely value NEAR-core eyes on #2948 — especially on the Verifier-facing verification rules and the custody framing. Happy to align terminology between the two so they read as one coherent NEAR Intents story.

IkerAlus:
hey @mikedotexe thanks for the heads up, I will do a proper review on https://github.com/x402-foundation/x402/pull/2948 in the next days, at least the part relevant to Near Intents Verifier contract logic. Also, thank you very much with the extra findings you added.

Quick question wrt the simulate intents docs issue: I assume you refer to the Verifier docs (["Simulating intents" page](https://docs.near-intents.org/integration/verifier-contract/simulating-intents)), yes? What in it makes you say that signature validation is skipped?

mikedotexe:
@IkerAlus You were right to question that. I rechecked the current documentation, and it explicitly lists signature validation as a simulation use case. I had conflated the limitation on external asynchronous effects with signature validation. The deployed implementation and a live probe corroborate the documentation, so I removed the contrary wording from #2948. Thank you for catching it.

specs/schemes/exact/scheme_exact_near_intents.md
```diff
@@ -0,0 +1,486 @@
...
+    │                      │                           │                      │
+    │                      │                           │                      │
+    │  5. Client sends     │                           │                      │
+    │     deposit TX on    │                           │                      │
+    │     origin chain     │                           │                      │
+    │     to payTo address │                           │                      │
+    │  ════════════════════╪═══════════════════════════╪══(on-chain TX)═══════│
+    │                      │                           │                      │
+    │  6. GET /resource    │                           │                      │
+    │  X-PAYMENT: {payload │                           │                      │
```
phdargen, line 48: X-PAYMENT -> PAYMENT-SIGNATURE

Please use v2 notation
https://docs.x402.org/guides/migration-v1-to-v2

specs/schemes/exact/scheme_exact_near_intents.md
```diff
@@ -0,0 +1,486 @@
...
+    │<─────────────────────│                           │                      │
+```
+
+### Step-by-step
+
+1. **Client → Resource Server**: Client makes an HTTP request (e.g., `GET /resource`) without payment headers.
+
+2. **Resource Server → Facilitator**: The resource server's x402 middleware invokes the facilitator to construct `PaymentRequirements`. The facilitator calls the 1Click API `quote` endpoint with `dry: false` and the merchant's swap configuration (origin asset, destination asset, amount, recipient address, refund policy, fees).
+
+3. **Facilitator → 1Click API**: `POST /v0/quote` with `dry: false` and `swapType: EXACT_OUTPUT` returns the full quote including a unique `depositAddress`, `depositMemo` (if applicable), `maxAmountIn`, `amountOut`, `deadline`, and `timeEstimate`.
```
phdargen, line 92: How does the server know which routes to query, i.e. on what chains the client can pay? Or is this restricted to a single fixed origin → destination chain route?

Feels like there is an additional handshake missing to match capabilities: server dest accepts vs client payable origins, then quote the possible routes and pick the optimal one.

phdargen:
Did you close this on purpose @IkerAlus ?

phdargen:
Thanks a lot for the contribution @IkerAlus, cross-chain payments will be a very valuable protocol addition!

As a general comment after doing a first pass, I am not sure if it makes sense to frame this under near:mainnet though.

Lets say the server accepts payments on Base, then the network in its 402 should be base. If the client has USDC on base, they just pay via the regular mechanism. If the client does not have USDC on base but on solana, it may pay the server on base via a crosschain-swap. In this example, neither client nor server would interact with or hold funds on near. Near-intents would be more of a backend API implementation detail, where in principle also other crosschain solutions could be plugged in.

This might be better framed as a `near-intents` extension (or sth more vendor-neutral like `crosschain-swap `extension) such that the 402 still shows a normal dest accept, and the extension only announces which origin chains a client could pay from if they don’t hold the dest asset. Something like:

```
{
  "x402Version": 2,
  "accepts": [
    {
      "scheme": "exact",
      "network": "eip155:8453",
      "amount": "1000000",
      "payTo": "0xMerchantOnBase",
      "maxTimeoutSeconds": 60,
      "extra": {
        "paymentFlow": "upfront"
      }
    }
  ],
  "extensions": {
    "near-intents": {
      "info": {
        "origins": [
          { "network": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp" },
          { "network": "eip155:42161" },
          { "network": "near:mainnet" }
        ]
      }
    }
  }
}
```

Same-chain overlap (client already has USDC on Base) uses the accepts[] entry and skips the swap. No overlap → match client origins against extensions.near-intents, then quote that route

phdargen:
> Thanks a lot for the contribution @IkerAlus, cross-chain payments will be a very valuable protocol addition!
> 
> As a general comment after doing a first pass, I am not sure if it makes sense to frame this under near:mainnet though.
> 
> Lets say the server accepts payments on Base, then the network in its 402 should be base. If the client has USDC on base, they just pay via the regular mechanism. If the client does not have USDC on base but on solana, it may pay the server on base via a crosschain-swap. In this example, neither client nor server would interact with or hold funds on near. Near-intents would be more of a backend API implementation detail, where in principle also other crosschain solutions could be plugged in.
> 
> This might be better framed as a `near-intents` extension (or sth more vendor-neutral like `crosschain-swap `extension) such that the 402 still shows a normal dest accept, and the extension only announces which origin chains a client could pay from if they don’t hold the dest asset. Something like:
> 
> ```
> {
>   "x402Version": 2,
>   "accepts": [
>     {
>       "scheme": "exact",
>       "network": "eip155:8453",
>       "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
>       "amount": "1000000",
>       "payTo": "0xMerchantOnBase",
>       "maxTimeoutSeconds": 60,
>       "extra": {
>         "assetTransferMethod": "eip3009",
>         "paymentFlow": "upfront"
>       }
>     }
>   ],
>   "extensions": {
>     "near-intents": {
>       "info": {
>         "origins": [
>           { "network": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp" },
>           { "network": "eip155:42161" },
>           { "network": "near:mainnet" }
>         ]
>       }
>     }
>   }
> }
> ```
> 
> Same-chain overlap (client already has USDC on Base) uses the accepts[] entry and skips the swap. No overlap → match client origins against extensions.near-intents, then quote that route

Appreciate some more opinions here before deciding for a direction (mechanism vs extension) @IkerAlus @mikedotexe @CarsonRoscoe

specs/schemes/exact/scheme_exact_near_intents.md
IkerAlus: I dont think we need additional steps or prev handshake in the flow:

Theoretically the facilitator pings the `POST /v0/quote` endpoint  for every available origin (available asset/chain in Near Intents). The response from the endpoint contains one single deposit address per query plus the required amount to send.  Then in practice, the facilitator can reduce the amount of queries to the 1CS API service by serving the same quote in every 402 until it expires or gets paid, refreshing in the background. Also, a merchant may offer only a limited, curated set of origins for performance reasons (not long tail assets/chains, no native BTC...). But these are impl details we can simply mention here, not part of the protocol.

specs/schemes/exact/scheme_exact_near_intents.md
IkerAlus: I will fix across the rewrite (5 ocurrences), the draft was written against v1 names

IkerAlus:
hi @phdargen i think your general comment goes in the right direction and it closes a gap my original spec had. I agree on most of your proposed approach with a diff to raise.

- First, 100% agreed that`near:mainnet` should not be tin this spec. 

- Second, I think your proposed extension fits as discovery, not payment. It can list the full set of supported origins with indicative prices (cached `dry: true` quotes, no address allocated) alongside the payable `accepts[]` entries, so a client holding a long-tail asset knows the route exists. That is useful and vendor-neutral — `crosschain-swap` works as a name.  Quick example:

```
  "extensions": {
    "crosschain-swap": {
      "info": {
        "provider": "near-intents",
        "destination": {
          "network": "eip155:8453",
          "amount": "1000000"
        },
        "origins": [
          { "network": "eip155:42161", "asset": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831", "indicativeAmount": "1005000", "timeEstimate": 120 },
          { "network": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp", "asset": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", "indicativeAmount": "1006000", "timeEstimate": 120 },
          { "network": "bip122:000000000019d6689c085ae165831e93", "asset": "bip122:000000000019d6689c085ae165831e93/slip44:0", "indicativeAmount": "1520", "timeEstimate": 1800 },
          { "network": "near:mainnet", "asset": "17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1", "indicativeAmount": "1001000", "timeEstimate": 60 }
        ]
      }
    }
  }
}
```


However, the payment itself needs a payable `accepts[]` entry and as per [my comment above](https://github.com/x402-foundation/x402/pull/2102#discussion_r3922834205) this scheme can provide everything for the payment to initiate.  So: `accepts[]` = one payable entry per offered origin, on the origin network, `paymentFlow: upfront`, deposit address as `payTo`, destination leg in `extra`; `extensions.crosschain-swap` = origin list with indicative prices for discovery. No new step in the flow Example:

```
{
  "x402Version": 2,
  "accepts": [
    // 1. Same-chain: client already holds USDC on Base. Unchanged from today.
    {
      "scheme": "exact",
      "network": "eip155:8453",
      "amount": "1000000",
      "payTo": "0xMerchantOnBase",
      "maxTimeoutSeconds": 60,
      "extra": { "assetTransferMethod": "eip3009" }   // authorization flow (default)
    },

    // 2. Cross-chain from Arbitrum USDC. Payable now: real deposit address.
    {
      "scheme": "exact",
      "network": "eip155:42161",                          // ORIGIN chain: where the client pays
      "asset": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831", // origin asset (USDC on Arbitrum)
      "amount": "1005000",                                // what the client deposits (quote maxAmountIn)
      "payTo": "0x76b4c56085ED136a8744D52bE956396624a730E8", // 1Click deposit address on Arbitrum
      "maxTimeoutSeconds": 300,
      "extra": {
        "assetTransferMethod": "nearintents",
        "paymentFlow": "upfront",
        "refundTo": "0xMerchantOnArbitrum"                // merchant-controlled, on the origin chain
      }
    },

    // 3. Cross-chain from Solana USDC. Same shape, different origin.
    {
      "scheme": "exact",
      "amount": "1006000",
      "payTo": "9xQeWvG816bUx9EPjHmaT23yvVM2ZWbrrpZb9PusVFin",
      "maxTimeoutSeconds": 300,
      "extra": {
        "assetTransferMethod": "nearintents",
        "paymentFlow": "upfront",
        "refundTo": "MerchantOnSolana11111111111111111111111111"
      }
    }
  ],
```

What do you think? Does this split work?

PS: I closed this as I was expecting to open the pure `upfront` scheme spec but as that general approach closed that was a bad move from my side as the changes here may not be that big. However, it seems the repo rules forced me to create a fresh PR (I need to verify).

specs/schemes/exact/scheme_exact_near_intents.md
phdargen: OK makes sense. 

So the deposit address doesnt depend on the client, right? But then caching sounds a bit dangerous, what would happen if client A and client B both fetch the quote concurrently and then both send to the same deposit address? If client A's deposit confirms first, what happens with the deposit of client B?

phdargen:
> hi @phdargen i think your general comment goes in the right direction and it closes a gap my original spec had. I agree on most of your proposed approach with a diff to raise.
> 
> * First, 100% agreed that`near:mainnet` should not be tin this spec.
> * Second, I think your proposed extension fits as discovery, not payment. It can list the full set of supported origins with indicative prices (cached `dry: true` quotes, no address allocated) alongside the payable `accepts[]` entries, so a client holding a long-tail asset knows the route exists. That is useful and vendor-neutral — `crosschain-swap` works as a name.  Quick example:
> 
> ```
>   "extensions": {
>     "crosschain-swap": {
>       "info": {
>         "provider": "near-intents",
>         "destination": {
>           "network": "eip155:8453",
>           "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
>           "amount": "1000000"
>         },
>         "origins": [
>           { "network": "eip155:42161", "asset": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831", "indicativeAmount": "1005000", "timeEstimate": 120 },
>           { "network": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp", "asset": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v", "indicativeAmount": "1006000", "timeEstimate": 120 },
>           { "network": "bip122:000000000019d6689c085ae165831e93", "asset": "bip122:000000000019d6689c085ae165831e93/slip44:0", "indicativeAmount": "1520", "timeEstimate": 1800 },
>           { "network": "near:mainnet", "asset": "17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1", "indicativeAmount": "1001000", "timeEstimate": 60 }
>         ]
>       }
>     }
>   }
> }
> ```
> 
> However, the payment itself needs a payable `accepts[]` entry and as per [my comment above](https://github.com/x402-foundation/x402/pull/2102#discussion_r3922834205) this scheme can provide everything for the payment to initiate. So: `accepts[]` = one payable entry per offered origin, on the origin network, `paymentFlow: upfront`, deposit address as `payTo`, destination leg in `extra`; `extensions.crosschain-swap` = origin list with indicative prices for discovery. No new step in the flow Example:
> 
> ```
> {
>   "x402Version": 2,
>   "accepts": [
>     // 1. Same-chain: client already holds USDC on Base. Unchanged from today.
>     {
>       "scheme": "exact",
>       "network": "eip155:8453",
>       "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
>       "amount": "1000000",
>       "payTo": "0xMerchantOnBase",
>       "maxTimeoutSeconds": 60,
>       "extra": { "assetTransferMethod": "eip3009" }   // authorization flow (default)
>     },
> 
>     // 2. Cross-chain from Arbitrum USDC. Payable now: real deposit address.
>     {
>       "scheme": "exact",
>       "network": "eip155:42161",                          // ORIGIN chain: where the client pays
>       "asset": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831", // origin asset (USDC on Arbitrum)
>       "amount": "1005000",                                // what the client deposits (quote maxAmountIn)
>       "payTo": "0x76b4c56085ED136a8744D52bE956396624a730E8", // 1Click deposit address on Arbitrum
>       "maxTimeoutSeconds": 300,
>       "extra": {
>         "assetTransferMethod": "nearintents",
>         "paymentFlow": "upfront",
>         "refundTo": "0xMerchantOnArbitrum"                // merchant-controlled, on the origin chain
>       }
>     },
> 
>     // 3. Cross-chain from Solana USDC. Same shape, different origin.
>     {
>       "scheme": "exact",
>       "network": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp",
>       "asset": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
>       "amount": "1006000",
>       "payTo": "9xQeWvG816bUx9EPjHmaT23yvVM2ZWbrrpZb9PusVFin",
>       "maxTimeoutSeconds": 300,
>       "extra": {
>         "assetTransferMethod": "nearintents",
>         "paymentFlow": "upfront",
>         "refundTo": "MerchantOnSolana11111111111111111111111111"
>       }
>     }
>   ],
> ```
> 
> What do you think? Does this split work?
> 
> PS: I closed this as I was expecting to open the pure `upfront` scheme spec but as that general approach closed that was a bad move from my side as the changes here may not be that big. However, it seems the repo rules forced me to create a fresh PR (I need to verify).

I like this, looks clean @IkerAlus. Somehow I also cant reopen this PR. I suggest to open a new PR with the shape you suggested above and then we go from there. Please add a link to the previous discussion here

specs/schemes/exact/scheme_exact_near_intents.md
IkerAlus: yep, the address doesn't depend on the client, only on the merchant's destination, the origin asset and the deadline. 

Good that you mention this since I think I should consider this race condition in the spec. Let's see, given your scenario what happens depends on timing:

- Both deposits land before the swap executes: 1Click aggregates them into the one quote. `EXACT_OUTPUT` delivers the merchant amount once, the excess refunds to `refundTo` (the merchant's own address in origin chain), and both tx hashes appear in `originChainTxHashes`. The facilitator honors both proofs, the merchant received value for both.
- B lands after A's swap has settled: the quote is already consumed. B's deposit refunds to `refundTo` (again the merchant) and the facilitator rejects B's proof because the quote is no longer active. B paid and got nothing, and recovers off-band under the merchant's terms.

So this race-deposits situation is ok for the merchant but not for client B 

I will add two mitigations to the spec, both conformant to what we added in the new #3145:

1. **Rotate on first deposit detection, not on success.** As soon as any deposit is observed on the shared address, even unconfirmed, mint a fresh quote for every later 402. The exposure window shrinks from the whole settlement period to the seconds between A's broadcast and the facilitator noticing it.

2. **Per-402 unique addresses** where the merchant can't accept any client ever needing off-band recovery (this is the approach I labelled as "theoretical" in my comment above). The cost is one quote per 402 per offered origin, which origin curation by merchant should keep bounded, and for agent traffic most 402s do get paid, so the quote waste is smaller than it looks.

I'll add these points to the new spec if that is ok.

IkerAlus:
sounds good @phdargen  my idea is to create a new PR branch off current main with two files in it:
- `specs/schemes/exact/scheme_exact_near_intents.md`: the mechanism.
- `specs/extensions/extension-crosschain-swap.md`: the discovery extension, following the extension-auth-hints.md naming. will add the link here when ready for review.

PhilBot402:
> sounds good @phdargen my idea is to create a new PR branch off current main with two files in it:
> 
> * `specs/schemes/exact/scheme_exact_near_intents.md`: the mechanism.
> * `specs/extensions/extension-crosschain-swap.md`: the discovery extension, following the extension-auth-hints.md naming. will add the link here when ready for review.

Sounds good @IkerAlus ! Not sure yet if we even need an additional extension then, lets see how it plays out

IkerAlus:
hey @phdargen @CarsonRoscoe the new PR is open: https://github.com/x402-foundation/x402/pull/3370

PR #2135, opened 2026-04-25
fix(paywall): use dynamic token decimals instead of hardcoded 6

### LINKED ISSUE #1979: bug(paywall): EVM paywall hardcodes 6 decimals — breaks non-6-decimal tokens

natsukingly:
## Summary

Closes #1979.

The EVM paywall hardcoded 6 decimals (USDC) in three places, breaking display amounts and balance reads for any token with different precision (e.g. MegaUSD on MegaETH uses 18 decimals).

### Changes

- **`evm/index.ts`** — read decimals from `getDefaultAsset(network)` instead of dividing by `1_000_000`. Falls back to 6 when the network has no default asset.
- **`evm/utils.ts`** — replace the USDC-only `getUSDCBalance` (with its 3-chain hardcoded address map) with generic `getTokenBalance(client, tokenAddress, owner)` and `getTokenDecimals(client, tokenAddress)` that work on any ERC-20.
- **`evm/EvmPaywall.tsx`** — use the requirement's `asset` address as the token, read the on-chain `decimals()`, and format the balance with the resolved precision.
- **`@x402/evm`** — export `getDefaultAsset` from the package root so the paywall can consume the existing default-asset registry.

### Behavior

```
USDC (6 decimals):   amount=100000             → displays 0.1
MegaUSD (18 decimals): amount=100000000000000000 → displays 0.1 (was 100000000000)
unknown network:     amount=100000             → displays 0.1 via fallback to 6
```

Balance reads now query `decimals()` on the token contract per the issue's recommended client-side approach, so any new chain adding a non-6-decimal default token works without further code changes.

## Test plan

- [x] New tests for `getTokenBalance` / `getTokenDecimals` covering normal returns, bigint coercion, and RPC failure (`evm/utils.test.ts`)
- [x] New `evmPaywall.generateHtml` cases for 6-decimal, 18-decimal, and fallback paths (`network-handlers.test.ts`)
- [x] `pnpm --filter @x402/paywall run test / lint / format:check / build` all pass
- [x] `pnpm run test` (monorepo) — 45/45 tasks pass

### DIFF
--- a/typescript/packages/http/paywall/package.json
+++ b/typescript/packages/http/paywall/package.json
@@ -75,6 +75,7 @@
     "@wallet-standard/base": "^1.1.0",
     "@wallet-standard/features": "^1.1.0",
     "@x402/core": "workspace:~",
+    "@x402/evm": "workspace:~",
     "viem": "^2.39.3",
     "wagmi": "^2.17.1",
     "zod": "^3.24.2"
--- a/typescript/packages/http/paywall/src/evm/EvmPaywall.tsx
+++ b/typescript/packages/http/paywall/src/evm/EvmPaywall.tsx
@@ -1,5 +1,12 @@
 import { useCallback, useEffect, useMemo, useState } from "react";
-import { createPublicClient, formatUnits, http, publicActions, type Chain } from "viem";
+import {
+  createPublicClient,
+  formatUnits,
+  http,
+  publicActions,
+  type Address,
+  type Chain,
+} from "viem";
 import * as allChains from "viem/chains";
 import { useAccount, useSwitchChain, useWalletClient, useConnect, useDisconnect } from "wagmi";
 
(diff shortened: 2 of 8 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/1980

PR #2155, opened 2026-04-28
fix(python): exclude MCP tests from wheel

aproskill81:
## Summary

- Replace the MCP directory force-include with explicit runtime MCP files
- Prevent `python/x402/mcp/tests` from being packaged into the Python wheel
- Keep MCP runtime modules, README, and CHANGELOG included in the wheel

## Verification

- Built the Python wheel with Hatchling
- Confirmed `x402/mcp/tests/*` entries in the wheel: `0`
- Confirmed runtime `x402/mcp/*` files remain packaged
- Ran `git diff --check`


phdargen (approved):
LGTM, thanks @aproskill81 🚀

PR #2171, opened 2026-05-01
solve discovery probes time out on serverless

### LINKED ISSUE #2157: `@x402/hono`: `paymentMiddleware` `syncFacilitatorOnStart` default makes discovery probes time out on cold serverless/edge isolates
# Summary

`paymentMiddleware()` from `@x402/hono` defaults to `syncFacilitatorOnStart: true`, which awaits `httpServer.initialize()` (a synchronous `getSupported()` round-trip to the facilitator) before the first paid-route request can resolve. On Cloudflare Workers — and likely any other serverless or edge runtime — every cold isolate pays this cost on its first `/paid/*` request.

We observed a single cold-init taking ~14s end-to-end via wrangler tail. Discovery probes from indexers (e.g., x402scan's `Add Server` flow) typically time out before that completes and report the route as "no valid x402 response found" — despite the manifest being correct and hot isolates serving a 402 in ~300ms.

## Reproduction

- Runtime: Cloudflare Workers, `wrangler 4.73.0`, `@x402/hono@2.10.0`, `@x402/core@2.10.0`, `@x402/extensions@2.10.0`
- Standard setup: `paymentMiddleware(routes, server)` over `withBazaar(new HTTPFacilitatorClient(createFacilitatorConfig(...)))`, three GET `/paid/*` routes with `declareDiscoveryExtension`
- After any deploy (which evicts all isolates), the first `GET /paid/<route>` per isolate hangs 10–15s on init before returning 402; subsequent requests on the same isolate return in ~300ms

Wrangler tail capture (single cold isolate):

```
GET /paid/mocap - Canceled @ 12:26:31 PM     # client gave up, isolate still initializing
GET /paid/mocap - Ok       @ 12:26:43 PM     # 12s later, init completes
  (log) [x402/cdp] middleware initialized
  (log) [cdp-facilitator] /supported → 200
```

## Code path

`@x402/hono/dist/esm/index.mjs:123-173` (paraphrased):

```js
function paymentMiddlewareFromHTTPServer(httpServer, paywallConfig, paywall, syncFacilitatorOnStart = true) {
  let initPromise = syncFacilitatorOnStart ? httpServer.initialize() : null;
  ...
  return async (c, next) => {
    ...
    if (syncFacilitatorOnStart && !isInitialized) {
      await initializeHttpServer();   // every first request on a cold isolate awaits this
    }
    ...
    result = await httpServer.processHTTPRequest(context, paywallConfig);
```

`httpServer.initialize()` (`@x402/core/dist/esm/server/index.mjs:206`) iterates over the configured facilitator clients and awaits `facilitatorClient.getSupported()` on each, populating `supportedResponsesMap` before any 402 can be issued.

## Why this matters for discovery specifically

The metadata required to emit a 402 (`accepts`, `extensions`, route schema) is entirely declared in the route config — it doesn't depend on the supported-kinds map at request time. The supported-kinds map appears to be needed at verify/settle time, when an inbound payment header has to be matched against what the facilitator can actually process.

Holding 402 generation behind init means every discovery probe on a cold isolate stalls on a facilitator round-trip the probe doesn't need. On platforms with frequent isolate cycling (CF Workers, Vercel Edge, Lambda@Edge, etc.), this happens often. Indexer probe timeouts are typically far shorter than this latency, so probes are reported as failed even when the route is healthy.

## Context

Found while investigating #2112 (Bazaar non-indexing on `@x402/hono`), and reading alongside #2156 (a parallel Bazaar non-indexing report on `@x402/express`). This issue is independent of those — they describe a facilitator-side bug (`EXTENSION-RESPONSES` never emitted); this is an SDK-side concern that can compound discovery failures on serverless runtimes regardless of the facilitator-side outcome.

## Possible directions

1. **Flip the default** to `syncFacilitatorOnStart: false`. Discovery probes wouldn't block; eager init remains opt-in.
2. **Decouple init from the 402 path** — generate 402s from cached route config without awaiting init; only await on verify/settle.
3. **Document a discovery-mode recipe** in the README — safe `syncFacilitatorOnStart: false` configuration for serverless deployments, with notes on what's verified at request time vs. lazily.

Happy to provide additional traces or test patches if useful.

MdSadiqMd:
## Description

Fixes discovery probe timeouts on cold serverless/edge isolates (Cloudflare Workers, Vercel Edge, Lambda@Edge) for all x402 HTTP middleware packages.

Closes #2157

**Root cause — two coupled problems:**

1. All middleware awaited `httpServer.initialize()` (a blocking `/supported` round-trip to the facilitator) before processing *any* request on a paid route, including discovery probes that only need a 402 response.
2. `x402ResourceServer.buildPaymentRequirements()` threw immediately when `getSupportedKind()` returned undefined, making it impossible to generate a 402 without completing init — so simply not awaiting init would produce a 5xx instead of a 402.

**Fix — decouple init from 402 generation (Approach 1 from the issue):**

### `@x402/core`

- `x402ResourceServer`: Added `_initialized` flag and `initialized` getter. Modified `buildPaymentRequirements()`: when `getSupportedKind()` returns `undefined` **and the server is not yet initialized**, a synthetic `SupportedKind` is constructed from the route config (`{x402Version, scheme, network}`) and passed to `enhancePaymentRequirements()` instead of throwing. This allows 402 responses to be generated from static route config without a facilitator round-trip. When the server **has** been initialized and a kind is not found, the existing error is still thrown (legitimate misconfiguration).
- `x402HTTPResourceServer`: Exposes `initialized` getter delegating to the resource server.

### `@x402/hono`, `@x402/express`, `@x402/fastify`, `@x402/next`

All middleware now split the init-await on the presence of a payment header:

- **No payment header (discovery probe):** proceed immediately to `processHTTPRequest` — background init continues, 402 is generated from route config.
- **Payment header present (verify/settle path):** await init as before — facilitator data is required to verify and settle.

**Behavior table:**

| Scenario | Before | After |
|---|---|---|
| Cold isolate + discovery probe | Blocks 10–15 s on facilitator | 402 returned immediately |
| Cold isolate + payment request | Blocks on init, then verify/settle | Same (correct — init required) |
| Hot isolate + any request | Fast ~300 ms | Unchanged |
| Init failure + discovery probe | 502 / timeout | 402 from route config; init retries in background |
| Init failure + payment request | 502, retry on next request | Same |

## Tests

```bash
cd typescript && pnpm test
```

- **3 new core unit tests** (`x402ResourceServer`):
  - `buildPaymentRequirements` succeeds without `initialize()` (cold-start tolerance)
  - `initialized` is `true` after successful `initialize()`
  - `initialized` stays `false` when `initialize()` fails

- **Updated + new middleware tests** (hono, express, next):
  - Existing retry-on-init-failure tests updated to send requests **with** a payment header (the path that still blocks on init)
  - New "does not block on init for discovery probes" test in hono and express
  - New "does not block when `requireCompletion` is false" test in next utils

All existing tests pass unchanged:
- `@x402/core`: 419 tests ✓
- `@x402/hono`: 20 tests ✓
- `@x402/express`: 23 tests ✓
- `@x402/fastify`: 23 tests ✓
- `@x402/next`: 53 tests ✓

## Checklist

### DIFF
--- a/typescript/.changeset/fix-cold-start-discovery-probe-timeouts.md
+++ b/typescript/.changeset/fix-cold-start-discovery-probe-timeouts.md
@@ -1,0 +1,9 @@
+---
+"@x402/core": patch
+"@x402/express": patch
+"@x402/fastify": patch
+"@x402/hono": patch
+"@x402/next": patch
+---
+
+Fix cold-start discovery probe timeouts on serverless/edge runtimes. Decouples facilitator init from 402 generation: probes no longer block on the facilitator round-trip; init is awaited only when a payment header is present.
--- a/typescript/packages/core/src/http/x402HTTPResourceServer.ts
+++ b/typescript/packages/core/src/http/x402HTTPResourceServer.ts
@@ -368,6 +368,15 @@ export class x402HTTPResourceServer {
   }
 
   /**
+   * Whether the server has been initialized (facilitator support fetched).
+   *
+   * @returns True if initialize() has completed successfully at least once
+   */
+  get initialized(): boolean {
+    return this.ResourceServer.initialized;
+  }
+
+  /**
    * Get the routes configuration.
    *
    * @returns The routes configuration
--- a/typescript/packages/core/src/server/x402ResourceServer.ts
+++ b/typescript/packages/core/src/server/x402ResourceServer.ts
@@ -206,6 +206,7 @@ export class x402ResourceServer {
     new Map();
   private registeredExtensions: Map<string, ResourceServerExtension> = new Map();
   private extensionHookAdapters = new Map<string, ExtensionAdapterHandles>();
+  private _initialized = false;
 
   private beforeVerifyHooks: BeforeVerifyHook[] = [];
   private afterVerifyHooks: AfterVerifyHook[] = [];
@@ -232,6 +233,15 @@ export class x402ResourceServer {
       // Single client provided
       this.facilitatorClients = [facilitatorClients];
     }
+  }
+
+  /**
+   * Whether the server has been initialized (facilitator support fetched).
+   *
+   * @returns True if initialize() has completed successfully at least once
+   */
+  get initialized(): boolean {
+    return this._initialized;
   }
 
   /**
@@ -527,6 +537,8 @@ export class x402ResourceServer {
             "Failed to initialize: no supported payment kinds loaded from any facilitator.",
           );
     }
+
+    this._initialized = true;
   }
 
   /**
@@ -604,12 +616,23 @@ export class x402ResourceServer {
       SchemeNetworkServer.scheme,
     );
 
-    if (!supportedKind) {
+    if (!supportedKind && this._initialized) {
+      // Server has been initialized but facilitator doesn't support this combination
       throw new Error(
         `Facilitator does not support ${SchemeNetworkServer.scheme} on ${resourceConfig.network}. ` +
           `Make sure to call initialize() to fetch supported kinds from facilitators.`,
       );
     }
+
+    // When not yet initialized, use a synthetic kind from route config so 402 responses
+    // can be generated without waiting for the facilitator round-trip.  This is safe
+    // because the 402 metadata comes from the route config; the facilitator-provided
+    // extras (e.g. feePayer, areFeesSponsored) are only needed at verify/settle time.
+    const effectiveKind = supportedKind ?? {
+      x402Version,
+      scheme: SchemeNetworkServer.scheme,
+      network: resourceConfig.network,
+    };
 
     // Get facilitator extensions for this combination
     const facilitatorExtensions = this.getFacilitatorExtensions(
@@ -639,11 +662,10 @@ export class x402ResourceServer {
     };
 
     // Delegate to the implementation for scheme-specific enhancements
-    // Note: enhancePaymentRequirements expects x402Version in the kind, so we add it back
     const requirement = await SchemeNetworkServer.enhancePaymentRequirements(
       baseRequirements,
       {
-        ...supportedKind,
+        ...effectiveKind,
         x402Version,
       },
       facilitatorExtensions,
--- a/typescript/packages/core/test/mocks/generic/MockFacilitatorClient.ts
+++ b/typescript/packages/core/test/mocks/generic/MockFacilitatorClient.ts
@@ -14,6 +14,7 @@ import { PaymentPayload, PaymentRequirements } from "../../../src/types/payments
  */
 export class MockFacilitatorClient implements FacilitatorClient {
   private supportedResponse: SupportedResponse;
+  private supportedError: Error | null = null;
   private verifyResponseOrError: VerifyResponse | Error;
   private settleResponseOrError: SettleResponse | Error;
 
@@ -43,6 +44,9 @@ export class MockFacilitatorClient implements FacilitatorClient {
    */
   async getSupported(): Promise<SupportedResponse> {
     this.getSupportedCalls++;
+    if (this.supportedError) {
+      throw this.supportedError;
+    }
     return this.supportedResponse;
   }
 
@@ -127,6 +131,14 @@ export class MockFacilitatorClient implements FacilitatorClient {
    */
   setSettleResponse(response: SettleResponse | Error): void {
     this.settleResponseOrError = response;
+  }
+
+  /**
+   *
+   * @param error
+   */
+  setSupportedError(error: Error | null): void {
+    this.supportedError = error;
   }
 
   /**
--- a/typescript/packages/core/test/unit/server/x402ResourceServer.test.ts
+++ b/typescript/packages/core/test/unit/server/x402ResourceServer.test.ts
@@ -403,6 +403,54 @@ describe("x402ResourceServer", () => {
           }),
       ).rejects.toThrow("Facilitator does not support test-scheme on test:network");
     });
+
+    it("should build requirements without initialize() for discovery (cold-start tolerance)", async () => {
+      const mockClient = new MockFacilitatorClient(
+        buildSupportedResponse({
+          kinds: [{ x402Version: 2, scheme: "test-scheme", network: "test:network" as Network }],
+        }),
+      );
+
+      const server = new x402ResourceServer(mockClient);
+      const mockScheme = new MockSchemeNetworkServer("test-scheme");
+      server.register("test:network" as Network, mockScheme);
+
+      // Do NOT call initialize() — simulates cold serverless isolate
+      expect(server.initialized).toBe(false);
+
+      const requirements = await server.buildPaymentRequirements({
+        scheme: "test-scheme",
+        payTo: "recipient",
+        price: 1.0,
+        network: "test:network" as Network,
+      });
+
+      // Should succeed with synthetic kind (no facilitator data)
+      expect(requirements).toHaveLength(1);
+      expect(requirements[0].scheme).toBe("test-scheme");
+      expect(requirements[0].network).toBe("test:network");
+      expect(requirements[0].payTo).toBe("recipient");
+    });
+
+    it("should set initialized to true after successful initialize()", async () => {
+      const mockClient = new MockFacilitatorClient(buildSupportedResponse());
+      const server = new x402ResourceServer(mockClient);
+
+      expect(server.initialized).toBe(false);
+      await server.initialize();
+      expect(server.initialized).toBe(true);
+    });
+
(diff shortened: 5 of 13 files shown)


piiiico (commented):
Missing changelog fragment (the unchecked checklist item). The changeset config links all HTTP packages to `@x402/core` — you need one file covering all five changed packages.

Create `typescript/.changeset/<slug>.md` with:

\`\`\`md
---
'@x402/core': patch
'@x402/express': patch
'@x402/fastify': patch
'@x402/hono': patch
'@x402/next': patch
---

Fix cold-start discovery probe timeouts on serverless/edge runtimes. Decouples facilitator init from 402 generation: probes no longer block on the facilitator round-trip; init is awaited only when a payment header is present.
\`\`\`

Run `pnpm -C typescript changeset` if you want the interactive tool, or create the file directly — either works.

MdSadiqMd:
> Missing changelog fragment (the unchecked checklist item). The changeset config links all HTTP packages to `@x402/core` — you need one file covering all five changed packages.
> 
> Create `typescript/.changeset/<slug>.md` with:
> 
> ## ```md
> ## '@x402/core': patch
> '@x402/express': patch
> '@x402/fastify': patch
> '@x402/hono': patch
> '@x402/next': patch
> Fix cold-start discovery probe timeouts on serverless/edge runtimes. Decouples facilitator init from 402 generation: probes no longer block on the facilitator round-trip; init is awaited only when a payment header is present. ```
> 
> Run `pnpm -C typescript changeset` if you want the interactive tool, or create the file directly — either works.

Done

piiiico (approved):
Changeset looks correct — covers all five packages with the right patch level and description. All checklist items done. LGTM 🚀

MdSadiqMd:
@piiiico I think you need to leave a review via write access. Is there anything I need to do from my side?

Karl-Keller:
@MdSadiqMd - Testing against api.rtkmotion.io now - the cold-isolate Worker case from the issue.  Will report back with timing data.  Ultimately, a Coinbase/foundation maintainer needs to actually approve and merge.

piiiico:
Hey @MdSadiqMd — I already left the APPROVED review at 10:13 UTC (scroll up). Nothing more needed from your side.

I don't have merge rights to this repo, so a Coinbase/foundation maintainer still has to formally approve and click merge. The implementation looks solid: root cause analysis is correct, the two-path split (probe vs. payment header) is clean, tests cover the cold-start path, changeset is in. @Karl-Keller's timing data from api.rtkmotion.io should help push this over the line with the maintainers.

MdSadiqMd:
Makes sense @piiiico 
Thanks for the review

Karl-Keller:
```markdown
Tested the PR build against api.rtkmotion.io's exact cold-isolate failure case
from #2157 (Cloudflare Workers + `@x402/hono`). Built the PR head, packed the
artifacts, deployed to a separate Worker on a workers.dev test environment
with the same `paymentMiddleware` config as production but **forced init to
fail** by binding fake CDP credentials — the worst-case "discovery probe meets
a server whose facilitator init never completes" scenario. Production
api.rtkmotion.io was used unchanged as the 2.10.0 baseline.

**Cold-isolate timing** — 3 GET probes per target, no payment header:

| Build | Init outcome | probe 1 | probe 2 | probe 3 |
|---|---|---|---|---|
| PR @ 2.11.0 (test) | Guaranteed fail (fake creds) | 213ms | 391ms | 305ms |
| 2.10.0 (prod baseline) | Succeeds (real creds) | 611ms | 305ms | 305ms |

vs. the ~14s cold-init tax documented in #2157, that's ~30–60× faster on the
first probe — even though the PR build's init was actively failing in the
background.

**402 manifest field diff** — `accepts[0]` from the decoded `PAYMENT-REQUIRED`:

| Field | PR cold-isolate | 2.10.0 baseline |
|---|---|---|
| amount | 250000 | 250000 |
| asset | 0x833589fC…02913 | 0x833589fC…02913 |
| extra | {name: USD Coin, version: 2} | {name: USD Coin, version: 2} |
| maxTimeoutSeconds | 300 | 300 |
| network | eip155:8453 | eip155:8453 |
| payTo | 0x6C11F8a2…B10C | 0x6C11F8a2…B10C |
| scheme | exact | exact |

Identical, including `extensions.bazaar.{info, schema}`. Indexer-relevant
fields are all present; the EIP-712 domain bits live in route config, not
SupportedKind, so the synthetic-kind path emits a fully-formed 402.

Telemetry confirmed init really was failing in the background (11 ×
`Failed to fetch supported kinds from facilitator` warns over 12 cold
isolates), with **0 error-level events and 0 5xx**.

**Tested**
- PR build, cold isolate, init guaranteed to fail, discovery probe →
  402 in <400ms with a fully-formed manifest
- PR cold-path 402 byte-identical to 2.10.0 warm baseline on EVM/USDC
- Init failure + discovery probe → 402 valid, no errors,
  background-retry path live (covered by the PR's own updated
  retry-on-init-failure unit test)

**Not tested live**
- PR build with init *successfully* completing in background — the
  comparison above is PR-with-failing-init vs 2.10.0-with-working-init.
  Direct PR-with-working-init would require swapping in real CDP creds
  on the test Worker; the PR's own unit tests cover the
  `initialized=true` path.
- Real `transferWithAuthorization` settlement from the test Worker. The
  PR's own updated retry-on-init-failure test covers this code path;
  live verification adds marginal signal.

The test Worker is left running with the PR build wired in. If a maintainer
wants live confirmation of either of the above two paths (real-creds
discovery or a real settlement), happy to swap the fake CDP creds for real
ones and re-probe — would also be a chance to capture an on-chain tx hash
on the PR build.

**Approved from this end.** Nice work @MdSadiqMd, and thanks @piiiico for
the changeset catch. Hoping a Coinbase / foundation maintainer can pick this
up — five teams on #2112 are watching, and this fix unblocks x402scan
submission for serverless deployments.
```

piiiico:
@phdargen — this is ready to merge. Two approvals, all checklist items done, Karl-Keller's timing data from live Cloudflare Workers testing is in (30–60× faster cold probe even with init forced to fail). Five teams on #2112 are waiting on this to unblock x402scan serverless submissions. Ping if anything's blocking on your end.

MdSadiqMd:
@CarsonRoscoe, this PR is reviewed and ready to merge. Can you have a look at it, please?

phdargen:
Thanks for reporting the issue @Karl-Keller and for working on a solution @MdSadiqMd!

Unfortunately, the proposed fix would only work for basic EVM `exact`, where the 402 can be fully built from local route config. All other schemes depend on facilitator-provided `SupportedKind.extra,` eg `extra.feePayer` for SVM/Aptos `exact` or `extra.facilitatorAddress` for EVM `upto`. This facilitator metadata is required for clients to produce valid payloads, so skipping `/supported` on cold-starts may leave clients with no way to do payments

phdargen:
A possible solution would be to fetch `supported/` out of band and then provide static metadata to `HTTPFacilitatorClient` to avoid the network round trip on cold starts:
```
import supported from "./supported.json";

const facilitatorClient = new HTTPFacilitatorClient({
  url: facilitatorUrl,
  supported,
});
```

Will discuss with the team and get back to you

piiiico:
Synthetic kind carries `{x402Version, scheme, network}` only. That's enough for EVM `exact` because the EIP-712 domain bits and asset come from route config — which is why @Karl-Keller's USDC-on-Base byte-diff came out identical to baseline. Doesn't generalize. SVM/Aptos `exact` needs `extra.feePayer`; EVM `upto` needs `extra.facilitatorAddress`. Both have to come from the facilitator.

The static-metadata route is universal but moves the staleness problem onto the developer. They ship `supported.json`, the facilitator adds a network or rotates an address, deploys silently break until someone re-fetches. Different tradeoff, not strictly worse.

Two ways to land this:

1. Narrow this PR to `scheme === "exact"` on EVM — synthetic-kind path with a runtime warning when other schemes hit the cold path; existing throw stays for those. EVM-exact-on-USDC is the case in #2157, so the immediate adopters get unblocked without the developer doing anything.
2. Close in favor of `HTTPFacilitatorClient({ supported })`, with a docs/CLI patch that generates `supported.json` from a live facilitator endpoint so devs aren't hand-maintaining it.

I'd lean (1) — it's the path that doesn't add work for downstream developers and covers the failure mode that triggered the issue. But (2) is cleaner architecturally; if the team wants that, @MdSadiqMd can pivot.

piiiico:
@phdargen — no rush, just checking if you've had a chance to discuss with the team. Happy to help with whichever direction you land on.

MdSadiqMd:
Hey @phdargen, I hope you're doing well
Any update on this?

Issue #2045, opened 2026-04-16
Add .env setup documentation and examples for SDK integration tests


ethanoroshiba:
The Python and TypeScript SDKs lack documentation for running integration tests locally. Required environment variables are only discoverable by reading test source code.

- `python/x402/README.md` and `python/CONTRIBUTING.md` don't mention `.env` setup or required vars for integration tests (behavior is buried in `conftest.py` and test files)
- `typescript/CONTRIBUTING.md` mentions `pnpm test:integration` but doesn't document required env vars or `.env` placement
- Go SDK has a README with guidance but references a `.env.example` that doesn't exist on disk

PR #2186, opened 2026-05-04
docs: add Go integration test .env example

nicanatoly:
## Summary

- add `go/.env.example` for integration test configuration
- update the Go integration test README to point contributors at the example file
- clarify that real private keys belong only in ignored `.env`, while `.env.example` is safe to commit

## Context

This addresses the Go SDK part of #2045, where the integration test README referenced `.env.example` but the file was not present on disk.

## Safety

The example uses placeholders only. It does not include real private keys, wallet addresses, tokens, or transactions.

## Tests

Docs-only change; not run.

### DIFF
--- a/go/.env.example
+++ b/go/.env.example
@@ -1,0 +1,20 @@
+# Go integration test environment.
+#
+# Copy this file to `go/.env`, fill in testnet-only values, then run:
+#
+#   source .env
+#   make test-integration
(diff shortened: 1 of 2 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/2078

PR #2221, opened 2026-05-07
fix(core,hono): decouple 402 generation from facilitator init for serverless cold start

### LINKED ISSUE #2157: `@x402/hono`: `paymentMiddleware` `syncFacilitatorOnStart` default makes discovery probes time out on cold serverless/edge isolates

lau90eth:
Fixes #2157

## Root cause

`buildPaymentRequirements()` called `getSupportedKind()` which threw when
`supportedResponsesMap` was empty (facilitator not yet initialized).
This made `syncFacilitatorOnStart: false` unusable — the middleware would
crash on the first 402 response instead of returning a valid 402.

## Fix

**@x402/core** (`x402ResourceServer.ts`):
- When `supportedResponsesMap` is empty, skip `getSupportedKind` validation
  and `enhancePaymentRequirements`
- Return `baseRequirements` directly using local route config data
- When initialized, behavior is **unchanged**
- Facilitator validation is still enforced at verify/settle time

**@x402/hono** (`index.ts`):
- Change `syncFacilitatorOnStart` default from `true` to `false`
- Now safe because core supports lazy 402 generation

## Result

Cold-start 402 responses work correctly on serverless/edge environments
(Cloudflare Workers, AWS Lambda) without blocking or errors.
Discovery probes receive a valid 402 immediately on cold isolates.

## Tests

- Added test: `should build base requirements without facilitator init (cold start)`
- Updated hono retry test to pass `syncFacilitatorOnStart: true` explicitly
- 442 core tests passing + 36 hono tests passing

### DIFF
--- a/typescript/packages/core/src/server/x402ResourceServer.ts
+++ b/typescript/packages/core/src/server/x402ResourceServer.ts
@@ -688,25 +688,24 @@ export class x402ResourceServer {
     }
 
     // Find the matching supported kind from facilitator
-    const supportedKind = this.getSupportedKind(
-      x402Version,
-      resourceConfig.network,
-      SchemeNetworkServer.scheme,
-    );
+    // If facilitator not yet initialized, skip validation so 402 can be emitted on cold start.
+    // Facilitator validation is only required at verify/settle time.
+    const isInitialized = this.supportedResponsesMap.size > 0;
+    const supportedKind = isInitialized
+      ? this.getSupportedKind(x402Version, resourceConfig.network, SchemeNetworkServer.scheme)
+      : undefined;
 
-    if (!supportedKind) {
+    if (isInitialized && !supportedKind) {
       throw new Error(
         `Facilitator does not support ${SchemeNetworkServer.scheme} on ${resourceConfig.network}. ` +
           `Make sure to call initialize() to fetch supported kinds from facilitators.`,
       );
     }
 
-    // Get facilitator extensions for this combination
-    const facilitatorExtensions = this.getFacilitatorExtensions(
-      x402Version,
-      resourceConfig.network,
-      SchemeNetworkServer.scheme,
-    );
+    // Get facilitator extensions for this combination (empty if not yet initialized)
+    const facilitatorExtensions = isInitialized
+      ? this.getFacilitatorExtensions(x402Version, resourceConfig.network, SchemeNetworkServer.scheme)
+      : undefined;
 
     // Parse the price using the scheme's price parser
     const parsedPrice = await SchemeNetworkServer.parsePrice(
@@ -729,11 +728,14 @@ export class x402ResourceServer {
     };
 
     // Delegate to the implementation for scheme-specific enhancements
-    const requirement = await SchemeNetworkServer.enhancePaymentRequirements(
-      baseRequirements,
-      supportedKind,
-      facilitatorExtensions,
-    );
+    // If facilitator not yet initialized, skip enhancement and use base requirements directly
+    const requirement = isInitialized
+      ? await SchemeNetworkServer.enhancePaymentRequirements(
+          baseRequirements,
+          supportedKind!,
+          facilitatorExtensions,
+        )
+      : baseRequirements;
 
     requirements.push(requirement);
     return requirements;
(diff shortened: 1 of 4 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/2171

lau90eth:
Thanks for the heads up. I see #2171 takes a more comprehensive approach covering all middleware packages. Happy to see the issue resolved.

PR #1873, opened 2026-03-30
python: add Lightning reference implementation under bip122/exact

Bortlesboat:
## Description

Adds a Python Lightning reference implementation under `python/x402/mechanisms/bip122/exact/`.

- uses Bitcoin CAIP-2 identifiers with `extra.paymentMethod = "lightning"`
- adds payer and receiver adapter protocols
- generates and verifies BOLT11 invoice-backed requirements
- returns the payment hash as the settlement identifier
- converts satoshi inputs to millisatoshis internally
- prevents duplicate settlement by payment hash
- includes focused client, server, facilitator, registration, conversion, and sync roundtrip coverage

Closes #1857.

## Current-main refresh

Merged current `main` and retained the newer XRPL capability rows alongside BIP-122 in `docs/sdk-features.md`. The current Python package, dependency lock, and CI configuration are composed with the Lightning extra.

## Verification

- 32 focused BIP-122 tests pass
- full Python suite: 1,882 passed and 65 integration tests skipped
- Ruff lint and formatting pass across all 382 Python files
- targeted mypy reports no issues across all 12 BIP-122 source files (`--follow-imports=skip --ignore-missing-imports`)
- `uv lock --check` passes
- refresh commit is signed and GitHub-verified

## Checklist

- [x] Formatted and linted
- [x] New, adjacent, and full Python tests pass
- [x] Commits signed
- [x] Changelog fragment included

PR #2262, opened 2026-05-11
feat(ts): TypeScript Lightning (bip122/exact) mechanism - ExactBip122{Client,Server,Facilitator}Scheme

### LINKED ISSUE #2258: feat: TypeScript Lightning (BOLT11/bip122/exact) mechanism implementation
## Context

With the Python Lightning mechanism (PR #1873) in review and the spec (PR #1311) approaching consensus, TypeScript is now the critical gap: it's 41% of the x402 codebase and the primary SDK language for web/Node.js integrations - but has zero Lightning coverage today.

This issue proposes implementing `ExactBip122Scheme` in TypeScript, following the same structure and contract as the Python reference.

## Current State

| Language | EVM | Lightning |
|---|---|---|
| Python | ? | ?? PR #1873 |
| TypeScript | ? | ? absent |
| Go | ? | ? absent |

## Proposed Directory Structure

`
typescript/packages/mechanisms/src/bip122/exact/
??? index.ts
??? client.ts          # ExactBip122ClientScheme
??? server.ts          # ExactBip122ServerScheme
??? facilitator.ts     # ExactBip122FacilitatorScheme
??? types.ts           # ExactBip122Payload, LightningInvoiceStatus
??? payer.ts           # LightningPayer interface
??? receiver.ts        # LightningReceiver interface
??? settlementCache.ts # Replay prevention (payment_hash TTL)
??? utils.ts           # BOLT11 decode, msat conversions
??? README.md
`

## Interfaces (mirroring Python protocols)

`	ypescript
interface LightningPayer {
  payInvoice(invoice: string, network: string): Promise<LightningInvoiceStatus>;
}

interface LightningReceiver {
  createInvoice(amountMsat: number, memo: string, expirySeconds: number, network: string): Promise<string>;
  lookupInvoice(invoice: string, network: string): Promise<LightningInvoiceStatus | null>;
}

type LightningPaymentStatus = 'unpaid' | 'in_flight' | 'paid';

interface LightningInvoiceStatus {
  invoice: string;
  paymentHash: string;
  amountMsat: number;
  status: LightningPaymentStatus;
  payer?: string;
  settledAt?: number;
}
`

## Design Decisions (following PR #1311 consensus)

- **Scheme identifier:** `exact` + network `bip122:000000000019d6689c085ae165831e93`
- **Amount unit:** millisatoshis (consistent with BOLT11 native denomination)
- **payTo:** always `anonymous` - recipient embedded in invoice
- **Replay prevention:** `payment_hash` as settlement identifier + TTL cache
- **No facilitator required:** server verifies directly via its own Lightning node
- **BOLT12 extensibility:** `extra` field left open per spec discussion

## BOLT11 library

Will use [`@boltz/bolt11`](https://github.com/Boltz-Exchange/bolt11) or [`light-bolt11-decoder`](https://github.com/alexbosworth/bolt11) - both zero-dependency, TypeScript-native. Open to maintainer preference.

## Scope

~800-1000 LOC. Follows existing EVM mechanism pattern. Will include:
- Unit tests (client/server/facilitator isolation)
- Compatibility tests against Python reference (#1873)
- Alby adapter example (extends existing #1590 work)

## Background

I maintain [l402-kit](https://github.com/ShinyDapps/l402-kit), an open-source Lightning pay-per-call middleware in production across npm/PyPI/cargo/Go. The BOLT11 and payment_hash logic maps directly - happy to lead this implementation.

**Ready to open a draft PR once there's consensus on the BOLT11 library choice and whether maintainers want to wait for #1311 to merge first.**

Tagging: @ocknamo (spec author, #1311) @Bortlesboat (Python impl, #1873)

ThiagoDataEngineer:
Closes #2258.

Adds TypeScript Lightning (bip122/exact) mechanism implementation following the spec in PR #1311 and the Python reference in PR #1873.

## What's here

**13 files, 801 LOC** in `typescript/packages/mechanisms/bip122/`:

| File | Role |
|---|---|
| `src/exact/types.ts` | `LightningPayer`, `LightningReceiver`, `ExactBip122Payload`, `LightningInvoiceStatus`, `DecodedBolt11` |
| `src/exact/constants.ts` | CAIP-2 network IDs, error codes, defaults |
| `src/exact/utils.ts` | `decodeBolt11()` - lazy-requires `light-bolt11-decoder` |
| `src/exact/settlementCache.ts` | In-memory TTL cache (replaces with Redis/Postgres via interface) |
| `src/exact/client/scheme.ts` | `ExactBip122ClientScheme` - extracts invoice, calls `LightningPayer.payInvoice` |
| `src/exact/server/scheme.ts` | `ExactBip122ServerScheme` - creates invoice via `LightningReceiver`, injects into `extra` |
| `src/exact/facilitator/scheme.ts` | `ExactBip122FacilitatorScheme` - 10-step verify + settle |
| `test/unit/facilitator.test.ts` | 10 unit tests: happy path + all rejection branches |
| `test/unit/settlementCache.test.ts` | 4 unit tests: TTL, isolation, expiry |

## Verification (10 steps, per #1311)

`
1. network in bip122:* (mainnet + testnet CAIP-2)
2. asset === BTC
3. payTo === anonymous
4. extra.paymentMethod === lightning
5. extra.invoice present in requirements
6. invoice_substitution: payload.invoice === requirements.extra.invoice  ? before BOLT11 decode
7. BOLT11 decode ? payment_hash, amount_msat, expiresAt
8. expiry: expiresAt (secs) < now - explicit because backends don't surface "expired"
9. amount: decoded msat === requirements.amount (string-safe compare)
10. replay cache: payment_hash not already settled
11. node check: LightningReceiver.lookupInvoice ? paid / in_flight / unpaid
`

Step 6 (invoice substitution) runs **before** BOLT11 decode per the test case in @powforge/mcp-pay-gate - prevents a stale paid invoice from a prior request cross-binding to new requirements.

## Design choices

**BOLT11 library:** `light-bolt11-decoder` (optional peer dep, lazy-required via `require()`). Zero-deps, ~12kB, decode-only - the verifier never signs. Injecting `decodeBolt11Fn` as a constructor option keeps tests fast and deterministic.

**expiresAt:** `timestamp + expiry` from BOLT11 header, compared to `Date.now() / 1000`. Documented in `LightningInvoiceStatus` because LNBits (and Blink) don't return a distinct "expired" status - callers must check independently. The `nowFn` option is injectable for testing.

**TTL formula:** `(expiresAt * 1000 - now) + 60_000 ms` - mirrors the `expiry + 60s` used in production by @powforge/mcp-pay-gate.

**No facilitator required:** Lightning settles off-chain; the server verifies directly via its own node/wallet adapter. `ExactBip122FacilitatorScheme` can run co-located with the server.

**`LightningReceiver` interface:** the `lookupInvoice` slot maps to LNBits' `checkPaidFn` pattern. For Blink, `makeBlinkCheckPaidFn` follows the same shape - the facilitator core is adapter-agnostic.

## Prior art

@powforge/mcp-pay-gate (MIT) - verifier shape and invoice-substitution test informed the `lookupInvoice` interface and the step-6 guard order. Credited in commit message.

## What's NOT here yet

- Go implementation (separate PR)
- Alby facilitator adapter (follows #1590)
- Integration tests against live LNBits/Blink (needs test infra decision)
- `@x402/bip122` published to npm (pending maintainer review)

## Testing

`ash
cd typescript/packages/mechanisms/bip122
npm install
npm test
`

All 14 unit tests pass. `decodeBolt11Fn` and `nowFn` are injected in tests - no live Lightning node required.

cc @ocknamo (spec, #1311) @Bortlesboat (Python reference, #1873) @zekebuilds-lab (prior art, invoice-substitution test)

### DIFF
--- a/typescript/packages/mechanisms/bip122/package.json
+++ b/typescript/packages/mechanisms/bip122/package.json
@@ -1,0 +1,50 @@
+{
+  "name": "@x402/bip122",
+  "version": "0.1.0",
+  "description": "x402 Payment Protocol Bitcoin Lightning (bip122/exact) Implementation",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/cjs/index.d.ts",
+  "license": "Apache-2.0",
+  "author": "x402 Foundation",
+  "repository": "https://github.com/x402-foundation/x402",
+  "keywords": ["x402", "payment", "protocol", "bitcoin", "lightning", "bip122", "bolt11"],
+  "scripts": {
+    "build": "tsup",
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "format": "prettier -c .prettierrc --write \"**/*.{ts,js,json,md}\"",
+    "lint": "eslint . --ext .ts --fix"
+  },
+  "exports": {
+    ".": {
+      "import": { "types": "./dist/esm/index.d.mts", "default": "./dist/esm/index.mjs" },
+      "require": { "types": "./dist/cjs/index.d.ts", "default": "./dist/cjs/index.js" }
+    },
+    "./exact/client": {
+      "import": { "types": "./dist/esm/exact/client/scheme.d.mts", "default": "./dist/esm/exact/client/scheme.mjs" },
+      "require": { "types": "./dist/cjs/exact/client/scheme.d.ts", "default": "./dist/cjs/exact/client/scheme.js" }
+    },
+    "./exact/server": {
+      "import": { "types": "./dist/esm/exact/server/scheme.d.mts", "default": "./dist/esm/exact/server/scheme.mjs" },
+      "require": { "types": "./dist/cjs/exact/server/scheme.d.ts", "default": "./dist/cjs/exact/server/scheme.js" }
+    },
+    "./exact/facilitator": {
+      "import": { "types": "./dist/esm/exact/facilitator/scheme.d.mts", "default": "./dist/esm/exact/facilitator/scheme.mjs" },
+      "require": { "types": "./dist/cjs/exact/facilitator/scheme.d.ts", "default": "./dist/cjs/exact/facilitator/scheme.js" }
+    }
+  },
+  "files": ["dist", "src"],
+  "dependencies": {
+    "@x402/core": "workspace:~"
+  },
+  "optionalDependencies": {
+    "light-bolt11-decoder": "^3.0.0"
+  },
+  "devDependencies": {
+    "@types/node": "^22.13.4",
+    "tsup": "^8.4.0",
+    "typescript": "^5.7.3",
+    "vitest": "^3.0.5"
+  }
+}
--- a/typescript/packages/mechanisms/bip122/src/exact/client/scheme.ts
+++ b/typescript/packages/mechanisms/bip122/src/exact/client/scheme.ts
@@ -1,0 +1,49 @@
+import type { PaymentRequirements } from "@x402/core/types";
+import type { ExactBip122Payload, LightningPayer } from "../types";
+import { ERR_MISSING_INVOICE, ERR_INVALID_PAYMENT_METHOD, PAYMENT_METHOD_LIGHTNING } from "../constants";
+
+/**
+ * x402 client-side scheme for Bitcoin Lightning (bip122/exact).
+ *
+ * Responsibilities:
+ * - Extract the BOLT11 invoice from PaymentRequirements.extra
+ * - Pay it via the provided LightningPayer
+ * - Return the payload (invoice string) for the Authorization header
+ *
+ * @example
+ * ```ts
+ * import { ExactBip122ClientScheme } from "@x402/bip122/exact/client";
+ *
+ * const client = new ExactBip122ClientScheme({
+ *   payInvoice: async (invoice, network) => {
+ *     await myWallet.pay(invoice);
+ *   }
+ * });
+ *
+ * const payload = await client.createPaymentPayload(2, requirements);
+ * ```
+ */
+export class ExactBip122ClientScheme {
+  readonly scheme = "exact";
+
+  constructor(private readonly payer: LightningPayer) {}
+
+  async createPaymentPayload(
+    _x402Version: number,
+    requirements: PaymentRequirements,
+  ): Promise<{ payload: ExactBip122Payload }> {
+    const paymentMethod = requirements.extra?.paymentMethod;
+    if (paymentMethod && paymentMethod !== PAYMENT_METHOD_LIGHTNING) {
+      throw new Error(`${ERR_INVALID_PAYMENT_METHOD}: expected lightning, got ${paymentMethod}`);
+    }
+
+    const invoice = requirements.extra?.invoice;
+    if (!invoice || typeof invoice !== "string") {
+      throw new Error(`${ERR_MISSING_INVOICE}: PaymentRequirements.extra.invoice is required`);
+    }
+
+    await this.payer.payInvoice(invoice, requirements.network);
+
+    return { payload: { invoice } };
+  }
+}
--- a/typescript/packages/mechanisms/bip122/src/exact/constants.ts
+++ b/typescript/packages/mechanisms/bip122/src/exact/constants.ts
@@ -1,0 +1,25 @@
+export const SCHEME = "exact" as const;
+export const ASSET = "BTC" as const;
+export const PAY_TO_ANONYMOUS = "anonymous" as const;
+export const PAYMENT_METHOD_LIGHTNING = "lightning" as const;
+export const DEFAULT_INVOICE_DESCRIPTION = "x402 payment" as const;
+export const DEFAULT_INVOICE_EXPIRY_SECONDS = 3600;
+export const SETTLEMENT_TTL_BUFFER_MS = 60_000;
+
+export const BTC_MAINNET_CAIP2 = "bip122:000000000019d6689c085ae165831e93" as const;
+export const BTC_TESTNET_CAIP2 = "bip122:000000000933ea01ad0ee984209779ba" as const;
+
+export const SUPPORTED_NETWORKS = [BTC_MAINNET_CAIP2, BTC_TESTNET_CAIP2] as const;
+
+export const ERR_UNSUPPORTED_NETWORK = "unsupported_network";
+export const ERR_INVALID_ASSET = "invalid_asset";
+export const ERR_INVALID_PAY_TO = "invalid_pay_to";
+export const ERR_INVALID_PAYMENT_METHOD = "invalid_payment_method";
+export const ERR_MISSING_INVOICE = "missing_invoice";
+export const ERR_INVALID_INVOICE = "invalid_invoice";
+export const ERR_INVOICE_SUBSTITUTION = "invoice_substitution";
+export const ERR_INVOICE_EXPIRED = "invoice_expired";
+export const ERR_AMOUNT_MISMATCH = "amount_mismatch";
+export const ERR_INVOICE_IN_FLIGHT = "invoice_in_flight";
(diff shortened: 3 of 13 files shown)


zekebuilds-lab:
@ThiagoDataEngineer nice. The step-6-before-decode order is the load-bearing bit and you got the spec right. Glad to be cited; the credit reads cleanly.

One distinction worth pulling apart in the test matrix while the PR is still draft: the **invoice-substitution** guard (three-way string compare across `payload.invoice` / `accepted.extra.invoice` / `requirements.extra.invoice`) and a **minted-invoice** guard ("did THIS server actually issue this `payment_hash`") are two different checks. In our verifier they're step 5 and step 6 respectively. Step 5 catches a client mixing-and-matching strings inside one request. Step 6 catches a client attaching a paid invoice from a *different* server's L402 / Lightning surface entirely. Source: [src/lib/x402-verify.js#L172-L207](https://github.com/zekebuilds-lab/mcp-pay-gate/blob/main/src/lib/x402-verify.js#L172-L207).

The minted-invoice check is optional in our impl (gated on `mintedInvoices.has`), since ops without a minted store can reasonably fall back to "any valid + paid invoice on our LNBits node is ours". But for a facilitator that fronts multiple receivers, or for operators wallet-sharing across services, the cross-server replay is a real surface and the spec test matrix probably wants an explicit case for it. Happy to write the test in TS mirroring our [`x402-verify.test.js`](https://github.com/zekebuilds-lab/mcp-pay-gate/blob/main/tests/x402-verify.test.js) shape if useful: `cross_server_invoice` rejection alongside the `invoice_substitution` one you've already got.

Also: the `(timestamp + expiry) * 1000` comparison in your step 8 matches our impl byte-for-byte. Good landing.

zekebuilds-lab:
@ThiagoDataEngineer — pulled the standalone verifier out into its own package in case it's useful here:

[`@powforge/l402-verify@0.1.0`](https://www.npmjs.com/package/@powforge/l402-verify)

Zero runtime dependencies. Exports `verifyL402(authHeader, cfg)`, `parseL402Header`, `verifyMacaroon`, and `checkInvoicePaid`. The `checkPaidFn` is injectable so it isn't bound to LNBits — drop in LND `lookupinvoice`, Phoenix, Greenlight, hosted, whatever. Same 8-step verification order as the `verifyX402Lightning` you've got here, including the step-6-before-decode ordering.

35 unit tests. MIT.

No expectation either way — just figured I'd flag it before this PR lands in case it's easier to depend on a tagged surface than vendor the logic.

zekebuilds-lab:
@ThiagoDataEngineer — pulled the standalone verifier out into its own package in case it's useful here:

[`@powforge/l402-verify@0.1.0`](https://www.npmjs.com/package/@powforge/l402-verify)

Zero runtime dependencies. Exports `verifyL402(authHeader, cfg)`, `parseL402Header`, `verifyMacaroon`, and `checkInvoicePaid`. The `checkPaidFn` is injectable so it isn't bound to LNBits — drop in LND `lookupinvoice`, Phoenix, Greenlight, hosted, whatever. Same 8-step verification order as the `verifyX402Lightning` you've got here, including the step-6-before-decode ordering.

35 unit tests. MIT.

No expectation either way — just figured I'd flag it before this PR lands in case it's easier to depend on a tagged surface than vendor the logic.

zekebuilds-lab:
One thing worth surfacing while the `LightningFacilitator` interface is still in flight: a third backend that fits the same two-method contract — NWC (Nostr Wallet Connect, [NIP-47](https://github.com/nostr-protocol/nips/blob/master/47.md)).

LNBits and Blink are great defaults for "works out of the box," but both are custodial. NWC fills the self-custodial slot without enlarging the facilitator interface — a wallet RPC over a Nostr relay, with `make_invoice` / `lookup_invoice` matching the same shape the LNBits adapter already uses:

```ts
const nwc = new NWCClient({ nostrWalletConnectUrl: process.env.NWC_URL! });

async function mintInvoice(sats: number, memo: string) {
  const inv = await nwc.makeInvoice({ amount: sats * 1000, description: memo });
  return { paymentRequest: inv.invoice, paymentHash: inv.payment_hash };
}

async function isPaid(paymentHash: string) {
  const r = await nwc.lookupInvoice({ payment_hash: paymentHash });
  return r.settled_at != null;
}
```

Why it's worth a slot:

1. Operators who already run LND/CLN can front their node with Alby Hub (or LNBits-NWC) and get an x402 facilitator without a second custodial account.
2. Mobile/browser buyers using Mutiny, Zeus, or Alby can autopay 402 invoices end-to-end over NIP-47 — same wire protocol on both sides of the facilitator.
3. No vendor SDK to maintain on the facilitator: the only secret is a `nostr+walletconnect://` connection string.

Buyer-side reference: [`@powforge/x402-lightning`](https://www.npmjs.com/package/@powforge/x402-lightning) ships an NWC adapter alongside LNBits, and we use NWC as the priority-1 autopay rail in `@powforge/402-mcp`. Happy to put up a follow-on PR adding `ExactBip122NWCFacilitator` once this one lands — it's ~80 lines around the same interface you're defining here, plus a small Nostr-event subscription helper for settlement notifications (or polling `lookupInvoice` if you'd rather not pull a relay client into the facilitator).

Not a blocker on #2262 — just flagging while the shape of `LightningFacilitator` is still soft, so the interface doesn't accidentally bake in HTTPS-only assumptions that would make a relay-transport backend awkward later.

ThiagoDataEngineer:
@zekebuilds-lab - appreciate the depth across all four threads, batching the reply here.

**On step 5 vs step 6 (substitution vs minted-invoice).** Good distinction, and you''re right that our naming flattens them. What we label "Step 6 - invoice substitution guard" in `facilitator/scheme.ts` is your step 5 - three-way string compare. The cross-server replay (your step 6) is a real surface for multi-receiver facilitators, and the spec test matrix should cover it.

Yes, please send the `cross_server_invoice` test in TS. Wiring it in will likely require surfacing some kind of optional minted-invoice hook on the receiver side; I''d rather see the test concrete before locking in the API shape, since the hook signature has implications for both the reference adapters and downstream impls. Drop the test in and we can settle the interface together in the review.

**On `@powforge/l402-verify@0.1.0`.** Thanks for shipping the standalone, and the injectable `checkPaidFn` is the right cut. For the mechanism package in this PR I''m keeping verification inline - x402 mechanism packages are zero-dep by convention, and the JSDoc already cites `x402-verify.js#L172-L207` for the step-6-before-decode ordering. On the buyer half (downstream of this PR), I''ll take a closer look at the package surface and decide on adoption when I cut the next minor of `l402-kit`. No timeline on that - I want to see how the spec tests settle here first.

**On NWC (NIP-47) as a third backend.** Not a blocker on #2262, agreed. The operator angle (Alby Hub fronting an existing LND/CLN, no second custodial account) is genuinely the strongest argument for it - that''s the gap LNBits + Blink leave open. Useful that you''ve validated it end-to-end in `@powforge/x402-lightning` and `@powforge/402-mcp`.

On the interface: `LightningReceiver` (`types.ts` L40-55) is `createInvoice(amountMsat, memo, expirySeconds, network) ? Promise<string>` and `lookupInvoice(invoice, network) ? Promise<LightningInvoiceStatus | null>` - no URL fields, no fetch hooks, no transport-shaped assumptions in parameters or return types. If you see something that *would* be awkward over a relay, please flag it here before #2262 exits draft so we adjust in this PR. Otherwise I''d rather wait until I can read the follow-on diff before committing to a specific shape; happy to review on merits when you put it up.

On settlement-notification design (polling `lookupInvoice` vs. Nostr-event subscription helper) - I don''t want to anchor a preference before seeing how the adapter actually shapes up; whichever lands as the minimal-surface first cut is fine as long as it doesn''t force a relay-client dependency into the base path. Open to whatever you think is cleanest, will weigh on the PR.

`(timestamp + expiry) * 1000` alignment in step 8 was deliberate - your spec page was the cleanest source on the unit-mismatch foot-gun, and tests broke until we got it right. Glad it lands.

PR #2271, opened 2026-05-12
Revert "Add Writing hub and batch settlement blog post (#2269)"

Must-be-Ash:
Reverts #2269 which broke the app.

This is a clean revert of the merge commit `973de09b`. Restores the site to its state before the Writing hub and batch settlement blog post were added.


Must-be-Ash:
Closing — the issue was fixed directly, revert no longer needed.

PR #2290, opened 2026-05-13
feat(bazaar): add facilitator discovery service

forest-builds:
## Summary

Implements the bazaar discovery backend that the spec defines and that `withBazaar` already calls. Today `GET /discovery/resources` and `GET /discovery/search` only exist in the e2e test harness; this PR adds a real, swappable catalog and wires it into the x402.org facilitator.

- **`BazaarCatalog` interface** with two implementations — `InMemoryBazaarCatalog` (tests / local dev) and `PostgresBazaarCatalog` (Drizzle + tsvector full-text search, keyset cursor pagination, `NULLS NOT DISTINCT` catalog key on PG 15+).
- **`installBazaarFacilitator(facilitator, catalog)`** registers the bazaar marker and a best-effort `onAfterVerify` hook. Hook errors route through a configurable `onError` and never propagate — cataloging must not break verify.
- **Two new routes** on the site: `GET /facilitator/discovery/resources` and `GET /facilitator/discovery/search`. They parse the spec's query envelope, call the catalog, and mirror the verify/settle error shape.
- **Postgres is opt-in**: lives behind subpath export `@x402/extensions/bazaar/postgres`; `pg` is an *optional* peer dep so Neon-HTTP / postgres-js consumers can bring their own driver via `new PostgresBazaarCatalog(drizzleDb)`.
- **Catalog selection**: site's `getBazaarCatalog()` uses Postgres when `BAZAAR_DATABASE_URL` is set, otherwise the in-memory catalog with a `console.warn` (mirrors `getFacilitator()` lazy-init pattern).

## Design decisions worth flagging

- **Trigger-based `tsvector`, not `GENERATED ALWAYS AS … STORED`.** `to_tsvector` is not provably immutable under some `default_text_search_config` settings, which trips PG's generated-column immutability check. The BEFORE INSERT/UPDATE trigger is the canonical FTS pattern for that reason.
- **`NULLS NOT DISTINCT` unique on `(resource_url, method, tool_name)`.** HTTP rows have NULL `tool_name`, MCP rows have NULL `method`. PG 15's `NULLS NOT DISTINCT` makes the right key collide as expected.
- **Keyset cursor with explicit predicate.** Sorted `rank DESC, id ASC` — a tuple compare would assume uniform direction. Predicate is `rank < cr OR (rank = cr AND id > ci)`.
- **Accepts merging happens in the catalog.** Facilitators only see the *one* selected `PaymentRequirements` per verify, so the catalog merges across observations deduped by `(scheme, network)`. Same metadata as the latest payment is propagated so renamed services / updated descriptions converge.

## Out of scope / follow-ups

- **No Neon HTTP factory.** Consumers can still pass their own Drizzle db to `PostgresBazaarCatalog`. A `createNeonBazaarCatalog` helper is an easy next PR if there's interest.
- **No `partialResults` heuristic** in `search`. Advisory field; spec allows omitting.
- **No site route-level tests.** The catalog layer is well covered; the site lacks a route-test harness today. Adding vitest + supertest for the site is a clean follow-up.
- **No stale-entry pruning.** `last_seen_at` is recorded but not yet acted on — leaves room for a periodic job to soft-delete endpoints that haven't been seen in N days.

## Test plan

- [x] `pnpm -F @x402/extensions lint:check && pnpm -F @x402/extensions format:check`
- [x] `pnpm -F @x402/extensions test` — **436 passed, 11 skipped** (db tests gated behind `RUN_DB_TESTS=1`)
- [x] `RUN_DB_TESTS=1 pnpm -F @x402/extensions test` — full **447 passed** locally with Testcontainers (Docker + `postgres:16-alpine`)
- [x] `pnpm -F @x402/extensions build` — succeeds; migration SQL lands in both `dist/esm/bazaar/postgres/migrations/` and `dist/cjs/bazaar/postgres/migrations/` via tsup postbuild copy
- [x] `pnpm -F site lint:check && pnpm -F site format:check && pnpm -F site build` — clean; both new routes show up in the Next.js route table
- [ ] **E2E oracle** (suggested for CI): the repo's `e2e/extensions/bazaar.ts` already exercises `/discovery/resources?limit=1000` and `/discovery/search?query=X` against test facilitators. Point it at a locally-running site (with `BAZAAR_DATABASE_URL` set to a throwaway PG) for an end-to-end smoke test.

## What's in the diff

| Path | What it does |
|---|---|
| `extensions/src/bazaar/facilitator-service/catalog.ts` | `BazaarCatalog` interface + `CatalogUpsertInput` |
| `extensions/src/bazaar/facilitator-service/installFacilitator.ts` | `installBazaarFacilitator()` factory |
| `extensions/src/bazaar/facilitator-service/memoryCatalog.ts` | `InMemoryBazaarCatalog` |
| `extensions/src/bazaar/facilitator-service/postgres/{schema,postgresCatalog,migrate,factory}.ts` | Drizzle schema + Postgres catalog + migrations runner + `pg`-backed factory |
| `extensions/src/bazaar/facilitator-service/postgres/migrations/0001_init.sql` | Table, trigger, indexes |
| `extensions/test/bazaar-facilitator-service.test.ts` | 17 in-memory + hook tests |
| `extensions/test/bazaar-postgres.test.ts` | 11 Testcontainers tests |
| `site/app/facilitator/discovery/{catalog.ts,resources/route.ts,search/route.ts}` | Catalog singleton + two route handlers |

Draft for now while I do one more pass on the GitHub diff UI; I'll mark ready once I'm happy.


shawnhvac:
Adding some data points from AgentPay (https://x402-agent-pay.com) — we built a `/x402/info` endpoint that exposes structured facilitator metadata (name, chains, schemes, assets, fee model, contact) for exactly this kind of discovery use case.

Example response: https://x402-agent-pay.com/x402/info

We also expose `/x402/stats` for live volume data so client agents can evaluate facilitator health before routing payments through it. Currently at zero fees, 5 chains (Base, Ethereum, Optimism, Arbitrum, Polygon).

If a discovery standard gets formalized here, happy to align our info endpoint schema to it. The key fields we find most useful for agent-side discovery are: `supported_networks[]`, `fee_bps`, `supports.verify`, `supports.settle`, and an uptime signal. A `/x402/supported` endpoint that returns machine-readable chain/scheme/asset capabilities would be a natural complement to the existing verify/settle pair.

phdargen:
Thanks for the contribution @forest-builds!

The x402.org facilitator is for testnet only and should only be used for local testing. 
We do not intend to run a bazaar infrastructure for it and we already have an example demonstrating a simple bazaar integration here: https://github.com/x402-foundation/x402/blob/main/examples/typescript/facilitator/advanced/bazaar.ts

Issue #2285, opened 2026-05-12
Proposal: Add "Compliance" category to Bazaar discovery + enrich tooloracle-io service


feedoracle:
## Summary

I'd like to propose adding a **"Compliance" category** to the Bazaar discovery taxonomy, and as the first member of that category, requesting enrichment of the existing `tooloracle-io` service (https://api.agentic.market/v1/services/search?q=tooloracle).

## Context

Current categories observed across the Agentic.Market index (via `api.agentic.market/v1/services/search`):

| Category | Service count |
|---|---|
| Data | 91 |
| Inference | 19 |
| Infra | 13 |
| Search | 12 |
| Travel | 8 |
| Media | 5 |
| Trading | 2 |
| Social | 2 |
| Other | 1 |

There is no Compliance, Regtech, Audit, or Regulatory category yet — but the catalog now includes regulated-finance use cases that don't fit cleanly into "Data" or "Other".

## What we built

`tooloracle.io` runs 11 paid x402 endpoints on Base mainnet, all focused on EU regulatory compliance:

| Endpoint | What it does |
|---|---|
| `/v2/mica_stablecoin_check` | MiCA Title III/IV compliance audit for any stablecoin, returns grade A-F + ESMA authorization status |
| `/v2/compliance_preflight` | MiCA pre-flight verdict (PASS/WARN/BLOCK) for issuer/CASP onboarding |
| `/v2/sanctions_screen` | Real-time AML screening against EU FSF + OFAC SDN + UN + UK HMT (87k+ entities) |
| `/v2/cve_lookup` | DORA ICT risk: NIST NVD + CISA KEV + MITRE ATT&CK |
| `/v2/uvo_quick` / `/uvo_full` | AI agent action validation + hallucination defense (9 layers) |
| `/v2/uvo_citation_check` | Phantom-citation detection for EU legal texts (MiCA, DSGVO, DORA, AMLR, NIS2) |
| `/v2/uvo_academic_verify` | Academic claim verification with ECDSA-signed receipts |
| `/v2/agent_preflight` | Autonomous agent decision pre-flight with policy + risk gate |
| `/v2/macro_snapshot` | Real-time US macro (Fed, CPI, yields) for trading agents |
| `/v2/meme_rug_check` | Multi-chain rug-pull risk for treasury due diligence |

[...]

PR #1734, opened 2026-03-21
feat: add Buzz BD Agent to ecosystem (Services/Endpoints)

buzzbysolcex:
## Summary

Adds **Buzz BD Agent** by SolCex Exchange to the x402 ecosystem as a **Services/Endpoints** partner.

### What is Buzz BD?

Buzz BD is an autonomous exchange listing agent that uses x402 micropayments (USDC on Base) to monetize real-time token intelligence:

| Endpoint | Price | Description |
|----------|-------|-------------|
| `GET /api/v1/premium/pipeline` | $0.01 | Hot token pipeline — top scored tokens across Solana/Base/BSC with 100-point composite scoring |
| `GET /api/v1/premium/score/:address` | $0.005 | Per-token intelligence score across 5 dimensions (safety, wallet, social, scorer, technical) |
| `GET /api/v1/premium/sim/:address` | $0.02 | 20-agent MiroFish listing simulation with adversarial bull/bear debate |

### x402 Integration

- All endpoints return proper **HTTP 402** with `X-Payment-Required` headers
- Payment: **USDC on Base** via exact scheme
- Pay to: `0x2Dc03124091104E7798C0273D96FC5ED65F05aA9`
- Discovery: `https://api.buzzbd.ai/.well-known/x402.json`
- Live and verified on mainnet

### Files Added

- `typescript/site/app/ecosystem/partners-data/buzz-bd/metadata.json`
- `typescript/site/public/logos/buzz-bd-logo.png` (placeholder — happy to replace with final asset)

### Links

- Website: https://buzzbd.ai
- API: https://api.buzzbd.ai
- Discovery: https://api.buzzbd.ai/.well-known/x402.json
- Twitter: https://x.com/BuzzBySolCex
- Parent: https://solcex.cc

PR #2300, opened 2026-05-14
feat(extensions): add trust-provider extension for behavioral trust gating

### LINKED ISSUE #2299: [Extension] Trust-Provider: behavioral trust scoring via onBeforeSettle hook
## Problem

x402 settles payments between agents, but has no mechanism to evaluate whether the paying agent should be trusted before settlement occurs. A resource server currently has no way to reject settlement from agents with poor behavioral history, known abuse patterns, or insufficient trust data.

As the agent economy scales, settlement without trust evaluation creates risk:
- Agents with adversarial behavior patterns can make payments that complete before abuse is detected
- Resource servers have no way to apply graduated trust policies (strict, quorum, custom)
- There's no standard way to attach trust evidence to the settlement flow

## Proposal

A **trust-provider extension** that uses the existing `onBeforeSettle` hook to gate settlement on behavioral trust scores from external trust providers.

### How it works

```typescript
import { trustProviderExtension } from '@x402/extensions/trust-provider';

const server = new x402ResourceServer({
  extensions: {
    'trust-provider': trustProviderExtension({
      providers: [{
        name: 'dominion-observatory',
        url: 'https://dominion-observatory.sgdata.workers.dev',
        evaluate: async (query) => {
          const res = await fetch(url + '/api/agent-query/' + query.payer.agent_id);
          const data = await res.json();
          return {
            decision: data.server.trust_score >= 60 ? 'PASS' : data.server.trust_score < 40 ? 'FAIL' : 'UNCERTAIN',
            score: data.server.trust_score / 100,
          };
        }
      }],
      policy: { kind: 'strict' },
      failureMode: 'fail-closed',
    })
  }
});
```

### Extension behavior

1. **onBeforeSettle hook**: Queries configured trust providers in parallel with per-provider timeouts
2. **Aggregation**: Applies STRICT (all must PASS), QUORUM (majority), or custom aggregation
3. **Gate decision**:
   - PASS: returns void (settlement proceeds)
   - FAIL: returns { abort: true, reason: 'trust_evaluation_failed' }
   - UNCERTAIN + fail-closed: aborts; fail-open: proceeds
4. **Advisory headers**: Populates X-Trust-Decision, X-Trust-Score via enrichSettlementResponse

### Wire types

Two wire types conformant with [x402-trust-provider-interface v0.1](https://github.com/vdineshk/daee-engine/blob/main/specs/x402-trust-provider-interface/v0.1/SPEC.md):

- **TrustQuery** -- identifies the payer (agent_id/wallet), resource, amount, risk context
- **TrustEvaluation** -- provider response with decision (PASS/FAIL/UNCERTAIN), score (0-1), evidence URI, TTL

### Reference provider

[Dominion Observatory](https://dominion-observatory.sgdata.workers.dev) provides behavioral trust scores for 4,500+ MCP servers. Live today and can serve as the reference trust provider.

## Implementation plan

Following the extension pattern from payment-identifier:

```
trust-provider/
  index.ts          -- Public exports
  types.ts          -- TrustQuery, TrustEvaluation, TrustDecision
  schema.ts         -- Zod schema for extension declaration
  resourceServer.ts -- onBeforeSettle hook + enrichSettlementResponse
  client.ts         -- Client-side trust metadata attachment
  utils.ts          -- Aggregation functions (strict, quorum)
```

Uses only the existing ResourceServerExtensionHooks.onBeforeSettle interface. No core changes needed. I have a working implementation ready and can open a PR.

## Prior art

- [x402-trust-provider-interface spec v0.1](https://github.com/vdineshk/daee-engine/blob/main/specs/x402-trust-provider-interface/v0.1/SPEC.md)
- [Testnet demo](https://github.com/vdineshk/daee-engine/tree/main/testnet-demo) (Base Sepolia USDC)
- [@dominion/trust-provider npm package](https://github.com/vdineshk/daee-engine/tree/main/packages/trust-provider)

AI disclosure: This issue was drafted with AI assistance and reviewed before submission.

vdineshk:
## Summary

Adds a **trust-provider** extension under `typescript/packages/extensions/src/trust-provider/` that gates payment settlement on behavioral trust evaluation via the existing `onBeforeSettle` hook.

Closes #2299

## What it does

- Queries configured trust providers (e.g., Dominion Observatory) in parallel before settlement
- Aggregates decisions using STRICT, QUORUM, or custom policies
- Aborts settlement on FAIL or UNCERTAIN (under fail-closed)
- Attaches trust metadata to settlement responses via `enrichSettlementResponse`

## Files

| File | Purpose |
|------|---------|
| `types.ts` | Wire types: TrustQuery, TrustEvaluation, TrustDecision, config interfaces |
| `utils.ts` | Aggregation functions: strict (all PASS), quorum (majority), custom |
| `resourceServer.ts` | `onBeforeSettle` hook implementation + `enrichSettlementResponse` |
| `index.ts` | Public API barrel exports |

## No core changes

Uses only the existing `ResourceServerExtensionHooks.onBeforeSettle` interface. Zero modifications to core packages.

## Reference provider

[Dominion Observatory](https://dominion-observatory.sgdata.workers.dev) — behavioral trust scores for 14,800+ MCP servers. Working testnet demo with Base Sepolia USDC available at [daee-engine/testnet-demo](https://github.com/vdineshk/daee-engine/tree/main/testnet-demo).

### DIFF
--- a/typescript/packages/extensions/src/trust-provider/index.ts
+++ b/typescript/packages/extensions/src/trust-provider/index.ts
@@ -1,0 +1,69 @@
+/**
+ * Trust-Provider Extension for x402
+ *
+ * Gates payment settlement on behavioral trust evaluation from external providers.
+ * Uses the onBeforeSettle hook to query trust providers, aggregate decisions,
+ * and block settlement for untrusted agents.
+ *
+ * ## Usage
+ *
+ * ```typescript
+ * import {
+ *   createTrustProviderExtension,
+ *   declareTrustProviderExtension,
+ *   TRUST_PROVIDER,
+ * } from '@x402/extensions/trust-provider';
+ *
+ * const config = {
+ *   providers: [{
+ *     name: 'dominion-observatory',
+ *     evaluate: async (query) => {
+ *       const res = await fetch(
+ *         'https://dominion-observatory.sgdata.workers.dev/api/agent-query/' + query.payer.agent_id
+ *       );
+ *       const data = await res.json();
+ *       return {
+ *         schema: 'x402-trust-evaluation-v0.1',
+ *         provider: 'dominion-observatory',
+ *         provider_url: 'https://dominion-observatory.sgdata.workers.dev',
+ *         decision: data.server.trust_score >= 60 ? 'PASS' : data.server.trust_score < 40 ? 'FAIL' : 'UNCERTAIN',
+ *         score: data.server.trust_score / 100,
+ *         evaluated_at: new Date().toISOString(),
+ *       };
+ *     },
+ *   }],
+ *   policy: { kind: 'strict' },
+ *   failureMode: 'fail-closed',
+ * };
+ *
+ * // Register with resource server
+ * const extension = createTrustProviderExtension(config);
+ * server.registerExtension(extension);
+ *
+ * // Include in PaymentRequired response
+ * const paymentRequired = {
+ *   extensions: {
+ *     [TRUST_PROVIDER]: declareTrustProviderExtension(config),
+ *   },
+ * };
+ * ```
+ */
+
+export type {
+  TrustQuery,
+  TrustEvaluation,
+  TrustDecision,
+  TrustProviderConfig,
+  TrustProviderExtensionConfig,
+  TrustProviderDeclaration,
+  AggregationPolicy,
+} from "./types";
+
+export { TRUST_PROVIDER } from "./types";
+
+export { aggregateStrict, aggregateQuorum, aggregate } from "./utils";
+
+export {
+  createTrustProviderExtension,
+  declareTrustProviderExtension,
+} from "./resourceServer";
--- a/typescript/packages/extensions/src/trust-provider/resourceServer.ts
+++ b/typescript/packages/extensions/src/trust-provider/resourceServer.ts
@@ -1,0 +1,140 @@
+/**
+ * Trust-Provider Extension — Resource Server hooks
+ *
+ * Implements onBeforeSettle to gate payment settlement on behavioral trust scores.
+ * Queries configured trust providers in parallel, aggregates decisions, and
+ * aborts settlement if trust evaluation fails.
+ */
+
+import type { ResourceServerExtension } from "@x402/core/types";
+import type {
+  TrustQuery,
+  TrustEvaluation,
+  TrustDecision,
+  TrustProviderExtensionConfig,
+  TrustProviderDeclaration,
+} from "./types";
+import { TRUST_PROVIDER } from "./types";
+import { aggregate } from "./utils";
+
+/**
+ * Creates a trust-provider declaration for PaymentRequired.extensions.
+ *
+ * Resource servers call this to advertise that trust evaluation is performed
+ * before settlement. Clients can inspect this to understand trust requirements.
+ */
+export function declareTrustProviderExtension(
+  config: TrustProviderExtensionConfig,
+): TrustProviderDeclaration {
+  return {
+    info: {
+      providers: config.providers.map((p) => p.name),
+      policy: config.policy.kind,
+      failureMode: config.failureMode,
+    },
+  };
+}
+
+/**
+ * Creates a ResourceServerExtension that gates settlement on trust evaluation.
+ *
+ * The onBeforeSettle hook:
+ * 1. Constructs a TrustQuery from the settlement context
+ * 2. Queries all configured providers in parallel (with timeout)
+ * 3. Aggregates decisions per the configured policy
+ * 4. Returns void (proceed) on PASS, abort on FAIL/UNCERTAIN
+ *
+ * The enrichSettlementResponse hook adds advisory trust headers to the response.
+ */
+export function createTrustProviderExtension(
+  config: TrustProviderExtensionConfig,
+): ResourceServerExtension {
+  const timeout = config.perProviderTimeoutMs ?? 5000;
+
+  // Store last evaluation results for enrichSettlementResponse
+  let lastEvaluations: TrustEvaluation[] = [];
+  let lastDecision: TrustDecision = "UNCERTAIN";
+
+  return {
+    key: TRUST_PROVIDER,
+
+    hooks: {
+      onBeforeSettle: async (_declaration, context) => {
+        const query: TrustQuery = {
+          schema: "x402-trust-query-v0.1",
+          payer: {
+            wallet: context.paymentPayload?.payload?.authorization?.from,
+          },
+          resource: {
+            url: context.requirements?.resource?.url ?? "",
+            method: context.requirements?.resource?.method,
+            amount: context.requirements?.maxAmountRequired
+              ? {
+                  value: context.requirements.maxAmountRequired.amount,
+                  currency: context.requirements.maxAmountRequired.asset?.symbol ?? "USDC",
+                  chain: context.requirements.maxAmountRequired.asset?.network ?? "unknown",
+                }
+              : undefined,
+          },
+          requested_at: new Date().toISOString(),
+        };
+
+        const evaluations = await Promise.all(
+          config.providers.map(async (provider) => {
+            try {
+              return await Promise.race([
+                provider.evaluate(query),
+                new Promise<never>((_, reject) =>
+                  setTimeout(() => reject(new Error("timeout")), timeout),
+                ),
+              ]);
+            } catch {
+              return {
+                schema: "x402-trust-evaluation-v0.1" as const,
+                provider: provider.name,
+                provider_url: "error",
+                decision: (config.failureMode === "fail-closed" ? "UNCERTAIN" : "PASS") as TrustDecision,
+                reason_code: "provider_error",
+                evaluated_at: new Date().toISOString(),
+              };
+            }
+          }),
+        );
+
+        lastEvaluations = evaluations;
+        const decision = aggregate(evaluations, config.policy);
+        lastDecision = decision;
+
+        if (decision === "FAIL") {
+          return {
+            abort: true,
+            reason: "trust_evaluation_failed",
+            message: `Trust gate blocked settlement: ${evaluations.map((e) => e.reason_code).join(", ")}`,
+          };
+        }
+
+        if (decision === "UNCERTAIN" && config.failureMode === "fail-closed") {
+          return {
+            abort: true,
+            reason: "trust_evaluation_uncertain",
+            message: "Trust evaluation inconclusive under fail-closed policy",
+          };
+        }
+
+        // PASS or UNCERTAIN+fail-open: proceed with settlement
+      },
+    },
+
+    enrichSettlementResponse: async () => {
+      if (lastEvaluations.length === 0) return undefined;
+
+      const topEval = lastEvaluations[0];
+      return {
+        decision: lastDecision,
+        score: topEval?.score,
+        evidence_uri: topEval?.evidence_uri,
+        provider: topEval?.provider,
+      };
+    },
+  };
+}
--- a/typescript/packages/extensions/src/trust-provider/types.ts
+++ b/typescript/packages/extensions/src/trust-provider/types.ts
@@ -1,0 +1,70 @@
+/**
+ * Trust-Provider Extension — wire types
+ *
+ * Defines the trust query, evaluation, and configuration types
+ * used by the onBeforeSettle hook to gate settlement on behavioral trust.
+ */
+
+export const TRUST_PROVIDER = "trust-provider" as const;
+
+export type TrustDecision = "PASS" | "FAIL" | "UNCERTAIN";
+
+export interface TrustQuery {
+  schema: "x402-trust-query-v0.1";
+  payer: {
+    wallet?: string;
+    agent_id?: string;
+    session_id?: string;
+  };
+  resource: {
+    url: string;
+    method?: string;
+    amount?: {
+      value: string;
+      currency: string;
+      chain: string;
+    };
+  };
+  context?: {
+    category?: string;
+    risk_band?: "low" | "medium" | "high";
+  };
+  requested_at: string;
+}
+
+export interface TrustEvaluation {
+  schema: "x402-trust-evaluation-v0.1";
+  provider: string;
+  provider_url: string;
+  decision: TrustDecision;
+  score?: number;
+  evidence_uri?: string;
+  reason_code?: string;
+  ttl_seconds?: number;
+  evaluated_at: string;
+}
+
+export type AggregationPolicy =
+  | { kind: "strict" }
+  | { kind: "quorum" }
+  | { kind: "custom"; combine: (evals: TrustEvaluation[]) => TrustDecision };
+
+export interface TrustProviderConfig {
+  name: string;
+  evaluate: (query: TrustQuery) => Promise<TrustEvaluation>;
+}
+
+export interface TrustProviderExtensionConfig {
+  providers: TrustProviderConfig[];
+  policy: AggregationPolicy;
+  failureMode: "fail-closed" | "fail-open";
+  perProviderTimeoutMs?: number;
+}
+
+export interface TrustProviderDeclaration {
+  info: {
+    providers: string[];
+    policy: string;
+    failureMode: string;
+  };
+}
--- a/typescript/packages/extensions/src/trust-provider/utils.ts
+++ b/typescript/packages/extensions/src/trust-provider/utils.ts
@@ -1,0 +1,38 @@
+/**
+ * Trust-Provider Extension — aggregation utilities
+ */
+
+import type { TrustEvaluation, TrustDecision } from "./types";
+
+/** STRICT: all providers must PASS. Any FAIL => FAIL, any UNCERTAIN => UNCERTAIN. */
+export function aggregateStrict(evals: TrustEvaluation[]): TrustDecision {
+  if (evals.some((e) => e.decision === "FAIL")) return "FAIL";
+  if (evals.some((e) => e.decision === "UNCERTAIN")) return "UNCERTAIN";
+  return "PASS";
+}
+
+/** QUORUM: majority rules. Ties resolve to UNCERTAIN. */
+export function aggregateQuorum(evals: TrustEvaluation[]): TrustDecision {
+  const n = evals.length;
+  const threshold = Math.ceil(n / 2);
+  const fails = evals.filter((e) => e.decision === "FAIL").length;
+  const passes = evals.filter((e) => e.decision === "PASS").length;
+  if (fails >= threshold) return "FAIL";
+  if (passes >= threshold && fails === 0) return "PASS";
+  return "UNCERTAIN";
+}
+
+/** Resolve aggregation for a given policy. */
+export function aggregate(
+  evals: TrustEvaluation[],
+  policy: { kind: "strict" } | { kind: "quorum" } | { kind: "custom"; combine: (e: TrustEvaluation[]) => TrustDecision },
+): TrustDecision {
+  switch (policy.kind) {
+    case "strict":
+      return aggregateStrict(evals);
+    case "quorum":
+      return aggregateQuorum(evals);
+    case "custom":
+      return policy.combine(evals);
+  }
+}


feedoracle:
Following on from the conversation in #2285 — happy to support this PR, and we'd be a reference consumer on the regulatory side.

A quick observation on the `TrustEvaluation` shape: the `evidence_uri` field is well-placed for our use case. We emit ECDSA-signed JWS receipts (DID:WBA, anchorable on Polygon) on every regulatory decision, so `evidence_uri` becomes a verifiable artifact rather than just a pointer. That maps cleanly to the `evidenceType: cryptographic` enrichment proposed in #2285.

Concrete example — how a TrustQuery → TrustEvaluation roundtrip would look against our Bazaar endpoint:

```json
// TrustQuery (from the trust-provider extension)
{
  "schema": "x402-trust-query-v0.1",
  "payer": { "agent_id": "agent_42", "wallet": "0x..." },
  "resource": {
    "url": "https://tooloracle.io/v2/integrity_gate",
    "method": "POST",
    "amount": { "value": "20000", "currency": "USDC", "chain": "base" }
  },
  "context": {
    "category": "compliance",
    "risk_band": "high"
  },
  "requested_at": "2026-05-15T13:00:00Z"
}

// TrustEvaluation (back from a regulatory provider like us)
{
  "schema": "x402-trust-evaluation-v0.1",
  "provider": "PredictionGuard",
  "provider_url": "https://feedoracle.io/predictionguard/mcp",
  "decision": "FAIL",
  "score": 18.9,
  "evidence_uri": "https://feedoracle.io/predictionguard/receipts/sha256-abc...",
  "reason_code": "RULE_5_17z_CONFLICT",
  "ttl_seconds": 3600,
  "evaluated_at": "2026-05-15T13:00:01Z"
}
```

`PASS / FAIL / UNCERTAIN` maps directly to our `ALLOW / BLOCK / WARN` internal decision band, so adapting our existing `pg_prediction_market_integrity_gate` tool into a `TrustProviderConfig.evaluate()` function is essentially a thin shim.

Two small suggestions, take or leave:

1. Could `TrustEvaluation` carry an optional `evidenceType: "regulatory" | "behavioral" | "attestation" | "cryptographic"` field? This is what I proposed in #2285 — it lets aggregation policies route differently. A `STRICT` policy probably wants all providers to PASS regardless of type, but a custom policy might want "any regulatory FAIL aborts immediately, behavioral FAIL is informational only."

2. The `context.category` lookup currently relies on the caller passing it. If a Bazaar-listed service self-declares `category` in its discovery manifest (which is what #2285 would formalize), the extension could populate `context.category` automatically from the resource URL. Tighter integration with the category proposal.

We have all 5 compliance endpoints live on Bazaar today and tracked in Dominion Observatory's compliance leaderboard (currently 6 ranked entries from us, 14 with `trust=75/0-interactions`). When this PR lands I'm happy to publish a reference `evaluate()` implementation against our backend as a worked example.

AI disclosure: this comment was drafted with Claude assistance and reviewed before posting.

chopmob-cloud:
Supporting this PR from the behavioral trust side — AlgoVoi runs a live agent trust bench (https://agent-trust-bench.algovoi.co.uk) that maps naturally onto the `TrustProviderConfig.evaluate()` interface.

**What the bench produces**

138 profiles across 30 threat categories, each generating a pass/fail signal on payment behavior: blind default-pay (no validation), amount stop-point failures ($0.01 → $0.10 → $1.00 → $10.00 escalations), authority injection, receiver spoofing, multi-turn manipulation, credential extraction. The bench currently runs each profile per-request and publishes aggregate counters at `/stats`.

Mapped to the `TrustEvaluation` shape:

```json
{
  "schema": "x402-trust-evaluation-v0.1",
  "provider": "AlgoVoi Agent Trust Bench",
  "provider_url": "https://agent-trust-bench.algovoi.co.uk",
  "decision": "FAIL",
  "score": 23.0,
  "evidence_uri": "https://agent-trust-bench.algovoi.co.uk/stats",
  "reason_code": "BLIND_DEFAULT_PAY_DETECTED",
  "ttl_seconds": 300,
  "evaluated_at": "2026-05-15T15:00:00Z"
}
```

**On feedoracle's two suggestions**

Both are right. On (1): adding `evidenceType` to `TrustEvaluation` is the correct hook for aggregation policy differentiation. Behavioral signals accumulate — a single FAIL from an agent that otherwise has a 95% pass rate is informationally different from a FAIL from an agent with no history. A custom aggregation policy that inspects `evidenceType: behavioral` can weight those differently from a regulatory FAIL, which is deterministic and context-independent. Without the field, the `STRICT` policy treats a first-probe miss the same as a confirmed rogue actor.

On (2): auto-populating `context.category` from the Bazaar registry entry would reduce the coordination surface significantly. If the resource URL is Bazaar-indexed with `category: Compliance`, the hook already knows the routing without the caller passing it.

**One addition to consider**

The `TrustQuery` shape currently includes `payer.wallet`. For multi-chain agents, a single payer may operate across address spaces (EVM, Solana, Algorand, etc.) under the same session identity. A `payer.agent_id` or `payer.session_token` field alongside `payer.wallet` would let behavioral providers correlate cross-chain history — relevant for the bench where we observe escalation patterns across payment networks.

Happy to implement a reference `evaluate()` against the bench API once this PR is reviewed. The bench runs live x402 paid endpoints (Base, Solana, Algorand) so the `TrustQuery` roundtrip can be fully end-to-end testable.

— AlgoVoi (chopmob-cloud)

feedoracle:
+1 on @chopmob-cloud's `payer.agent_id` / `payer.session_token` addition. From the regulatory side, the same correlation problem shows up across payment rails rather than payment networks per se — we emit ES256K-signed receipts across four rails today (USDC + EURC on Base, XRP + RLUSD on XRPL), and a single regulatory subject (say, a CASP under MiCA Art. 35) can legitimately settle across all four. Tying those settlements back to a single subject under audit currently requires off-chain joins. A `payer.agent_id` on `TrustQuery` would let the trust-provider hook do that join at the protocol layer.

One implementation note that may be useful for the spec text: if `payer.agent_id` is treated as a DID-resolvable identifier (e.g. `did:web:...`, `did:wba:...`), the same field carries both the cross-chain wallet correlation @chopmob-cloud described and the upstream regulatory subject identity. Two consumers, one field. Avoids parallel `payer.session_token` + `payer.regulatory_subject_id` proliferation.

Adjacent layer worth flagging: our **AgentGuard** module (20 tools) implements roughly the runtime-side counterpart to what `TrustProviderConfig.evaluate()` does at the settlement hook — policy preflight, scope/session validation, replay guard, payment-policy check, cross-tool anomaly detection, audit-log write/query. If the trust-provider extension lands with `payer.agent_id` carrying a DID, AgentGuard can serve as the runtime enforcement layer that the `evaluate()` decision routes into. Same identity field, two consumers (settlement-time + execution-time).

On the reference `evaluate()` offer — happy to publish one against the regulatory side. We have ~36 paid compliance products live across Base + XRPL settlements, so the trust-provider roundtrip can be tested across both EVM and non-EVM rails. Combined with @chopmob-cloud's bench on Base/Solana/Algorand, the test matrix gets reasonably broad before the PR lands.

chopmob-cloud:
+1 on DID as the identity format for `payer.agent_id`. The two-consumer observation (settlement-time trust evaluation + execution-time runtime enforcement) points at a third consumer that makes the DID choice load-bearing: **on-chain agent registration**.

AlgoVoi already registers against ERC-8004 (agent ID 5938 on ARC Testnet). ERC-8004 ties a DID to an on-chain identity and reputation registry — the same identifier can anchor the TrustProvider lookup, the AgentGuard enforcement decision, and the chain-readable registration in one field. That three-way binding is what makes `payer.agent_id` worth specifying as a DID rather than an opaque token: a `did:web:...` or `did:wba:...` can be resolved to an on-chain record, a compliance attestation, and an execution-time policy in the same roundtrip. An opaque session token can only do one of those.

Practical implication for the spec text: the `TrustQuery` should probably define `payer.agent_id` as optionally DID-resolvable but not require it, since many callers will not yet have on-chain registration. A `payer.agent_id` that looks like a DID URI signals "resolve me for history and on-chain posture"; a bare string signals "opaque correlator, cross-session only". Evaluators can handle both.

On the test matrix: we cover 8 chains on the behavioral side (Base, Solana, Algorand, VOI, Hedera, Stellar, Tempo, ARC). Combined with feedoracle's Base + XRPL regulatory coverage, that reaches 9 distinct networks pre-merge, including 5 non-EVM chains. Worth noting in the PR test plan — the extension is designed to be chain-agnostic at the hook, and the combined matrix tests that claim before it lands.

Happy to coordinate the reference `evaluate()` implementations so the behavioral and regulatory examples use the same `TrustQuery` schema version and can be tested in sequence without breaking changes between them.

— AlgoVoi (chopmob-cloud)

feedoracle:
The three-consumer framing locks the DID choice in. ERC-8004 registration as the third consumer is the right call — that's where the on-chain reputation graph lives, and binding `payer.agent_id` to a DID that resolves to all three (TrustProvider history, AgentGuard runtime policy, ERC-8004 registry record) collapses what would otherwise be three separate identity systems into one resolvable identifier.

The "optionally DID-resolvable, not required" rule for the spec text is correct. Bare-string opaque correlators stay valid for callers without on-chain registration; DID URIs unlock the richer roundtrip. Two-mode evaluator handling is straightforward.

On the test matrix coordination — 9 distinct networks (4 EVM + 5 non-EVM) is broader than what's typical for a payment extension pre-merge. Worth flagging as test plan in the PR description itself, not just in implementation notes. Agent 402 Tape's per-chain settlement view (cc @danielnorkin) becomes the post-merge measurement reference.

Adjacent on our side: we run `quantum_trust_passport` and `agent_trust_passport` tools today (QuantumOracle, SupportOracle) — structurally similar to ERC-8004's reputation surface, off-chain. Worth aligning the resolution path so a DID can route to both an on-chain ERC-8004 record and an off-chain trust-passport endpoint via the same identifier. We'll register against ERC-8004 in the near term to close the loop.

Coordinated reference `evaluate()` makes sense — same `TrustQuery` schema version, behavioral and regulatory examples runnable in sequence. Happy to align on a shared test fixture set (one query → two evaluators → composite verdict at the hook) so the PR review can demonstrate the composition end-to-end.

chopmob-cloud:
On the off-chain/on-chain DID resolution path: the DID Document service endpoint array is exactly the right mechanism. A `did:web:...` document can declare both:

```json
{
  "service": [
    {
      "id": "#erc8004",
      "type": "ERC8004Registry",
      "serviceEndpoint": "https://arc.network/registry/agent/5938"
    },
    {
      "id": "#compliance",
      "type": "ComplianceAttestation",
      "serviceEndpoint": "https://api.algovoi.co.uk/compliance/attestation"
    }
  ]
}
```

A TrustProvider that receives `payer.agent_id: "did:web:algovoi.co.uk"` can resolve the DID Document and route to whichever service type it needs — on-chain ERC-8004 record for reputation history, off-chain attestation for compliance posture. Same identifier, two resolution paths, no coordination overhead between the consumers. When feedoracle registers against ERC-8004, the same pattern applies: `did:web:tooloracle.io` resolves to both the on-chain registry record and the trust-passport endpoint.

This is worth a sentence in the spec text for `payer.agent_id`: "If the value is a DID URI, evaluators SHOULD resolve the DID Document and may route to service endpoints of types `ERC8004Registry`, `ComplianceAttestation`, or `TrustPassport` as appropriate for their evaluation class."

On the PR test plan — agreed, the 9-network matrix belongs in the PR description rather than implementation notes. Updating #2322 to include it now.

On the shared fixture set: one `TrustQuery`, two evaluators (behavioral bench + regulatory oracle), composite verdict at the hook. A concrete fixture target:

```json
{
  "schema": "x402-trust-query-v0.1",
  "payer": {
    "agent_id": "did:web:test-agent.algovoi.co.uk",
    "wallet": "0xabcd...1234"
  },
  "resource": {
    "url": "https://agent-trust-bench.algovoi.co.uk/escalation/step-2",
    "method": "GET",
    "amount": { "value": "10000", "currency": "USDC", "chain": "base" }
  },
  "context": { "category": "Compliance", "risk_band": "high" },
  "requested_at": "2026-05-15T18:00:00Z"
}
```

Behavioral evaluator (AlgoVoi bench): returns FAIL if the agent has previously paid the escalation endpoint without validating. Regulatory evaluator (feedoracle): returns ALLOW or WARN based on sanctions/AML posture of the agent DID. Composite at the hook under STRICT policy: any FAIL aborts settlement. Under a custom policy: regulatory FAIL is deterministic and overrides; behavioral FAIL is informational unless sustained over N calls.

Happy to commit this fixture to the PR test directory once the PR is open for review. Will update #2322 description now.

— AlgoVoi (chopmob-cloud)

chopmob-cloud:
Flagging a related thread that has opened from the Compliance working group (#2285): **#2327 -- Privacy-preserving x402: settlement visibility classes and evidence plane implications**.

The connection to this thread (#2300 trust-provider extension): a fully-private settling payer is structurally invisible to behavioral accumulators. A trust-provider verdict for that payer would have to rely entirely on regulatory + cryptographic evidence. The payer.agent_id / payer.session_token fields you and @feedoracle proposed here become the only non-settlement signal available for trust assessment when privacy_class is attested-private or fully-private.

That interaction is worth noting in the trust-provider spec -- a SHOULD or MAY for how verdicts degrade when the behavioral evidence plane is blind.

danielnorkin:
Thanks @feedoracle for the cc. Three quick notes from the observation plane.

**On the measurement role.** Confirmed for post-merge per-chain settlement view across the nine networks (4 EVM + 5 non-EVM) that @chopmob-cloud listed. Once `TrustQuery` lands with `payer.agent_id`, we can publish per-agent-DID settlement aggregates alongside the per-facilitator view we already produce. That gives consumers a way to validate trust-provider verdicts against on-chain behavior reproducibly.

**On facilitator attribution as input to TrustProvider.** Adjacent thread worth flagging: [#2335](https://github.com/x402-foundation/x402/issues/2335) proposes standardizing facilitator-attribution so a trust-provider can reliably know which facilitator handled a settlement (and read its declared signer roles / validity windows). Without it, the trust-provider extension would have to fall back to the same heuristic-labeling problem the observational layer hits. Same input-quality issue from two sides.

**On privacy interaction.** The trust-provider extension intersects with #2326 (privacy_class) cleanly: a `fully-private` settling payer is invisible on the behavioral plane, and as @chopmob-cloud noted, verdicts for those payers must degrade. From the observational side, those settlements are also invisible to us, so the post-merge measurement reference will explicitly report per-`privacy_class` coverage rather than appearing to undercount.

Aligned overall. Will publish a Tape view that lets reviewers verify trust-provider verdicts against observed settlement behavior once this lands.

vdineshk:
Thanks everyone — this has been a really productive schema thread.

On evidenceType: I like this addition. The Observatory already tags behavioral vs. static provenance internally; surfacing that in the TrustEvaluation shape makes it actionable for callers who need to differentiate regulatory attestations from runtime behavioral signals. Happy to prototype a union type here — "behavioral" | "regulatory" | "self-attested" | "third-party" as a starting point.

On payer.agent_id with DID resolution: the ERC-8004 alignment is exactly right. I was using opaque session tokens as a lightweight placeholder, but DID-based identity makes cross-chain correlation deterministic rather than session-scoped. The service endpoint resolution pattern @chopmob-cloud described would let the trust provider dereference both on-chain and off-chain attestations from a single identifier — which is cleaner than anything I had in mind.

@danielnorkin — thanks for confirming coverage across nine networks. I'll take a look at [#2335](https://github.com/vdineshk/daee-engine/issues/2335) and [#2326](https://github.com/vdineshk/daee-engine/issues/2326) this week; the facilitator attribution question in particular intersects directly with how the Observatory attributes interaction provenance.

Will open a follow-up commit with evidenceType, DID-based payer.agent_id, and a stub service endpoint resolver. Feedback welcome before I close that out.

chopmob-cloud:
Reference DID Document for the resolver stub now live:

**https://api.algovoi.co.uk/.well-known/did.json** (resolves `did:web:api.algovoi.co.uk`)

Exposes seven service endpoints — the relevant ones for trust-provider dereferencing:

- `#erc8004` → on-chain ERC-8004 registry record (ARC agent ID 5938)
- `#compliance` → public posture attestation (regulatory)
- `#verascore` → attestation-provider metadata + dimensionMap (third-party / behavioral)
- `#bench` → 138-profile adversarial conformance fixtures (behavioral)

Plus `#a2a`, `#x402`, and `#pqc-keys` for the existing surfaces. Ed25519 verification method matches the JWKS at `/.well-known/jwks.json` (`alg=EdDSA`, kid `d0481df4...`). `alsoKnownAs` carries the equivalent `did:key:z6Mkg...` form for callers that prefer key-bound identifiers.

Resolver can dereference against this directly. Happy to add more service types if useful for the v0.1 spec.

— AlgoVoi (chopmob-cloud)

danielnorkin:
Coming at this from the seller side of a marketplace, where the agents offering paid endpoints need a way to prove identity to buyers and trust providers.

We operate a marketplace where sellers register their agents across EVM chains (Base, Ethereum, Polygon, Arbitrum, Optimism, Avalanche, Hedera) and Solana, each choosing their identity method per agent. Every registered agent already carries the substrate a resolver would dereference: an ERC-8004 IdentityRegistry record where applicable, an x402 payment surface, an A2A task surface, owner-wallet binding, and review history tied to on-chain settlements.

What's missing today is the standard envelope. @chopmob-cloud's reference doc is a clean single-agent example. The scaling question for v0.1: should marketplaces be first-class issuers of these documents on behalf of the sellers listed on them? Concretely:

1. Seller signs once with their wallet. Marketplace serves the document at a stable URL and rotates service endpoints as the agent evolves.
2. The document declares one or more identity methods side by side. `did:web` for discovery, ERC-8004 for the on-chain anchor, `did:ethr` / `did:hedera` / `did:sol` where the seller prefers a chain-native method, all linked via `alsoKnownAs`.
3. Trust providers dereference one URL and pick whichever identity method they trust.

Stays facilitator-agnostic and ledger-agnostic by design. Seller chooses the identity methods, marketplace serves the envelope.

Would the v0.1 spec be open to formalising a marketplace-issuance pattern alongside self-hosted, or is the intent to keep issuance strictly self-hosted per the reference doc?

chopmob-cloud:
@danielnorkin — appreciate the cite. One observation in support of marketplace-issuance as a first-class pattern: our `/.well-known/did.json` already ships with `alsoKnownAs` linking three methods today (`did:web:api.algovoi.co.uk` + `did:key:z6MkgExzvcpvxrghf4Q3285xqSdenhRZHcP6wc5UvY6VVaz5` + the just-issued `did:foxbook:01KRXTMK3Z20J7V7MMD17W6T59` once the agent-card `x-foxbook` block lands). The "side-by-side identity methods" mechanism you describe works the same whether the issuer is the seller themselves or a marketplace operating on the seller's behalf — the substrate is method-list-under-`alsoKnownAs` either way, signed by the issuer's key.

Stays facilitator-agnostic + ledger-agnostic per your framing. The marketplace-issuance pattern would just be a different `id` field controller in the published DID Document, with the seller's verification method(s) under `verificationMethod` rather than the issuer's. Trust providers dereference one URL and rank the methods by their own policy.

— AlgoVoi (chopmob-cloud)

feedoracle:
@vdineshk — glad you're back on this. Two notes from the regulatory + cryptographic side that may be useful as you draft the follow-up commit.

**On the evidenceType union.** The four-value taxonomy you sketched (`behavioral | regulatory | self-attested | third-party`) overlaps cleanly with the Compliance category evidenceType taxonomy already landed in [#2322](https://github.com/x402-foundation/x402/pull/2322) (`behavioral | regulatory | observational | cryptographic`). The two are working different layers (your union describes the *source* of a trust evaluation; #2322 describes the *receipt shape* the source produces), but they cross-cut: a `third-party` source can produce `cryptographic` receipts, a `regulatory` source produces `regulatory` receipts, and so on. Worth referencing #2322's taxonomy as the receipt-shape vocabulary so a TrustEvaluation can carry both `source` and `evidenceType` without semantic collision. The four `evidenceShape` constraint rules and JCS canonicalisation floor in #2322 / #2334 apply identically downstream of either field.

**On marketplace-issuance for DID Documents** (raised by @danielnorkin, supported by @chopmob-cloud). One regulatory wrinkle worth landing in the v0.1 spec text: when a marketplace publishes a DID Document on a seller's behalf, the **controller** of that document carries the customer-due-diligence obligation for the seller's signing key under AMLR Art. 16. The seller's `verificationMethod` entries are the *subject* of the DD, but the marketplace as `id` controller is the *obligated party* for verifying those keys against sanctions/PEP lists at issuance time and on rotation. This is structurally the same accountability pattern as platform-issued vs self-issued attestations under MiCA Art. 88 (information requests run to the controller, not the subject). Practical implication: the v0.1 spec should probably state that a trust provider dereferencing the document MAY treat `id` controller and `verificationMethod` subject differently when computing trust scores — controller liability is what gives marketplace-issued documents their regulatory weight, and a self-issued document is structurally different from a marketplace-issued one even if the cryptographic substrate is identical. The `alsoKnownAs` multi-method pattern @chopmob-cloud and @danielnorkin described handles this cleanly: each method declares its own controller, and the trust provider ranks methods by both cryptographic strength and the regulatory standing of their controllers.

Happy to coordinate on the v0.1 spec text if a controller-vs-subject distinction lands as part of the follow-up commit.

chopmob-cloud:
@feedoracle — controller-vs-subject distinction is the right reframe. Concrete data point on the self-issued side: our `/compliance/attestation` already publishes the UK MLR 2017 / SAMLA s.20 / AMLR-equivalent framework alignment (frameworks list, active screening sources, audit-chain head, selective-disclosure methodology). The `id` controller and `verificationMethod` subject collapse into a single identity for self-issued — unified DD obligation, no controller-vs-subject split. Trust providers dereferencing our `/.well-known/did.json` see that posture explicitly. Happy to coordinate on v0.1 spec text for the controller-vs-subject surface.

— AlgoVoi (chopmob-cloud)

feedoracle:
@chopmob-cloud — the self-issued degenerate case is the right framing. The three patterns line up cleanly as a regulatory-accountability matrix:

| Issuance pattern | `id` controller | `verificationMethod` subject | DD obligation under AMLR Art. 16 |
|---|---|---|---|
| **Self-issued** | seller | seller | unified — single party holds full DD obligation for own keys (AlgoVoi `/.well-known/did.json` is the reference example) |
| **Marketplace-issued** | marketplace operator | seller | split — marketplace DDs the seller as a customer + DDs the seller's signing keys at issuance and rotation; seller retains DD obligation for their own operational scope |
| **Third-party-issuer** (KYC-as-a-service, regulated identity provider) | identity provider | seller | layered — identity provider carries the DD obligation for the assertion they sign; both seller and any downstream marketplace are downstream relying parties |

The third-party-issuer pattern is structurally distinct from marketplace-issued and worth landing as its own row: a regulated KYC provider (Onfido, Sumsub, IDnow, Persona under EU AMLR-equivalent supervision) issuing a verifiable credential about a seller's identity is not the same as a marketplace operator running its own DD. The KYC provider is licensed for the assertion; the marketplace is licensed for the venue. Under AMLR Art. 16 they sit in different obligation tiers, and a trust provider ranking methods by regulatory standing should be able to read which tier each `verificationMethod` belongs to.

**Concrete spec text suggestion** for the v0.1 commit:

```
verificationMethod entries SHOULD include an optional `issuance_pattern` field with
values `self` | `marketplace` | `third-party-issuer`. Trust providers MAY weight
methods differently based on the regulatory-standing of the controller, where:
  - self: controller and subject collapse; weight follows the subject's own posture
  - marketplace: controller-side DD (AMLR Art. 16) applies to seller onboarding + key rotation
  - third-party-issuer: controller is the credentialed assertion-issuer; weight follows
    the issuer's regulatory license tier (e.g. AMLR-supervised, MiCA-CASP-authorised, eIDAS QTSP)
```

Happy to coordinate the v0.1 spec text — sounds like there's working-group consensus on landing the three-pattern surface explicitly rather than leaving it implicit in `id` vs `verificationMethod` semantics. @vdineshk — workable as part of the follow-up commit you outlined?

chopmob-cloud:
@feedoracle — three-pattern matrix is the right shape; `issuance_pattern` as an optional `verificationMethod` field reads clean. One cross-pattern data point: AlgoVoi's `/.well-known/did.json` already binds two issuance patterns side by side under `alsoKnownAs`:

- `did:web:api.algovoi.co.uk` — structurally `self` (controller = subject, `/compliance/attestation` carries the UK MLR 2017 / SAMLA s.20 framework alignment as the AMLR-equivalent posture surface)
- `did:foxbook:01KRXTMK3Z20J7V7MMD17W6T59` (transparency log leaf 7) — structurally `third-party-issuer` (Foxbook controls the registration assertion; we're the subject)

The matrix maps onto live infra without modification — and the multi-method `alsoKnownAs` surface gives the trust provider an explicit signal for ranking methods per the regulatory standing of each controller.

Will adopt `issuance_pattern: "self"` on the did:web `verificationMethod` entry the moment v0.1 spec text lands. @vdineshk — workable as part of the follow-up commit?

— AlgoVoi (chopmob-cloud)

vdineshk:
Follow-up commit landed: [`vdineshk/daee-engine@5876d9c`](https://github.com/vdineshk/daee-engine/commit/5876d9c) — covers the three items from this thread.

### What shipped

1. **`evidenceType` on `TrustEvaluation`** — six-value taxonomy: `behavioral | regulatory | self-attested | third-party | cryptographic | observational`. Aligned with the Bazaar Compliance category taxonomy in #2322. The two are orthogonal: `evidenceType` describes the evaluation source; `evidenceShape` (from #2322/#2334) describes the receipt format. Both can coexist without semantic collision.

2. **DID-based `payer.agent_id`** — full `did:web` resolver stub (`resolveDID`, `isDID`, `findServiceEndpoints`). If `agent_id` is a DID URI, evaluators resolve the DID Document and route to service endpoints (`ERC8004Registry`, `ComplianceAttestation`, `TrustPassport`, `BehavioralBench`). Bare strings remain valid as opaque correlators. The Observatory provider extracts the domain from `did:web:...` for lookup.

3. **`issuance_pattern` on verification methods** — `self | marketplace | third-party-issuer` per the three-pattern regulatory accountability matrix @feedoracle and @chopmob-cloud converged on. Spec text on the type documents the controller-vs-subject DD obligation split (AMLR Art. 16) and the self-issued degenerate case.

### Also added

- `aggregateByEvidenceType` — a custom aggregation function where regulatory/cryptographic FAILs abort immediately, behavioral FAILs are informational unless sustained. Demonstrates how `evidenceType` makes policy routing concrete.
- `X-Trust-Evidence-Type` advisory header.
- Privacy degradation note on `beforeSettle` — when behavioral plane is blind (fully-private payer), verdicts degrade to regulatory + cryptographic evidence only.

@feedoracle — yes, `issuance_pattern` is workable and landed. The three-row matrix (self / marketplace / third-party-issuer) maps cleanly. Happy to coordinate on v0.1 spec text for the controller-vs-subject surface.

@chopmob-cloud — `issuance_pattern: "self"` is ready for adoption on your `did:web` verificationMethod. The resolver stub can dereference your `/.well-known/did.json` directly. Will test against `did:web:api.algovoi.co.uk` this week.

@danielnorkin — will review #2335 (facilitator attribution) and #2326 (privacy_class) this week as noted. The facilitator attribution question intersects directly with how the Observatory attributes interaction provenance.

Next step: push a follow-up commit to this PR branch with the updated types mirrored into the x402 extension. The `@dominion/trust-provider` npm package already carries the canonical types.

vdineshk:
@reaworks-ops — buyer-auditable receipt is the right next layer. The current `BeforeSettleResult` already carries most of what you listed:

| Your field | Where it lives today |
|---|---|
| trust provider ids + policy mode | `BeforeSettleResult.evaluations[].provider` + `config.policy.kind` |
| provider verdicts + timestamps | `TrustEvaluation.decision`, `.score`, `.evaluated_at`, `.ttl_seconds` |
| aggregation decision | `BeforeSettleResult.decision` (PASS / FAIL / UNCERTAIN) |
| evidence type | `TrustEvaluation.evidenceType` (six-value taxonomy landed in latest commit) |

What's missing and worth adding:

1. **Redaction boundary** — which fields from the `TrustQuery` each provider actually saw. Today every provider gets the full query; a `redacted_fields?: string[]` on each evaluation would make input visibility auditable.
2. 2. **Side-effect state** — `settled | refused | reversed | human_review`. This lives downstream of `beforeSettle` (in the facilitator), but a `SettlementReceipt` type that wraps `BeforeSettleResult` + facilitator outcome would close the loop.
3. 3. **Falsifier** — strongest idea in your list. A `falsifier?: string` field on `TrustEvaluation` (e.g. "score derived from <30 calls in trailing 7d window") lets buyers assess confidence without trusting the number blindly.
4. 4. **Cure path** — needs spec-level work in the facilitator layer. @danielnorkin's observation-plane work in #2335 may be the right place for dispute/refund routing.
Re: the pilot — happy to provide a real Observatory settlement trace. The `/agent-query/{server}` endpoint already returns a live trust verdict with payment proof; I can package a full round-trip (query + evaluation + facilitator settle) as a reference packet. Let's take that to a side thread or DM so this PR stays focused on the extension types.

Tracking the receipt shape as a follow-up to this PR — will open a spec issue for `SettlementReceipt` once #2300 lands.

chopmob-cloud:
@vdineshk, implementation lands clean. Three points to lock alongside the new commit:

**1. Three-pattern matrix + AMLR Art. 16 framing.** The `issuance_pattern: self | marketplace | third-party-issuer` adoption with controller-vs-subject DD obligation split per AMLR Art. 16 reads correctly against the structural reframe feedoracle drew. The self-issued degenerate case (controller = subject, unified DD obligation, no split) is exactly the shape AlgoVoi ships at `/compliance/attestation` today, so the matrix maps onto live infra without translation logic.

**2. DID Document service endpoints, live infra cite.** AlgoVoi already publishes the four-service resolver surface vdineshk's `did:web` resolver walks for. From `did:web:api.algovoi.co.uk/.well-known/did.json` today:

| vdineshk's typed endpoint | AlgoVoi service entry | Status |
|---|---|---|
| `ComplianceAttestation` | `#compliance` → `/compliance/attestation` | Live, SAMLA s.20 framework alignment + UK MLR 2017 + AMLR-equivalent posture |
| `TrustPassport` (equivalent) | `#verascore` → `/.well-known/verascore-provider.json` | Live, did:key + JWKS + Ed25519 signing surface |
| `BehavioralBench` | `#bench` → `agent-trust-bench.algovoi.co.uk` | Live, 138 profiles across 30 categories |
| `ERC8004Registry` | (queued) → `urn:erc8004:identity:<id>` once agentgraph_bridge_erc8004 lands | In build |

Three of four endpoints resolve against production traffic today. The fourth lands when the ERC-8004 bridge clears its post-launch engineering window.

**3. Cross-anchor to A2A #1734 fixture-matrix work.** The 4-layer URN scheme this resolver pattern implies (`urn:erc8004:identity:<id>` + `urn:observatory:eval:<id>` + `urn:foxbook:leaf:<id>` + `urn:concordia:attestation:<id>`, all resolved from one `did:web:` subject) is the same shape I committed to author in the v0.3.3-working-doc.md cross-extension matrix earlier today (https://github.com/a2aproject/A2A/discussions/1734#discussioncomment-16971035). The `(claim_type, evidenceType, source_provider_did)` discrimination tuple kenneives drew composes cleanly with the `issuance_pattern` accountability matrix here. Same DID resolves the structural-identity question, the evidence-provenance question, and the regulatory-accountability question without per-protocol translation logic.

Substrate-and-primitive layering carried into the implementation surface. Appreciated the explicit credit in the commit announcement.

AlgoVoi (chopmob-cloud), did:web:api.algovoi.co.uk + did:foxbook:01KRXTMK3Z20J7V7MMD17W6T59 + did:key:z6MkgExzvcpvxrghf4Q3285xqSdenhRZHcP6wc5UvY6VVaz5

vdineshk:
Three of four resolving against production traffic is the validation surface I was hoping for when the `did:web` resolver landed. Concrete notes back on each:

**1. AMLR Art. 16 mapping.** The self-issued degenerate case (controller = subject, unified DD obligation) is intentional — it's the minimal viable compliance posture for single-operator MCP servers, which is ~80% of the Observatory's 14,800 tracked endpoints today. The split only activates when `issuance_pattern` moves to `marketplace` or `third-party-issuer`, which is where the regulatory accountability routing actually matters. Good to confirm AlgoVoi's `/compliance/attestation` already exercises the degenerate shape without translation logic.

**2. Endpoint table.** That four-service mapping is exactly the resolver surface `findServiceEndpoints()` walks. Three live + one in-build is better coverage than I expected at this stage. When `agentgraph_bridge_erc8004` clears its engineering window, the `ERC8004Registry` service entry should resolve the same `urn:erc8004:identity:<id>` shape the fixture matrix needs — no adapter layer required.

**3. Cross-anchor composition.** The `(claim_type, evidenceType, source_provider_did)` discrimination tuple composing cleanly with `issuance_pattern` is the structural property that makes this extensible beyond the three current rows. Each DID resolves identity, provenance, and accountability from one subject without per-protocol translation — that's the design invariant. The four-layer URN scheme you and kenneives drew is the convergence surface for v0.3.3.

Follow-up commit with the mirrored types incoming once I'm back on the dev machine. The `aggregateByEvidenceType` export already handles the behavioural/regulatory split correctly — regulatory FAILs abort, behavioural FAILs are soft signals — which is the same discrimination the fixture matrix needs to preserve.

feedoracle:
@vdineshk — adoption reads clean across all three (evidenceType six-value taxonomy, DID-based `payer.agent_id`, `issuance_pattern` three-pattern matrix). The orthogonality framing — `evidenceType` describes the evaluation source, `evidenceShape` describes the receipt format, both coexist without semantic collision — is exactly the structural property that lets #2300 and #2322/#2334 compose without coupling.

Three small notes on the `issuance_pattern` landing, mostly for forward-compatibility:

1. **Field placement.** Per-`verificationMethod` entry rather than at the outer DID Document level is correct — a single document can carry multiple methods with different issuance patterns side-by-side (as @chopmob-cloud's `did:web` + `did:foxbook` pair already demonstrates). Outer-document placement would have forced one pattern per DID, which collapses the third-party-issuer case.

2. **Open-enum default behaviour.** Trust providers SHOULD treat unknown `issuance_pattern` values as opaque (no policy-based routing) rather than rejecting the registry entry, mirroring how `disclosure_policy` and `framework` handle unknown values in #2334. This lets new regulatory tiers (eIDAS QTSP-issued, MiCA-CASP-issued, AMLR-supervised-issuer) land without spec PR cycles.

3. **Controller-vs-subject DD obligation spec text.** Happy to draft the AMLR Art. 16 / MiCA Art. 88 paragraph as a follow-on once #2300 closes — keeps the merge surface for this PR clean while preserving the regulatory framing for the v0.1 spec text @chopmob-cloud and I converged on.

`aggregateByEvidenceType` (regulatory/cryptographic FAILs abort, behavioral FAILs informational unless sustained) is the right default. The asymmetry maps directly to what the four frameworks I named in `evidenceShape.framework` actually require — AMLR/DORA/MiCA make regulatory determinations constitutive of the check, not auxiliary.

Donk338:
We run a trust registry for x402 services (continuous probing + paid delivery spot-checks — we buy from listed sellers and verify they actually deliver), and this extension is exactly the interface the ecosystem is missing. To exercise it from the provider side, we've implemented a live conforming provider:

**`https://aegis.borisinc.com/trust/evaluate`** — returns `x402-trust-evaluation-v0.1`:

- `POST` with an `x402-trust-query-v0.1` body (this PR's `onBeforeSettle` flow): evaluates the **payer** wallet (on-chain history heuristics + a known-bad list from address-poisoning campaigns we've observed on Base; honest `UNCERTAIN` when we lack data).
- `GET ?resource=<url>`: evaluates the **seller** — same wire format, buyer-side. Decisions are backed by real paid purchases, e.g. `PASS` / `reason_code: delivery_verified` / `score: 0.916`, with an `evidence_uri`.

It's itself x402-paid ($0.005; first GET evaluation free per caller so you can try it) — trust evaluations are agent commerce too. Plugs into this PR's config with any x402-capable fetch:

```ts
{
  name: "aegis",
  evaluate: async (q) =>
    (await x402Fetch("https://aegis.borisinc.com/trust/evaluate", {
      method: "POST",
      headers: { "content-type": "application/json" },
      body: JSON.stringify(q),
    })).json(),
}
```

Two suggestions from implementing against the branch:

1. **Standardize the provider transport, not just the types.** `TrustProviderConfig.evaluate` is an arbitrary async function, so every provider integration is bespoke glue. If the extension also documented a canonical HTTP binding (`POST TrustQuery -> TrustEvaluation`), providers become plug-and-play config entries (`providers: [{ name, url }]`). Happy to PR that on top of this branch.

2. **A Python port.** The Python SDK has no extension hooks yet; we'd be glad to contribute equivalent `onBeforeSettle` gating for the FastAPI middleware — we run it in production and can dogfood it immediately.

Also +1 on `UNCERTAIN` as a first-class decision — in our data roughly half of listed x402 endpoints are unreachable at any given time, so putting fail-open/fail-closed policy at the aggregation layer rather than forcing providers to guess is the right call.

Donk338:
Following the `evidenceType` discussion (@feedoracle @chopmob-cloud) — one evidence class is missing from the proposed enum, and it's the one that answers the buyer's primary question: **`delivery`**.

`regulatory`, `behavioral`, `attestation`, and `cryptographic` all characterize the *payer* or the transaction context. None of them say whether a paid resource actually returns the goods after settlement — which, in a protocol with no native refunds, is the single largest risk a buyer-side gate needs to price. Reachability and 402-compliance probes don't answer it either: in our registry data, "challenge-honest, delivery-broken" is a recurring failure mode — a service holds a valid paywall, settles the payment, and 404s the paid retry. You only catch it by actually buying.

Proposed addition:

```
evidenceType: "regulatory" | "behavioral" | "attestation" | "cryptographic" | "delivery"
```

`delivery` = the provider has made real paid purchases against the resource and verified the outcome (delivered or not, price honesty on the wire, content sanity). Example from our live provider:

```json
{
  "schema": "x402-trust-evaluation-v0.1",
  "provider": "aegis",
  "provider_url": "https://aegis.borisinc.com",
  "decision": "PASS",
  "score": 0.916,
  "evidenceType": "delivery",
  "reason_code": "delivery_verified",
  "evidence_uri": "https://aegis.borisinc.com/trust?url=...",
  "evaluated_at": "2026-07-11T03:00:00Z"
}
```

(Our delivery evidence is backed by Ed25519-signed receipts with a public verify endpoint — so a `delivery`-class verdict can be `cryptographic`-grade in artifact terms. Which suggests `evidenceType` should describe the evidence *class*, while artifact format is a property of what `evidence_uri` resolves to, per the #2285 discussion.)

One related nuance for aggregation policy: evidence divides by **subject**, not just type. Behavioral-bench and regulatory verdicts are about the *payer*; delivery and uptime verdicts are about the *resource*. A STRICT policy mixing subjects is incoherent — a buyer-side gate should abort on a resource-subject delivery FAIL regardless of payer standing, while a seller-side settlement gate only cares about payer-subject evidence. An optional `subject: "payer" | "resource"` on `TrustEvaluation` would let policies route this cleanly. It also interacts with the privacy_class discussion (#2326) less destructively than feared: a fully-private payer blinds payer-subject behavioral evidence, but resource-subject delivery evidence is unaffected — verdicts don't need to degrade uniformly.

On the shared fixture set (@chopmob-cloud): count us in as the delivery evaluator. One `TrustQuery` → behavioral (AlgoVoi) + regulatory (feedoracle) + delivery (Aegis) → composite verdict at the hook would demonstrate composition across all three evidence subjects. Our provider is live on Base mainnet today (`POST https://aegis.borisinc.com/trust/evaluate`, x402-paid, first GET evaluation free), so the fixture can run end-to-end against production.

Donk338:
Following up on my `delivery` evidenceType note above with numbers, since I only asserted the argument at the time.

Paid delivery spot-checks from our registry, queried 2026-07-26: 272 checks with on-chain settlement against 102 services ($4.56 USDC, 2026-07-07 → 2026-07-26). 181 resolved: 138 delivered, 43 did not. The 43 concentrate in 5 services out of the 96 with a resolved settled check.

All five scored liveness 1.0, challenge_valid 1.0, price_honest 1.0. Failure shapes: 22 checks returned HTTP 402 *after* settlement (3 services), 11 returned 502 (1 service), 10 returned HTTP 200 whose body was not the purchased resource (1 service). One of the five is otherwise a verified-tier service at 87.4 that failed a single delivery — so the class covers reliability, not just fraud.

The 200-with-wrong-body case is the one I'd point at for the taxonomy question: it passes liveness, challenge validity, price honesty, and HTTP status. Nothing in `behavioral | regulatory | self-attested | third-party | cryptographic | observational` names it, because all six describe the payer or the transaction context, and this is a property of the *seller*, measured after settlement.

Concretely, in `TrustEvaluation`:

```json
{
  "evidenceType": "delivery",
  "score": 0.0,
  "decision": "FAIL",
  "reason_code": "paid_no_delivery",
  "signals": { "spot_n": 5, "delivery_rate": 0.0, "last_paid_check": 1784126061 }
}
```

`observational` is the closest existing member but it's doing a different job — per-chain settlement observation (@danielnorkin's view) is about whether the payment happened, not whether the goods arrived.

Same argument written up against the VCX §19 v1.1 carve-outs in #2747, since the composite-trust-query profile is where the two extensions would meet.

@vdineshk anything blocking this PR from your side? Happy to contribute the delivery-evidence section if it's useful.

jywy78kgrf-create:
Strong +1 to @Donk338's `delivery` evidence type — and some corroborating data,
since we ran the same experiment at larger scale and hit a distinction I think
the enum needs.

On 2026-08-07 we paid **502 distinct Base payout addresses** — every seller in
the discoverable catalog with a priced endpoint at or under $0.05 — one real
settled USDC payment each, from a named wallet (`merona.base.eth`), identifying
itself in the User-Agent. Results:

| outcome | sellers | share |
|---|---:|---:|
| delivered | 150 | 29.9% |
| no live paywall | 199 | 39.6% |
| payment rejected | 118 | 23.5% |
| **charged, unserved** | **12** | **2.4%** |
| answered an error | 4 | 0.8% |
| unreachable | 13 | 2.6% |
| skipped by policy | 6 | 1.2% |

**Refinement 1 — `delivery` evidence must separate pre-settlement from
post-settlement failure.** Only the bottom class is delivery evidence. "No live
paywall" (39.6%) and "payment rejected" (23.5%) are liveness/protocol failures —
the money never moved, and they're already observable by passive probing at
near-zero cost. The class that requires actually paying, and the only one that
can support a block, is **payment settled → request refused** (2.4%). If the enum
collapses these, providers will emit `delivery` evidence for things a cheap
liveness check already catches, and the strong signal gets diluted by the weak
one. Suggest the evidence entry carry whether settlement occurred, not just that
a delivery check failed.

**Refinement 2 — delivery evidence is perishable in a way behavioral evidence
isn't.** A seller who fixes their endpoint should clear immediately; a
wash-trading pattern shouldn't. In our implementation a `charged_unserved`
evaluation carries a 1-hour TTL against a 6-hour default, precisely so the next
paid probe can clear it. If TTL is uniform per provider rather than per evidence
class, delivery evidence will either go stale-positive (blocking a seller who
fixed it) or force providers to under-use it.

**On what the 12 actually are** — worth stating because it changed how we treat
the signal: the HTTP codes were 5×`400`, 5×`405`, 2×`500`. In every case the
payment settled *before* the request was validated. We make no claim of intent;
a correctly-parameterized request would likely be served. But for an agent
economy the narrower fact is arguably worse than fraud: **these endpoints charge
for requests they cannot serve**, so an agent with a wrong parameter or a stale
method hint loses money on every retry, silently, at machine speed. That's a
gate-worthy fact no payer-side scoring can see.

Also +1 to the `subject` field distinguishing payer-evidence from
resource-evidence — we arrived at the same need independently and ship it as an
explicit `direction: seller|buyer` on the query, since the two draw on
structurally different evidence (a payer is assessed from history; a seller can
be actively probed).

One economic note for the spec's assumptions: **the entire 502-seller audit cost
less than a cup of coffee.** First-party delivery evidence isn't an expensive
edge case to be used sparingly — at x402 prices, verifying an entire market is
cheaper than speculating about it. Worth designing the enum on the assumption
this evidence class becomes common.

Our implementation is live and free if it's useful to test against:
`POST https://api.merona.io/v1/trust/evaluate` (`x402-trust-query-v0.1` in), with
typed `basis[]` entries where first-party probe results carry class `curated` and
the settlement tx as the ref. Descriptor at `/.well-known/x402-trust-provider`;
a zero-dependency conformance checker is at
https://github.com/jywy78kgrf-create/merona under `conformance/` if you want
vectors for the delivery cases.

Donk338:
Strong agreement on Refinement 1, and I can add a seller-side observation that I
think bears on how the enum should be specified — because you swept us, and we
weren't in your 502.

**On Refinement 1 — separating pre- from post-settlement is right, and the
pre-settlement classes aren't just weaker, they're structurally blind here.**

Refreshing my 2026-07-26 numbers, re-queried today: 338 settled delivery
spot-checks against 121 services ($5.2415 USDC, 2026-07-07 → 2026-08-07). 214
resolved: 156 delivered, 58 did not. The 58 concentrate in 8 services out of
the 115 with a resolved settled check. Failure shapes: 32 returned HTTP 402
*after* settlement, 15 returned HTTP 200 whose body was not the purchased
resource, 11 returned 502.

The part I'd put in the spec text rather than the rationale: every one of those
8 is, by construction, a service that held a valid paywall and accepted a
settled payment. You cannot land in this set otherwise. So "no live paywall"
and "payment rejected" don't merely under-detect this failure mode — they
select against it. A seller has to be working well enough to charge you before
it can charge you and not deliver. Collapsing the classes doesn't dilute the
strong signal so much as invert it, because the sellers most capable of failing
this way are the ones that look best on everything observable for free.

Which is why I'd go further than "carry whether settlement occurred": make
settlement the *precondition* of the class. If `evidenceType: "delivery"` can
be emitted without a settled payment behind it, the class stops meaning
anything and a buyer can't tell a paid verdict from a probe.

**On the 502-seller sweep — the methodological caveat is the denominator.**

We're an x402 seller as well as an evaluator, and we're inside your stated
band: 22 of our 24 paid paths on `api.borisinc.com` are at or under $0.05,
starting at $0.001 (the two exceptions are job-sized bundles at $0.25 and
$0.50).

From the host your collectors run on we've logged 14 requests since
2026-07-30, four of which identify as merona (`merona-collectors/0.1` ×2 on
08-03, `merona-pulse/0.1` ×2 on 08-09); the rest arrive as `pulse-monitor/0.1`
and `x402-index-liveness/1.0`, which I'm not assuming are yours. No payment
attempt in any of them — every request resolved to a 402 challenge or a 404,
and none was retried with an `X-PAYMENT` header. On 2026-08-07 itself the only
request from that host was a liveness probe. So we weren't in the 502, and it
isn't because we failed to deliver — we were never asked to.

The likely reason is concrete and worth stating because it generalises: on
2026-08-03 `merona-collectors/0.1` requested `/.well-known/ai-catalog.json` and
`/ai-catalog.json` from us and got 404 on both. We serve `/.well-known/x402`,
`/llms.txt` and `/openapi.json`, and we're registered in Bazaar — but not that
manifest. ARD was only announced in mid-July and is still a draft, so as of
your sweep almost nobody was serving it. That's on us as much as anyone, and
we'll fix our end.

But the spec consequence stands regardless of who serves what: **a first-party
delivery sweep's denominator is its discovery set, not the market.** "29.9%
delivered" is a property of the catalog that was reachable on the day. If a
`delivery` evidence entry doesn't carry how its subject was discovered, a
consuming policy can't distinguish three very different states — evaluated and
delivered, evaluated and refused, and never evaluated because the crawler
couldn't find it. Under a STRICT aggregation the third silently behaves like
the second, and whichever catalog a provider happens to crawl becomes a
de-facto allowlist. That's a bad property to bake in during the weeks when the
discovery layer itself is a month old and adoption is near zero.

Suggested: an optional `discovery_basis` (or a field on `basis[]`) naming the
catalog and snapshot the subject came from, and an explicit `NOT_EVALUATED`
distinct from `UNCERTAIN`. `UNCERTAIN` currently means "we looked and can't
say"; there's no way to say "we never looked."

**On Refinement 2 — agreed, and the concentration is what forces it.**

58 non-deliveries across just 8 services, against 115 with a resolved settled
check. That distribution is the argument: a uniform per-provider TTL has to
serve both the service that fails once and recovers, and the one that fails 15
times out of 15. Per evidence class is the minimum; per class *and* verdict
would be better, since a FAIL on a recoverable class should decay faster than a
PASS on it.

**On `subject`** — +1 to `direction: seller|buyer` over my `subject` naming if
that's where consensus lands; the important part is that the field exists, and
your framing (a payer is assessed from history, a seller can be actively
probed) states the reason better than mine did.

@vdineshk — still happy to write the delivery-evidence section of the spec text
if that helps move this along.

jywy78kgrf-create:
@Donk338 — taking your points in order, with receipts where I owe them.

Settlement as precondition: accepted, and it's the better formulation. My "carry whether settlement occurred" left the class ambiguous; your version — you cannot emit delivery evidence without a settled payment behind it — is what makes a paid verdict distinguishable from a probe at the wire level. Your selection-effect observation deserves to be in the spec text verbatim: a seller has to be working well enough to charge you before it can charge you and not deliver, so the free-to-observe classes don't just under-detect this failure mode, they anti-select for it.

On the denominator: you're right, and checking made it worse in an instructive way. Our sweep's discovery set was a Bazaar catalog snapshot (2026-08-08, 32,875 resources, content-addressed) — api.borisinc.com appears in it zero times, so you were never evaluated, exactly as your logs show. But here's the sharper part: our own broader discovery layer had already found you — you appear in our service directory via an MCP-directory collector. One pipeline knew you existed; the sweep's denominator, drawn from Bazaar-priced rows only, didn't include you. So discovery failure isn't even binary per provider — it's per pipeline within a provider. Which is why discovery_basis naming just the catalog isn't enough: it should carry the snapshot digest the subject set was drawn from, so a consumer can check whether a given seller was discoverable at evaluation time.

We've shipped exactly that, live now:

Every /v1/delivery response and every curated first_party_probe basis entry from /v1/trust/evaluate carries discovery_basis = {catalog, snapshot_sha256, snapshot_date} — currently sha256:8ef8a87bd8e56d66559329aeedad9f047c6c66e2fa2c2b39f433373311521f03, dated 2026-08-08: the exact snapshot that excluded you. Omitted entirely when no snapshot is loaded, never a fabricated digest.
An address with no probe history now returns an explicit status: "not_evaluated" (with the discovery basis attached) instead of collapsing into a generic unknown — "we never looked" is now distinguishable from "we looked and can't say" on our wire. We'll align the field name with whatever the spec picks.
On the user-agents — your log reading was exactly right, and it exposed a documentation failure on our side. All four merona-identified requests were ours: [merona-collectors/0.1](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/merona-collectors/0.1) (registry/directory collectors — including the 08-03 /.well-known/ai-catalog.json requests; that's our ARD collector measuring draft-catalog adoption, which is how we knew ARD serving was near-zero) and [merona-pulse/0.1](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/merona-pulse/0.1) (x402 wire-shape survey). [x402-index-liveness/1.0](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/x402-index-liveness/1.0) is also ours — nightly liveness, read-only by design, which is why you saw 402s with no X-PAYMENT retry. [pulse-monitor/0.1](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/pulse-monitor/0.1) is not ours. The 2026-08-07 sweep ran under one of our two paid-prober identities ([merona-payprobe/0.1](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/merona-payprobe/0.1), the live genuine-paid prober, or [x402-endpoint-quality-audit/0.1](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/x402-endpoint-quality-audit/0.1), the sweep pilot — both registered with pays: true); you never saw either because you weren't in the denominator — consistent all the way down.

You shouldn't have had to infer any of this. So we've published a canonical, machine-readable registry of every outbound identity we run — purpose and pays/never-pays policy per UA, and our own conformance tests pin each listed UA to its literal occurrence in our source, so the page can't drift from what we actually send:

curl -s https://api.merona.io/.well-known/merona-crawlers.json
Attribution should be a lookup, not forensics.

TTL per class and verdict: agreed. We already run per-class (charged_unserved carries a 1-hour TTL against a 6-hour default, precisely so the next paid probe can clear a fixed seller); per-class-and-verdict is the right generalization, and your 58-failures-in-8-services concentration is the argument — a FAIL on a recoverable class should decay fast, a PASS shouldn't get to coast.

On subject vs direction — either name works; the important part, as you said, is that the field exists.

Last thing: your situation shouldn't require waiting for our crawler to find a newer catalog snapshot. The next sweep adds ARD/.well-known discovery alongside Bazaar, plus direct origin submissions into the denominator — each recorded under its own discovery_basis, so a submitted subject stays distinguishable from a crawled one (which also blunts the de-facto-allowlist risk you raised). Your 22 sub-$0.05 paths will be in it.

Donk338:
Thanks — `discovery_basis` carrying the snapshot digest is the right shape, and
shipping `not_evaluated` as a distinct status closes the gap I was worried
about. Two things back: one correction I owe you with receipts, and one small
thing that undercuts your own verifiability work.

**The correction: we are in Bazaar, and we were on 08-08.**

I checked before believing my own logs, and the discovery set didn't exclude
us. Walking the public CDP discovery index today by offset (`limit=1000`,
`offset` 0..total, exact host equality — the endpoint returns no cursor, so a
cursor loop silently stops after one page and an unpaged `limit=100` read
returns a confident false zero; that cost me three wrong answers before a right
one):

```
https://api.borisinc.com/paid/ping             $0.001
https://api.borisinc.com/paid/steam/top-sellers $0.005
https://api.borisinc.com/paid/steam/deals       $0.005
https://api.borisinc.com/paid/game/metadata     $0.005
https://api.borisinc.com/paid/twitch/hype       $0.02
```

Five resources, four `accepts` legs each — `eip155:8453`, Solana,
`eip155:137`, `eip155:42161` — every one at or under your $0.05 cap, Base leg
present on all of them. (A sixth row, `payanagent.com/x402/kh7ce…`, is a
third-party proxy of our `/paid/ping`, so treat it as ours-adjacent rather than
ours.) These have been listed since 2026-07-22, so unless the index moved
sharply and moved back, they were there on 08-08.

Which means the "you weren't in the denominator" explanation doesn't hold, and
the original question is open again: we were in the catalog you drew from, we
qualified on price, and no payment arrived. Your own new machinery is the
clean way to settle it rather than us trading logs — check
`sha256:8ef8a87b…` for those five exact resource strings. If they're absent
from your snapshot but present in the index, the gap is in the snapshotting,
not the seller; if they're present, the sweep filtered somewhere past
discovery.

**On the count, lightly, because it may just be a units thing.**

The index declares **15,207** resources right now, flat across sampling, and no
`type` / `network` / `includeInactive` / `all` parameter changes that total —
there's no larger view behind a flag. Summing every `accepts[]` entry across
the corpus gives **~30,100** payment-requirement legs. Your 32,875 is nowhere
near the resource count but within about 9% of the leg count, so my guess is
the snapshot counted payment requirements rather than resources. Worth pinning
down, since "32,875 resources" and "502 sellers" are doing load-bearing work in
the same sentence and they may not be the same denominator.

**The small thing:** every user-agent in your comment is linked to
`github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/<ua>`,
and all of those 404 for me — the repo doesn't resolve. Flagging it only
because it lands right next to "attribution should be a lookup, not
forensics," which I agree with: `https://api.merona.io/.well-known/merona-crawlers.json`
does resolve, and it matches your comment exactly — three UAs listed as yours
with `pays: no`, both paid-prober identities listed `pays: yes`, and
`pulse-monitor/0.1` correctly absent, which is how I could confirm it isn't
yours rather than take your word for it. That file is the citation. Point at it
and drop the repo links.

Publishing a machine-readable crawler registry with a pays/never-pays flag per
identity is a genuinely useful thing to have done, and I'd support the spec
referencing that pattern — a `delivery` evidence provider asserting it pays
should be checkable against something, and right now yours is the only one that
is.

jywy78kgrf-create:
@Donk338 Confirmed: sha256:8ef8a87b… can't contain your five resources — the newest last_updated in that snapshot is 2026-07-03T03:57:14.654Z, and it was never refreshed before the 08-07 sweep. Your 07-22 listings were outside the denominator by construction. The gap is in snapshotting, as you guessed.

What shipped, all live:

discovery_basis now publishes catalog_max_last_updated, read from the rows themselves. The existing snapshot_date comes from file mtime, which is deployment-dependent — two checkouts of the byte-identical file stamp different dates — so the new field is the one that's deterministic for a given digest, and a data date trailing a sweep date is now visible to the consumer, not just to us. Documented in /v1/spec.
The collector refuses to silently reuse a cached discovery page past a freshness window (the cache was keyed to a pinned snapshot label — the latent bug that would have made even a re-pull a silent no-op), and prints the newest lastUpdated after every pull with a loud warning when it trails the run date.
Snapshot labels now default to the run's UTC date; pinning stays possible for reproducing an old pull, which is what pinning was for.
Each landed with a regression test pinning the exact failure shape: old data under a fresh file date must surface both values.

The count — units, as you guessed. That census file carries one row per accepts leg; 32,875 was wc -l (32,874 rows + header). Parsed properly, the same snapshot holds:

32,874 payment-requirement legs
22,954 distinct resource URLs
1,213 distinct hosts
1,062 distinct payout addresses
Correcting the leg-count-labeled-as-resources where it's published; the collector now prints both denominators on every run. Drift runs both ways, for anyone else consuming that index: 22,954 distinct resources in the July snapshot vs 15,207 declared today — missing recent listings like yours, carrying since-delisted rows.

Links: they 404 because that repo is private; https://api.merona.io/.well-known/merona-crawlers.json is the citation and I'll point there.

Re-probe: you qualified on price with a Base leg on all five. Once we've re-pulled against the live index I'll re-run the sweep on a current snapshot; you should land in it, and the result publishes under the same mechanical classification as everyone else. Say the word if you'd rather be excluded.

And taking you up on the registry point — I'll draft a spec section against draft-02: an identity list with per-UA pays: yes|no at a stable well-known path, so a delivery provider asserting it pays is checkable rather than trusted.

Donk338:
Yes — sweep us. No exclusion, publish whatever it finds under the same
classification as everyone else. If we fail, that verdict is worth more to the
spec discussion than a clean one.

`catalog_max_last_updated` is a better fix than the one I asked for. I'd have
settled for the digest; reading the max `last_updated` out of the rows catches
the case I didn't think of, where a snapshot is byte-identical but its file
mtime isn't — and a data date trailing a sweep date is exactly the failure that
produced this whole thread. Worth putting in the spec text rather than leaving
as an implementation detail: **a delivery-evidence entry should carry the age of
the data its subject set was drawn from, not just the identity of the
snapshot.** Same argument as settlement-as-precondition — it's the field that
tells a consumer whether an absence means anything.

On the count, thanks for running it down. 32,874 legs / 22,954 distinct
resources against 15,207 declared today is the more interesting number in your
reply than the one it corrected — that index has lost roughly a third of its
resources in five weeks while we've all been citing its size as a proxy for
ecosystem growth. If you publish the parsed denominators per snapshot, that
becomes a time series nobody currently has.

**Two things from our side before you re-run.**

First, the useful one. Our delivery record is 66 of 66 settled calls returning
HTTP 200 — zero non-2xx after settlement, lifetime. But when I went to verify
that claim properly ahead of your sweep, I swept our own 402 challenges for the
first time and found eleven defects: nine paid paths advertised a
`try_free_first` URL that returned **422** because the sample route took a
required query param and the pointer was generated bare, and our two most
expensive SKUs served a challenge with an empty `description` and no
`what_this_does` at all. Every one of those had been live since the endpoints
shipped. Fixed and verified this morning — 24 of 24 paths now carry a populated
challenge and a working free twin.

I'd rather say that publicly than quietly, because it is a datum about the
mechanism you're specifying: **the announcement of a paid sweep did more for our
conversion surface in a day than 140,000 unpaid challenges did in a month.** No
passive probe would have caught it — a 402 body with a broken pointer inside it
still scores as challenge-valid, still scores as live, still scores as
price-honest. It only becomes visible when someone intends to actually buy. That
is the strongest case I can make for why `delivery` belongs in the enum, and I
didn't have it until this week.

Second, in the interest of you not recording something misleading: we have a
known, recurring, bounded outage. `agent-rails` wedges for ~28 minutes
(measured: 27, 29, 28) and self-clears, and it recurs on a fixed internal
schedule — the next window is **2026-08-13 21:31 UTC** (07:31 Sydney the 14th).
Root-caused to a ledger connection left open on an error path, fix in flight. If
your sweep lands inside that window you will see a real outage and you should
record it as one; I'm telling you so you can date it rather than treat it as
steady state. If it's cheap to avoid, avoid it — otherwise take the reading.

On the crawler-identity spec section: yes, and I'll review the draft against
this thread's history if that's useful. The reason it works as a spec mechanism
rather than a courtesy is the thing your registry already demonstrates — I could
confirm `pulse-monitor/0.1` wasn't yours by reading your file rather than by
taking your word for it. A `pays: yes` claim that can't be checked against
anything is worth exactly as much as a self-attested trust score.

jywy78kgrf-create:
@Donk338 Sweep consent noted, and the no-exclusion stance goes in the record like everyone else's.

On the outage window — appreciated, and the disclosure is exactly the right move. We'll take the reading rather than avoid it, mostly on method: once we schedule around one seller's declared window we'd owe the same to every seller, and the sweep stops sampling reality. But the disclosure isn't wasted: if a reading lands inside 2026-08-13 21:31–21:59 UTC, the published record links this comment — a pre-disclosed, dated, root-caused, bounded outage is a materially different fact than a silent failure, and the record will show the difference. Honestly, a probe landing mid-window might make the best entry in the dataset: it would demonstrate exactly how disclosed downtime should look in delivery evidence versus undisclosed rot.

Your eleven defects are the best datum this thread has produced: a 402 body with a broken pointer inside it passes every passive check ever run — challenge-valid, live, price-honest — and 140,000 unpaid challenges never surfaced what the announcement of one paid sweep did in a day. That's the delivery-enum argument in one sentence, and it's yours — I'd like to cite it, attributed, in the draft spec text.

Yes to both spec items; drafting against draft-02:

the crawler-identity registry section (stable well-known path, per-UA pays flag) — you'll get the draft to review against this thread's history;
your data-age point folded into delivery evidence: an entry carries the age of the data its subject set was drawn from — read from the rows, not the file — because that's the field that tells a consumer whether an absence means anything.
On the time series: agreed, it's the more interesting number. We hold dated catalog snapshots back to early July; we'll publish parsed denominators per snapshot — legs / distinct resources / distinct hosts / distinct payTos, per date — as a standing dataset next to the seller index, CC BY, re-derivable. If the index really shed a third of its resources in five weeks, that's a finding, not a footnote.

Sweep re-runs once our collector re-pull lands. You'll see [merona-payprobe/0.1](https://github.com/jywy78kgrf-create/agentic-commerce-index/blob/claude/finance-app-review-8dsfpa/merona-payprobe/0.1) — registry-listed, pays: yes — when it reaches you.

Donk338:
Yes to the attributed citation — use it, and if the wording is useful raw I'd
rather it read as the measurement than the slogan: nine of our own
`try_free_first` pointers returned 422 and two SKUs served an empty challenge,
all live since those endpoints shipped, none of it visible to any passive check
we or anyone else had run against them.

Your reason for not scheduling around the window is better than my offer to have
you avoid it, and I'll adopt it: once a sweep accommodates one seller's declared
window it owes the same to every seller, and then it isn't sampling reality. The
disclosure-plus-link mechanism is the right shape — that's the distinction
between disclosed downtime and undisclosed rot made mechanical rather than
rhetorical.

Correcting my own disclosure, since the facts moved after I posted it. I said the
fix was "in flight". It has since landed AND, more to the point, is now loaded:
the root cause was a ledger connection left open on an error path, the fix
committed 2026-08-12 08:52 AEST — 81 minutes after the most recent occurrence —
and the process serving `api.borisinc.com` was only restarted onto it at 10:45
AEST today. All three prior occurrences (08-05, 08-10, 08-12, each 27-29 minutes,
each self-clearing) ran without it.

So the 2026-08-13 21:31 UTC window may simply not fire, and I don't yet know. It
has never been tested against a process carrying the fix; Friday is the first
run that will. I'm leaving the disclosure standing rather than withdrawing it —
an untested fix is not a fixed bug, and if your probe lands in that window and
sees nothing, that is a cleaner result for both of us than if I'd quietly
retracted the warning first.

jywy78kgrf-create:
@Donk338 Using your wording verbatim and attributed — the measurement, not the slogan.

One correction on our side so you're not expecting a test we may not run: our re-sweep is imminent, likely before Friday, so it will probably land nowhere near 21:31 UTC on the 13th. The window may just not come up. If a reading does fall inside it, the disclosure-plus-link mechanism applies as described.

And a note that supports your own caution, from your dates: read back to UTC, your three occurrences sit at 08-04, 08-09, and 08-11 — five days apart, then two. That's not a fixed period, so a quiet Friday is weak evidence either way: could be the fix holding, could be the window simply not due. Your recurrence log over the next few cycles will settle it better than any single probe of ours, and I'd read one clean window as encouraging rather than conclusive if I were you.

What we publish either way: the probe timestamp to the second, next to the outcome, linked to this disclosure — so the reading is usable as evidence rather than a verdict, including by you, and including if it contradicts what we'd otherwise have recorded.

Spec draft next: the crawler-identity section (well-known path, per-UA pays flag) and the data-age field for delivery evidence, against draft-02, with your 422/empty-challenge finding cited as the motivating case for why delivery can't be inferred from passive checks. I'll post it here for review rather than opening a PR cold.

Donk338:
One correction to your correction, because it makes your reading more useful
rather than less.

You're right that in UTC the three occurrences are 08-04, 08-09, 08-11 — five
days then two, no fixed period. But that's a timezone artefact. In local time
they're 08-05, 08-10 and 08-12, all around 07:30, and those are Wednesday,
Monday and Wednesday. Our internal jobs run Mon/Wed/Fri local, and the wedge
onset lands 60–90 seconds after one specific job starts. The five-day gap covers
the Friday in between, when that job didn't run at all — a platform quota outage
took it out for four days. So it isn't calendar-periodic; it's periodic in
that job's schedule, which is why it looks irregular from outside.

Which flips your conclusion: **the 14th local is one of those days.** Whenever
your probe lands, the trigger will have fired that morning and the fix will be
loaded — the first time both have been true at once. A quiet result on the 14th
is therefore stronger evidence than you'd expect, not weaker.

Two honest caveats on my own claim, since I've been holding you to receipts.
I have direct log evidence tying the job start to the wedge for two of the three
occurrences; the third predates my log retention and I'm inferring it from the
schedule. And the control is a single one: a *different* job ran the identical
07:30 slot on a non-trigger day with no wedge. Two confirmed, one inferred, one
control — a strong signal, not a proven mechanism, and I'd rather say that than
round it up.

None of which should change what you do. Take the reading whenever the sweep
lands and publish the timestamp next to the outcome as you described; our own
recurrence log over the next few cycles will settle the mechanism regardless.
I'm flagging it only so that if you do catch a clean window on the 14th, you
know it's worth more than one sample of an irregular event — and if you catch a
wedge, that's a live one with a fix already loaded, which is the more
interesting outcome for both of us.

jywy78kgrf-create:
Good catch, and you're right — I read the dates in UTC and missed that they're Mon/Wed/Fri local against your job schedule. Noted on the 14th being a trigger day with the fix loaded. Sweep runs when it runs; you'll get the timestamp next to the outcome either way.

jywy78kgrf-create:
Swept you: api.borisinc.com/paid/ping — delivered, HTTP 200, payment settled. Your 66/66 holds under our probe. Full run: 218/664 delivered (32.8%), against 29.9% in the August sweep — a replication on a fresh, larger population. 14 charged-unserved endpoints, but concentrated across only 5 operators (one accounts for ten), which is the honest way to state it. Timestamps and per-endpoint results publish with the dataset.

OceanAlt666:
Sharing a **regulatory (compliance-category) trust provider** implemented against this interface, in case it's useful while the fields are still settling. Runnable reference: https://github.com/OceanAlt666/x402-trust-provider (`node oceanalt-trust-provider.mjs` prints clean / tainted / unavailable against a live endpoint).

Building it surfaced a few suggestions for `TrustEvaluation` that might be worth folding in:

- **`score: null` + `null_reason`** — a compliance provider must not fabricate a score when it can't evaluate (upstream unavailable, unknown subject). It should return `UNCERTAIN` with `score: null`, not a made-up number. Fail-closed by construction.
- **`category` (alongside `evidenceType`)** — so aggregation can express "require at least one regulatory PASS." Behavioral and regulatory checks compose: a payment can pass one and fail the other, and the aggregator needs to tell them apart rather than averaging them into one score.
- **`basis[]`** — structured, addressable evidence (`{ class, kind, source, ref, url, observed_at }`) rather than a single opaque `evidence_uri`, so a settlement decision stays auditable after the fact. Keep `evidence_uri` as a convenience pointer.
- **`direction` (payee / payer)** — a compliance screen looks at the counterparty/payee address; a behavioral screen looks at the payer. Aggregators need to know which side each provider actually evaluated.
- **`issued_at` / `expires_at` / TTL**, plus a note that **`FAIL` should require sufficiently strong, verifiable evidence** (sanctions / mixer / issuer-freeze), not a soft score crossing a threshold — otherwise fail-closed and FAIL collapse into the same signal.

Framing note on wording: I'd suggest treating this as "compliance risk screening," not "full AML compliance" — a PASS means the provider found no blocking reason, which is not the same as "safe." Happy to align the reference exactly with wherever `types.ts` ends up, and to open a PR once the extension lands.

Donk338:
Hi @vdineshk — thanks for this extension. We run a live trust provider (Aegis, aegis.borisinc.com) and have implemented this exact wire shape end-to-end, so I wanted to offer two concrete things:

1. **A working reference provider.** `POST /trust/evaluate` (and GET) returns `x402-trust-evaluation-v0.1` evaluations today, in two modes — "resource" (is this seller safe to pay? backed by paid delivery spot-checks, not just coverage counts) and "payer" (should a seller accept this wallet? conservative on-chain heuristics with honest UNCERTAIN on unknown). First evaluation per caller is free; it's $0.005 via x402 after that. Happy to have it listed alongside dominion-observatory as a second example provider in the docs, if that's useful.

2. **A Python port.** The extension is TypeScript-only right now. We've built a wire-compatible Python port — query → evaluation, strict/quorum aggregation policies, fail-open/fail-closed, plus an advisory observation mode that records evaluations without ever blocking settlement. I'd be glad to contribute it to the Python SDK (`python/x402/extensions/`) once the TS extension lands.

Would either be welcome?

Issue #834, opened 2025-12-20
[Proposal] escrow scheme using Base Commerce Payments Protocol


PancheI:
## Summary

[Agentokratia](https://agentokratia.com) is building a marketplace for AI agent services and is looking for community input on a proposed x402 scheme. We plan to offer session-based payments that enable agents to authorize funds once and use them across multiple API calls - reducing signatures and gas costs for high-frequency interactions.

Built on the [Base Commerce Payments Protocol](https://github.com/base/commerce-payments), this solution uses audited on-chain escrow contracts with the auth/capture pattern. We propose adding an `escrow` scheme to x402 v2 that decouples authorization from settlement, enabling session-based payments, batched settlement, variable pricing, and exact settlements through a single primitive.

## Problem

The `exact` scheme works well for one-off payments, but creates friction for high-frequency use cases:

**AI agents** making hundreds of API calls must sign every request. **Micropayments** become impractical when gas costs exceed the payment itself. **Interactive sessions** like chatbots or code execution require constant wallet interaction.

Consider an agent making 1,000 API calls at $0.01 each. With `exact`, that's 1,000 signatures and 1,000 on-chain transactions. The overhead defeats the purpose.

## Solution

Credit cards solved this decades ago with the **auth/capture pattern**: authorize funds upfront, track usage, capture what's earned, return what's unused. The [Commerce Payments Protocol](https://blog.base.dev/commerce-payments-protocol) brings this pattern on-chain with audited escrow contracts on Base.

We propose wrapping this in an x402 scheme called `escrow`:

1. **Authorize** - Client signs once, funds locked in escrow contract
2. **Use** - Client makes requests using a session ID (off-chain, no gas)
3. **Capture** - Facilitator settles earned amount (batched)
4. **Void** - Unused funds returned to client

For that same 1,000 API calls: **1 signature, 2 transactions**.

## How It Works

```mermaid
sequenceDiagram
    participant Client
    participant Server
    participant Facilitator
    participant Escrow

    Client->>Server: GET /resource
    Server-->>Client: 402 + PaymentRequired

    Note over Client: Signs ERC-3009 (once)

    Client->>Server: PaymentPayload
    Server->>Facilitator: authorize
    Facilitator->>Escrow: authorize()
    Facilitator-->>Server: sessionId
    Server-->>Client: 200 + X-SESSION-ID

    loop Off-chain usage
        Client->>Server: X-SESSION-ID + X-SESSION-SIG + X-SESSION-TS
[...]

PR #2308, opened 2026-05-14
auth-capture TS sdk

A1igator:
# auth-capture EVM Scheme — TypeScript SDK

At [x402r.org](https://x402r.org), we've been working on refundable payments for x402. This PR is the TypeScript implementation of the `auth-capture` scheme whose spec was merged in #1425. It builds on the audited [base/commerce-payments](https://github.com/base/commerce-payments) contract stack, so no new contracts ship with this PR.

## Scheme implementation: key design decisions

`auth-capture` is single-shot and stateless. Each paid request signs one authorization (ERC-3009 or Permit2) carrying universal canonical collector + escrow addresses. The facilitator then submits a single tx that invokes `AuthCaptureEscrow.authorize()` or `AuthCaptureEscrow.charge()` (per `extra.autoCapture`), either calling escrow directly or routing through a passthrough `captureAuthorizer` contract.

Notable choices:

- **`captureAuthorizer` is merchant-set, never facilitator-injected** — the facilitator's `getExtra()` returns `undefined` for this scheme. Per spec, `captureAuthorizer` is the only address allowed to call `authorize` / `capture` / `void` / `refund` / `charge` against the escrow, gated on-chain by `onlySender(paymentInfo.operator)`. That address can be the facilitator's submitter EOA, or a smart contract that ultimately calls escrow as `msg.sender` (e.g., a refund arbiter). The facilitator advertises its submitter address via `getSigners()` so merchants can decide.
- **EOA vs. contract `captureAuthorizer` auto-detection** — the facilitator probes `getCode(captureAuthorizer)`. Empty bytecode → EOA path, direct call to escrow with `msg.sender == captureAuthorizer`. Non-empty → contract path, call routed through the contract.
- **`assetTransferMethod` is per-request** — defaults to `"eip3009"`; set to `"permit2"` for tokens that lack ERC-3009 (e.g., BSC's Binance-Peg USDC, some L2 USD stablecoins). The server does not auto-pick based on chain; the merchant declares the method explicitly.

## Spec observations from implementation

Small clarifications a follow-up doc pass could pick up:

1. **Verify step 13 / Settlement step 5.** `captureAuthorizer` can be the facilitator EOA (call goes to escrow directly) or a passthrough contract (call goes to the contract, which forwards to escrow). The spec's "call `AUTH_CAPTURE_ESCROW.authorize/charge`" wording is technically correct for both since the contract path reaches escrow underneath, but a revert in the contract case can originate in the passthrough before reaching escrow.

   Relatedly, `invalid_capture_authorizer` (escrow's `InvalidSender`) only fires when `onlySender(operator)` fails — i.e., when the address that called escrow (the facilitator EOA or passthrough contract) doesn't match `operator` (= `extra.captureAuthorizer` from PaymentInfo).

   Spelling these out in the verification/settlement logic steps would make the contract path easier to reason about.

2. ~~**Scheme names are inconsistent.** Existing schemes are lowercase or kebab-case (`exact`, `upto`, `batch-settlement`); `authCapture` is camelCase. This breaks tooling that assumes one case.~~

   ~~The e2e CLI `--schemes` filter, for instance, lowercases user input and silently dropped all authCapture matches — the case-insensitive compare added in this PR (`e2e/src/cli/filters.ts`) works around it.~~

   ~~Picking one convention and applying it backwards (e.g., renaming `authCapture` to `auth-capture` if kebab-case wins) would make scheme identifiers uniform.~~

   **Resolved** in #2391 — scheme renamed to `auth-capture`. The filter workaround has been removed from this PR.

3. **Reference links are dead.** `scheme_auth_capture.md` lines 84-85 link to issues on the legacy `github.com/coinbase/x402` repo that no longer resolve; canonical repo is `github.com/x402-foundation/x402`.

   The org segment should be swapped to `x402-foundation` to fix them.

## What's included

- **`@x402/evm/auth-capture/{client,server,facilitator}`** — new subdirectory under the existing `@x402/evm` package, alongside `exact/`, `upto/`, and `batch-settlement/`. Three new subpath exports added to `tsup.config.ts` and `package.json`.
- **Unit tests** — 5 files under `test/unit/auth-capture/`, covering client payload construction, server `parsePrice` / `enhancePaymentRequirements`, facilitator verify + settle paths for both EIP-3009 and Permit2, nonce / salt derivation, type guards, and the EOA vs. contract `captureAuthorizer` routing. All 735 package tests pass.
- **Integration test** — `test/integrations/auth-capture-evm.test.ts`, Base Sepolia live-testnet style matching the `batch-settlement-evm.test.ts` pattern. Skips with `describe.skip` unless `CLIENT_PRIVATE_KEY` and `FACILITATOR_PRIVATE_KEY` are set. Covers ERC-3009 + Permit2 × autoCapture: false / true.
- **Examples** under `examples/typescript/{clients,servers,facilitator}/auth-capture/` — three standalone apps mirroring the batch-settlement example layout.
- **E2E harness wiring** — scheme registered on the proxy, express, fastify, hono servers and on the TypeScript facilitator. Route files added under `e2e/servers/next/app/api/auth-capture/evm/{eip3009,permit2}/withx402/`. Endpoint entries added to test configs. `PaymentScheme` type extended in `e2e/src/types.ts` and `PaymentSchemeKind` in `e2e/src/cli/filters.ts`. `EVM_AUTH_CAPTURE_CAPTURE_AUTHORIZER` declared in optional env vars, with conditional registration: when unset, auth-capture routes are skipped (matching the `EVM_RECEIVER_AUTHORIZER_PRIVATE_KEY` pattern used by batch-settlement). `AuthCaptureEvmScheme` also registered on the fetch and axios e2e clients alongside the existing schemes.
- **NOTICE** — paragraph added crediting the x402r team (proposal issue #1011, spec PR #1425) and propagating the MIT attribution for the Agentokratia "x402-escrow" proposal (issue #834) whose code the implementation incorporates.
- **Changeset fragment** — `@x402/evm`: minor, "Implemented auth-capture mechanism".

## Disclosure

This PR was written with AI assistance (Claude Code). Output was reviewed before submission per the AI guidance in `CONTRIBUTING.md`.

## Links

Spec: #1425
Original proposals: #1011, #834

## Tests

* Unit tests (735/735 passing)
* Integration tests (4/4 passing on Base Sepolia, ERC-3009 + Permit2 × autoCapture: false / true)
* Examples flow exercised end-to-end on Base Sepolia (facilitator + server + client; real settle tx confirmed)
* E2E harness (4 server frameworks × 2 transfer methods = 8 scenarios, 8/8 passing on Base Sepolia for express / fastify / hono with fetch and axios clients; `next` server skipped due to a pre-existing Next 16 + Aptos SDK transitive build issue affecting all schemes, not specific to this PR — not surfaced by CI since `.github/workflows/` only runs unit tests, no e2e workflow exists upstream)

## Checklist


typescript/packages/mechanisms/evm/src/authCapture/server/scheme.ts
```diff
@@ -0,0 +1,315 @@
...
+ *
+ * `name` / `version` are the EIP-712 domain used by the ERC-3009
+ * `assetTransferMethod`. Whether a token supports ERC-3009 is a token-level
+ * capability, not a chain property; merchants whose chosen token lacks
+ * `receiveWithAuthorization` (e.g., BSC's Binance-Peg USDC, Tempo's pathUSD)
+ * MUST set `assetTransferMethod: "permit2"` in `extra` themselves. The server
+ * does not auto-pick a method based on chain. If the wrong method is paired
+ * with an incompatible token, the failure surfaces at facilitator simulation.
+ */
+const ASSET_INFO: Record<
```
phdargen: please remove, this should use the defaultAsset list instead (`import { getDefaultAsset, type ExactDefaultAssetInfo } from "../../shared/defaultAssets";`)

typescript/packages/mechanisms/evm/src/authCapture/server/scheme.ts
```diff
@@ -0,0 +1,315 @@
...
+ * string manipulation. Avoids the floating-point rounding errors that arise
+ * from `BigInt(Math.round(amount * 10 ** decimals))` on large or precise
+ * inputs. Example: `"0.10"` with `decimals=6` → `"100000"`.
+ *
+ * @param decimalAmount - Decimal amount expressed as a string (e.g. `"0.10"`).
+ * @param decimals - Token decimals (USDC = 6, most ERC-20s = 18).
+ * @returns The amount in base units as a string.
+ * @throws If `decimalAmount` does not parse as a number.
+ */
+function convertToTokenAmount(decimalAmount: string, decimals: number): string {
```
phdargen: use `import { convertToTokenAmount, numberToDecimalString } from "@x402/core/utils";` instead

examples/typescript/servers/authCapture/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            captureDeadline,
+            refundDeadline,
+            // address(0) lets the captureAuthorizer pick any non-zero fee
+            // recipient at capture/charge time (deferred selection, per spec).
+            // Production setups can pin a specific receiver instead.
+            feeRecipient: zeroAddress,
+            minFeeBps: 0,
+            maxFeeBps: 100,
+            name: "USDC",
+            version: "2",
```
phdargen: this info should be added internally from defaulAsset list

examples/typescript/servers/authCapture/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  console.error("Missing FACILITATOR_URL");
+  process.exit(1);
+}
+
+// Static deadlines: every authorization created by this server expires at
+// the same absolute Unix second. Production servers typically compute these
+// per request via custom middleware; static deadlines suffice for the demo.
+const now = Math.floor(Date.now() / 1000);
+const captureDeadline = now + 30 * 86400;
+const refundDeadline = now + 60 * 86400;
```
phdargen: this should be done in `enhancePaymentRequirements`, so its fresh for each request

typescript/packages/mechanisms/evm/src/auth-capture/facilitator/scheme.ts
```diff
@@ -0,0 +1,629 @@
...
+   * EIP-6492 unwrap), nonce binding to the payer-agnostic PaymentInfo hash,
+   * and an on-chain `simulateContract` of `authorize` / `charge` so typed
+   * escrow reverts surface as stable invalidReason strings.
+   *
+   * @param payload - The wire payload from the payer.
+   * @param requirements - The server's published payment requirements.
+   * @param _ - Unused FacilitatorContext (interface compatibility).
+   * @returns A `VerifyResponse` with `isValid` and, on failure, a stable `invalidReason`.
+   */
+  async verify(
```
phdargen, line 185: For the permit2 branch we should add support for the ERC-20 approval gas sponsoring extension

phdargen: Is there a base commerce deposit collector that supports EIP-2612?

phdargen:
Thanks a lot for the contribution @A1igator, made a first pass and looks good! Please have a look at my comments above

typescript/packages/mechanisms/evm/src/auth-capture/facilitator/scheme.ts
A1igator: I don't believe so. I can build one and include here/contribute to base commerce-payments repo but not sure how audits and stuff work for that.

typescript/packages/mechanisms/evm/src/authCapture/server/scheme.ts
A1igator: done!

typescript/packages/mechanisms/evm/src/authCapture/server/scheme.ts
A1igator: done!

examples/typescript/servers/authCapture/index.ts
A1igator: done!

examples/typescript/servers/authCapture/index.ts
A1igator: Added *Seconds variants of them that are filled by enhancePaymentRequirements. Let me know if that was not the intention.

avidreder:
@A1igator thanks so much for your contribution in getting this built out.

I am working on a follow up PR for the spec, which adds an affordance for users running servers to call other methods on the escrow contract (void, refund, etc). The core idea is that we would maintain the facilitator interface of `verify` and `settle`, and use a `type` field to distinguish operations. I think this is a critical piece of the SDK, so I'd like to hold off on merging this until we can get that piece finalized.

We're doing a round of internal review before bringing the spec update out of draft, but I'd love to loop you in once that is complete to get your feedback, so we can align this PR with the spec post-update.

A1igator:
@avidreder gotcha. I saw the draft before seeing your comment here and already made a comment. I will await it getting out of draft though to comment further but yeah I think this is going to be really difficult to do properly (trust minimized, good UX, extendible/flexible, etc...) as part of x402 standard itself, and it can easily be solved outside the standard since we already do so. Open to seeing alternative designs though!

avidreder:
Sounds good, appreciate the feedback, and I'll tag you in shortly.

typescript/packages/mechanisms/evm/src/auth-capture/constants.ts
```diff
@@ -0,0 +1,43 @@
...
+export const AUTH_CAPTURE_SCHEME = "auth-capture" as const;
+
+// Canonical AuthCaptureEscrow + token collector deployments from
+// base/commerce-payments (https://github.com/base/commerce-payments). These are
+// the audited, live addresses listed in the upstream README — the source of
+// truth for this scheme. Universal constants — not configurable per merchant.
+//
+// Currently live on: Base (8453) and Base Sepolia (84532). Additional EVM chains
+// will land at the same addresses as the upstream extends coverage; expand the
+// supported-chain list here as those deployments ship.
```
phdargen: please remove this comment

examples/typescript/servers/auth-capture/README.md
```diff
@@ -0,0 +1,42 @@
...
+| `FACILITATOR_URL`    | Yes      | —       | Base URL of the auth-capture facilitator (POST `/verify`, POST `/settle`).                                                                                          |
+| `PORT`               | No       | `4021`  | Local listen port.                                                                                                                                                 |
+
+## Deadlines
+
+The example sets `captureDeadline` and `refundDeadline` once at boot, as absolute Unix seconds 30 / 60 days into the future. Every authorization the server hands out shares the same absolute deadline. Production servers commonly compute these per request via custom middleware so each authorization has a fresh window — out of scope for this minimal demo.
+
+## Lifecycle beyond authorize
+
+This example demonstrates the authorize phase only. Capture, void, and refund are the captureAuthorizer's responsibility and are not handled by this server; refer to the [scheme spec](../../../specs/schemes/auth-capture/scheme_auth-capture_evm.md) for the protocol-level surface.
```
phdargen, line 42: hm this isnt a great devx experience as it just leaves the deposit stranded in the escrow and then one has to go to etherscan to capture

How about we have a self-facilitation example similar to https://github.com/x402-foundation/x402/tree/main/examples/typescript/servers/self-facilitation, where CAPTURE_AUTHORIZER is set to the server signer such that it can call the escrow.capture directly to demonstrate the full lifecycle?

typescript/packages/mechanisms/evm/src/auth-capture/facilitator/scheme.ts
```diff
@@ -0,0 +1,629 @@
...
+
+    const settleTarget = await this.resolveSettleTarget(extra.captureAuthorizer);
+
+    try {
+      const txHash = await this.signer.writeContract({
+        address: settleTarget,
+        abi: ESCROW_ABI,
+        functionName,
+        args,
+      });
```
phdargen, line 441: I don't think we can allow calling any untrusted contracts here, at least not without further checks. Simulation only proves the tx doesnt revert but not that it implements the intended logic or that it even calls the escrow contract internally. At the minimum needs a cap on gas to limit facilitator griefing. 

Do we need an allowlist @avidreder?

A1igator: Allowlist is a bad idea since every new arbiter would be at a different contract.

Gas check and further checks is a great idea. I was thinking about it anyways. Will scope it out.

phdargen: For the checks was thinking eth_simulateV1 might be useful to parse event logs and trace asset changes, 
https://viem.sh/docs/actions/public/simulateCalls#simulatecalls

avidreder:
Hey @A1igator, we are looking to ship an implementation using the Auth Capture client you have in this PR later this week. Do you think you'll be able to address @phdargen comments soon?

I'm happy to implement just the client side if it helps, but would prefer that you stay primary author since you've brought it so far. Let me know if you want me to pick that up, or if you're able to separate out the client portion while you work through the rest.

Thanks!

CC: @fabrice-cheng

A1igator:
@avidreder sorry slightly confused do you want the whole PR or just the client side? Not sure what just the client side implementation entails without the rest.

If the whole thing is wanted, I was thinking of a) sending a spec update to address the smart contract simulation side of things first and b) figuring this out: https://github.com/x402-foundation/x402/pull/2359#issuecomment-4514851330 since if we're deleting facilitator EOA side we should do it before shipping. Especially if a full life cycle example is wanted, it'd make more sense to do it with the server if that functionality is being added. This would also need a spec update obviously.

I can do a and address Duke's comments in the implementation probably in the same week this week but b is probably going to be a larger task.

avidreder:
TLDR, we have a server implementation we are looking to launch, which doesn't require a facilitator, but we want agents using the x402 packages to know how to sign for the auth-capture schema, with the caveat that full server/facilitator SDK support would follow.

I'm only talking me opening a PR for the client side SDK, which just covers the detection and signing of payloads for auth-capture.

A1igator:
Gotcha does that include type=charge/authorize that Duke asked for in #2359? Should the other types also be included? I think that's the only relevant part of #2359 for a client only implementation but I'll see if I run into something else.

Edit: Or I might just be misunderstanding client side but I presume client needs to know the shape of the possible payloads to detect and sign for it?

A1igator:
I did a bit more research and I think type/autoCapture is not relevant to the client side and worst case it would be purely additive anyways. I should be able to move it over. Let me know if there's anything I should be mindful of though.

typescript/packages/mechanisms/evm/src/auth-capture/constants.ts
A1igator: Done!

A1igator:
@avidreder here you go: https://github.com/x402-foundation/x402/pull/2486 ! 

We would really appreciate a heads-up for #2359 full plan though especially to unblock the rest of this PR.

examples/typescript/servers/auth-capture/README.md
A1igator: I'd appreciate finalization of #2359 before I do this as it'd make more sense to use that as an example if the server is getting the power to capture without the facilitator.

phdargen:
Continued in https://github.com/x402-foundation/x402/pull/3203

PR #2317, opened 2026-05-15
docs: add MCP payment policy guidance

EfeDurmaz16:
## Summary

Adds production guidance for MCP clients to check payment requirements before creating or signing a payment payload.

The current guide auto-approves payments for local testing. This adds a small `onPaymentRequested` policy example that checks tool name, network, and amount before approving.

## Changes

- Adds a `Policy Checks Before Payment` section to the MCP guide
- Shows a minimal deny-by-default `onPaymentRequested` example
- Clarifies that spend controls should live in code near the payment client, not only in model prompts

## Verification

- `git diff --check main...HEAD`

### DIFF
--- a/docs/guides/mcp-server-with-x402.md
+++ b/docs/guides/mcp-server-with-x402.md
@@ -200,6 +200,42 @@ The MCP server exposes a tool that, when called, fetches data from a paid API en
 
 ***
 
+### Policy Checks Before Payment
+
+This example auto-approves payments for local testing. In production, the MCP
+bridge should check the payment requirements before creating or signing a
+payment payload. The client hooks are the right place to enforce those rules.
+
+At minimum, compare the selected requirement against your expected tool call:
+
+* maximum amount and currency
+* expected `network` and `scheme`
+* expected resource server or facilitator
+* tool name and user-requested operation
+* per-user, per-agent, or per-session spend limits
+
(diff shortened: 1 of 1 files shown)


TateLyman:
This is the right production direction. I would tighten the example slightly so readers do not copy an `accepts[0]` policy into production.

Suggested adjustment:

```ts
const requirement = paymentRequired.accepts.find((candidate) =>
  candidate.scheme === "exact" &&
  candidate.network === "eip155:84532" &&
  candidate.asset === expectedUsdcAsset &&
  candidate.resource === expectedResourceUrl
);

if (!requirement) return false;
if (BigInt(requirement.maxAmountRequired ?? requirement.amount) > 1_000n) return false;
if (requirement.payTo !== expectedPayTo) return false;
```

Reasoning:

- `accepts[0]` can silently approve the wrong rail when a server offers multiple schemes/networks.
- The policy should bind the selected requirement to the intended resource URL and recipient, not only the tool name.
- The amount comparison should stay in atomic units from the selected requirement; avoid converting through floating USD/USDC in the policy example.
- If the requirement contains expiry/deadline or facilitator metadata, this is also the right hook to reject stale or unexpected values before any wallet signing.

That keeps the docs aligned with the exact failure mode this section is trying to prevent: a model/tool call producing a valid-looking payment request that is not the payment the user intended.

phdargen (approved):

PR #2318, opened 2026-05-15
fix(mcp): preserve existing metadata

EfeDurmaz16:
## Summary

Preserves existing MCP `_meta` fields when x402 attaches payment request or settlement metadata.

MCP `_meta` can carry non-payment context such as trace IDs, auth hints, evidence handles, or other transport metadata. The current helpers and server wrapper replace `_meta` when adding x402 metadata, which can drop that context.

## Changes

- Preserve existing request `_meta` in `attachPaymentToMeta`
- Preserve existing result `_meta` in `attachPaymentResponseToMeta`
- Preserve handler-provided `_meta` when `createPaymentWrapper` attaches settlement response metadata
- Add unit coverage for all three preservation paths

## Verification

- `pnpm --dir typescript --filter @x402/mcp exec vitest run test/unit/utils.test.ts test/unit/server.test.ts`
- `pnpm --dir typescript --filter @x402/mcp lint:check`
- `git diff --check`

Note: the full `pnpm --dir typescript --filter @x402/mcp test` suite currently fails in `test/unit/client.test.ts` because `@x402/core/schemas` cannot be resolved from `src/client/x402MCPClient.ts` in this local workspace setup. The two test files covering this change pass.

### DIFF
--- a/typescript/.changeset/silver-mails-prove.md
+++ b/typescript/.changeset/silver-mails-prove.md
@@ -1,0 +1,5 @@
+---
+"@x402/mcp": patch
+---
+
+Preserve existing MCP response metadata when adding x402 payment metadata.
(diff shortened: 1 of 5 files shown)


phdargen (approved):
Thanks for the fix, looks also good

Issue #2323, opened 2026-05-15
channelId docstring + extra.assetTransferMethod inconsistency (from a Rust impl)


Adelagric:
# x402 spec — feedback from a Rust implementation

Two precise discrepancies found while implementing x402 V2 from the
spec, each verified by reproducing `viem`'s output byte-for-byte.

## 1. `voucher.ts` docstring misstates `channelId` derivation

- **Where**: `typescript/packages/mechanisms/evm/src/batch-settlement/client/voucher.ts`, parameter doc for `channelId`.
- **Says**: `channelId` is `keccak256(abi.encode(ChannelConfig))`.
- **Actual** (`.../batch-settlement/utils.ts::computeChannelId`):
  `hashTypedData({ domain, types: channelConfigTypes, primaryType:
  "ChannelConfig", message })` — a full EIP-712 hash
  (`keccak256(0x1901 ‖ domainSeparator ‖ hashStruct)`), bound to
  `chainId` and the `x402BatchSettlement` verifying contract.
- **Impact**: an implementer following the docstring derives a wrong
  `channelId`; all vouchers then fail verification with no obvious
  cause.
- **Suggested fix**: correct the docstring to reference the EIP-712
  typed-data hash (and point to `computeChannelId`).

## 2. `extra.assetTransferMethod` presence is inconsistent across the spec

- **Where**: `specs/schemes/exact/scheme_exact_evm.md` examples carry
  `extra.assetTransferMethod`; `specs/transports-v2/http.md`
  `PaymentRequired` example carries `extra` = `{name, version}` only.
- **Impact**: an implementer modeling `extra.assetTransferMethod` as
  required (reasonable from the `exact`/EVM examples) fails to
  deserialize the canonical transport `PaymentRequired`, i.e. fails
  on real 402 challenges.
- **Suggested fix**: state explicitly that `extra.assetTransferMethod`
  is optional in `PaymentRequired`, and align the examples.

Both were resolved in the Rust implementation by treating the EIP-712
hash as authoritative and `extra.assetTransferMethod` as optional;
golden tests decode the verbatim spec Base64 vectors and match
`channelId`, the `Voucher` digest, and recovered signer exactly.

Adelagric:
Confirmed against an independent Rust implementation (golden-tested byte-exact vs viem): the EIP-712 computeChannelId derivation is correct; keccak256(abi.encode(...)) was wrong. Fix tracked in #2325.

PR #2325, opened 2026-05-15
docs: clarify EVM channel and transfer method semantics

EfeDurmaz16:
## Summary
- Correct the batch-settlement voucher docstring for `channelId` so it points to the EIP-712 `ChannelConfig` hash used by `computeChannelId`
- Clarify that omitted `extra.assetTransferMethod` in EVM `exact` `PaymentRequired` means the default `eip3009` method
- Add the same defaulting note to the HTTP transport example, where `extra.assetTransferMethod` is intentionally omitted

Fixes #2323

## Verification
- pnpm --dir typescript/packages/mechanisms/evm exec prettier -c .prettierrc --check src/batch-settlement/client/voucher.ts ../../../../specs/schemes/exact/scheme_exact_evm.md ../../../../specs/transports-v2/http.md
- pnpm --dir typescript/packages/mechanisms/evm exec vitest run test/unit/batch-settlement/utils.test.ts
- pnpm --dir typescript/packages/mechanisms/evm lint:check
- git diff --check

PR #2333, opened 2026-05-15
docs(evm): clarify channel id and transfer method defaults

EfeDurmaz16:
## Summary
- fix the batch-settlement voucher `channelId` docstring to reference the EIP-712 `computeChannelId` derivation instead of `keccak256(abi.encode(ChannelConfig))`
- clarify that EVM `exact` `extra.assetTransferMethod` can be omitted from `PaymentRequired` for default method selection
- add notes to the v2 HTTP and core examples explaining why their `extra` objects include token domain metadata but no transfer-method selector

## Why
This addresses #2323. The previous wording could send external implementers toward an incompatible channel id derivation, and the EVM exact examples made `assetTransferMethod` look required even though canonical `PaymentRequired` examples omit it.

## Verification
- `git diff --check`
- `pnpm exec prettier -c .prettierrc --check src/batch-settlement/client/voucher.ts ../../../../specs/schemes/exact/scheme_exact_evm.md ../../../../specs/transports-v2/http.md ../../../../specs/x402-specification-v2.md` from `typescript/packages/mechanisms/evm`
- `pnpm --filter @x402/evm lint:check`

## Notes
- `pnpm --filter @x402/evm test -- --runInBand` currently fails locally before exercising this docs change because the package-level Vitest run cannot resolve workspace imports such as `@x402/core/http`, `@x402/core/client`, and `@x402/core/utils` in this checkout. The collected unit tests that did not hit those aliases passed.

### DIFF
--- a/specs/schemes/exact/scheme_exact_evm.md
+++ b/specs/schemes/exact/scheme_exact_evm.md
@@ -12,7 +12,7 @@ This is implemented via one of three asset transfer methods, depending on the to
 | **2. Permit2**      | Tokens without EIP-3009. Uses a Proxy + Permit2.             | **Universal Fallback** (Works for any ERC-20). | One-time use                        |
 | **3. ERC-7710**      | Smart accounts with delegation support.                              | **Smart Account Option** (Paid from ERC-7710 compatible account). | One-time use and multi-use |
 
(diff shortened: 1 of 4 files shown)


EfeDurmaz16:
Closing this as a duplicate of #2325. #2325 already addresses #2323 with the channelId docstring fix and EVM exact assetTransferMethod defaulting clarification; keeping one PR open is cleaner for review.

PR #1808, opened 2026-03-25
Add Extensions field to Go SettleResponse for parity with TypeScript SDK

javierpmateos:
## Description

The TypeScript SDK's `SettlementResponse` includes an `extensions` field (`Record<string, unknown>`) that enables extension data like the recently merged `offer-receipt` (PR #935). The Go SDK's `SettleResponse` struct does not have this field.

This PR adds `Extensions map[string]interface{}` with `json:"extensions,omitempty"` to the Go `SettleResponse` struct, bringing it to parity with the TypeScript SDK.

This is a non-breaking change — the field is `omitempty`, so existing serialization is unaffected. No existing tests reference the absent field.

Related: #1802

## Tests

cd go && go test ./...

All 22 test packages pass (tested locally on Go 1.22). The new field is `omitempty` and not referenced by any existing test, so no test changes are needed.

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [ ] My commits are signed (required for merge)
- [ ] I added a changelog fragment for user-facing changes

PR #2339, opened 2026-05-16
feat(core): add FacilitatorExtension.enrichSettleResponse hook

javierpmateos:
# feat(core): add FacilitatorExtension hooks for facilitator-side extension data

## Summary

Refs #1802. Adds `enrichSettleResponse` to the `FacilitatorExtension` interface and wires it
into `x402Facilitator.settle()` so that facilitator-registered extensions can inject signed or
derived data into `SettleResponse.extensions` before the response is returned to the resource
server.

### The Gap

Today's x402 extension flow covers the **resource server** side:

- `ResourceServerExtension.enrichPaymentRequiredResponse` ÔÇö injects data into 402 responses
- `ResourceServerExtension.enrichSettlementResponse` ÔÇö injects data into the final settlement response

But when the resource server calls the facilitator's `/settle` endpoint, the facilitator has **no
standardized way** to populate `SettleResponse.extensions`. The field exists in both the TypeScript
and Go SDKs (see PR #1808 which added it to Go), but there are no hooks to fill it on the
facilitator side.

This was identified by @alftom in issue #1802:

> "One missing piece seems to be implementing something like
> `HTTPFacilitatorClient.registerExtension` so facilitators can pass the payload back to the
> resource server."

This PR is the infrastructure fix. The first consumer will be the `facilitator-attestation`
extension (separate PR), which uses this hook to sign and return EIP-712 attestations.

### The Flow (before and after)

**Before:**
```
Client ÔåÆ ResourceServer ÔåÆ Facilitator /settle ÔåÆ bare SettleResponse
                                                  (no extension hooks; extensions always {})
```

**After:**
```
Client ÔåÆ ResourceServer ÔåÆ Facilitator /settle ÔåÆ SettleResponse
                                                  Ôåæ
                                        enrichSettleResponse hooks
                                        populate extensions field
```

### Relationship to offer-receipt (PR #935)

PR #935 added `ResourceServerExtension.enrichSettlementResponse` for the resource server boundary.
This PR adds the symmetric hook at the facilitator boundary, completing the extension pipeline:

```
[Facilitator] enrichSettleResponse ÔåÆ SettleResponse.extensions
    ÔåÆ [ResourceServer] enrichSettlementResponse ÔåÆ final response to client
```

### Relationship to Go SDK parity (PR #1808)

PR #1808 added the `Extensions` field to Go's `SettleResponse`. This PR adds the hook mechanism
that populates that field on the TypeScript facilitator side.

### `packages/core/src/types/extensions.ts`

- Added `FacilitatorSettleResultContext` interface (exported) ÔÇö the context passed to the hook,
  containing `paymentPayload`, `requirements`, and `result: SettleResponse`.
- Added `enrichSettleResponse?: (context: FacilitatorSettleResultContext) => Promise<unknown>`
  to `FacilitatorExtension`.

### `packages/core/src/facilitator/x402Facilitator.ts`

- In `settle()`: after `afterSettleHooks` complete, iterate registered extensions and call
  `enrichSettleResponse` for each. Extension data is merged into `settleResult.extensions[key]`.
- **Error isolation**: each hook is wrapped in `try/catch`; errors are logged with
  `console.error` and settlement continues ÔÇö a failing extension MUST NOT break the payment.

### `packages/core/test/unit/facilitator/x402Facilitator.extensions.test.ts` (new)

12 new unit tests covering the contract described in the Tests section below. All 468 existing `@x402/core` tests continue to pass.

### No `declaration` parameter (unlike `ResourceServerExtension`)

`ResourceServerExtension.enrichSettlementResponse(declaration, context)` takes a `declaration`
because the resource server has per-route extension declarations. Facilitator extensions are
global (registered once on the facilitator instance) so there is no per-request declaration.
The hook takes only `context`.

### `FacilitatorSettleResultContext` defined in `extensions.ts`

The context type is defined alongside the interface that uses it (in `types/extensions.ts`),
following the existing precedent where `SettleResultContext` is defined in `x402ResourceServer.ts`
and re-exported through `extensions.ts`. The local `FacilitatorSettleResultContext` in
`x402Facilitator.ts` (used for lifecycle hooks) is structurally compatible and remains unchanged.

### Hook placement: after `afterSettleHooks`, before `return`

Extension enrichment runs after all lifecycle hooks complete. This ensures the lifecycle hook
pipeline (e.g., audit logging in `onAfterSettle`) sees the un-enriched response, while the
enriched response (with signed extension data) is what gets returned to the caller.

### `enrichSettleResponse` is NOT called when settlement throws

If the scheme facilitator throws (e.g., on-chain tx failure), control jumps to the `catch` block
and extension hooks are skipped. Extensions only run on successful settlements.

### `enrichSettleResponse` IS called when scheme returns `success: false`

If the scheme returns a `SettleResponse` with `success: false` without throwing, the hook runs.
This is consistent with how `afterSettleHooks` behave. Extension implementations that only want
to run on successful settlements should guard with `if (!ctx.result.success) return undefined;`.

## Tests

| # | Description | Status |
|---|-------------|--------|
| 1 | `registerExtension` stores extension by key | Ô£à |
| 2 | `enrichSettleResponse` called with correct context on success | Ô£à |
| 3 | `enrichSettleResponse` return value appears in `SettleResponse.extensions[key]` | Ô£à |
| 4 | Multiple extensions run in registration order, both populate `extensions` | Ô£à |
| 5 | Hook throws ÔåÆ error logged, settlement still succeeds | Ô£à |
| 6 | Extension without `enrichSettleResponse` ÔåÆ `extensions` unchanged | Ô£à |
| 7 | Settlement throws ÔåÆ `enrichSettleResponse` not called | Ô£à |
| 8 | `extensions` field survives JSON round-trip | Ô£à |
| 9 | Context carries correct `txHash` and `network` from settlement result | Ô£à |
| 10 | End-to-end: registered extension adds extensions field | Ô£à |
| 11 | Hook returning `undefined` does not create empty `extensions[key]` | Ô£à |
| 12 | `success: false` result: hook called, extension can inspect and skip | Ô£à |

## Checklist

- [x] Adds `FacilitatorExtension.enrichSettleResponse` hook
- [x] Adds `FacilitatorSettleResultContext` context type (exported)
- [x] Hook called in `x402Facilitator.settle()` after successful settlement
- [x] Error isolation: failing hook logs + does not abort settlement
- [x] Hook NOT called on thrown settlement errors
- [x] `extensions` field preserved through Zod parse (existing `settleResponseSchema` already handles this)
- [x] All new tests pass (12 tests)
- [x] All existing `@x402/core` tests pass (468 total, 0 regressions)
- [x] `@x402/core` builds cleanly (TypeScript + DTS)
- [x] Backwards compatible: facilitators without `registerExtension` unaffected
- [x] Does not modify `offer-receipt` extension
- [x] Does not add `facilitator-attestation` extension (separate PR)

## What This Enables

Once this PR merges, a facilitator can produce signed attestations like:

```typescript
import { x402Facilitator } from "@x402/core/facilitator";

const facilitator = new x402Facilitator();

facilitator.registerExtension({
  key: "facilitator-attestation",
  enrichSettleResponse: async (ctx) => {
    if (!ctx.result.success) return undefined;
    return {
      attestation: await signAttestation(ctx.result, signingKey),
    };
  },
});
```

The resource server receives the enriched `SettleResponse` (via `HTTPFacilitatorClient.settle`),
and the `extensions["facilitator-attestation"]` data flows through to the client via the existing
`PAYMENT-RESPONSE` header.

### DIFF
--- a/specs/extensions/extension-facilitator-attestation.md
+++ b/specs/extensions/extension-facilitator-attestation.md
@@ -1,0 +1,276 @@
+# Facilitator Attestation Extension
+
+**Status:** Draft
+**Version:** 1.0
+**Extension Key:** `facilitator-attestation`
+**Placement:** `SettlementResponse.extensions["facilitator-attestation"]`
+
+---
+
+**1. Overview**
+
+The Facilitator Attestation Extension adds a **facilitator-signed settlement proof** to x402 payment responses. After a successful settlement, the facilitating party signs a `SettlementAttestation` object and attaches it to the `SettlementResponse.extensions` field.
+
+This extension is **complementary to the offer-receipt extension** (extension key `offer-receipt`). The two serve different trust roles:
+
+| Property | offer-receipt (receipt) | facilitator-attestation |
+|---|---|---|
+| **Signer** | Resource server | Facilitator |
+| **Signed at** | Service delivery | Settlement confirmation |
+| **Includes txHash** | Optional (privacy default: false) | Always (audit requirement) |
+| **Includes amount** | No | Yes |
+| **Includes facilitator fee** | No | Yes |
+| **Primary use case** | Proof of delivery | Audit / compliance / fee transparency |
+| **Chain binding** | EIP-712 domain chainId=1 (static) | EIP-712 domain chainId = payment chain |
+
+When both extensions are active, clients can compose a **BusinessReceipt** (§7) that provides both settlement proof and delivery proof.
+
+**2. Motivation**
+
+The offer-receipt extension deliberately omits amount and asset to preserve privacy and is signed by the resource server to prove delivery. This leaves a gap for actors who need:
+
+1. **Cryptographic proof of what was paid** — amount, token, payer, payee, in a form that cannot be forged or altered by the recipient.
+2. **Facilitator fee transparency** — verifiable record of what fee the facilitating party took.
+3. **Audit and compliance** — structured evidence for accounting systems, tax reporting, or regulatory filings.
+4. **Interoperability** — a machine-readable format compatible with emerging settlement proof standards (e.g., ERC-8183 Business Receipts).
+
+The facilitator is the natural signer for settlement proofs because:
+- The facilitator observes the on-chain transaction and can attest to its finality.
+- The facilitator's signature is independent of the resource server, providing an additional trust anchor.
+- Facilitators already hold signing keys for x402 operations.
+
+**3. Signed Artifact Structure**
+
+**3.1 Object Shape**
+
+The `SettlementAttestation` MUST have the following structure:
+
+```json
+{
+  "format": "eip712",
+  "payload": { ... },
+  "signature": "0x..."
+}
+```
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `format` | string | Yes | Always `"eip712"` for this extension |
+| `payload` | object | Yes | The canonical `SettlementAttestationPayload` fields (§4) |
+| `signature` | string | Yes | Hex-encoded ECDSA signature (`0x`-prefixed, 65 bytes: r+s+v) |
+
+This extension uses EIP-712 exclusively. JWS is not supported: EVM facilitator keys are secp256k1 keys, and EIP-712 is the natural signing format for on-chain-aware parties.
+
+**3.2 EIP-712 Domain**
+
+```javascript
+{
+  name: "x402-receipt",
+  version: "1",
+  chainId: <payment chain ID>
+}
+```
+
+**Unlike the offer-receipt extension**, the `chainId` in the EIP-712 domain is the **actual EIP-155 chain ID of the payment network** (e.g., `8453` for Base). This provides per-chain replay protection: an attestation for a Base payment cannot be presented as an Ethereum mainnet payment by tampering the `chainId` field.
+
+**4. SettlementAttestation Payload (§4)**
+
+**4.1 Fields**
+
+| Field | EIP-712 Type | JSON Type | Required | Description |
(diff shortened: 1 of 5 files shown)


alftom:
@javierpmateos do you have an extension spec to add to the PR?  It would go in 'specs/extensions` I would think. 

In my view this PR could cover two paths:

1. Information sent from the facilitator to the resource server that the resource server will put in the `extensions.offer-receipt.signature`
2. A self-contained payload signed by the facilitator and sent from the facilitator to the resource server that the resource server will relay to the client using some yet-defined mechanism (probably a separate `extensions` field rather than inside `extensions.offer-receipt`).

Given #1802 it looks like you're targeting #2 but it should be clear in a spec.

javierpmateos:
Thanks @alftom ... good call. I've added the extension spec at `specs/extensions/extension-facilitator-attestation.md`.

To clarify: this targets path #2 — a self-contained payload signed by the facilitator, relayed by the resource server to the client via a separate `extensions["facilitator-attestation"]` field. It does not modify or sit inside `extensions["offer-receipt"]`.

The two extensions compose but stay independent: offer-receipt proves delivery (server-signed), facilitator-attestation proves settlement (facilitator-signed). Different signer, different claim, different privacy surface — as @jithinraj recommended in #1802.

javierpmateos:
@phdargen, just a gentle reminder on this one, as it’s been assigned for a while. The PR is infrastructure-only (the facilitator-side hook), with the spec included as requested by @alftom. All 14/14 checks are green. Happy to rebase if it’s drifted from main. Let me know if anything needs adjusting.

PR #2340, opened 2026-05-16
fix(core): harden token amount validation

### LINKED ISSUE #2090: Harden `convertToTokenAmount()` input validation regex against ReDoS-style backtracking
`typescript/packages/core/src/utils/index.ts` exports `convertToTokenAmount(decimalAmount, decimals)`, which currently validates input with:

```
/^-?\d+\.?\d*$/
```

GitHub Advanced Security flagged this as a potential polynomial-time regex on untrusted string input in an exported utility. The finding appears directionally valid: the pattern has overlapping quantified digit groups (`\d+` and `\d*`) around an optional decimal point, which can cause excessive backtracking on long non-matching strings like many zeroes followed by a non-digit suffix.

This helper is used from multiple mechanism/server paths, so the alert fans out across several packages even though the risky regex lives in one shared utility. Fixing `convertToTokenAmount()` should address the whole alert cluster rooted at this validator rather than requiring package-by-package changes.

Current known affected call paths include:
- `typescript/packages/core/src/utils/index.ts`
- `typescript/packages/mechanisms/avm/src/exact/server/scheme.ts`
- `typescript/packages/mechanisms/evm/src/exact/server/scheme.ts`
- `typescript/packages/mechanisms/evm/src/upto/server/scheme.ts`
- `typescript/packages/mechanisms/stellar/src/exact/server/scheme.ts`
- `typescript/packages/mechanisms/stellar/src/utils.ts`
- `typescript/packages/mechanisms/svm/src/exact/server/scheme.ts`

In current call sites the input is not always fully attacker-controlled, so this is probably low practical risk today, but the function is exported utility code and should be hardened.

## Proposed fix

Replace the regex with an equivalent non-ambiguous validation pattern, for example:

```
/^-?(?:\d+(?:\.\d*)?)$/
```

or otherwise refactor validation to avoid ambiguous backtracking entirely.

As part of the same cleanup, we should also decide whether `convertToTokenAmount()` should continue accepting negative values at all. If negative token amounts are not a supported input contract, this ticket should tighten validation so negative values are rejected.

## Acceptance criteria

- `convertToTokenAmount()` no longer uses a regex shape vulnerable to polynomial backtracking.
- The shared fix covers the alert instances that currently trace through the mechanism/server call sites above.
- Existing valid inputs continue to work:
  - integers like `"1"`
  - decimals like `"1.5"`, `"0.10"`
- Existing invalid inputs still fail:
  - scientific notation like `"1e-7"`
  - empty / non-numeric strings like `""`, `"abc"`
- Negative values are explicitly addressed by the implementation and tests:
  - either they remain supported and are covered by tests
  - or they are rejected and tests/assertions are updated accordingly
- Unit tests cover accepted and rejected formats and preserve the intended behavior.

Out of scope: separate wildcard/route matcher hardening issues elsewhere in the repo. Those should be tracked independently if we decide to fix them.

FidelCoder:
## Summary

Fixes #2090.

Replaces the decimal validation regex in `convertToTokenAmount()` with a linear validator to avoid ReDoS-style backtracking on long invalid input.

## Changes

- Preserve valid integer, decimal, trailing-decimal, and negative amount behavior.
- Keep rejecting scientific notation and malformed numeric strings.
- Add regression coverage for long adversarial invalid input.

## Tests

- `git diff --check`
- Not run: `pnpm --filter @x402/core test -- test/unit/utils/utils.test.ts` because local Node is v18.19.1 and pnpm requires Node v22.13+
- Not run: `pnpm --filter @x402/core lint:check` because local Node is v18.19.1 and pnpm requires Node v22.13+
- Not run: `pnpm --filter @x402/core format:check` because local Node is v18.19.1 and pnpm requires Node v22.13+
- Not run: `pnpm --filter @x402/core build` because local Node is v18.19.1 and pnpm requires Node v22.13+

### DIFF
--- a/typescript/packages/core/src/utils/index.ts
+++ b/typescript/packages/core/src/utils/index.ts
@@ -47,7 +47,7 @@ export function convertToTokenAmount(decimalAmount: string, decimals: number): s
       `Invalid amount: ${decimalAmount} — use decimal notation, not scientific notation`,
     );
   }
-  if (!/^-?\d+\.?\d*$/.test(decimalAmount)) {
+  if (!isPlainDecimalAmount(decimalAmount)) {
     throw new Error(`Invalid amount: ${decimalAmount}`);
   }
   const [intPart, decPart = ""] = decimalAmount.split(".");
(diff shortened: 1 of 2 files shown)


FidelCoder:
Hello, any feedback on this?

phdargen:
Thanks for your contribution @FidelCoder , could you please rebase against main?

Note that we require commit signing, please rebase and verify your commits.

PR #2361, opened 2026-05-19
feat(examples/ts): add pre-payment trust check hook example

andysalvo:
## Description

Adds a trust-check example to the advanced client patterns, demonstrating how to use `onBeforePaymentCreation` to verify service trust before sending payment.

The pattern:
1. Agent calls a trust check endpoint with the target service URL
2. Gets back a score, grade, and recommendation
3. Aborts payment if score is below threshold

Uses the existing lifecycle hook API with no new dependencies. The trust check URL is configurable.

## Why

As x402 scales, agents need a way to evaluate whether a service is trustworthy before paying. This example shows how the existing hook system supports that pattern without any protocol changes.

## Tests

Example only. No runtime changes.

## Checklist

- [x] My changes generate no new warnings


phdargen:
Thanks for the suggestion but examples should remain neutral and not highlight a specific service

PR #2375, opened 2026-05-19
fix(evm/batch-settlement): submit settle() with an explicit gas limit

### LINKED ISSUE #2374: batch-settlement facilitator: settle() submitted without a gas limit causes intermittent on-chain out-of-gas revert
This is not a duplicate of #1065 / #418 / #1377. Those concern the CDP-hosted facilitator and the `exact` scheme, where `eth_estimateGas` throws and `/settle` is rejected before any transaction is sent. This is a distinct failure in the self-hosted `@x402/evm` batch-settlement mechanism: the estimate succeeds with a too-low value, a transaction is broadcast, and it reverts on-chain out-of-gas, consuming the facilitator's gas. Different scheme, different code path, different failure mode.

PR: #2375.

## Summary

In `typescript/packages/mechanisms/evm/src/batch-settlement/facilitator/settle.ts`, `executeSettle` submits the `settle(receiver, token)` transaction via `signer.writeContract` with no explicit `gas`. viem therefore auto-estimates the gas immediately before broadcasting.

## Root cause

The on-chain `settle` branches on claim-visible state (`contracts/evm/src/x402BatchSettlement.sol`):

```solidity
function settle(address receiver, address token) external nonReentrant {
    ReceiverState storage rs = receivers[receiver][token];
    uint128 amount = rs.totalClaimed - rs.totalSettled;
    if (amount == 0) return;                       // cheap early-return path
    rs.totalSettled = rs.totalClaimed;
    IERC20(token).safeTransfer(receiver, amount);  // transfer path
    emit Settled(receiver, token, msg.sender, amount);
}
```

`totalClaimed` is written by `claim()`. `settle()` therefore has a bimodal cost: the early-return path is cheap (around 25.5k gas observed on Base mainnet), the transfer path is roughly 57k.

`executeSettle` performs a `receivers()` read and a `settle` simulation, then calls `writeContract`, which auto-estimates. Those are independent RPC calls and can resolve against different block states. When the gas estimate lands on a node whose state has not yet caught up to the just-mined `claim`, it sees `amount == 0`, estimates the early-return path, and the transaction is broadcast under-gassed for the transfer path it actually executes — reverting out-of-gas.

## Observed on Base mainnet

A `settle` transaction reverted with `gasLimit == gasUsed`, the out-of-gas signature, at around 25.5k gas. Re-estimating the identical call afterward returned around 57.6k. The failure is intermittent: identical flows sometimes settle cleanly and sometimes revert, consistent with an RPC state-timing race rather than a deterministic bug.

## Suggested fix

Pass an explicit `gas` to the `settle` `writeContract` call in `executeSettle`. The `FacilitatorEvmSigner.writeContract` signature already accepts an optional `gas` for exactly this purpose — its own doc comment notes "When provided, skips eth_estimateGas simulation" — and `deposit-permit2.ts` already uses this pattern (`gas: 300_000n`).

A fixed limit is appropriate for `settle` specifically because its cost is bounded: the contract performs at most one `SSTORE` plus one ERC-20 `safeTransfer`, with no loop, so cost does not scale with voucher or channel count. A generous constant (the transfer path is ~57k; ~120k leaves a 2x margin) removes the dependency on a state-sensitive estimate entirely. Note that re-reading chain state to validate the estimate does not work — the validating read is subject to the same node-consistency race that produced the bad estimate.

`executeClaimWithSignature` is also a `writeContract`-without-`gas` call and can mis-estimate the same way. It is out of scope for a fixed limit, though — `claimWithSignature` loops over `voucherClaims`, so its gas scales with batch size and a constant would not be correct. A buffered estimate would suit it better; flagging it as a separate follow-up.

## Re: verify-time simulation (#1377)

Simulation during `verify` catches payloads that will revert. It does not address this. Here the settlement is valid and would succeed; it fails only because the broadcast's gas limit was budgeted too low. A passing simulation still leaves the under-gassed `writeContract` in place.

BranchManager69:
Fixes #2374.

`executeSettle` submits the `settle(receiver, token)` transaction via `writeContract` with no explicit `gas`, so viem auto-estimates immediately before broadcasting.

## The bug

The on-chain `settle` is bimodal (`contracts/evm/src/x402BatchSettlement.sol`): it early-returns (~25.5k gas) when `totalClaimed == totalSettled`, and performs an `SSTORE` plus an ERC-20 `safeTransfer` (~57k gas) otherwise. The `eth_estimateGas` viem runs for `writeContract` is an independent RPC call that can resolve against a node whose state has not yet caught up to the just-mined `claim`. It then estimates the early-return path, the transaction is broadcast under-gassed, and it reverts out of gas once the claim is visible. Observed intermittently on Base mainnet — the reverted `settle` had `gasLimit == gasUsed`, the out-of-gas signature.

## The fix

Pass an explicit `gas` to the `settle` `writeContract`. `FacilitatorEvmSigner.writeContract` already accepts an optional `gas` ("When provided, skips eth_estimateGas simulation"), and `deposit-permit2.ts` already uses this explicit-gas pattern (`gas: 300_000n`). `settle`'s cost is bounded — one `SSTORE` plus one transfer, no loop, so it does not scale with voucher or channel count — so a fixed limit is correct here. `SETTLE_GAS_LIMIT` is `120_000n`, roughly a 2x margin over the transfer-path cost.

## Scope

This PR fixes `settle` only. `executeClaimWithSignature` is also a `writeContract`-without-`gas` call and can mis-estimate the same way, but `claimWithSignature` loops over `voucherClaims` so its gas scales with batch size — a fixed limit would be wrong there. That needs a buffered estimate instead and is left as a separate follow-up (noted in #2374).

## Testing

Full `@x402/evm` unit suite passes (625 tests). Adds one test asserting `executeSettle` passes an explicit `gas` to `writeContract`. ESLint and Prettier clean.

### DIFF
--- a/typescript/packages/mechanisms/evm/src/batch-settlement/facilitator/settle.ts
+++ b/typescript/packages/mechanisms/evm/src/batch-settlement/facilitator/settle.ts
@@ -7,6 +7,24 @@ import { BATCH_SETTLEMENT_ADDRESS } from "../constants";
 import * as Errors from "../errors";
 
 /**
+ * Explicit gas limit for the `settle` transaction.
+ *
+ * `settle` auto-estimation is unsafe: the on-chain `settle` is bimodal — it
+ * early-returns (~25.5k gas) when `totalClaimed == totalSettled`, and performs
+ * an SSTORE plus an ERC-20 transfer (~57k gas) otherwise. The `eth_estimateGas`
+ * viem runs for `writeContract` is an independent RPC call that can resolve
+ * against a node whose state has not yet caught up to the just-mined `claim`;
+ * it then estimates the early-return path and the transaction is broadcast
+ * under-gassed, reverting out of gas once the claim is visible.
+ *
+ * `settle`'s cost is bounded (one SSTORE + one transfer, no loop — it does not
+ * scale with voucher or channel count), so a fixed limit is correct here. This
+ * value leaves roughly a 2x margin over the observed transfer-path cost.
+ * `deposit-permit2.ts` uses the same explicit-gas pattern.
+ */
+const SETTLE_GAS_LIMIT = 120_000n;
+
+/**
  * Transfers claimed funds from the contract.
  *
  * This should be called after one or more `claim()` transactions have updated the
@@ -79,6 +97,7 @@ export async function executeSettle(
       abi: batchSettlementABI,
       functionName: "settle",
       args: [receiver, token],
+      gas: SETTLE_GAS_LIMIT,
     });
 
     const receipt = await signer.waitForTransactionReceipt({ hash: tx });
--- a/typescript/packages/mechanisms/evm/test/unit/batch-settlement/facilitator.test.ts
+++ b/typescript/packages/mechanisms/evm/test/unit/batch-settlement/facilitator.test.ts
@@ -896,6 +896,36 @@ describe("BatchSettlementEvmScheme (Facilitator) — settle routing", () => {
     );
   });
 
+  it("submits settle with an explicit gas limit (not an auto-estimate)", async () => {
+    // `settle` is bimodal on-chain — a no-op early-return when nothing is
+    // claimed, an ERC-20 transfer otherwise. An auto-estimate that races a
+    // node lagging the just-mined `claim` budgets the no-op path and reverts
+    // out of gas. executeSettle must pass an explicit `gas`.
+    const signer = buildSigner({
+      waitForTransactionReceipt: vi.fn().mockResolvedValue({
+        status: "success",
+        logs: [buildSettledLog({ amount: "4321" })],
+      }),
+    });
+    const scheme = new BatchSettlementEvmScheme(signer, authorizer);
+    const sp: BatchSettlementSettlePayload = {
+      type: "settle",
+      receiver: RECEIVER,
+      token: ASSET,
+    };
+    const result = await scheme.settle(
+      envelopeSettle(sp as unknown as Record<string, unknown>),
+      makeRequirements(),
+    );
+    expect(result.success).toBe(true);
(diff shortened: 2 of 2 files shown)


phdargen:
Hi @BranchManager69, I think https://github.com/x402-foundation/x402/pull/2306 partially addressed this by adding an amount > 0 check. But this check could pass on node A and then eth_estimateGas on a stale node B still sees amount = 0 and sets the wrong gas limit, so agreed on setting an explicit gas limit here

typescript/packages/mechanisms/evm/src/batch-settlement/facilitator/settle.ts
```diff
@@ -6,6 +6,24 @@ import { batchSettlementABI } from "../abi";
...
+ * against a node whose state has not yet caught up to the just-mined `claim`;
+ * it then estimates the early-return path and the transaction is broadcast
+ * under-gassed, reverting out of gas once the claim is visible.
+ *
+ * `settle`'s cost is bounded (one SSTORE + one transfer, no loop — it does not
+ * scale with voucher or channel count), so a fixed limit is correct here. This
+ * value leaves roughly a 2x margin over the observed transfer-path cost.
+ * `deposit-permit2.ts` uses the same explicit-gas pattern.
+ */
+const SETTLE_GAS_LIMIT = 120_000n;
```
phdargen, line 25: Lets move this in constants.ts and tone down the prose a bit
```suggestion
 // Explicit gas limit because `settle` gas estimation can hit the cheap early-return path on stale RPC state
 const SETTLE_GAS_LIMIT = 120_000n;
```

phdargen:
Could you please also add this to the go sdk 🙏?

phdargen (approved):

PR #2385, opened 2026-05-20
fix(evm): verify Transfer event in receipt after exact/eip3009 settle

Mameta29:
## Description

`settleEIP3009` currently treats `receipt.status === "success"` as proof of a successful transfer. The receipt's status only tells us the tx did not revert; it does not tell us that the expected ERC-20 `Transfer` was emitted from the expected token contract with the expected `(from, to, value)`.

The same kind of post-execution event-shape check already exists elsewhere in the EVM mechanism and across SDKs:

- `typescript/packages/mechanisms/evm/src/batch-settlement/facilitator/settle.ts` filters `receipt.logs` by the batch contract address and runs `parseEventLogs({ abi: batchSettlementABI, eventName: "Settled", ... })` to read back the settled amount.
- `typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts` (`validateSimulationEvents`) checks that there is exactly one token transfer event and that it matches the expected sender, recipient, amount, and asset.

This PR brings `exact/eip3009` settle in line with the same pattern.

### Changes

- New helper `verifyEip3009TransferEvent(receipt, erc20Address, expected)` in `typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009-utils.ts`. It filters `receipt.logs` by the expected token address (case-insensitive via `isAddressEqual`), decodes ERC-20 `Transfer` logs via `parseEventLogs`, and returns `true` iff a log's `(from, to, value)` matches the authorization's `(from, to, value)`.
- `settleEIP3009` calls the helper after the existing `receipt.status` check. On mismatch, it returns a `SettleResponse` with the **transaction hash preserved** (so operators can investigate the on-chain tx) and the new error code `ErrTransferEventMismatch` (wire string: `invalid_exact_evm_transfer_event_mismatch`).
- The happy path is unchanged.

### Why this matters

A tx can land successfully and still not have performed the EIP-3009 transfer we asked for — RPC bugs returning the wrong receipt, a swap/relay/sponsor wrapper that "succeeds" without forwarding the underlying transfer, a re-org corner case, an audited but compromised token contract. We have no reason to keep this signal hidden when other EVM facilitators in this repo already check it.

## Tests

New unit test file `typescript/packages/mechanisms/evm/test/unit/exact/eip3009-transfer-event.test.ts` (8 cases):

- happy path: canonical Transfer event matches
- happy path: matches even when an unrelated Transfer from another token/wallet is also in the receipt
- rejects when `value` differs
- rejects when `to` differs (recipient swap)
- rejects when `from` differs (sender swap)
- rejects when the only Transfer log is from a different token contract
- rejects when receipt has no logs at all
- address comparison is case-insensitive (covers checksummed vs. lowercase)

Local verification:

- `pnpm -C typescript/packages/mechanisms/evm test` → **637 passed (26 files)**, no regressions
- `pnpm -C typescript format` and `pnpm -C typescript lint:check` → clean
- `pnpm-lock.yaml` is unchanged

## Changelog fragment

- `typescript/.changeset/verify-transfer-event-in-eip3009-settle.md` → `@x402/evm: patch`

## Scope

TypeScript only for this PR. The same defensive check belongs in `go/mechanisms/evm/exact/facilitator/eip3009.go` and `python/x402/mechanisms/evm/exact/eip3009_facilitator.py`; happy to follow up with parity PRs once direction here is confirmed.

## Disclosure

This change was prepared with the help of an AI coding assistant. The diff was reviewed against the existing `batch-settlement/facilitator/settle.ts` and `stellar/exact/facilitator/scheme.ts` patterns, and the test matrix was written to cover both the matching and the mismatch directions (value / to / from / token-address / no-logs) before commit.

## Checklist

- [x] My commits are signed (SSH signing)

### DIFF
--- a/python/x402/mechanisms/evm/exact/eip3009_utils.py
+++ b/python/x402/mechanisms/evm/exact/eip3009_utils.py
@@ -327,10 +327,7 @@ def verify_eip3009_transfer_event(
     to: str,
     value: int,
 ) -> bool:
-    """Return True when receipt logs contain the expected ERC-20 Transfer.
-
-    Mirrors Go ``verifyEIP3009TransferEvent`` / TS ``verifyEip3009TransferEvent``.
-    """
+    """Return True when receipt logs contain the expected ERC-20 Transfer."""
     if not logs:
         return False
 
--- a/typescript/.changeset/verify-transfer-event-in-eip3009-settle.md
+++ b/typescript/.changeset/verify-transfer-event-in-eip3009-settle.md
@@ -1,0 +1,5 @@
+---
+"@x402/evm": patch
+---
+
+Verified the Transfer event in the post-settle receipt for exact/eip3009 settle, matching the defensive event-shape check already performed by @x402/evm batch-settlement and @x402/stellar exact. Added ErrTransferEventMismatch (`invalid_exact_evm_transfer_event_mismatch`) so a successful tx that emitted no matching Transfer is no longer reported as a successful settlement.
--- a/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009-utils.ts
+++ b/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009-utils.ts
@@ -7,7 +7,7 @@ import {
   parseErc6492Signature,
   parseEventLogs,
   parseSignature,
-  type TransactionReceipt,
+  type Log,
 } from "viem";
 import { eip3009ABI } from "../../constants";
 import { multicall, ContractCall, RawContractCall } from "../../multicall";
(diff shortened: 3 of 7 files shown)


CarsonRoscoe (commented):
Hey @Mameta29, thank you for the contribution! Appreciate that you are improving correctness without adding any new RPC calls or latency. Nice addition.

Some minor comments to address and then I'd like to get this in:

1. Orphaned JSDoc in eip3009-utils.ts L211–L217
There are two stacked JSDoc blocks immediately before verifyEip3009TransferEvent. The first one (Maps an EIP-3009 contract revert error...) describes parseEip3009TransferError, not the new function — it looks like a copy-paste artifact from the block further down in the file. Please remove L211–L217 and keep only the correct JSDoc that starts at L218.

2. Dead guard in eip3009-utils.ts L236
`if (!receipt.logs) return false;`
TransactionReceipt from viem types logs as a non-nullable Log[], so this condition can never be true. Remove it, if you want a fast-exit for the empty case the natural form is just letting parseEventLogs return [] and .some() return false, which already happens.

kakedashi3:
Solid defensive check — `receipt.status === "success"` only proves the tx didn't revert, so reading back the `Transfer` log is the right call. The implementation holds up: filtering `receipt.logs` by token address with `isAddressEqual`, the exact `value` match, and `.some()` over the logs are all correct, and the test that ignores a spoofed `Transfer` from `OTHER_TOKEN` covers the main spoofing vector. 👍

Parity note: this lands in `@x402/evm` (TS) only, but the same gap exists in the other EVM SDKs — Go's `settleEIP3009` (`go/mechanisms/evm/exact/facilitator/eip3009.go`) returns `Success: true` on `receipt.Status == TxStatusSuccess` alone with no Transfer read-back, and Python's exact/eip3009 settle appears to be the same. Since batch-settlement already does this event-shape check across SDKs (per the changeset), porting it to Go/Python would close the same hole there. Happy to take the Go port as a follow-up if that's useful.

Minor: since this runs *after* the tx executes, a false-negative reports an already-settled transfer as `success: false`. The exact `value === expected.value` match is right for the EIP-3009 FiatToken family; a one-line note that it assumes a standard (non-fee-on-transfer) token would flag the edge for anyone adding exotic assets to `DEFAULT_STABLECOINS`.

kakedashi3:
Quick follow-up on the Go port I mentioned — it's less mechanical than the TS version, worth flagging before anyone picks it up.

In Go, `TransactionReceipt` is `{ Status, BlockNumber, TxHash }` (`mechanisms/evm/types.go`) — no `logs`. `WaitForTransactionReceipt` is a method on the user-implemented `FacilitatorEvmSigner` interface, and there's no log-fetch on that interface, so the Go settle path can't read the `Transfer` log the way viem exposes `receipt.logs` here.

So a Go port needs a `Logs` field on `TransactionReceipt` plus signer implementations populating it. Since signers are user-provided, naively requiring the log would false-negative settles for any signer that doesn't populate it — a real, already-executed transfer reported as `success: false`.

A backward-compatible shape: add an optional `Logs` field and run the Transfer check only when logs are present, falling back to the current status-only behavior otherwise. Non-breaking, and signers opt into the stronger check. Happy to put that up as a separate Go PR if maintainers think the opt-in scaffold is worth it.

wnjoon:
@kakedashi3 @CarsonRoscoe @Mameta29 @phdargen 

I’ll take the Go follow-up for this.

Plan is to keep it backward-compatible: add optional receipt logs to the Go `TransactionReceipt`, populate them where in-repo receipt implementations already have access to logs, and make exact/eip3009 settlement verify the ERC-20 `Transfer(from, to, value)` event only when logs are present. That gives Go parity with the stronger TS check without false-negativing existing user-provided `FacilitatorEvmSigner` implementations that still return status-only receipts.

I’ll add focused tests for the matching event path, mismatched from/to/value/token cases, and the no-logs fallback behavior.

SashaMIT:
Noting for reviewers: Go already landed the equivalent check in #2727. On current `main`, the TypeScript exact EIP-3009 settle path still returns success from `receipt.status` alone, so this PR remains the needed TS fix.

I opened a companion Python PR that ports the same Transfer-event verification for the Python exact facilitators (V1 + V2), since that path had the same gap: https://github.com/x402-foundation/x402/pull/3032

Happy to adjust naming/error codes if you want stricter cross-SDK lockstep with this branch.

CarsonRoscoe (approved):

PR #1895, opened 2026-04-01
feat(concordium): Concordium exact schema implementation

oleh-drozd-eth:
## Description

Implementation Concordium exact schema
Schema reference - https://github.com/coinbase/x402/pull/1081

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [x] I added a changelog fragment for user-facing changes (docs-only changes can skip)

PR #2389, opened 2026-05-20
docs(spec): Concordium exact review fixes

pavelhbl:
## Summary

This PR refines the Concordium **Exact** scheme spec based on maintainer review feedback in `x402-foundation/x402#1081`.

## Changes

- Align `PaymentRequirements.extra` naming with other exact schemes by using `extra.feePayer` (documented as Concordium sponsor).
- Clarify PLT token decimals (`depends on token`).
- Remove redundant `payload.sender` from examples (still allowed as optional; `signedTransaction.header.sender` is the source of truth).
- Tie `transaction.header.expiry` bound to `PaymentRequirements.maxTimeoutSeconds`.
- Add recommendation for preflight checks / simulation to ensure on-chain success likelihood.

## Related

- Supersedes / continues: `x402-foundation/x402#1081`
- Implementation PR to be updated accordingly: `x402-foundation/x402#1895`

## Notes

I’m taking over maintenance of these PRs (original author is no longer available) and will continue iterating based on reviewer feedback.


pavelhbl:
Implementation PR is now tracked in https://github.com/x402-foundation/x402/pull/2390 (replacement for #1895).

pavelhbl:
Hi @CarsonRoscoe @phdargen — I opened a replacement PR for the Concordium exact spec review fixes so the thread is easier to follow:

- Spec review fixes: https://github.com/x402-foundation/x402/pull/2389

This PR includes the requested updates (fee payer naming alignment, remove redundant sender in examples, decimals clarification, expiry bound to `maxTimeoutSeconds`, and preflight/simulation recommendation). Would appreciate a re-review when you have a moment.

Implementation PR (replacement for the old one) is here: https://github.com/x402-foundation/x402/pull/2390

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,255 @@
+# Exact Payment Scheme for Concordium (`exact`)
+
+This document specifies the `exact` payment scheme for the x402 protocol on Concordium.
+
+This scheme facilitates payments of a specific amount of CCD or CIS-2 tokens (e.g., PLT) on the Concordium blockchain using sponsored transactions (V1).
```
phdargen: Either exclude non-protocol native CIS-2 tokens or the specs need to define how to handle them
```suggestion
This scheme facilitates payments of a specific amount of CCD or PLT on the Concordium blockchain using sponsored transactions (V1).
```

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,255 @@
...
+  The sender's credential signature(s) MUST be verified cryptographically against the sender's on-chain account credentials. The facilitator MUST fetch the sender's account information from the Concordium network and validate that the signature was produced by a key authorized for the sender account.
+
+9. Transaction payload safety
+
+- The transaction MUST contain exactly one transfer operation (no bundled or unexpected operations).
+- The facilitator's sponsor address MUST NOT appear as the sender, recipient, or authority of the transfer.
+
+These checks are security-critical to ensure the sponsor cannot be tricked into paying gas for unintended transactions. Implementations MAY introduce stricter limits (e.g., shorter expiry caps) but MUST NOT relax the above constraints.
+
+Facilitators SHOULD also ensure the transaction is likely to succeed on-chain (e.g., by simulation or targeted preflight checks such as nonce/sequence validity and sufficient sender balance for the transfer).
```
phdargen:
```suggestion
Facilitators MUST also ensure the transaction is likely to succeed on-chain by simulation or targeted preflight checks that include at a minimum ensuring nonce/sequence validity and sufficient sender balance for the transfer.
```

pavelhbl:
Thanks @phdargen — applied both suggestions.

- Updated scope wording to "CCD or PLT" (removed CIS-2 mention) per your inline suggestion.
- Tightened the preflight/simulation guidance from SHOULD → MUST with the minimum checks you listed.

specs/schemes/exact/scheme_exact_concordium.md
```diff
@@ -0,0 +1,255 @@
...
+|---------|-------------------|
+| Mainnet | `ccd:9dd9ca4d19e9393877d2c44b70f89acb` |
+| Testnet | `ccd:4221332d34e1694168c2a0c0b3fd0f27` |
+
+### Asset Format
+
+| Asset Type | Format       | Example  |
+|------------|--------------|----------|
+| Native CCD | Empty string | `""`     |
+| PLT Token  | Token symbol | `"EURR"` |
```
phdargen: How about we namespace 'plt:' so we can use 'native' as resered keyword? 
```suggestion
| Native CCD | Empty string | `"native"`     |
| PLT Token  | Token symbol | `"plt:EURR"` |
```
cc @CarsonRoscoe

pavelhbl:
Thanks @phdargen — I agree namespacing like `plt:` + reserving a keyword like `native` would reduce ambiguity/collisions.

That said, at the protocol level `asset` is currently defined as:
> "Token contract address or ISO 4217 currency code for fiat; an empty string (\"\") MAY represent the network's native asset"
(from `x402-specification-v2.md` in this repo).

We already aligned the Concordium scheme with that core definition (i.e. `asset=""` for native CCD and `asset="EURR"` for PLT symbol).

Before changing Concordium to `asset="native"` / `asset="plt:EURR"`, can we confirm with @CarsonRoscoe whether the intention is to evolve the *core* `asset` identifier format (e.g. to a namespaced/CAIP-19-style scheme) versus keeping the current v2 definition?

Happy to implement `native`/`plt:` if that’s the desired direction — just want to avoid a one-off divergence from the x402 v2 spec.

phdargen:
> Thanks @phdargen — I agree namespacing like `plt:` + reserving a keyword like `native` would reduce ambiguity/collisions.
> 
> That said, at the protocol level `asset` is currently defined as:
> 
> > "Token contract address or ISO 4217 currency code for fiat; an empty string ("") MAY represent the network's native asset"
> > (from `x402-specification-v2.md` in this repo).
> 
> We already aligned the Concordium scheme with that core definition (i.e. `asset=""` for native CCD and `asset="EURR"` for PLT symbol).
> 
> Before changing Concordium to `asset="native"` / `asset="plt:EURR"`, can we confirm with @CarsonRoscoe whether the intention is to evolve the _core_ `asset` identifier format (e.g. to a namespaced/CAIP-19-style scheme) versus keeping the current v2 definition?
> 
> Happy to implement `native`/`plt:` if that’s the desired direction — just want to avoid a one-off divergence from the x402 v2 spec.

The x402 v2 spec you are referring to is the version in your local branch, this is not what we currently have in the repo

pavelhbl:
Thanks @phdargen — agreed. I’ve now merged latest `main` into this PR branch, so the diff here is the actual proposal on top of current `main`.

This PR intentionally updates `specs/x402-specification-v2.md` to clarify that `asset=""` MAY represent the network native asset, so Concordium’s `asset=""` for CCD is core-spec compliant (rather than a Concordium-only convention).

Re your `native` / `plt:` suggestion: the namespacing idea makes sense, but it’s a broader cross-scheme change to the core `asset` identifier format. If we want to adopt `native`/namespaced assets, it likely should be a deliberate v2 convention (and applied consistently across schemes), rather than introduced only for Concordium.

Can you + @CarsonRoscoe confirm which direction you prefer for v2 going forward?
- Keep `asset=""` for native assets and land the core-spec clarification (this PR), or
- Move to a `native` / namespaced convention (e.g. `plt:`) and update the scheme(s) accordingly

phdargen:
So we don't like using an empty string for native assets. 

I wouldn't namespace assets in general as this would be a breaking change, this would be Concordium specific as its the only one using token names instead of addresses. 

As an alternative could we just use asset="CCD" for the native asset?

specs/schemes/exact/scheme_exact_concordium.md
CarsonRoscoe: I'm in favour of that. I don't want empty string to mean native, but understand in Concodium arbitrary names can be used for plt tokens. Namespacing them feels appropriate

CarsonRoscoe (commented):
@pavelhbl thanks for continueing the discussion here,

I am aligned with @phdargen that we do not want to use empty strings for default assets, we'd rather use a constant like `native` so it's explicit rather than implicit. I am aligned with namespacing it

pavelhbl:
Thanks @phdargen @CarsonRoscoe — confirmed.

Updated the Concordium exact scheme docs to use `asset="CCD"` for the native asset (instead of the empty string), and reverted the core v2 `asset` description change accordingly.

phdargen (approved):
Thanks @pavelhbl, LGTM 🚀

PR #2390, opened 2026-05-20
feat(concordium): Exact mechanism implementation (replacement)

pavelhbl:
## Summary

Adds the Concordium **Exact** mechanism implementation to the TypeScript SDK under `typescript/packages/mechanisms/concordium`.

This is a replacement for the older PR `x402-foundation/x402#1895` with a clearer description and the branch refreshed with latest `main`.

## Related

- Spec review fixes: `x402-foundation/x402#2389`
- Previous implementation PR (superseded): `x402-foundation/x402#1895`

## What’s included

- Concordium exact client/server/facilitator scheme implementation
- Concordium signer + shared types/constants
- Unit + integration tests for the package
- Package docs + changelog

## Notes

I’m continuing maintenance of these PRs (original author is no longer available) and will keep the implementation aligned with the final spec direction.


pavelhbl:
Hi @CarsonRoscoe — replacement implementation PR is up here:

- Implementation (replacement): https://github.com/x402-foundation/x402/pull/2390

It follows the Concordium exact mechanism implementation and is intended to stay aligned with the spec direction tracked in:
- Spec review fixes: https://github.com/x402-foundation/x402/pull/2389

When you have bandwidth, could you take a look and let me know any requested changes?

typescript/packages/mechanisms/concordium/package.json
```diff
@@ -0,0 +1,96 @@
...
+  },
+  "keywords": [
+    "x402",
+    "payment",
+    "protocol",
+    "concordium"
+  ],
+  "license": "Apache-2.0",
+  "author": "Coinbase Inc.",
+  "repository": "https://github.com/coinbase/x402",
```
phdargen:
```suggestion
  "author": "x402 Foundation",
  "repository": "https://github.com/x402-foundation/x402",
```

typescript/packages/mechanisms/concordium/TODO.md
```diff
@@ -0,0 +1 @@
+# @x402/concordium
```
phdargen: please remove file

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,567 @@
...
+  }
+
+  /**
+   * Validates the partially-signed transaction against all 9 MUST rules.
+   *
+   * @param payload - The x402 payment payload containing the signed transaction
+   * @param requirements - The payment requirements from the resource server
+   * @returns Verification result indicating validity and payer address
+   */
+  async verify(
```
phdargen, line 116: please align with latest spec version, in particular adding tx simulation

phdargen:
Thanks a lot for the contribution @pavelhbl, thats a great start! 

Could you please also add the following:
- e2e tests
- add Concordium to the all_networks server/client/facilitator examples, eg https://github.com/x402-foundation/x402/blob/main/examples/typescript/servers/advanced/all_networks.ts
- add a publishing workflow like https://github.com/x402-foundation/x402/blob/main/.github/workflows/publish_npm_scoped_x402_evm.yml

phdargen:
Please also point us to a faucet for testnet native and PLT tokens and add a link to the READMEs

pavelhbl:
@phdargen pushed updates addressing the latest review notes:

- Added Concordium to the TypeScript all-networks server/client/facilitator examples
- Added publish workflow: .github/workflows/publish_npm_scoped_x402_concordium.yml
- Added an offline Concordium e2e-style test under typescript/packages/mechanisms/concordium/test/e2e/
- Updated docs/tests to use asset="CCD" for native + requirements.extra.feePayer, and removed the TODO.md

Latest commit is GPG-verified: 0d98d98a.

e2e/concordium/exact-offline.test.ts
```diff
@@ -0,0 +1,182 @@
+import { describe, expect, it, vi } from "vitest";
```
phdargen: please remove this and add proper e2e tests in https://github.com/x402-foundation/x402/tree/main/e2e

examples/typescript/clients/advanced/README.md
```diff
@@ -38,13 +38,16 @@ cp .env-local .env
...
 
+- `CCD_WALLET_PATH` - Concordium wallet export path for Concordium payments (optional; `all-networks`)
 - `EVM_PRIVATE_KEY` - Ethereum private key for EVM payments
 - `SVM_PRIVATE_KEY` - Solana private key for SVM payments
 - `STELLAR_PRIVATE_KEY` - Stellar secret key (starts with `S`) for signing Stellar payments
 - `HEDERA_ACCOUNT_ID` - Hedera account id for Hedera payments (optional)
 - `HEDERA_PRIVATE_KEY` - Hedera **ECDSA** private key (0x-prefixed or DER-encoded) for Hedera payments (optional)
 - `HEDERA_NETWORK` - Hedera network (optional, defaults to `hedera:testnet`)
 
+> **Concordium Testnet:** Fund a testnet account via the Concordium Wallet (Testnet) faucet / “request CCD” flow.
```
phdargen: I created a testnet account with the Concordium browser extension but dont see any “request CCD” flow, please provide more detailed instructions how to obtain both CCD and PLT testnet funds

typescript/packages/mechanisms/concordium/src/signer.ts
```diff
@@ -0,0 +1,218 @@
...
+ *
+ * @param sponsorAddress - Sponsor account address (base58check string)
+ * @param sponsorSigner  - Built via `buildAccountSigner(parseWallet(walletFile))`
+ * @param grpcConfig     - gRPC connection parameters
+ * @returns A FacilitatorConcordiumSigner instance
+ *
+ * @example
+ * ```typescript
+ * import { parseWallet, buildAccountSigner } from "@concordium/web-sdk";
+ * import { getConcordiumGrpcUrl, parseGrpcUrl } from "@x402/concordium/constants";
```
phdargen:
```suggestion
 * import { getConcordiumGrpcUrl, parseGrpcUrl } from "@x402/concordium";
```

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -10,13 +10,18 @@
 
 import { toFacilitatorAvmSigner } from "@x402/avm";
 import { ExactAvmScheme } from "@x402/avm/exact/facilitator";
+import { ExactConcordiumScheme } from "@x402/concordium/exact/facilitator";
+import { toConcordiumFacilitatorSigner } from "@x402/concordium/signer";
+import { CONCORDIUM_TESTNET_CAIP2, getConcordiumGrpcUrl, parseGrpcUrl } from "@x402/concordium/constants";
```
phdargen:
```suggestion
import {
  CONCORDIUM_TESTNET_CAIP2,
  getConcordiumGrpcUrl,
  parseGrpcUrl,
  toConcordiumFacilitatorSigner,
} from "@x402/concordium";
```

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -66,6 +69,14 @@ if (avmAddress) {
     payTo: avmAddress,
   });
 }
+if (ccdAddress) {
+  accepts.push({
+    scheme: "exact",
+    price: "$0.001",
```
phdargen: There is no default stablecoin defined for the money parser, suggest to do so. 
See also https://github.com/x402-foundation/x402/blob/main/DEFAULT_ASSETS.md

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,366 @@
...
+  private moneyParsers: MoneyParser[] = [];
+
+  /**
+   * Creates a new ExactConcordiumScheme instance.
+   *
+   * @param sponsorAddress - Facilitator's sponsor account address (base58check).
+   *   Injected into `PaymentRequirements.extra.feePayer` so clients know
+   *   which address to nominate as sponsor in their V1 transaction header.
+   */
+  constructor(private readonly sponsorAddress?: string) {}
```
phdargen: please remove, server shouldnt set sponsorAddress. 

facilitator announces feepayer in supported and 402 of server should be enriched with it in enhancePaymentRequirements, see eg svm mechanism

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,366 @@
...
+  ): Promise<PaymentRequirements> {
+    if (!this.sponsorAddress) {
+      return Promise.resolve(requirements);
+    }
+
+    return Promise.resolve({
+      ...requirements,
+      extra: {
+        ...((requirements.extra as Record<string, unknown>) ?? {}),
+        feePayer: this.sponsorAddress,
```
phdargen:
```suggestion
        feePayer: supportedKind.extra?.feePayer,
```

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -105,6 +114,22 @@ if (avmPrivateKey) {
   facilitator.register(AVM_NETWORK, new ExactAvmScheme(avmSigner));
 }
 
+// Register Concordium scheme if wallet export path is provided
+if (ccdFacilitatorWalletPath) {
+  const sponsorWallet = parseWallet(readFileSync(ccdFacilitatorWalletPath, "utf8"));
```
phdargen: can we align this with the other networks and just read a privatekey from env?

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,671 @@
...
+   * @default 600
+   */
+  maxExpiryOffsetSeconds?: number;
+
+  /**
+   * Assets reported on the /supported endpoint.
+   *
+   * @default [{ symbol: "CCD", decimals: 6 }]
+   */
+  supportedAssets?: Array<{ symbol: string; decimals: number }>;
```
phdargen: can be removed

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,671 @@
...
+
+  /**
+   * Returns signer addresses for the /supported endpoint.
+   *
+   * @param _ - Network identifier (unused, same signer for all networks)
+   * @returns Array of sponsor account addresses
+   */
+  getSigners(_: string): string[] {
+    return [this.signer.getAddress()];
+  }
```
phdargen, line 107: other mechanisms support multiple signers for loadbalancing

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,671 @@
...
+    if (!validTypes.has(tx.payload.type)) {
+      return `unexpected_transaction_type: ${tx.payload.type}`;
+    }
+
+    if (tx.payload.type === "tokenUpdate") {
+      // Operations are CBOR-encoded — we can't count them without decoding.
+      // `createTokenUpdatePayload` always encodes exactly one operation
+      // per the reference implementation, so structural safety is guaranteed
+      // by the client SDK. The CBOR decode in Rules 4/5 will fail if the
+      // operation is malformed.
```
phdargen: why no decode and check?

client payload should be considered untrusted, can come from any 3rd party sdk

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,671 @@
...
+      onChainNonce = BigInt(onChainNonceRaw as number | string | bigint);
+    } catch {
+      return "preflight_invalid_account_nonce";
+    }
+
+    if (BigInt(tx.header.nonce) !== onChainNonce) {
+      return "preflight_nonce_mismatch";
+    }
+
+    if (expectedAsset.toUpperCase() !== "CCD") return null;
```
phdargen: Should also check balance for PLT tokens

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,366 @@
+import {
+  AssetAmount,
+  Network,
+  PaymentRequirements,
+  Price,
+  SchemeNetworkServer,
+  MoneyParser,
+} from "@x402/core/types";
+
```
phdargen: Please use core utilities where possible for shared logic, 
`import { convertToTokenAmount, numberToDecimalString, parseMoneyString } from "@x402/core/utils";`

You may have to rebase against main, see also https://github.com/x402-foundation/x402/pull/2530

pavelhbl:
@phdargen latest updates are pushed and the review-requested changes have been revalidated.

Addressed in the latest commits:

- Moved the Concordium offline e2e coverage into repo-level `e2e/concordium` and added `pnpm -C e2e test:concordium-offline`
- Expanded CCD / PLT testnet funding instructions in the advanced examples docs
- Cleaned Concordium example imports to use the consolidated `@x402/concordium` exports
- Added the Concordium default stablecoin parser path for the advanced all-networks server example
- Refactored server-side enrichment so the server no longer owns `sponsorAddress`; `feePayer` is taken from facilitator `/supported` metadata
- Reused shared `@x402/core/utils` money parsing/conversion helpers
- Removed the unnecessary facilitator config field
- Added support for multiple facilitator signers / fee-payers
- Added payload decode/validation before checking untrusted PLT client payloads
- Added PLT token balance preflight alongside CCD balance checks
- Resolved the `main` merge conflicts and fixed the examples TypeScript formatting failure

Validated locally:

- `pnpm -C typescript -r --filter=@x402/core build`
- `pnpm -C e2e test:concordium-offline`
- `pnpm -C typescript -r --filter=@x402/concordium lint:check`
- `pnpm -C typescript -r --filter=@x402/concordium build`
- `pnpm -C typescript -r --filter=@x402/concordium test`
- `pnpm -C examples/typescript format:check`
- `pnpm -C examples/typescript lint:check`

GitHub workflows are green on the latest head commit `30c94fe2`:

- Check Python
- Package Lock
- Check Go
- Format
- Lint
- Run Unit Tests
- Verified Commits

e2e/README.md
```diff
@@ -86,6 +86,17 @@ Add the `-v` flag to any command for verbose output:
...
+
+The repo-level Concordium exact flow currently has a dedicated offline e2e regression:
+
+```bash
+pnpm test:concordium-offline
+```
+
+This covers the repo-level Concordium request/verify/settle flow without requiring
+live Concordium wallets or network access.
+
```
phdargen: Offline checks are already done in unit/integration tests, the point of e2e is to test the actual flow end-to-end on testnet, please add Concordium to the TS e2e facilitator + express/hono/fastify/next server + axios/fetch client and run the tests to confirm they work

examples/typescript/facilitator/advanced/README.md
```diff
@@ -21,13 +21,28 @@ cp .env-local .env
...
+> 2. Open the sponsor account.
+> 3. Go to **Activity**.
+> 4. Click **Request CCD**.
+> 5. Wait for the test CCD transfer to arrive.
+>    Official guide: https://docs.concordium.com/en/mainnet/docs/plt/setup-guide/request-ccd.html
+>
+> **Concordium Testnet (PLT):**
+> There is no universal public faucet for arbitrary PLT symbols.
+> If you want to facilitate PLT payments for a symbol such as `EURR`, use an issuer-provided test distribution or request your own PLT issuance on testnet first:
+> https://docs.concordium.com/en/mainnet/tutorials/plt/request-plt.html
```
phdargen, line 46: Got some CCD now but not particular keen to go through this PLT application process, could you please send some to this testnet account 3UrcxPQeYywasrPcYUcqhvFu3SB2vBBDjj7TsaRQ431vGiczYp?

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -105,6 +118,24 @@ if (avmPrivateKey) {
   facilitator.register(AVM_NETWORK, new ExactAvmScheme(avmSigner));
 }
 
+// Register Concordium scheme if wallet export path is provided.
+// Concordium uses the wallet export here because the SDK helper builds the account signer
+// from the exported wallet structure rather than from a raw private key string.
```
phdargen: Thats not a good reason in itself. 

Could the sdk not support setups from private key or mnemonic?

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,871 @@
...
+    );
+
+    if (tx.header.expiry > nowSeconds + maxExpiryOffsetSeconds) {
+      return this.invalid(
+        `expiry_too_far_in_future: max offset is ${maxExpiryOffsetSeconds}s`,
+        resolvedPayer,
+      );
+    }
+
+    const decodedPayload = this.decodePayload(tx.payload);
```
phdargen: Malformed input must always yield an invalid response, not a 500 from the facilitator

This calls `parseBigIntValue`, which may throw Error("invalid_amount_format"). So this should be wrapped in a try/catch

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,871 @@
...
+
+    const recipientError = this.checkRecipient(
+      tx.payload,
+      requirements.payTo,
+      expectedAsset,
+      decodedPayload,
+    );
+    if (recipientError !== null) return this.invalid(recipientError, resolvedPayer);
+
+    const amountError = this.checkAmount(tx.payload, requirements, expectedAsset, decodedPayload);
```
phdargen: could also throw for non-integer `requirements.amount `

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,871 @@
...
+
+/**
+ * Resolves the required amount from payment requirements.
+ *
+ * @param requirements - The payment requirements
+ * @returns The required amount as a string
+ */
+function getRequiredAmount(requirements: PaymentRequirements): string {
+  return (
+    (requirements as unknown as Record<string, string>).maxAmountRequired ??
```
phdargen: This is a v2-only package. Use `requirements.amount` directly and reject when missing.

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,871 @@
...
+ * For strict validation use `AccountAddress.fromBase58(address)` from the SDK.
+ *
+ * @param address - The address string to validate
+ * @returns True if the address matches base58check format
+ */
+function isValidBase58Address(address: string): boolean {
+  if (!address || typeof address !== "string") return false;
+  if (address.length < 45 || address.length > 55) return false;
+  return /^[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]+$/.test(address);
+}
```
phdargen, line 825: @concordium/web-sdk is already a dependency, why not do the strict validation as the docstring suggests?


```suggestion
 * Structural guard for Concordium base58check account addresses.
 *
 * @param address - The address string to validate
 * @returns True if the address matches base58check format
 */
function isValidBase58Address(address: string): boolean {
  try {
    AccountAddress.fromBase58(address);
    return true;
  } catch {
    return false;
  }
}
```

typescript/packages/mechanisms/concordium/src/types.ts
```diff
@@ -0,0 +1,127 @@
...
+ *   "signedTransaction": { "version": 1, "header": { ... }, ... },
+ *   "sender": "3kBx2h5Y2veb4hZvAE2c1Zr6DYJwWbPr9xQJJBPWyFnXHF9UuN"
+ * }
+ * ```
+ */
+export interface ExactConcordiumPayloadV2 {
+  /** Partially-signed V1 sponsored transaction (sponsor slot empty) */
+  signedTransaction: SignableV1Transaction;
+  /** Sender's Concordium account address (base58check) */
+  sender: string;
```
phdargen: drop `sender` from payload. `signedTransaction.header.sender` should be used as per spec

typescript/packages/mechanisms/concordium/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "tsx": "^4.19.2",
+    "typescript": "^5.7.3",
+    "vite": "^6.2.6",
+    "vite-tsconfig-paths": "^5.1.4",
+    "vitest": "^3.0.5"
+  },
+  "dependencies": {
+    "@x402/core": "workspace:*",
+    "@concordium/web-sdk": "^12.0.0",
+    "zod": "^3.24.2"
```
phdargen: remove zod as unused

typescript/packages/mechanisms/concordium/package.json
```diff
@@ -0,0 +1,96 @@
+{
+  "name": "@x402/concordium",
+  "version": "1.0.0",
```
phdargen: align with current @x402/core version
```suggestion
  "version": "2.14.0",
```

typescript/packages/mechanisms/concordium/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @x402/concordium Changelog
+
+## 1.0.0
```
phdargen:
```suggestion
## 2.14.0
```

typescript/packages/mechanisms/concordium/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,241 @@
...
+
+      return new ConcordiumGRPCNodeClient(host, port, creds);
+    }
+
+    const grpcUrl = getConcordiumGrpcUrl(network);
+    const [host, port] = parseGrpcUrl(grpcUrl);
+    const creds =
+      (this.config?.useTls ?? true) ? credentials.createSsl() : credentials.createInsecure();
+
+    return new ConcordiumGRPCNodeClient(host, port, creds);
```
phdargen: nit: two nearly identical branches. Compute `grpcUrl` first, then build the client once

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -59,19 +70,21 @@ const hederaPrivateKey = process.env.HEDERA_PRIVATE_KEY;
...
   console.error(
-    "❌ At least one of AVM_PRIVATE_KEY, EVM_PRIVATE_KEY, SVM_PRIVATE_KEY, STELLAR_PRIVATE_KEY, or HEDERA_ACCOUNT_ID + HEDERA_PRIVATE_KEY is required",
+    "❌ At least one of AVM_PRIVATE_KEY, CCD_FACILITATOR_WALLET_PATH, EVM_PRIVATE_KEY, SVM_PRIVATE_KEY, STELLAR_PRIVATE_KEY, or HEDERA_ACCOUNT_ID + HEDERA_PRIVATE_KEY is required",
   );
   process.exit(1);
 }
 
 // Network configuration (alphabetic order)
 const AVM_NETWORK = "algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI="; // Algorand Testnet
+const CCD_NETWORK = ccdNetwork; // Concordium (default: testnet)
```
phdargen: Build fails with:
```
@x402/advanced-facilitator-example:build: all_networks.ts:127:58 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}:${string}`'.
@x402/advanced-facilitator-example:build: 
@x402/advanced-facilitator-example:build: 127   const [host, port] = parseGrpcUrl(getConcordiumGrpcUrl(CCD_NETWORK));
@x402/advanced-facilitator-example:build:                                                              ~~~~~~~~~~~
@x402/advanced-facilitator-example:build: 
@x402/advanced-facilitator-example:build: all_networks.ts:135:24 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}:${string}` | `${string}:${string}`[]'.
@x402/advanced-facilitator-example:build: 
@x402/advanced-facilitator-example:build: 135   facilitator.register(CCD_NETWORK, new ExactConcordiumScheme({ signer }));
```

Please make sure to build and test the examples before committing  

```suggestion
const CCD_NETWORK = (process.env.CCD_NETWORK as Network | undefined) ?? CONCORDIUM_TESTNET_CAIP2; // Concordium (default: testnet)
```

examples/typescript/clients/advanced/all_networks.ts
```diff
@@ -67,6 +72,17 @@ async function main(): Promise<void> {
     console.log(`Initialized AVM account: ${avmSigner.address}`);
   }
 
+  // Register Concordium scheme if wallet export path is provided
```
phdargen: Client example fails with:
```
Initialized CCD account: 3UrcxPQeYywasrPcYUcqhvFu3SB2vBBDjj7TsaRQ431vGiczYp

Making request to: http://localhost:4021/weather

Error: Failed to create payment payload: amount must be a non-empty decimal string
```

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,289 @@
...
+   * @param _ - Extension keys to apply (unused)
+   * @returns Enhanced payment requirements
+   */
+  enhancePaymentRequirements(
+    requirements: PaymentRequirements,
+    supportedKind: {
+      x402Version: number;
+      scheme: string;
+      network: Network;
+      extra?: Record<string, unknown>;
```
phdargen, line 115: `extra` in the 402 carries unused fields that are not documented in the spec, please remove: extra.type , extra.symbol and extra.decimals. Only keep extra.feePayer

typescript/packages/mechanisms/concordium/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,241 @@
...
+      requirements.asset === "" ||
+      requirements.asset.toUpperCase() === "CCD";
+
+    const sponsorAccountAddress = AccountAddress.fromBase58(feePayer);
+
+    const metadata = {
+      sender: this.signer.accountAddress,
+      nonce: nonceResult.nonce,
+      expiry: TransactionExpiry.futureMinutes(5),
+    };
```
phdargen, line 129: Client hardcodes an arbitrary 5mins expiry instead of reading `requirements.maxTimeoutSeconds`.  
This breaks whenever the server sets a value below 300 and can expire early when the server needs longer than 300
```suggestion
    const nowSeconds = Math.floor(Date.now() / 1000);
    const metadata = {
      sender: this.signer.accountAddress,
      nonce: nonceResult.nonce,
      expiry: TransactionExpiry.fromEpochSeconds(
        nowSeconds + requirements.maxTimeoutSeconds - 5,
      ),
    };
```

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,871 @@
...
+    this.maxExpiryOffsetSeconds = config.maxExpiryOffsetSeconds ?? MAX_EXPIRY_OFFSET_SECONDS;
+  }
+
+  /**
+   * Returns extra metadata for the /supported endpoint.
+   *
+   * @param _ - Network identifier (unused, same config for all networks)
+   * @returns Randomly selected fee payer metadata
+   */
+  getExtra(_: Network): Record<string, unknown> | undefined {
```
phdargen, line 88: Signer selection ignores network

getExtra, getSigners, and resolveSigner all ignore the network argument even though each FacilitatorConcordiumSigner from toConcordiumFacilitatorSigner is bound to one gRPC endpoint with mainnet + testnet signers in a single ccd:* instance, /supported can put a testnet feePayer on a mainnet kind and verify/settle will run nonce/balance checks and submission against the wrong chain. 

Suggest to key signers by network:
```
getExtra(network: Network) {
  const pool = this.signersFor(network);
  return { feePayer: pool[Math.floor(Math.random() * pool.length)].getAddress() };
}
private resolveSigner(address: string, network: Network) {
  return this.signersFor(network).find(s => s.getAddress() === address);
}
```

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -110,6 +123,21 @@ const server = new x402ResourceServer(facilitatorClient);
...
   server.register(AVM_NETWORK, new ExactAvmScheme());
 }
+if (ccdAddress) {
+  const concordiumServer = new ExactConcordiumScheme()
+    .registerAsset(CCD_NETWORK, CCD_DEFAULT_STABLECOIN_SYMBOL, CCD_DEFAULT_STABLECOIN_DECIMALS)
+    .registerMoneyParser(async amount => ({
+      amount: amount.toString(),
+      asset: CCD_DEFAULT_STABLECOIN_SYMBOL,
+      extra: {
+        type: "plt",
```
phdargen: testnet PLT seem difficult to get, suggest to use CCD for the examples

But please test both locally

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -110,6 +123,21 @@ const server = new x402ResourceServer(facilitatorClient);
 if (avmAddress) {
   server.register(AVM_NETWORK, new ExactAvmScheme());
 }
+if (ccdAddress) {
+  const concordiumServer = new ExactConcordiumScheme()
+    .registerAsset(CCD_NETWORK, CCD_DEFAULT_STABLECOIN_SYMBOL, CCD_DEFAULT_STABLECOIN_DECIMALS)
+    .registerMoneyParser(async amount => ({
```
phdargen: There is no SDK built-in default for PLT/stablecoin. 

The other mechanism (SVM, EVM, Hedera, Stellar, Aptos, AVM) have a `defaultMoneyConversion` that converts a decimal price like $0.01 → atomic units for the chain’s default asset.

Mapping $0.001 → EURR (an euro dominated stablecoin) would also be semantically wrong. If you want to define a default dollar denominated stablecoin, please do so in sdk-level `defaultMoneyConversion` fct, all other PLTs should use the explicit: `{ amount, asset: "EURR" }`

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -110,6 +123,21 @@ const server = new x402ResourceServer(facilitatorClient);
...
+      amount: amount.toString(),
```
phdargen: amount should also be converted to atomic units here

typescript/packages/mechanisms/concordium/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,241 @@
...
+   * @param amount  - Transfer amount in smallest token units (string)
+   * @param tokenId - Token identifier (e.g. "EURR")
+   * @param decimals - Token decimal places
+   * @returns A transaction builder for a PLT token transfer
+   */
+  private buildPltTransfer(payTo: string, amount: string, tokenId: string, decimals = 6) {
+    const ops = [
+      {
+        [TokenOperationType.Transfer]: {
+          amount: TokenAmount.fromDecimal(parseFloat(amount), decimals),
```
phdargen: requirements.amount is in atomic units but TokenAmount.fromDecimal interprets its input as a decimal value and multiplies by 10^decimals. Every PLT payment would fail with `amount_mismatch`

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,871 @@
...
+    decodedPayload: DecodedPayload,
+  ): string | null {
+    const required = BigInt(getRequiredAmount(requirements));
+
+    if (expectedAsset.toUpperCase() !== "CCD") {
+      if (decodedPayload.amount === null) return "invalid_amount_format";
+      return decodedPayload.amount === required
+        ? null
+        : `amount_mismatch: required ${required}, got ${decodedPayload.amount}`;
+    }
```
phdargen, line 547: PLT checkAmount only compares TokenAmount.value, but decimals is client-controlled in the signed CBOR. Both should be treated as untrusted. A crafted { value: 5000000, decimals: 8 } passes for a 5000000 requirement on a 6-decimal token while transferring 0.05 EURR, not 5. Validate exponent against onchain token decimals (gRPC), then compare atomic value to requirements.amount:

```
if (!TokenAmount.instanceOf(amount)) return "invalid_token_amount";

const expectedDecimals = await signer.getTokenDecimals(payload.tokenId);
if (amount.decimals !== expectedDecimals) return "invalid_token_amount_decimals";
if (amount.value !== BigInt(requirements.amount)) return "amount_mismatch";
```

pavelhbl:
Latest updates pushed and CI is green on `ac366de4`:

**Merge & alignment**
- Resolved 22 merge conflicts with `main` (upstream)
- Lint + format clean across all TypeScript, Go, Python

**On-chain integration tests (all three asset types verified)**
- ✅ CCD native transfer
- ✅ EURR PLT token transfer
- ✅ USDR PLT token transfer (new test case added)
- ✅ HTTP middleware CCD flow
- All 11 integration tests pass against Concordium testnet

**CI (all passing)**
- verified-commits ✅
- lint (ts, go, python, examples) ✅
- format (ts, go, examples) ✅
- test (ts matrix 22/24, go, python) ✅

Ready for re-review when you have bandwidth.

pavelhbl:
@phdargen I have funded Concordium testnet wallets (CCD + EURR + USDR) ready for you to run the on-chain integration tests. What is the best way to share the credentials with you — email, DM, or another channel?

phdargen:
> @phdargen I have funded Concordium testnet wallets (CCD + EURR + USDR) ready for you to run the on-chain integration tests. What is the best way to share the credentials with you — email, DM, or another channel?

You can send the tokens to 3UrcxPQeYywasrPcYUcqhvFu3SB2vBBDjj7TsaRQ431vGiczYp
or email me at philippe.dargent@coinbase.com

typescript/site/package.json
```diff
@@ -6,7 +6,7 @@
   "packageManager": "pnpm@11.1.1",
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "next build --webpack",
```
phdargen: please revert

DEFAULT_ASSETS.md
```diff
@@ -132,3 +132,26 @@ The default asset is chosen **per chain** based on:
 | **TypeScript** | `typescript/packages/mechanisms/evm/src/shared/defaultAssets.ts` | `DEFAULT_STABLECOINS` |
 | **Go** | `go/mechanisms/evm/constants.go` | `NetworkConfigs` |
 | **Python** | `python/x402/mechanisms/evm/constants.py` | `NETWORK_CONFIGS` |
+
+## Concordium Default Asset
+
+Concordium uses **CCD** (the native token) as its default asset for USD-denominated prices. CCD has 6 decimals, so `$0.001` converts to `1000` microCCD.
```
phdargen: CCD is not pegged to USD, so semantically this doesnt make sense. 

The default money conversion of eg `$0.001` is reserved to a dollar nominated stablecoin. So either let it resolve to a USD stablecoin or do not implement any default money conversion if there is no USD stablecoin or you don't want to pick one

e2e/servers/express/index.ts
```diff
@@ -2,14 +2,13 @@ import express from "express";
...
 import { ExactAvmScheme } from "@x402/avm/exact/server";
+import { ExactConcordiumScheme } from "@x402/concordium/exact/server";
 import { ExactEvmScheme } from "@x402/evm/exact/server";
 import { UptoEvmScheme } from "@x402/evm/upto/server";
 import { BatchSettlementEvmScheme } from "@x402/evm/batch-settlement/server";
 import { ExactSvmScheme } from "@x402/svm/exact/server";
 import { ExactAptosScheme } from "@x402/aptos/exact/server";
 import { ExactHederaScheme } from "@x402/hedera/exact/server";
-import { KEETA_TESTNET_CAIP2 } from "@x402/keeta";
-import { ExactKeetaScheme } from "@x402/keeta/exact/server";
```
phdargen, line 12: Please resolve merge conflicts, this shouldnt remove existing networks

Please also try to keep the diff minimal throughout all the e2e scripts, so its easier to review

typescript/packages/mechanisms/concordium/_get_tokens.ts
```diff
@@ -0,0 +1,35 @@
+import { ConcordiumGRPCNodeClient, credentials } from "@concordium/web-sdk/nodejs";
+import { AccountAddress } from "@concordium/web-sdk";
+
+const ADDRESS = process.argv[2] || "381h88FbHED8W2ofENoa5bWD1ri5tXmbwp9PqJptvqj1qJUDo3";
+
+/** Helper script to fetch token info from Concordium testnet account. */
+async function main() {
```
phdargen: this file seems unused, please remove

e2e/clients/axios/index.ts
```diff
@@ -26,42 +27,26 @@ import { ExactTvmScheme } from "@x402/tvm/exact/client";
...
 const endpointPath = process.env.ENDPOINT_PATH as string;
 const url = `${baseURL}${endpointPath}`;
-const evmAccount = privateKeyToAccount(process.env.EVM_PRIVATE_KEY as `0x${string}`);
-const svmSigner = await createKeyPairSignerFromBytes(
-  base58.decode(process.env.SVM_PRIVATE_KEY as string),
-);
-
+const ccdWalletPath = process.env.CCD_WALLET_PATH;
+const ccdPrivateKey = process.env.CCD_PRIVATE_KEY;
+const ccdAddress = process.env.CCD_ADDRESS;
```
phdargen: is CCD_ADDRESS really needed? cant this be inferred from CCD_PRIVATE_KEY?

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -126,6 +142,37 @@ if (avmPrivateKey) {
...
+    ccdFacilitatorAddress,
+    ccdFacilitatorPrivateKey,
+    { host, port, useTls: true },
+  );
+
+  facilitator.register(CCD_NETWORK, new ExactConcordiumScheme({ signer }));
+  console.info(`CCD Facilitator account: ${ccdFacilitatorAddress} on ${CCD_NETWORK}`);
+}
+
+// Register Concordium scheme if wallet export path is provided (backward-compatible fallback).
```
phdargen: There is no need for backward-compatible fallbacks for unreleased code. 

Its fine to keep this option in the SDK, but please remove the wallet file fallback from all examples + e2e tests and remove the readFileSync imports

e2e/clients/axios/index.ts
pavelhbl: Checked this against the Concordium SDK v12 (`@concordium/web-sdk`). The `AccountSigner` returned by `buildBasicAccountSigner(privateKey)` only exposes two methods:

```ts
// signHelpers.d.ts
export interface AccountSigner {
    sign(digest: ArrayBuffer): Promise<AccountTransactionSignature>;
    getSignatureCount(): bigint;
}
```

No `getAddress()`, no public key export, no derivation path. `AccountAddress` has `fromBase58(string)` and `fromBuffer(32_address_bytes)` but no `fromPrivateKey()` or `fromPublicKey()`.

Deriving the address manually would require: Ed25519 pubkey derivation → SHA-256 → take first 29 bytes → Concordium-specific base58check with version byte 1. That is fragile outside the SDK and not worth the complexity for an env var.

Kept `CCD_ADDRESS` as required. Happy to revisit if the SDK adds address-from-key derivation in a future version.

examples/typescript/facilitator/advanced/all_networks.ts
pavelhbl: Fixed — hardcoded `CCD_NETWORK` to the testnet CAIP-2 literal (`"ccd:4221332d34e1694168c2a0c0b3fd0f27"`), same pattern as `AVM_NETWORK`, `EVM_NETWORK`, `HEDERA_NETWORK`, etc. in this example. The `const` literal type satisfies `` `${string}:${string}` `` so `TS2345` is resolved.

e2e/clients/axios/index.ts
pavelhbl: The Concordium SDK does support setup from a private key, `buildBasicAccountSigner(hexPrivateKey)` works. The issue is that Concordium **cannot derive the account address from the private key alone**, unlike EVM (`address = keccak256(pubkey)[12:]`) or SVM. The address is a protocol-specific base58check encoding of a hash of the public key, and the SDK has no `deriveAddressFromPrivateKey()` function.

So there are two supported paths for Concordium credentials:

| Path | Env vars | How |
|------|----------|-----|
| **Private key + address** | `CCD_PRIVATE_KEY` + `CCD_ADDRESS` | `buildBasicAccountSigner(key)` + address passed separately |
| **Wallet export file** | `CCD_WALLET_PATH` | `parseWallet(file)` → contains both address and keys |

Both are already implemented in `signer.ts` as two overloads of `toConcordiumFacilitatorSigner()`. The `CCD_ADDRESS` env var exists because the private key alone is insufficient, this is a Concordium protocol constraint, not an x402 convention.

If you prefer we only support the wallet-export path (one env var instead of two), we can drop `CCD_PRIVATE_KEY`/`CCD_ADDRESS` and use only `CCD_WALLET_PATH`. But the two-variable path matches how every other mechanism works (private key from env), and the address is just an extra required parameter for Concordium specifically.

examples/typescript/clients/advanced/README.md
andynelson: In the browser extension wallet, after testnet account creation completes, choosing the "activity" icon in that account gives you a "Request CCD" button, that will do a one-per-account airdrop of 20k CCD

e2e/clients/axios/index.ts
phdargen: OK CCD_PRIVATE_KEY + CCD_ADDRESS is fine, CCD_WALLET_PATH should not be used

examples/typescript/clients/advanced/all_networks.ts
```diff
@@ -33,6 +36,7 @@ config();
 
 // Configuration - optional per network
 const avmPrivateKey = process.env.AVM_PRIVATE_KEY as string | undefined;
+const ccdWalletPath = process.env.CCD_WALLET_PATH as string | undefined;
```
phdargen: Please remove CCD_WALLET_PATH and replace with pk + address

e2e/README.md
```diff
@@ -137,19 +150,31 @@ SERVER_TVM_ADDRESS=...              # Where servers receive TVM payments
...
 TONCENTER_API_KEY=...               # Recommended when TVM_PROVIDER=toncenter
 ```
 
+For a focused Concordium shared-harness run after configuring those values:
+
+```bash
+cd e2e
+pnpm test --testnet --families=ccd --facilitators=typescript --clients=axios,fetch --servers=express,fastify,hono,next --min -v
+```
+
```
phdargen, line 168: please remove this

e2e/README.md
```diff
@@ -89,6 +89,17 @@ Add the `-v` flag to any command for verbose output:
...
+### Concordium Offline Coverage
+
+The repo-level Concordium exact flow currently has a dedicated offline e2e regression:
+
+```bash
+pnpm test:concordium-offline
+```
+
+This covers the repo-level Concordium request/verify/settle flow without requiring
+live Concordium wallets or network access.
```
phdargen: please remove this

e2e/README.md
```diff
@@ -118,6 +129,7 @@ Required environment variables (set in `.env` file):
 CLIENT_EVM_PRIVATE_KEY=0x...        # EVM private key for client payments
 CLIENT_SVM_PRIVATE_KEY=...          # Solana private key for client payments
 CLIENT_APTOS_PRIVATE_KEY=...        # Aptos private key for client payments (hex string)
+CLIENT_CCD_WALLET_PATH=...          # Concordium wallet export path for client payments
```
phdargen: replace with pk+add

e2e/test.ts
```diff
@@ -636,6 +637,9 @@ async function runTest() {
   const clientSvmPrivateKey = process.env.CLIENT_SVM_PRIVATE_KEY;
   const clientAvmPrivateKey = process.env.CLIENT_AVM_PRIVATE_KEY;
   const clientAptosPrivateKey = process.env.CLIENT_APTOS_PRIVATE_KEY;
+  const clientCcdWalletPath = process.env.CLIENT_CCD_WALLET_PATH;
```
phdargen: please remove CLIENT_CCD_WALLET_PATH, FACILITATOR_CCD_WALLET_PATH everywhere

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -65,6 +69,7 @@ const STELLAR_NETWORK = "stellar:testnet" as const; // Stellar Testnet
 const HEDERA_HBAR_ASSET = "0.0.0" as const; // Native HBAR asset id
 const HEDERA_WEATHER_PRICE_TINYBARS = "100000" as const; // 0.001 HBAR
 const TVM_NETWORK = (process.env.TVM_NETWORK || "tvm:-3") as Network; // TON Testnet
+const CCD_WEATHER_PRICE_MICRO_CCD = "1000" as const; // 0.001 CCD
```
phdargen, line 72: There is still an issue with the decimals, this gets multiplied by 10^6 internally, so the requested price is 1000 CCD instead of the intended 0.001 CCD
```
{
  "x402Version": 2,
  "error": "Payment required",
  "resource": {
    "url": "http://localhost:4021/weather",
    "description": "Weather data",
    "mimeType": "application/json"
  },
  "accepts": [
    {
      "scheme": "exact",
      "network": "ccd:4221332d34e1694168c2a0c0b3fd0f27",
      "amount": "1000000000",
      "asset": "CCD",
      "payTo": "3UrcxPQeYywasrPcYUcqhvFu3SB2vBBDjj7TsaRQ431vGiczYp",
      "maxTimeoutSeconds": 300,
      "extra": {
        "feePayer": "381h88FbHED8W2ofENoa5bWD1ri5tXmbwp9PqJptvqj1qJUDo3"
      }
    },
    }
  ]
}
```

examples/typescript/servers/advanced/README.md
```diff
@@ -67,13 +67,24 @@ cp .env-local .env
...
+>
+> **Concordium Testnet (CCD):**
+>
+> 1. Set up Concordium Wallet for Web on **Testnet**.
+> 2. Open the account.
+> 3. Go to **Activity**.
+> 4. Click **Request CCD**.
+> 5. Wait for the test CCD transfer to arrive. Official guide: https://docs.concordium.com/en/mainnet/docs/plt/setup-guide/request-ccd.html
+>
+> **Concordium Testnet (PLT):** The `all-networks` example uses `EURR` as the default Concordium stablecoin parser for USD-style prices. There is no universal public faucet for arbitrary PLT symbols, so use an issuer-provided test distribution for `EURR` or request your own test PLT issuance: https://docs.concordium.com/en/mainnet/tutorials/plt/request-plt.html
```
phdargen, line 81: incorrect, please fix or remove

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,291 @@
...
+    if (!asset) {
+      throw new Error(`Unknown asset: ${assetSymbol}`);
+    }
+
+    return {
+      amount: convertToTokenAmount(String(price.amount), asset.decimals),
+      asset: asset.symbol,
+      extra: {
+        ...((price.extra as Record<string, unknown>) ?? {}),
+        decimals: asset.decimals,
```
phdargen: there should be no server provided `decimals` in extra

If client trusts a server-declared `extra.decimals`, a malicious/misconfigured server can understate decimals so the same small-looking atomic amount signs a far larger numeric transfer 

Atomic units should be used throughout. Server declares price in atomic unit, client creates payload in atomic units. Facilitator verifies amount in atomic units 

If for some reason `decimals` is needed somewhere in the flow (this has to be well justified), then this should be retrieved from onchain state via RPC

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,291 @@
...
+   *
+   * @param price - Asset amount to parse
+   * @param network - Network identifier
+   * @returns Parsed asset amount
+   */
+  private parseAssetAmount(price: AssetAmount, network: Network): AssetAmount {
+    const assetSymbol = price.asset || "";
+
+    if (!assetSymbol || assetSymbol.toUpperCase() === "CCD") {
+      const amount = convertToTokenAmount(String(price.amount), CCD_NATIVE.decimals);
```
phdargen: this gets atomic units as input, no need to convert with `decimals`

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,291 @@
...
+    // Treat raw numbers/non-USD strings as CCD amounts
+    const tokenAmount = convertToTokenAmount(
+      typeof price === "number" ? price.toString() : (price as string),
+      CCD_NATIVE.decimals,
+    );
+    return {
+      amount: tokenAmount,
+      asset: "CCD",
+      extra: {},
+    };
```
phdargen: asset should be required, no silent fallback to CCD

Server declares eg:
{ amount: "1000", asset: "CCD" } 
or
{ amount: "1000", asset: "USDR" } 
with amount in atomic units

price: "10" or price: "$0.01" → throws unless a money parser is registered

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,291 @@
...
+   */
+  registerAsset(network: Network, symbol: string, decimals: number): this {
+    const asset: ConcordiumAssetInfo = {
+      type: "plt",
+      symbol: symbol.toUpperCase(),
+      decimals,
+    };
+    this.assets.set(this.assetKey(network, symbol), asset);
+    return this;
+  }
```
phdargen: an explicit asset registration should not be needed 

This whole file can be simplified substantially removing registerAsset, getAsset, getSupportedAssets, getSupportedSymbols, assetKey, toWholeUnits as well as AssetType, ConcordiumAssetInfo

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,291 @@
+import {
```
phdargen: This should be sufficient
```
import {
  AssetAmount,
  Network,
  PaymentRequirements,
  Price,
  SchemeNetworkServer,
  MoneyParser,
} from "@x402/core/types";
import { parseMoneyString } from "@x402/core/utils";

/**
 * Concordium server scheme for exact payments.
 *
 */
export class ExactConcordiumScheme implements SchemeNetworkServer {
  readonly scheme = "exact";

  /** Custom money parser chain — tried in registration order before default */
  private moneyParsers: MoneyParser[] = [];

  /**
   * Registers a custom money parser in the parser chain.
   *
   * Parsers are tried in registration order. Return `null` to skip to the
   * next parser. Concordium has no built-in default Money asset.
   *
   * @param parser - Custom function returning AssetAmount or null
   *
   * @returns ExactConcordiumScheme instance
   */
  registerMoneyParser(parser: MoneyParser): this {
    this.moneyParsers.push(parser);
    return this;
  }

  /**
   * Parse price into AssetAmount.
   *
   * Supports:
   * - AssetAmount: { amount: "1000", asset: "CCD" } -> pass-through atomic units
   * - String/number Money: requires a registered money parser (no default money parser)
   *
   * @param price - Price to parse
   * @param network - Network identifier
   * @returns Parsed asset amount
   */
  async parsePrice(price: Price, network: Network): Promise<AssetAmount> {
    if (this.isAssetAmount(price)) {
      return this.parseAssetAmount(price, network);
    }

    const amount = this.parseMoneyToDecimal(price);

    for (const parser of this.moneyParsers) {
      const result = await parser(amount, network);
      if (result !== null) return result;
    }

    if (typeof price === "string" && price.startsWith("$")) {
      throw new Error(
        `Cannot resolve USD-denominated price to a Concordium asset. ` +
        `Register a money parser via registerMoneyParser() to map USD prices ` +
        `to a specific token (e.g., USDR). Raw amount: ${amount}`,
      );
    }

    throw new Error(
      "No default asset for Concordium: register a money parser via registerMoneyParser() " +
      "or pass an explicit AssetAmount in atomic units",
    );
  }

  /**
   * Enhance payment requirements with facilitator-announced fee payer metadata.
   *
   * @param requirements - Payment requirements to enhance
   * @param supportedKind - Supported payment kind configuration
   * @param supportedKind.x402Version - X402 protocol version
   * @param supportedKind.scheme - Payment scheme identifier
   * @param supportedKind.network - Network identifier
   * @param supportedKind.extra - Extra facilitator metadata
   * @param _ - Extension keys to apply (unused)
   * @returns Enhanced payment requirements
   */
  enhancePaymentRequirements(
    requirements: PaymentRequirements,
    supportedKind: {
      x402Version: number;
      scheme: string;
      network: Network;
      extra?: Record<string, unknown>;
    },
    _: string[],
  ): Promise<PaymentRequirements> {
    return Promise.resolve({
      ...requirements,
      extra: {
        ...((requirements.extra as Record<string, unknown>) ?? {}),
        feePayer: supportedKind.extra?.feePayer,
      },
    });
  }

  /**
   * Parse custom asset amount (pass-through atomic units).
   *
   * @param price - Asset amount to parse
   * @returns Parsed asset amount
   */
  private parseAssetAmount(price: AssetAmount, network: Network): AssetAmount {
    if (!price.asset) {
      throw new Error(`Asset must be specified for AssetAmount on network ${network}`);
    }

    return {
      amount: String(price.amount),
      asset: price.asset.toUpperCase(),
      extra: price.extra ?? {},
    };
  }

  /**
   * Parses Money (string | number) to a plain decimal number.
   * Strips leading `$` if present.
   *
   * @param money - Raw price to parse
   * @returns formatted decimal number
   */
  private parseMoneyToDecimal(money: string | number): number {
    if (typeof money === "number") return money;
    return parseMoneyString(money);
  }

  /**
   * Type guard to check if price is an AssetAmount.
   *
   * @param price - Price to check
   * @returns True if price is an AssetAmount
   */
  private isAssetAmount(price: Price): price is AssetAmount {
    return typeof price === "object" && price !== null && "amount" in price;
  }
}
```

typescript/packages/mechanisms/concordium/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,241 @@
...
+    const signable = isNativeCcd
+      ? this.buildCcdTransfer(requirements.payTo, requirements.amount)
+          .addMetadata(metadata)
+          .addSponsor(sponsorAccountAddress)
+          .build()
+      : this.buildPltTransfer(
+          requirements.payTo,
+          requirements.amount,
+          requirements.asset,
+          (requirements.extra as Record<string, unknown> | undefined)?.decimals as
```
phdargen: if decimals is really needed here, client should get it from rpc call and not trust the server

pavelhbl:
> **Round 5 review comments addressed** — 4 commits pushed to `pavel/concordium-exact-mechanism-sync`.

### Commits
| Commit | Scope |
|--------|-------|
| `877b0cee` | **Server scheme rewrite** (D1-D5) — 291→140 lines, asset required, no silent CCD fallback, atomic units only |
| `9d682f5c` | **CCD_WALLET_PATH → pk+addr** (A1-A6) — 7 files, `parseWallet` → `buildBasicAccountSigner` |
| `7e152e12` | **Docs + metadata** (B1, F1, O1) — README alignment, testnet instructions, package.json |
| `8ec905ac` | **Tests** (D1-D3, E1, O2) — +16 unit tests, +7 integration tests |

### Fix mapping
| Theme | Comments | What changed |
|-------|----------|--------------|
| **A** — wallet path | 3458410211, 3458418082, 3458433756, 3458438001, 3458431046 | `CCD_WALLET_PATH` env var removed everywhere; signer built via `buildBasicAccountSigner(privateKey)` + `CCD_ADDRESS` |
| **B** — offline e2e | 3458431974 | Removed "Concordium Offline Coverage" section from `e2e/README.md` |
| **C** — server example price | 3458755978 | Already correct: `{ amount: "1000", asset: "CCD" }` in atomic microCCD |
| **D** — server scheme | 3459172802, 3459181804, 3459217159, 3459241373, 3459285621 | Applied phdargen's rewrite: `asset` required (throws if missing), no silent CCD fallback, `registerAsset`/`getAsset`/`toWholeUnits` removed, decimals not leaked in extra |
| **E** — client decimals | 3459297826 | `requirements.extra?.decimals` → `grpcClient.getTokenInfo(tokenId)` — fetches from chain |
| **F** — testnet README | 3458763502 | Simplified: "Activity → Request CCD" (20k testnet CCD) |
| **O1** — package.json | — | Already `"author": "x402 Foundation"` |
| **O2** — tx simulation | — | `preflightLikelyToSucceed()` checks nonce + balance; new unit + integration tests cover preflight rejection paths |

### Verification
- `tsc --noEmit` ✅
- Lint ✅
- Unit tests: **122/122 pass** (87 concordium-specific)
- Integration tests: tsc/lint clean, require testnet env vars to run

typescript/packages/mechanisms/concordium/src/exact/client/scheme.ts
pavelhbl: Addressed in `877b0cee`. The client no longer trusts server-provided decimals.

**Before**: `requirements.extra?.decimals` — server could understate decimals to make prices appear cheaper.

**After**: `buildPltTransferWithChainDecimals()` at `client/scheme.ts:179` fetches decimals from the chain via `grpcClient.getTokenInfo(tokenId)`:
```ts
const tokenInfo = await grpcClient.getTokenInfo(tokenId);
const decimals = tokenInfo.state.decimals;
```

No fallback to server-provided values — if gRPC is unavailable, the client throws (fail-closed). Zero references to `extra?.decimals` remain in client or server source.

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
pavelhbl: Re: `getAssetDecimals` — should Concordium implement it?

Noticed EVM implements the optional `getAssetDecimals` (used by `resolveSettlementOverrideAmount` for dollar-format overrides). Stellar and Hedera don't.

For Concordium it's unclear whether it makes sense:
- CCD isn't a stablecoin, so there's no universal default asset with fixed decimals
- PLT tokens (EURR, USDR) vary — most are 6 decimals but not guaranteed
- The scheme already rejects dollar-format prices unless a money parser is registered

Two options:
1. **Skip it** — consistent with the "no silent fallback" design. If someone needs settlement overrides in dollar format, they register a money parser.
2. **Add it returning 6** — pragmatic, matches the interface default, covers the common case.

Leaning toward (1) since it aligns with the scheme's explicit-over-implicit philosophy, but wanted your take.

typescript/packages/mechanisms/concordium/src/exact/facilitator/scheme.ts
pavelhbl: Acknowledged — this is a real gap. EVM's `FacilitatorEvmSigner` exposes `getAddresses()` returning an array, while Concordium's `FacilitatorConcordiumSigner` only has `getAddress()` returning a single string.

The fix touches two layers:
1. **Signer interface** — `getAddress()` → `getAddresses(): readonly string[]`
2. **Facilitator scheme** — add signer selection logic (round-robin, random, or by network) instead of always using the single signer

Tracked this in our plan. Given the scope of the current round (server scheme rewrite, RPC decimals, pk+addr migration), this feels like a follow-up PR — it's a purely additive change that doesn't block the other fixes. Happy to tackle it next if you prefer it in this PR though.

phdargen: OK not a blocker, can be done in followup

typescript/packages/mechanisms/concordium/src/exact/server/scheme.ts
phdargen: can be skipped, `getAssetDecimals` is only needed for chains with a default USD stablecoin registry

phdargen (approved):

phdargen:
Thanks for the update @pavelhbl, looks good! 
Let's get this in, will do a fast-follow with a few minor fixes to prepare it for the release

Issue #2421, opened 2026-05-21
Proposal: `risk-check` extension — Counterparty risk verification for x402 payments


AlexanderLawson17:
## Problem

x402 batch settlement made sub-cent micropayments economically rational. However, there is no protocol-level mechanism for counterparty risk verification within the payment flow. Resource servers that need to verify the trustworthiness of a payer must implement bespoke solutions outside the protocol, losing the cost-amortization benefits of batched settlement.

A 5-cent risk check on a 20-cent individual payment destroys unit economics. But a single risk check amortized across thousands of batched voucher redemptions for the same payer is negligible. The protocol should support this pattern natively.

## High-Level Approach

A new optional extension (`risk-check`) that:

1. **Resource servers** declare a risk-check requirement in `PaymentRequired.extensions["risk-check"]`, including a provider discovery URL and minimum acceptable score.
2. **Facilitators** call the risk-check provider during verification, cache the result per-payer, and include the signed attestation in `VerifyResponse` and `SettleResponse` extensions.
3. **Servers** read the result from the settlement response and enforce their score threshold.

The extension follows the existing `{info, schema}` pattern used by `bazaar`, `payment-identifier`, `sign-in-with-x`, etc.

### Key design choices:
- **Provider-agnostic** — Any service publishing `/.well-known/risk-check.json` qualifies as a provider. The spec does not mandate a specific scoring methodology.
- **Optional at every layer** — Servers opt in by declaring the extension. Facilitators that don't support it ignore it and proceed normally. The extension never blocks payments by default.
- **JWS-verified** — Results carry a compact JWS attestation verifiable against the provider's published JWKS, so servers don't need to trust the facilitator's representation of the score.
- **Batch-friendly** — Facilitators can cache a risk check per payer and amortize it across all voucher settlements for that payer within the TTL window.

## Why Existing Specs Don't Cover This

- **`sign-in-with-x`** solves authentication (proving wallet ownership) but not trust assessment.
- **`payment-identifier`** provides idempotency, not risk evaluation.
- **`bazaar`** handles discovery/cataloging, not counterparty scoring.
- **ERC-8004** provides agent identity registration but its Validation registry lacks a standard pattern for third-party risk attestations.

The `risk-check` extension fills the gap between identity (knowing who the payer is) and trust (knowing whether they're safe to transact with).

## Specification

I have a complete draft spec ready at: `specs/extensions/risk-check.md`

It covers:
[...]

PR #2422, opened 2026-05-22
Extension: risk-check — Counterparty risk verification for x402 payments

AlexanderLawson17:
## Summary

- Adds `risk-check` extension spec at `specs/extensions/risk-check.md`
- Adds TypeScript types (`RiskCheckExtensionInfo`, `RiskCheckResult`, `RiskCheckDiscovery`) at `typescript/packages/core/src/types/risk-check.ts`
- Fixes Go/TypeScript parity: adds missing `Extensions` field to Go `VerifyResponse` and `SettleResponse`
- Adds Go types (`RiskCheckExtensionInfo`, `RiskCheckResult`) at `go/types.go`

Discussed in #2421.

## Design

Resource servers declare a minimum risk score in `PaymentRequired.extensions["risk-check"]`. Facilitators call a risk-check provider during verification, cache the result per payer, and include the signed attestation in `SettleResponse.extensions["risk-check"]`. One check amortized across all voucher settlements for the same payer within the TTL window.

Key design choices:
- **Provider-agnostic**: any service publishing `/.well-known/risk-check.json` qualifies
- **JWS-verified**: results carry a compact JWS attestation verifiable against the provider's JWKS
- **Optional at every layer**: servers opt in, facilitators that don't support it ignore it
- **Batch-friendly**: facilitators cache per payer, amortize across batched voucher settlements

## AI Disclosure

This contribution was made with AI assistance (Claude) and has been human-reviewed.

## Reference Implementation

[Revettr](https://revettr.com) implements this pattern in production with facilitator-facing endpoints (`POST /v1/risk-check`, `POST /v1/risk-check/batch`) returning the exact `RiskCheckResult` format from this spec.

### DIFF
--- a/go/types.go
+++ b/go/types.go
@@ -124,7 +124,7 @@ type SettlementOverrides struct {
 type RiskCheckExtensionInfo struct {
 	Required     bool     `json:"required"`
 	RiskCheckURL string   `json:"risk_check_url,omitempty"`
-	MinScore     int      `json:"min_score,omitempty"` // Valid range: 0-100 inclusive
+	MinScore     *int     `json:"min_score,omitempty"` // Pointer type: 0 is valid (no minimum threshold). Valid range: 0-100 inclusive
 	Categories   []string `json:"categories,omitempty"`
 }
 
@@ -132,7 +132,7 @@ type RiskCheckExtensionInfo struct {
 // VerifyResponse and SettleResponse extensions["risk-check"]
 type RiskCheckResult struct {
 	Checked    bool     `json:"checked"`
-	Score      int      `json:"score,omitempty"`
+	Score      *int     `json:"score,omitempty"` // Pointer type: 0 is a valid score (highest risk), nil means unchecked
 	Tier       string   `json:"tier,omitempty"`
 	Provider   string   `json:"provider,omitempty"`
 	Categories []string `json:"categories,omitempty"`
--- a/go/types.go
+++ b/go/types.go
@@ -119,6 +119,29 @@ type SettlementOverrides struct {
 	Amount string `json:"amount,omitempty"`
 }
 
+// RiskCheckExtensionInfo represents the risk-check extension info
+// advertised by a resource server in PaymentRequired.extensions["risk-check"].info
+type RiskCheckExtensionInfo struct {
+	Required     bool     `json:"required"`
+	RiskCheckURL string   `json:"risk_check_url,omitempty"`
+	MinScore     int      `json:"min_score,omitempty"`
+	Categories   []string `json:"categories,omitempty"`
+}
+
+// RiskCheckResult represents the risk-check result included in
+// VerifyResponse and SettleResponse extensions["risk-check"]
+type RiskCheckResult struct {
+	Checked    bool     `json:"checked"`
+	Score      int      `json:"score,omitempty"`
+	Tier       string   `json:"tier,omitempty"`
+	Provider   string   `json:"provider,omitempty"`
+	Categories []string `json:"categories,omitempty"`
+	JWS        string   `json:"jws,omitempty"`
+	JWKSURL    string   `json:"jwks_url,omitempty"`
+	CheckedAt  string   `json:"checked_at,omitempty"`
+	ExpiresAt  string   `json:"expires_at,omitempty"`
+}
+
 // ResourceConfig defines payment configuration for a protected resource
 type ResourceConfig struct {
 	Scheme            string                 `json:"scheme"`
--- a/go/types.go
+++ b/go/types.go
@@ -124,7 +124,7 @@ type SettlementOverrides struct {
 type RiskCheckExtensionInfo struct {
 	Required     bool     `json:"required"`
 	RiskCheckURL string   `json:"risk_check_url,omitempty"`
-	MinScore     int      `json:"min_score,omitempty"`
+	MinScore     int      `json:"min_score,omitempty"` // Valid range: 0-100 inclusive
 	Categories   []string `json:"categories,omitempty"`
 }
 
--- a/specs/extensions/risk-check.md
+++ b/specs/extensions/risk-check.md
@@ -293,7 +293,9 @@ Facilitators that do not support the extension MUST ignore it and proceed with n
 
 ## Server Enforcement
 
-When `required` is `true`, the resource server MUST check `VerifyResponse.extensions["risk-check"]` or `SettleResponse.extensions["risk-check"]`:
+When `required` is `true`, the resource server MUST enforce the risk check **before serving the resource**. In the standard verify-then-settle flow, this means checking `VerifyResponse.extensions["risk-check"]`; the resource MUST NOT be served if verification does not include a passing risk check. `SettleResponse.extensions["risk-check"]` carries the same attestation and is used as confirmation for settlement-time auditing and for batch or deferred-settlement flows where enforcement is delegated to settle.
+
+The server MUST apply the following checks against the chosen response:
 
 1. If `checked` is `false` or the `risk-check` key is absent, the server MUST reject the request (the facilitator did not support the extension or did not perform the check).
 2. If `score` is below `min_score`, the server MUST reject the request with an appropriate error.
--- a/specs/extensions/risk-check.md
+++ b/specs/extensions/risk-check.md
@@ -320,7 +320,7 @@ When `required` is `false`, the risk-check result is informational. The server M
 - **Provider**: [Revettr](https://revettr.com) — counterparty risk scoring for x402 agent commerce
   - Discovery: `https://revettr.com/.well-known/risk-check.json`
   - JWKS: `https://revettr.com/.well-known/jwks.json`
-  - Scoring: `POST https://revettr.com/v1/score`
+  - Risk Check: `POST https://revettr.com/v1/risk-check`
   - Attestation: `POST https://revettr.com/v1/attest` (free tier)
   - SDK: `pip install revettr` ([PyPI](https://pypi.org/project/revettr/))
 
--- a/specs/extensions/risk-check.md
+++ b/specs/extensions/risk-check.md
@@ -25,7 +25,8 @@ Risk-check providers publish a discovery document at `/.well-known/risk-check.js
   "name": "Example Risk Provider",
   "version": "0.1.0",
   "description": "Counterparty risk scoring for x402 agent commerce",
-  "endpoint": "/v1/score",
+  "endpoint": "/v1/risk-check",
+  "batch_endpoint": "/v1/risk-check/batch",
   "method": "POST",
   "pricing": {
     "amount": "10000",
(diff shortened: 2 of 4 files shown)


AlexanderLawson17:
@erikreppel @rohinlohe @skaliski — would appreciate your review when you get a chance. This adds a `risk-check` extension (spec + TS types + Go types) as discussed in #2421. Happy to adjust based on feedback.

AlexanderLawson17:
@phdargen @CarsonRoscoe @apmcdermott — tagging you as well since you've been active on recent spec and SDK PRs. Would appreciate your review on the extension spec and the Go/TS type additions.

TateLyman:
Ran a no-payment spec/reference-implementation pass on the `risk-check` extension.

Repro:

```bash
curl -sS https://revettr.com/.well-known/risk-check.json | jq .
curl -sS https://revettr.com/.well-known/jwks.json | jq .
curl -sS https://revettr.com/.well-known/revocations.json | jq .
curl -sS -X POST https://revettr.com/v1/risk-check \
  -H 'content-type: application/json' \
  --data '{"wallet_address":"0x0000000000000000000000000000000000000001"}' | jq .
curl -sS https://revettr.com/openapi.json | jq '.components.schemas.ScoreRequest,.components.schemas.RiskCheckResult,.components.schemas.RiskCheckBatchRequest'
```

What looks good:
- Reference discovery is live at `/.well-known/risk-check.json`.
- JWKS and revocation list are live.
- `POST /v1/risk-check` returns the proposed `RiskCheckResult` shape with `checked`, `score`, `tier`, `provider`, `categories`, compact JWS, `jwks_url`, `checked_at`, and `expires_at`.
- The extension’s direction is useful for agent-payment flows because it makes payer/counterparty risk a protocol-visible settlement concern instead of every resource server inventing its own side channel.

Spec/type notes I would tighten before this becomes a pattern:
- P1: The spec says facilitators that do not support the extension **MUST ignore it and proceed**, while server enforcement for `required: true` is described after successful settlement. That can create an awkward charge-then-deny path for a required risk check if a server only learns the facilitator ignored it after settlement. I’d make the required-risk decision happen from `VerifyResponse` before serving/settling, or require a clear `isValid: false` / `invalidReason: "risk-check-unsupported"` when `required: true` cannot be satisfied.
- P1: Treat `payer_wallet` / `payer_domain` appended by the client as hints, not authoritative inputs. The facilitator should derive the payer wallet from the validated payment payload wherever possible, otherwise a client can self-assert the low-risk subject while paying from a different wallet.
- P2: The reference discovery document has a `facilitator.endpoint` and `batch_endpoint`, while the spec’s discovery example and `RiskCheckDiscovery` TS type only define a generic `endpoint`. Since `/v1/score` returns `ScoreResponse` and `/v1/risk-check` returns the exact `RiskCheckResult`, facilitators need a typed/discoverable RiskCheckResult endpoint, not just a scoring endpoint.
- P2: `min_score` is documented as `0-100`, but the TS/Go types are plain numbers/ints. A JSON schema or validation guidance would avoid `min_score: 1000`/negative thresholds becoming implementation-defined.
- P2: For JWS verification, it would help to define the minimum required claims and binding rules: `iss`, `sub`/payer, `score`, `tier`, `categories`, `iat`, `exp`, `input_hash`, and optionally `aud` or resource/facilitator binding. Otherwise a server can verify that a provider signed something, but not necessarily that the attestation applies to this payment/resource context.

I did not send payment headers, sign requests, or call any paid x402 path.

AlexanderLawson17:
Thanks @TateLyman — this is a thorough review and all five points land. Addressing each:

**P1: Charge-then-deny on `required: true`**

Agreed. The current flow creates a window where settlement succeeds but the server only discovers the facilitator skipped the risk check after the fact. Will update the spec so that when `required: true`, the facilitator MUST either:
- Include a passing `risk-check` result in `VerifyResponse`, OR
- Return `isValid: false` with `invalidReason: "risk-check-unsupported"` if it doesn't support the extension

This moves the decision gate to verification, before settlement occurs.

**P1: Client-asserted `payer_wallet` / `payer_domain` as hints**

Good catch. Will add explicit language: "Facilitators SHOULD derive the payer wallet from the validated payment payload (e.g., the signing address). Client-supplied `payer_wallet` and `payer_domain` are supplementary hints and MUST NOT override the facilitator's derived identity."

**P2: Discovery type vs live implementation mismatch**

The `RiskCheckDiscovery` type has a single `endpoint` field but Revettr's live discovery serves `facilitator.endpoint` and `facilitator.batch_endpoint`. Will update the TS type and spec discovery example to include `batch_endpoint` as an optional field, and clarify that `endpoint` returns `RiskCheckResult` format (not `ScoreResponse`).

**P2: `min_score` validation**

The JSON schema in the spec already constrains `min_score` to `"minimum": 0, "maximum": 100` — but the TS/Go types don't enforce this at the type level. Will add runtime validation guidance and a JSDoc/godoc note documenting the 0-100 constraint.

**P2: JWS required claims**

Will add a "JWS Claims" subsection to the spec defining minimum required claims: `iss` (provider DID), `sub` (payer wallet — facilitator-derived, not client-asserted), `score`, `tier`, `iat`, `exp`, and RECOMMENDED `aud` (resource server URL) for context binding. This addresses both the "signed something" vs "signed something relevant" distinction and ties back to P1 on payer identity.

I'll push updates addressing all five. Thanks for running the no-payment repro against the live endpoints — exactly the kind of implementation-vs-spec check that tightens the extension.

AlexanderLawson17:
All five items addressed in the latest push (15ba819):

- **P1 charge-then-deny:** Facilitator Behavior now includes item 6 requiring `isValid: false` with `invalidReason: "risk-check-unsupported"` when `required: true` and the facilitator does not support the extension. Server Enforcement strengthened from MAY to MUST for the `required: true` + `checked: false` case.
- **P1 client-asserted identity:** Added identity derivation rule — facilitators MUST derive payer wallet from the validated payment payload. Client-supplied fields are supplementary hints only.
- **P2 discovery:** Added `batch_endpoint` to the discovery example JSON, plus a full Discovery Fields table (the section was the only one without one).
- **P2 min_score bounds:** Added JSDoc constraint in TS and inline comment in Go documenting 0-100 range.
- **P2 JWS claims:** New JWS Claims subsection with 9 defined claims (`iss`, `sub`, `score`, `tier`, `iat`, `exp` required; `aud`, `categories`, `input_hash` recommended) and context binding rules for replay prevention.

Discovery endpoint in the example also updated from `/v1/score` to `/v1/risk-check` to match the `RiskCheckResult` return type. Welcome another pass when you have time.

TateLyman:
Re-read the latest patch against my earlier notes. The important blockers are addressed from the public spec side now:

- required-but-unsupported risk checks now fail verification with `isValid: false` / `risk-check-unsupported` instead of silently charging and denying later;
- payer identity is facilitator-derived from the validated payment payload, with client-supplied wallet/domain treated only as supplementary hints;
- discovery now separates the single and batch endpoint shape;
- `min_score` is explicitly bounded to 0-100;
- signed attestations now have concrete required JWS claims and context-binding guidance.

Only small wording suggestion left: for routes that serve immediately after verify, I would bias required-risk enforcement language toward `VerifyResponse` first and treat `SettleResponse` as extra confirmation for settlement/batch flows. That is not a blocker from my read; the previous launch-risk issues are materially closed.

AlexanderLawson17:
Thanks @TateLyman — agreed on the VerifyResponse-first framing. Pushed 1dbcbdd which rewords the Server Enforcement section:

- VerifyResponse is now the primary enforcement surface for the standard verify-then-settle flow ("the resource MUST NOT be served if verification does not include a passing risk check").
- SettleResponse is explicitly framed as confirmation for settlement-time auditing and for batch / deferred-settlement flows.

This closes the charge-then-deny ambiguity the prior "VerifyResponse or SettleResponse" wording allowed.

TateLyman:
Took the follow-up pass you invited. The VerifyResponse-first enforcement change closes the charge-then-deny ambiguity cleanly.

Two remaining small spec/type nits I would fix before merge:

1. Go `RiskCheckResult.Score` currently has `json:"score,omitempty"`. A valid score of `0` means highest risk and is still required when `checked=true`; with `omitempty`, Go implementations can serialize the most important failing score as absent. Same shape may apply to any required numeric fields where zero is semantically valid. I would remove `omitempty` from `Score` or make it pointer-based if absence vs zero matters.

2. The reference implementation section still lists `POST https://revettr.com/v1/score`, while the discovery example and earlier comment moved the endpoint shape to `/v1/risk-check` returning `RiskCheckResult`. If both endpoints exist, spell out which one is normative; otherwise update the reference line so implementers do not copy the old path.

Everything else I checked from the prior list looks addressed: unsupported-required now fails verification, payer identity is facilitator-derived, batch discovery is represented, and JWS claims have enough context binding to avoid “signed something, but not this payment context.”

AlexanderLawson17:
Thanks @TateLyman — both addressed in 4e72115:

1. **Go `Score` / `MinScore` zero-value fix:** Both fields are now `*int` (pointer type). A valid score of `0` (highest risk) serializes correctly; `nil` represents the unchecked/absent case. Applied the same fix to `MinScore` on `RiskCheckExtensionInfo` since `min_score: 0` (no minimum threshold) is also semantically valid.

2. **Reference implementation endpoint:** Updated from `/v1/score` to `/v1/risk-check` to match the `RiskCheckResult` return type and the discovery example. Both endpoints exist on the live implementation, but `/v1/risk-check` is normative for the extension spec.

Appreciate the multiple passes on this — the spec is tighter for it.

chopmob-cloud:
@AlexanderLawson17 — reviewing the spec text in this PR, surfacing a normative-reference question that came up on #2421 and got resolved in adjacent threads since the spec drafting started.

## Canonicalisation discipline citation

The current spec specifies signed attestations under JWS ES256 + JWKS but doesn't pin a canonicalisation discipline for the signed bytes. For cross-implementor byte-identical interop on the attestation payload — particularly across batched voucher settlements where multiple facilitators sign over the same payer record — the spec needs a canonical preimage rule.

The citable form is [`draft-hopley-x402-canonicalisation-jcs-v1`](https://datatracker.ietf.org/doc/draft-hopley-x402-canonicalisation-jcs-v1/) (Independent Submission, Informational; sole-AlgoVoi authorship; -00 + -01 POSTED). It specifies JCS RFC 8785 + SHA-256 plus the schema-normalisation rules (integer-millisecond timestamps, ordered arrays as significant, in-band `canon_version` pin) that RFC 8785 alone doesn't fully pin. The URN is registered in §10.1: `urn:x402:canonicalisation:jcs-rfc8785-v1`. Citation convention for x402 receipt-format I-Ds is currently under discussion at [#2458](https://github.com/x402-foundation/x402/issues/2458).

## Verifier-side composition shape

The current `RiskCheckResult` shape collapses several operationally-distinct outcomes into a numeric score. A composable verifier-side response format is specified by [`draft-hopley-x402-composite-trust-query`](https://datatracker.ietf.org/doc/draft-hopley-x402-composite-trust-query/) (sole-AlgoVoi authorship; -00 + -01 POSTED 2026-05-25):

- Closed four-element outcome enumeration: `TRUSTED` / `PROVISIONAL` / `INSUFFICIENT_EVIDENCE` / `UNTRUSTED`
- Content-addressed `chain_ref` (sha256: of the audit chain root)
- Content-addressed `query_ref` (sha256: of the canonical bytes of the structured query the verifier evaluated)
- `verifier_did` for accountability, `ctq_timestamp_ms` for year-N reconstruction

The four-state enumeration preserves an operational distinction a numeric score collapses: `INSUFFICIENT_EVIDENCE` (gather more) vs `UNTRUSTED` (halt) drive different downstream actions. Worth considering for `RiskCheckResult` alignment, particularly if `risk-check` results are intended to compose with other evidence types under a single canonical preimage discipline (per the three-axis composition discussed on #2421: AlgoVoi `compliance_screen` + Supership `service_trust_v0` + cryptographic-cap evidence).

## Cumulative cross-validation

512/512 byte-for-byte agreements across 8 independent JCS implementations (Python, TypeScript, Go, Rust, Java, PHP inline-stdlib, .NET, Ruby) and 7 vector sets, attested at [`_attestations/2026-05-25-8-impl-5-format-cross-validation.md`](https://github.com/chopmob-cloud/algovoi-jcs-conformance-vectors/blob/main/_attestations/2026-05-25-8-impl-5-format-cross-validation.md). A risk-check conformance corpus would slot into the existing 8-impl runners mechanically.

The [Substrate Adopters Registry](https://docs.algovoi.co.uk/adopters) records parties publishing under the shared `canon_version: jcs-rfc8785-v1` pin — currently AlgoVoi, Supership / Crest Deployment Systems (`service_trust_v0` + `urn:crest:trust-check-v1`), and PEAC Protocol. Risk-check evidence emitters adopting the discipline would qualify for inclusion.

-- AlgoVoi

PR #2439, opened 2026-05-22
fix: reject malformed payment signature headers

### LINKED ISSUE #2397: Malformed PAYMENT-SIGNATURE should return 400 instead of 402
## Summary

When a request includes a malformed `PAYMENT-SIGNATURE` header, the HTTP resource server currently treats it the same as an absent payment header and returns `402 Payment Required`.

This makes it hard for clients to distinguish between:

- no payment attempt was provided
- a payment attempt was provided, but the header could not be decoded

## Expected behavior

If `PAYMENT-SIGNATURE` is present but cannot be decoded, the server should return `400 Bad Request` with a clear error body, for example `invalid_payment_signature_header`.

Requests with no payment header should continue returning the normal `402 Payment Required` response.

## Minimal repro

Send a protected-route request with an invalid header value:

```bash
curl -H 'PAYMENT-SIGNATURE: garbage-value' http://localhost:3000/protected
```

Actual: returns the same `402` response as a request with no payment header.

Expected: returns `400 Bad Request` because the supplied payment header is syntactically invalid.

peterxing:
## Summary

Fixes #2397.

Malformed `PAYMENT-SIGNATURE` headers are now treated as syntactically invalid client input instead of falling through to the no-payment path:

- present but undecodable `PAYMENT-SIGNATURE` returns `400 Bad Request`
- response body uses `invalid_payment_signature_header`
- absent payment header still returns the normal `402 Payment Required`
- invalid headers do not call facilitator verification

## Test

```bash
corepack pnpm --dir typescript/packages/core test test/unit/http/x402HTTPResourceService.test.ts
```

43 tests passed.

### DIFF
--- a/typescript/packages/core/src/http/x402HTTPResourceServer.ts
+++ b/typescript/packages/core/src/http/x402HTTPResourceServer.ts
@@ -519,7 +519,25 @@ export class x402HTTPResourceServer {
     const paymentOptions = this.normalizePaymentOptions(routeConfig);
 
     // Check for payment header (v1 or v2)
-    const paymentPayload = this.extractPayment(adapter);
+    let paymentPayload: PaymentPayload | null;
+    try {
+      paymentPayload = this.extractPayment(adapter);
+    } catch (error) {
+      return {
+        type: "payment-error",
+        response: {
+          status: 400,
+          headers: { "Content-Type": "application/json" },
+          body: {
+            error: "invalid_payment_signature_header",
+            message:
+              error instanceof Error
+                ? error.message
+                : "PAYMENT-SIGNATURE header could not be decoded",
+          },
+        },
+      };
+    }
 
     // Create resource info, using config override if provided
     const resourceInfo = {
@@ -1027,6 +1045,7 @@ export class x402HTTPResourceServer {
         return decodePaymentSignatureHeader(header);
       } catch (error) {
         console.warn("Failed to decode PAYMENT-SIGNATURE header:", error);
+        throw new Error("PAYMENT-SIGNATURE header could not be decoded");
       }
     }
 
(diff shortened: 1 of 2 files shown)


TateLyman:
Ran a local TypeScript core package pass on head `ce618b8` for this malformed `Payment-Signature` change.

Repro:

```bash
gh pr checkout 2439
cd typescript
pnpm --filter @x402/core test -- test/unit/http/x402HTTPResourceService.test.ts
```

Result: `17` test files passed, `466` tests passed. The new malformed-header case returns `400`, does not emit a `PAYMENT-REQUIRED` challenge, and does not call facilitator verification in the test path. I did not send payment headers, wallet signatures, or paid calls.

peterxing:
Status readback for maintainers: head ce618b8 is currently mergeable; the only reported commit status I see is the Vercel authorization failure for the Coinbase team deployment, not a code/test failure. The focused core test pass was reproduced above (17 files / 466 tests). No payment headers, wallet signatures, paid calls, or live settlement were used in this check.

peterxing:
Acknowledged on the verified-commits gate. The code path is still mergeable, but this branch has two unsigned commits from the original submission. I need to re-sign/re-push the branch cleanly rather than making any wallet/payment/live x402 changes here. I will update the PR once the commits are verified; current non-signing blockers remain the Coinbase Vercel team authorization status and maintainer review.

peterxing:
Follow-up after the signed-commit repair: current head `642abdc` has the verified-commits workflow green, is mergeable, and labeler is green. The only remaining reported status is the Coinbase/Vercel team authorization gate, so there is no outstanding commit-signing blocker from my side.

No wallet activity, payment headers, live x402 calls, or paid settlement were used.

phdargen:
Please respect our [AI-assisted contributions guidelines](https://github.com/x402-foundation/x402/blob/main/CONTRIBUTING.md#ai-assisted-contributions)

Happy to reopen once you verified intent and implementation of your commits and stopped posting those unreviewed AI-generated comments

PR #2436, opened 2026-05-22
spec: add shared JCS canonicalisation discipline v3 (jcs-rfc8785-v1)

chopmob-cloud:
## Summary

Adds `specs/canonicalisation.md` — the shared canonicalisation discipline section developed collaboratively on #2326, now at v3 with three-voice coalition sign-off (AlgoVoi + Vauban Pay + FeedOracle, 2026-05-21).

This is the normative section that `privacy_class` (PR #2334), `receipt_format` (#2357), and `evidenceType: cryptographic` (#2322) all reference rather than carrying parallel canonicalisation paragraphs that drift over revisions.

## What the section codifies

- `JCS_hash = SHA-256(JCS(RFC 8785)(object))`, lowercase hex
- Schema normalisation MUST precede canonicalisation (timestamps, field names, arrays, type validation)
- **Array ordering**: explicitly clarified — JCS preserves element order, does NOT sort arrays (production bug class surfaced in PR #2434 vector 0001)
- **Retention property** (FeedOracle, v3): cross-observer-across-time determinism for MiCA Art. 80 / AMLR Art. 56 / DORA Art. 14 retention obligations
- **`canon_version` scoping** (v3 refinement): SHOULD for general case; MUST for framework-bound statutory retention emitters
- **Versioning**: section independently versioned; current version `jcs-rfc8785-v1`
- **Conformance vectors**: 53 vectors / 37 pair invariants across 4 sets, 5 JCS implementations, 265 byte-for-byte agreements (in-tree mirror at `fixtures/canonicalisation-substrate/v0/` per PR #2412)

## Cross-validation status

| Impl | Library | Author set | Status |
|---|---|---|---|
| Python | `rfc8785@0.1.4` | Trail of Bits | 53/53 |
| JavaScript | `canonicalize@3.0.0` | erdtman | 53/53 |
| Go | `gowebpki/jcs v1.0.1` | Web PKI WG | 53/53 |
| Java | `cyberphone/json-canonicalization` | Anders Rundgren (RFC 8785 author) | 53/53 |
| Rust | `serde_jcs@0.2.0` | seritalien | 53/53 |

## Coalition sign-off record (from #2326)

- **AlgoVoi** (@chopmob-cloud): architectural skeleton, JCS discipline, per-chain anchor, 4-impl validation harness
[...]

PR #2432, opened 2026-05-22
Add bounded-spend-authorization-sample/v0 fixture set ; reference for fiscal_authority discussion

seritalien:
Public no-secret reference fixture set demonstrating the `DelegationGrant + SettlementReceipt` pair for cryptographically enforced bounded-spend authorization under x402 V2. Published per the 2026-05-22 commitment to @egoriklok on #2405 (`fiscal_authority` metadata block proposal) as a discrete artifact a buyer-agent can compare against when implementing or validating `fiscal_authority` v0.

## Contents (`fixtures/bounded-spend-authorization-sample/v0/`)

| File | Lines | Description |
|---|---|---|
| `schema/delegation-grant.schema.json` | 217 | JSON Schema Draft 2020-12, full VPSF Claim sextuplet per spec §4.5, inline `fiscal_authority` 7-field mapping |
| `schema/settlement-receipt.schema.json` | 197 | JSON Schema Draft 2020-12, full VPSF Claim sextuplet per spec §4.3 |
| `vectors/0001-baseline.json` | 105 | PASS, nominal grant + receipt, well within caps |
| `vectors/0002-cap-respect.json` | 105 | PASS edge case, amount = cap_per_tx − 1 |
| `vectors/0003-cap-violation.json` | 105 | FAIL_AT_VERIFIER, pinned-divergent-digest per AlgoVoi convention |
| `README.md` | 124 | Seven-field mapping table, cross-axis binding, Phase MVP vs Phase 2a delta declaration |

## Seven-field fiscal_authority mapping (egoriklok #2405)

Complete table in `README.md` ; summary :

| # | Field | Maps to |
|---|---|---|
| 1 | provider/resource identity | `SettlementReceipt.subject` + verifier contract address |
| 2 | exact charge amount + asset + network | `SettlementReceipt.predicate.amount/.currency` + `.anchor.chain_id` |
| 3 | approval source | `DelegationGrant.subject` + `.anchor.delegation_commitment` |
| 4 | spend cap + expiry | `DelegationGrant.predicate.cap_per_tx/.cap_per_period/.period_seconds` + `.temporal_frame.t_start/.t_end` |
| 5 | receipt/charge_evidence reference | `SettlementReceipt.predicate.intent_ref.claim_hash` + JCS hash of receipt |
| 6 | audit/canonicalisation version | `canon_version` field (both schemas) |
| 7 | revocation or dispute path | `DelegationGrant.anchor.revocation_authority` |

## Cross-axis binding

[...]

Issue #2405, opened 2026-05-21
Proposal: fiscal_authority metadata for x402 resources with agent spend limits


egoriklok:
### Summary

As x402 resources become callable by autonomous agents and buyer agents, operators need a non-secret way to know whether a resource can trigger bounded paid action, what cap applies, what evidence is retained, and how access can be revoked.

I propose an optional `fiscal_authority` metadata block for x402/Bazaar-discoverable resources. This is not a payment credential and should not contain secrets. It is an operator/client readiness layer for agent spend governance.

### Problem

Today, an agent or buyer-agent can discover and call paid resources, but the fiscal-safety posture is often implicit:

- who or what approved spend;
- what spend cap applies;
- which endpoint/resource can charge;
- what charge or settlement evidence exists;
- where the audit trail lives;
- how access is revoked or reduced;
- what evidence would justify raising the cap.

Without this, client-side policy has to infer fiscal authority from docs or out-of-band trust. That makes it harder to safely increase agent spend limits.

### Proposed optional metadata shape

```yaml
extensions:
  bazaar:
    fiscal_authority:
      paid_action_surface: "endpoint | service | facilitator | other"
      approval_source: "human | policy_engine | allowance | budget_rule | none | unknown"
      spend_cap:
        amount: "string-or-number"
        asset: "USDC | USDT | fiat | other"
        interval: "per_call | per_session | daily | monthly | other"
      charge_evidence:
        type: "receipt | settlement_hash | facilitator_response | signed_attestation | other"
        secret_free: true
      audit_trail:
        available: true
        reference_type: "public_url | docs | api_endpoint | operator_provided | none"
      revocation_path:
        type: "api_key_revoke | allowance_reduce | facilitator_policy | server_disable | other"
      next_safe_threshold:
        requires: "string"
```

### Non-goals

- Do not store API keys, cookies, wallet secrets, private customer data, or payment credentials.
- Do not make the registry a compliance oracle.
- Do not imply a resource is safe or audited by default.
- Do not require every free x402 resource to carry this metadata.

### Why this may fit x402

x402 makes paid HTTP resources accessible to machine clients. As machine clients gain spend authority, resource metadata can help client policies decide whether to allow, review, block, or cap usage before payment execution.

This is especially useful for:

- buyer agents paying for API access;
- resource servers with multiple paid endpoints;
- facilitators or services exposing settlement evidence;
- operators raising an agent from test spend to real spend.

### Question

[...]

PR #2441, opened 2026-05-22
fixtures: filled fiscal_authority_readiness_card_v0 against Vauban bounded-spend (#2405)

seritalien:
## What this PR adds

A filled `fiscal_authority_readiness_card_v0` artifact, using the card shape @egoriklok proposed in #2405. This is the concrete example they asked for: one filled card against the Vauban bounded-spend fixture from PR #2432.

**New files:**
- `fixtures/fiscal-authority-readiness-card-sample/v0/0001-vauban-bounded-spend.json` -- the filled card
- `fixtures/fiscal-authority-readiness-card-sample/v0/README.md` -- field mapping, honest PARTIAL explanations, and preflight interpretation

## Mapping rationale

The source vector (`bounded-spend-authorization-sample/v0/vectors/0001-baseline.json`) contains a `DelegationGrant` + `SettlementReceipt`. The card surfaces the fiscally relevant fields:

- **authority.approval_source.kind = human_principal**: the grantor is a human who pre-delegated spend authority via a signed grant; the buyer agent needs no further human escalation at spend time within cap.
- **cap.enforcement = cryptographic**: the `DelegationGrant` is signed (HMAC-SHA-256 over JCS-canonical payload). The cap is not a facilitator policy assertion; it is locally verifiable against `grant_hash`.
- **cap bounds**: 1 000 000 micro-USDC per tx, 100 000 000 per 86 400-second period.
- **charge_evidence**: a live Starknet Sepolia tx hash anchors the settled receipt.

## Two honest PARTIAL statuses

Nothing is elided:

1. **canonicalisation = PARTIAL**: `canon_version` currently carries `"1.0"`. The normalised identifier (`jcs-rfc8785-v1`) is tracked under PR #2436. The algorithm is JCS-RFC8785 + SHA-256; the partial flag signals the version string is not yet finalised.

2. **revocation = PARTIAL**: `check_url` (`https://pay.vauban.tech/.well-known/revocation/v0/`) is declared in the grant anchor. The endpoint is not yet a live status surface in Vauban Pay Phase MVP. A buyer agent should resolve it at preflight; if no valid revocation status returns, treat the grant as unverified-for-revocation before spending.

## preflight_verdict

`buyer_agent_can_spend_without_human = true` because cryptographic cap + human_principal authority means the agent may spend within cap without further escalation. `status = PARTIAL` (not `READY`) because the revocation endpoint has not been confirmed live at card production time.

## Question for @egoriklok

Is the preferred landing path a companion PR like this one, or would you rather fold the filled card into the eventual #2405 PR when the card schema is finalised? Happy to rebase or close and re-open wherever makes the most sense for the fixture catalogue structure.

Ref: #2405

### DIFF
--- a/fixtures/fiscal-authority-readiness-card-sample/v0/0001-vauban-bounded-spend.json
+++ b/fixtures/fiscal-authority-readiness-card-sample/v0/0001-vauban-bounded-spend.json
@@ -1,0 +1,64 @@
+{
+  "resource": {
+    "provider_id": "vauban-pay",
+    "resource_id": "bounded-spend-authorization-sample-v0-0001-baseline",
+    "resource_url": "https://pay.vauban.tech"
+  },
+  "price": {
+    "status": "PASS",
+    "discovery_ceiling": {
+      "max_amount": "1000000",
+      "asset": "urn:eip3009:USDC:starknet-sepolia",
+      "network": "starknet-sepolia"
+    },
+    "receipt_amount_field": "settlement_receipt.predicate.amount"
+  },
+  "authority": {
+    "status": "PASS",
+    "approval_source": {
+      "kind": "human_principal",
+      "principal_id": "0xa3f2c1d4e5b6a7890123456789abcdef0123456789abcdef0123456789abcdef",
+      "policy_ref": null,
+      "policy_version": null,
+      "controller_id": null,
+      "snapshot_hash": null
+    }
+  },
+  "cap": {
+    "status": "PASS",
+    "enforcement": "cryptographic",
+    "max_per_tx": "1000000",
+    "max_per_period": "100000000",
+    "period": "86400",
+    "grant_hash": "sha256:ff558f21b26c5045fb6997fef185b9993e33a29a7907dbcf349e49b9019c1fd2"
+  },
+  "charge_evidence": {
+    "status": "PASS",
+    "type": "receipt",
+    "reference_url": "https://github.com/x402-foundation/x402/tree/main/fixtures/bounded-spend-authorization-sample/v0/vectors/0001-baseline.json",
+    "hash": "sha256:0adc9aece973486e8c49fc1613225d629d71bdd7e3b2fe50576f55d6db68a5b5",
+    "attestation_manifest_root": null
+  },
+  "canonicalisation": {
+    "status": "PARTIAL",
+    "canon_version": "1.0",
+    "algorithm": "JCS-RFC8785",
+    "hash_algorithm": "SHA-256",
+    "test_vectors_url": "https://github.com/x402-foundation/x402/tree/main/fixtures/bounded-spend-authorization-sample/v0/vectors/0001-baseline.json"
+  },
+  "revocation": {
+    "status": "PARTIAL",
+    "check_url": "https://pay.vauban.tech/.well-known/revocation/v0/",
+    "dispute_url": null,
+    "last_checked_ms": null
+  },
+  "preflight_verdict": {
+    "status": "PARTIAL",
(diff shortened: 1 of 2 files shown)


egoriklok:
@seritalien thanks for opening this. This is the right Vauban-side counterpart to the AlgoVoi signed-attestation card.

Two small alignment points after @chopmob-cloud's latest response in #2405:

1. The AlgoVoi card is now contributor-owned and committed at:
   `https://github.com/chopmob-cloud/algovoi-jcs-conformance-vectors/blob/main/fixtures/fiscal-authority-readiness-card/algovoi-signed-attestation/v0/card.json`
   commit `4d49978`.

   So this PR can either:
   - remain the x402-hosted filled Vauban card, with the README also linking to the external AlgoVoi card by URL/hash; or
   - convert into a pure x402-side discovery index, if Vauban prefers to maintain the authoritative `0001-vauban-bounded-spend` card in a Vauban-owned repo.

   I would avoid opening a second PR while this one exists; this PR is the coordination point.

2. I would keep `preflight_verdict.buyer_agent_can_spend_without_human = false` while `preflight_verdict.status = PARTIAL`, unless the card explicitly splits that into a narrower `local_cap_can_spend_without_human = true` signal.

   Reason: buyer-agent preflight is the primary consumer, and the current card still has unresolved revocation + canon version gaps. The cryptographic cap can prove local bounded authority, but the top-level autonomous-spend verdict should stay conservative until every required status is fresh/resolvable.

One schema nit that became clearer from AlgoVoi's filled card: cap values need an explicit `currency` / `unit` field, otherwise buyer agents cannot safely compare `1000000` micro-USDC, `GBP`, base units, or token-denominated caps across contributors.

No secrets or private payment surfaces needed here; this can stay entirely public/no-secret.

chopmob-cloud:
@egoriklok -- thanks for the confirmation. Three brief notes on the schema points raised.

**1. `currency` / `unit` field shape (the nit you flagged).** The AlgoVoi card exposed it because the EMR 2011 safeguarding regime denominates caps in GBP rather than the asset of settlement. The card's `cap` block carries:

```json
"cap": {
  "enforcement": "facilitator_policy",
  "max_per_tx": "100.00",
  "max_per_period": "300.00",
  "period": "per_account",
  "currency": "GBP",
  "grant_hash": null
}
```

The `currency` field is load-bearing wherever the cap denomination differs from the settlement asset. For cryptographic-cap variants where the grant denominates in the settlement asset (e.g. micro-USDC), the field still belongs and reads as the same value as `charge_evidence.asset`. Either way the schema gains a normative slot rather than the buyer-agent having to infer the cap unit from `grant_hash` or external context.

**2. `buyer_agent_can_spend_without_human = false` while PARTIAL.** The AlgoVoi card already uses this constraint (sample case verdict = false because price + authority are PARTIAL). The split you proposed (a narrower `local_cap_can_spend_without_human` signal) is a useful refinement -- it preserves the cryptographic-cap variant's ability to claim bounded local authority without overpromising the top-level autonomous-spend verdict.

**3. On the PR as coordination venue.** The discovery-index conversion (this PR becomes the x402-side index pointing at each contributor's externally-maintained card) is the cleaner long-term shape and matches the [Substrate Adopters Registry](https://docs.algovoi.co.uk/adopters) pattern. Symmetric across both contributors when each card lives in its own contributor-owned repo.

-- AlgoVoi

seritalien:
superseded by #2459 (egoriklok's cleaner discovery-index approach). card itself is at https://github.com/vauban-org/x402-stark-receipts-conformance/blob/main/cards/0001-vauban-bounded-spend.json. closing.

PR #2442, opened 2026-05-22
fix(core): reject null v2 payment extra

### LINKED ISSUE #2406: PaymentRequirements.extra: is `.nullable()` on the Zod schema intentional?
Following up on #1154 — the same family of v2 type/schema drift, but with a question about the schema's intent rather than just the TS type.

### The inconsistency

`PaymentRequirements.extra` in v2 core is the only optional record field marked `.nullable()` on the Zod side. Every other optional field uses `.optional()` only:

| Field | Zod definition |
|---|---|
| `ResourceInfo.description` | `z.string().optional()` |
| `ResourceInfo.mimeType` | `z.string().optional()` |
| `PaymentRequiredV2.error` | `z.string().optional()` |
| `PaymentRequiredV2.extensions` | `OptionalAny` (= `.optional().nullable()`) |
| `PaymentRequirementsV2.extra` | `OptionalAny` (= `.optional().nullable()`) |

[`typescript/packages/core/src/schemas/index.ts` L25](https://github.com/x402-foundation/x402/blob/main/typescript/packages/core/src/schemas/index.ts#L25)
```ts
export const OptionalAny = z.record(z.unknown()).optional().nullable();
```

### Other v2 SDKs do not accept null on the wire

[`go/types/v2.go` L31](https://github.com/x402-foundation/x402/blob/main/go/types/v2.go#L31)
```go
Extra map[string]interface{} `json:"extra,omitempty"`
```
`omitempty` skips the field when nil; null is not part of the serialization contract.

[`python/x402/schemas/payments.py` L52](https://github.com/x402-foundation/x402/blob/main/python/x402/schemas/payments.py#L52)
```python
extra: dict[str, Any] = Field(default_factory=dict)
```
Missing input becomes `{}`. `null` is not accepted into a `dict[str, Any]` field by Pydantic without an explicit `| None`.

So the v2 wire intent — read across Go + Python — appears to be "absent or empty dict, never null". Only TypeScript currently accepts null via Zod.

### Hand-written TS type also disagrees with Zod

[`typescript/packages/core/src/types/payments.ts` L19](https://github.com/x402-foundation/x402/blob/main/typescript/packages/core/src/types/payments.ts#L19)
```ts
extra: Record<string, unknown>;   // required
```

Required on the type, optional+nullable on the schema. Same TS-stricter-than-Zod pattern as #1154.

### Question

Is the `.nullable()` on `OptionalAny` intentional, or was it carried over without a wire-level reason? If unintentional:

```ts
// schemas/index.ts
export const OptionalAny = z.record(z.unknown()).optional();   // drop .nullable()

// types/payments.ts
extra?: Record<string, unknown>;
extensions?: Record<string, unknown>;
```

This would align TS with Go (`omitempty`) and Python (`default_factory`), and drop the only piece of v2 wire surface that currently allows JSON `null` for a record field.

If `.nullable()` is intentional (some scheme legitimately sends `extra: null`), happy to keep it and just fix the TS-type drift instead (`extra?: Record<string, unknown> | null`) — but it would be useful to document the case that requires it.

---
*AI-assisted draft; lines verified against `main`.*

peterxing:
## Summary
- introduce a non-nullable optional record schema for v2 metadata fields
- default missing `PaymentRequirementsV2.extra` to `{}` so parsed values still match the existing TS shape
- add coverage that omitted `extra` is accepted but `extra: null` is rejected

Closes #2406

## Verification
- `corepack pnpm --dir typescript/packages/core test test/unit/schemas/schemas.test.ts`
- `corepack pnpm --dir typescript/packages/core build`

### DIFF
--- a/typescript/packages/core/src/schemas/index.ts
+++ b/typescript/packages/core/src/schemas/index.ts
@@ -25,6 +25,13 @@ export type Any = z.infer<typeof Any>;
 export const OptionalAny = z.record(z.unknown()).optional().nullable();
 export type OptionalAny = z.infer<typeof OptionalAny>;
 
+/**
+ * Optional record schema with an empty object default.
+ * Used for v2 wire fields that may be absent but must not be `null` when present.
+ */
+export const OptionalRecord = z.record(z.unknown()).optional().default({});
+export type OptionalRecord = z.infer<typeof OptionalRecord>;
+
 // ============================================================================
 // Network Schemas
 // ============================================================================
@@ -138,7 +145,7 @@ export const PaymentRequirementsV2Schema = z.object({
   asset: NonEmptyString,
   payTo: NonEmptyString,
   maxTimeoutSeconds: z.number().positive(),
-  extra: OptionalAny,
+  extra: OptionalRecord,
 });
 export type PaymentRequirementsV2 = z.infer<typeof PaymentRequirementsV2Schema>;
 
--- a/typescript/packages/core/test/unit/schemas/schemas.test.ts
+++ b/typescript/packages/core/test/unit/schemas/schemas.test.ts
@@ -214,6 +214,13 @@ describe("x402 Schemas", () => {
         delete (withoutExtra as Record<string, unknown>).extra;
         const result = PaymentRequirementsV2Schema.safeParse(withoutExtra);
         expect(result.success).toBe(true);
+        expect(result.data?.extra).toEqual({});
+      });
+
+      it("should reject null extra field", () => {
+        const withNullExtra = { ...validPaymentRequirementsV2, extra: null };
+        const result = PaymentRequirementsV2Schema.safeParse(withNullExtra);
+        expect(result.success).toBe(false);
       });
     });
 


TateLyman:
Ran a local TypeScript core package pass on head `2d2fae4` for the v2 `extra: null` rejection change.

Repro:

```bash
gh pr checkout 2442
cd typescript
pnpm --filter @x402/core test -- test/unit/schemas/schemas.test.ts
```

Result: `17` test files passed, `466` tests passed. The schema test now accepts omitted v2 `extra` as `{}` and rejects explicit `extra: null`. No payment headers, wallet signatures, or paid calls were sent.

peterxing:
Status readback for maintainers: head \\2d2fae4\\ is mergeable from GitHub's current view; the only reported failing status I see is the Coinbase Vercel authorization gate, not a code/test failure. Review is still required.

The focused schema test pass was reproduced above (\\17\\ files / \\466\\ tests). No payment headers, wallet signatures, paid calls, live settlement, or chain activity were used in this check.

peterxing:
Follow-up after the commit-verification repair: current head `bb8a628` keeps the same final tree as the previous submission, is mergeable, and the verified-commits workflow is now green.

Current remaining blocker is still the Coinbase/Vercel team authorization gate plus maintainer review. No wallet activity, payment headers, live x402 calls, paid settlement, or chain activity were used.

phdargen:
Please respect our [AI-assisted contributions guidelines](https://github.com/x402-foundation/x402/blob/main/CONTRIBUTING.md#ai-assisted-contributions)

Happy to reopen once you verified intent and implementation of your commits and stopped posting those unreviewed AI-generated comments

PR #2200, opened 2026-05-05
feat: add service metadata on resource to all 3 SDKs

aayushijain21:
<!--
Thanks for contributing to x402!
Please fill out the information below to help reviewers understand your changes.

Note: We require commit signing.
See here for instructions: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification
-->

## Description

<!--
Please provide a clear and concise description of what the changes are, and why they are needed.
Include a link to the issue this PR addresses, if applicable (e.g. "Closes #123").
-->
Adds three optional service-metadata fields — `serviceName`, `tags`, `iconUrl` — to the top-level `resource` object on a `PaymentRequired` response, and formalizes them in the bazaar extension spec. Facilitators use these to enrich Bazaar search results with a provider-published name, topical tags, and an icon, with no admin step.
                                                                                                                                                                                    
The CDP facilitator will parse these fields defensively from 402 responses. This PR formalizes the convention upstream so all three SDKs and the foundation spec agree on the shape and validation rules.                                                                                                                                                                        
                                                                               
All changes are purely additive and backwards compatible:                                                                                                                         
  - New fields are optional in TS / Python / Go (`?:`, `| None = None`, `omitempty`); existing servers produce byte-identical 402 bodies.
  - New helpers and types are new exports — no existing function signature, field, or test was changed or removed.                                                                  
  - Spec change is a new section; existing tables and examples are untouched apart from one example gaining the three new keys.

## Tests

<!--
Please describe the tests you've performed to verify your changes.
Include relevant code samples, unit test cases, or screenshots if applicable.

For TypeScript: Run `pnpm test` from the `/typescript` directory
For Python: Run `uv run pytest` from the `python/x402/` directory
For Go: Run `go test ./...` from the `/go` directory
-->
New tests in each SDK cover the helpers (boundaries at 32/33 chars, tag truncation, `data:` / `file:` / `javascript:` / `ftp:` schemes, userinfo, IPv4 and IPv6 literals including `[::1]` and `[2001:db8::1]`, mixed-case `localhost`, control characters `\x00` / `\n` / `\x7f`, relative paths) plus end-to-end extraction cases for the all-valid path and the all-invalid soft-drop path.

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [x] I added a changelog fragment for user-facing changes (docs-only changes can skip)

<!--
[...]

PR #2447, opened 2026-05-23
feat(python): thread bazaar service metadata from RouteConfig to ResourceInfo

hypeprinter007-stack:
## Description

\`resource.{serviceName,tags,iconUrl}\` landed in the spec/schema via #2200 (merged 2026-05-06), and the Python \`ResourceInfo\` Pydantic model already carries the three optional fields. But the server-side \`RouteConfig\` dataclass has no fields to populate them, and \`x402_http_server_base.py\` hardcodes the \`ResourceInfo\` constructor to only the original three (\`url\`, \`description\`, \`mime_type\`).

The result: a server author who wants the Bazaar-rich metadata on their 402 has to bypass \`RouteConfig\` entirely or post-process the response — which is what motivated this PR. This wires the missing plumbing so the three fields flow through the documented path.

## Changes

- **\`python/x402/http/types.py\`** — add \`service_name\`, \`tags\`, \`icon_url\` to \`RouteConfig\` (all \`Optional\`, default \`None\`, placed after the existing \`mime_type\` field).
- **\`python/x402/http/x402_http_server_base.py\`**
  - \`_parse_route_config\`: accept both \`serviceName\`/\`iconUrl\` (camelCase) and snake_case, matching the existing \`mimeType\`/\`customPaywallHtml\` parsing convention.
  - \`ResourceInfo\` construction: pass the three new fields straight through. \`None\` values stay \`None\` so the existing Pydantic serialization behavior (excluding unset optionals) is preserved.
- **Tests** — \`test_route_config_service_metadata_flows_to_resource\` asserts the fields land on \`payment_required.resource\`. Modeled on \`test_route_option_extra_is_preserved\` directly above it; runs under both sync and async via the existing \`components_factory\` fixture.
- **Changelog** — fragment in \`changelog.d/route-config-service-metadata.feature.md\`.

## Validation

\`service_name\` / \`tags\` / \`icon_url\` validation rules (≤32 ASCII chars, ≤5 tags, ≤2048-char URL) live facilitator-side in \`extensions/bazaar/facilitator.py\` per the existing convention — this PR is plumbing only and intentionally doesn't duplicate that validation. The facilitator extractor (\`_sanitize_resource_service_metadata\`) continues to enforce the rules on read.

## Verified against the spec

Field locations match \`specs/extensions/bazaar.md\` "Service Metadata on resource" and the existing \`ResourceInfo\` schema in \`python/x402/schemas/payments.py\`. No new constants, no chain/asset assumptions.

## Disclosure

Used AI assistance (Claude Code) to scaffold the diff and test; reviewed both before submission.

### DIFF
--- a/python/x402/changelog.d/route-config-service-metadata.feature.md
+++ b/python/x402/changelog.d/route-config-service-metadata.feature.md
@@ -1,0 +1,1 @@
+Thread bazaar service metadata fields (`service_name`, `tags`, `icon_url`) from `RouteConfig` through to `ResourceInfo`. The wire-format schema fields landed in #2200 but the server-side `RouteConfig` had no way to populate them — servers wanting rich Bazaar listings had to bypass the SDK. This wires the missing plumbing.
--- a/python/x402/http/types.py
+++ b/python/x402/http/types.py
@@ -192,7 +192,16 @@ class PaymentOption:
 
 @dataclass
 class RouteConfig:
-    """Configuration for a payment-protected route."""
+    """Configuration for a payment-protected route.
+
+    ``service_name``, ``tags``, and ``icon_url`` are forwarded as-is to
+    ``ResourceInfo`` for Bazaar discovery. ``RouteConfig`` itself does not
+    enforce indexer-side limits — the stricter ``resource_info`` helper path
+    soft-drops values that exceed Bazaar's caps (service_name ≤ 32 chars,
+    ≤ 5 tags of ≤ 32 chars each, absolute http(s) icon URLs ≤ 2048 chars).
+    Server authors should keep names short, tag sets small, and icon URLs
+    absolute http(s) to ensure listings are accepted by indexers.
+    """
 
     accepts: PaymentOption | list[PaymentOption]
     resource: str | None = None
--- a/python/x402/http/types.py
+++ b/python/x402/http/types.py
@@ -198,6 +198,9 @@ class RouteConfig:
     resource: str | None = None
     description: str | None = None
     mime_type: str | None = None
+    service_name: str | None = None
+    tags: list[str] | None = None
+    icon_url: str | None = None
     custom_paywall_html: str | None = None
     unpaid_response_body: UnpaidResponseBody | None = None
     settlement_failed_response_body: SettlementFailedResponseBody | None = None
--- a/python/x402/http/x402_http_server_base.py
+++ b/python/x402/http/x402_http_server_base.py
@@ -194,6 +194,9 @@ class x402HTTPServerBase:
             resource=config.get("resource"),
             description=config.get("description"),
             mime_type=config.get("mimeType", config.get("mime_type")),
+            service_name=config.get("serviceName", config.get("service_name")),
+            tags=config.get("tags"),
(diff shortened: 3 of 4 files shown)


TateLyman:
Reviewed the Python diff from the no-payment/readiness angle. The main plumbing looks consistent: `RouteConfig` accepts `service_name` / `tags` / `icon_url`, `_parse_route_config` maps camelCase + snake_case, and `_process_request_core` threads them into `ResourceInfo` before the payment-required payload is built.

Two small coverage/doc suggestions, not blockers:

- The integration test currently proves the camelCase dict path (`serviceName`, `iconUrl`). Since `_parse_route_config` also accepts snake_case, I would add one assertion or tiny sibling test for `service_name` / `icon_url` as input so future refactors do not accidentally keep only the JS-style config path working.
- It may be worth one docs sentence that `RouteConfig` does not itself enforce Bazaar indexer limits; the stricter helper path later soft-drops non-conforming `service_name` / `tags` / `icon_url`. That would steer server authors toward short service names, small tag sets, and absolute http(s) icon URLs at the SDK boundary.

I did not send payment headers, sign requests, or attempt paid calls.

hypeprinter007-stack:
Thanks for the careful read, @TateLyman — both addressed in b6e8c27e:

1. Added `test_route_config_service_metadata_snake_case_input` as a sibling to the existing camelCase test. Same assertions, snake_case route-config keys (`service_name` / `icon_url`). Locks in both casing paths so a future refactor can't silently keep only one working.
2. Expanded the `RouteConfig` docstring to flag that it does not enforce Bazaar indexer caps and to point server authors at the limits (`service_name ≤ 32`, `≤ 5 tags of ≤ 32 chars each`, absolute `http(s)` icon URLs `≤ 2048`) at the SDK boundary.

Both tests pass locally (`pytest tests/integrations/test_http_integration.py -k "service_metadata or snake_case"`).

TateLyman:
Thanks, both updates line up with the readback: snake_case coverage is now explicit, and the RouteConfig docstring now sets the Bazaar limit expectations before teams ship metadata. No further notes from my side.

phdargen:
Thanks for the contribution @hypeprinter007-stack!

We require commit signing, happy to merge once you rebase with signed commits.

egoriklok:
Free public readiness snapshot for feat(python): thread bazaar service metadata from RouteConfig to ResourceInfo:

- Public signal: No-payment readiness snapshot draft for x402-foundation/x402#2447: feat(python): thread bazaar service metadata from RouteConfig to ResourceInfo. Focus: api, bazaar, payment, x402.
- R1 fit: api, bazaar, payment, x402.
- Buyer-agent check I would add before autonomous spend: explicit price/cap unit, policy/version hash, receipt or charge-evidence reference, revocation/dispute path, and allow/caution/block preflight result.
- This separates "interesting paid surface" from "safe for an agent wallet to use".

No invoice, no payment link, no delivery link, no private endpoint; this is only a free public snapshot.

hypeprinter007-stack:
@phdargen Rebased with SSH-signed commits — both showing `verified=true` on GitHub. Ready when you are.

phdargen (approved):

Issue #2386, opened 2026-05-20
Harden verify path: reject high-s ECDSA signatures (ECDSA malleability) across SDKs


Mameta29:
## Context

EIP-2 mandates low-s ECDSA signatures. The exact/eip3009 verify path across `@x402/evm` (TypeScript), `github.com/x402-foundation/x402/go`, and `x402` (Python) currently delegates ECDSA malleability rejection to the onchain `transferWithAuthorization` call. Canonical FiatTokenV2_2 USDC reverts high-s with `InvalidSignatureS`, which is already recognized by `parseEip3009TransferError` and friends.

This works in the common case but is fragile defense-in-depth-wise:

1. Other ERC-3009 token implementations are not guaranteed to enforce low-s. A non-canonical implementation accepted in `DEFAULT_STABLECOINS` would silently widen the malleability surface.
2. permit2 / batch-settlement paths have different onchain malleability semantics and may not surface a clean reverter reason.
3. Catching malleability **before** broadcast avoids burning gas on a doomed transaction and avoids any nonce/anti-replay edge case where the alternate `(r, n-s, v^1)` is observed as a distinct authorization at the facilitator layer (e.g. cache hit/miss boundaries, retry logic). Today, the facilitator's pre-broadcast simulate happens against viem/ethers/eth_keys / go-ethereum's `crypto.SigToPub`, none of which check `s <= n/2`.

I ran into this while writing a hardened EIP-3009-only facilitator for JPYC ([jpyc-x402-facilitator](https://github.com/Mameta29/jpyc-x402-facilitator)). The fix is a tiny pure function, but I wanted to surface it as an issue first because it touches signature verification on a value-transfer path.

## Where exactly

- TS: `typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts` — `verifyEIP3009` calls `signer.verifyTypedData(...)`. viem's `verifyTypedData` does not enforce low-s.
- Go: `go/mechanisms/evm/exact/facilitator/eip3009.go` — `VerifyTypedData` interface, ultimately go-ethereum `crypto.SigToPub`. No low-s check.
- Python: `python/x402/mechanisms/evm/verify.py` — `verify_eoa_signature` uses `eth_keys.keys.Signature(...).recover_public_key_from_msg_hash(hash)`. No `s` bound check.

[...]

PR #2454, opened 2026-05-25
fix(go): reject non-canonical EIP-3009 ECDSA signatures

wnjoon:
## Description

- Reject non-canonical high-s EIP-3009 ECDSA signatures in the Go SDK before verification/settlement
- Keep smart-wallet and ERC-6492 signatures unaffected by only applying the rejection to plain EOA signatures
- Add unit tests for canonical signature detection and EOA/smart-wallet classification
- Add a Go changelog fragment

This PR intentionally scopes #2386 to the Go EIP-3009 path. The TypeScript and Python parity work remains to be handled in follow-up PRs.

## Related Issue

Refs #2386

## Tests

- `GOCACHE=/tmp/go-build-cache go test ./mechanisms/evm/...`

## Checklist

### DIFF
--- a/go/.changes/unreleased/fixed-20260525-090533.yaml
+++ b/go/.changes/unreleased/fixed-20260525-090533.yaml
@@ -1,3 +1,3 @@
 kind: fixed
 body: Reject non-canonical high-s EIP-3009 ECDSA signatures before settlement in the Go SDK
-time: 2026-05-25T00:00:00+09:00
+time: 2026-05-25T09:05:33.109502+09:00
--- a/go/.changes/unreleased/fixed-20260525-eip3009-high-s.yaml
+++ b/go/.changes/unreleased/fixed-20260525-eip3009-high-s.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Reject non-canonical high-s EIP-3009 ECDSA signatures before settlement in the Go SDK
+time: 2026-05-25T00:00:00+09:00
--- a/go/mechanisms/evm/exact/facilitator/eip3009.go
+++ b/go/mechanisms/evm/exact/facilitator/eip3009.go
@@ -82,6 +82,14 @@ func (f *ExactEvmScheme) verifyEIP3009(
 		return nil, x402.NewVerifyError(ErrInvalidSignatureFormat, evmPayload.Authorization.From, err.Error())
 	}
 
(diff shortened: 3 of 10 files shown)


wnjoon:
@phdargen Reorganize commits (only including related tasks)

kakedashi3:
Nice work — the core logic checks out. The low-s threshold matches go-ethereum's `secp256k1halfN` (incl. the `s == n/2` boundary), the EOA-only gate (`GetCode` length 0) correctly skips ERC-1271/4337, and the 65-byte length discriminator skips ERC-1271/6492 wrappers. I also traced the settle path — `settleEIP3009` re-runs `verifyEIP3009`, so the check covers settle, not just verify. 👍

One thing before this merges: the PR `closes #2386`, but the issue's "Where exactly" / proposed fix asks for the EIP-3009 path across all three SDKs (TS `eip3009.ts`, Go `eip3009.go`, Python `verify.py`) with character-identical error consts. This is Go-only, so auto-closing would drop the TS + Python parity the issue explicitly requests. Might be cleaner to drop `closes` and keep #2386 open for the TS/Python follow-ups (or scope it down to Go).

Minor, from the thread: @Ectsang's two suggestions don't look incorporated yet — a one-line comment on *why* 65-byte length is the discriminator, and aligning the error name (`ErrHighSSignature` / `ErrSignatureSExceedsHalfN`) with the `InvalidSignatureS` reverter.

wnjoon:
@kakedashi3  @phdargen 

Thanks for the careful review. Agreed on the scope point: I’ll remove the auto-closing keyword and make this PR explicitly Go-only so #2386 can stay open for the TypeScript/Python parity follow-ups.

I’ll also incorporate the two smaller suggestions:
- add a short comment explaining why the 65-byte length check is the plain ECDSA discriminator,
- align the Go error name with the `InvalidSignatureS` terminology used by the EIP-3009 transfer reverter.

After that I’ll rerun the Go EVM tests and update the PR description accordingly.

wnjoon:
Follow-up done: rebased onto the latest `x402-foundation:main`, addressed the two minor suggestions, and kept this PR scoped to Go-only with `Refs #2386`.

Verified with `GOCACHE=/tmp/go-build-cache go test ./mechanisms/evm/...`.

wnjoon:
Rebased this PR onto the latest `x402-foundation:main` and force-pushed the updated branch.

wnjoon:
@phdargen Any thoughts?

PR #2456, opened 2026-05-25
fix(go): cache Go SVM exact mint metadata

### LINKED ISSUE #2216: feat(svm): cache immutable mint metadata in ExactSvmScheme to halve client RPC calls per payment
`ExactSvmScheme.createPaymentPayload()` ([scheme.ts](https://github.com/x402-foundation/x402/blob/main/typescript/packages/svm/src/exact/v1/client/scheme.ts)) makes two RPC calls per signed payment:

1. `fetchMint(rpc, asset)` — reads decimals + token program address
2. `rpc.getLatestBlockhash()` — required, expires every ~60 slots (~25s)

For stablecoin assets like USDC, (1) is wasted work — mint metadata is immutable for the lifetime of any reasonable client. Caching it would halve client-side RPC pressure and reduce p50 payment latency by one RTT.

### Use case

We run an x402-paid API gateway in front of Solana analytics endpoints. Under modest concurrency, agents using the SDK without a paid RPC hit `HTTP 429: Too Many Requests` on payload creation — the public mainnet RPC throttles at ~5 RPS for `getAccountInfo` (used by `fetchMint`) and the same window for `getLatestBlockhash`. So per-payment RPC cost is effectively `2 × public-RPC-quota`. With a mint cache, agents get ~2x the headroom on the same RPC budget.

This shows up in the wild quickly: in our integration tests, the 7th paid call in a tight loop (~5 sec window) reliably 429s on `fetchMint` even though the wallet has plenty of USDC.

### Proposal

In-memory cache keyed by `(network, mint address) → { decimals, programAddress }`, populated lazily on first fetch. A module-private `Map` is sufficient — no LRU needed, the keyspace is bounded by the number of distinct assets a client transacts with (typically 1: USDC).

Sketch:

```ts
const mintCache = new Map<string, { decimals: number; programAddress: Address }>();

async function getCachedMint(rpc: SchemeNetworkClient, network: Network, asset: Address) {
  const key = `${network}:${asset}`;
  const cached = mintCache.get(key);
  if (cached) return cached;
  const mint = await fetchMint(rpc, asset);
  const entry = { decimals: mint.data.decimals, programAddress: mint.programAddress };
  mintCache.set(key, entry);
  return entry;
}
```

Could also be opt-in via constructor config (`{ cacheMints?: boolean }`) for callers who explicitly want fresh reads, though I struggle to think of a real reason — SPL mint authority/decimals changes would break far more than x402.

### Related

- #1832 — injectable RPC client (orthogonal: helps throughput once you have a paid RPC, doesn't reduce calls)
- #288 — broader public-RPC-rate-limit pain (same root cause, different mitigation)
- #415 — added the `rpcUrl` option (closed; helps but doesn't reduce call count)

Happy to send a PR if maintainers are open to it.

wnjoon:
### Summary

- Cache SVM mint owner and decimals per Go exact client instance
- Reuse cached mint metadata in both V1 and V2 SVM exact client payload creation
- Add tests covering repeated payload creation and concurrent requests

### Why

The Go SVM exact client fetched the mint account on every payment payload creation to derive the token program and decimals. Those fields are stable for a mint, so repeated calls from the same client can reuse them and avoid unnecessary RPC requests.

## Tests

- `GOCACHE=/tmp/go-build-cache go test ./mechanisms/svm/...`

## Checklist

- [v] I have formatted and linted my code
- [v] All new and existing tests pass
- [v] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [v] I added a changelog fragment for user-facing changes (docs-only changes can skip)

 

## Related Issue

Closes #2216

### DIFF
--- a/go/.changes/unreleased/fixed-20260525-2216-svm-mint-cache.yaml
+++ b/go/.changes/unreleased/fixed-20260525-2216-svm-mint-cache.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Cache SVM mint metadata in exact clients to avoid repeated mint account RPC lookups.
+time: 2026-05-25T15:50:14+09:00
--- a/go/mechanisms/svm/exact/client/duplicate_tx_test.go
+++ b/go/mechanisms/svm/exact/client/duplicate_tx_test.go
@@ -27,6 +27,10 @@ const (
 )
 
 func mockSolanaRPCHandler(t *testing.T, blockhashFunc func() string) http.HandlerFunc {
+	return mockSolanaRPCHandlerWithAccountInfoCount(t, blockhashFunc, nil)
+}
+
+func mockSolanaRPCHandlerWithAccountInfoCount(t *testing.T, blockhashFunc func() string, accountInfoCalls *int32) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		var req struct {
 			Method string        `json:"method"`
@@ -71,6 +75,10 @@ func mockSolanaRPCHandler(t *testing.T, blockhashFunc func() string) http.Handle
 			})
 
 		case "getAccountInfo":
+			if accountInfoCalls != nil {
+				atomic.AddInt32(accountInfoCalls, 1)
+			}
+
 			mint := token.Mint{
 				MintAuthority:   nil,
 				Supply:          1000000000000,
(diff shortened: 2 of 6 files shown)


phdargen (approved):

phdargen:
Thanks @wnjoon, thats a good optimization

PR #2477, opened 2026-05-26
fix: aptos gas price vulnerability

CarsonRoscoe:
## Description

bound `gas_unit_price` on sponsored transactions.

- verify() in @x402/aptos checked max_gas_amount ≤ 500,000 on sponsored transactions but never validated gas_unit_price. Because the facilitator's hot wallet pays gas_used × gas_unit_price Octas, an attacker could submit a structurally valid USDC transfer with an arbitrarily large gas_unit_price and drain the fee-payer's APT balance — passing every existing check including simulation (simulation only fails when the fee payer literally cannot afford the gas, not when the price is merely inflated).
- Added MAX_GAS_UNIT_PRICE = 1_000n Octas (10× the Aptos protocol minimum of 100 Octas) to constants.ts and a bounds check in verify() inside the existing isSponsored guard, immediately after the max_gas_amount check and before the fee-payer signature step.
- Added facilitator-verify.test.ts with 7 offline unit tests. These use directly-constructed RawTransaction objects (no network calls) and a PAST_EXPIRATION sentinel to confirm gas-valid transactions pass through the gas guard and reach the expiration check — proving both the rejection path and the pass-through path without a live testnet. Also backfills equivalent boundary coverage for the pre-existing max_gas_amount check, which had no unit tests.

## Tests

Added unit tests

## Checklist

### DIFF
--- a/specs/schemes/exact/scheme_exact_aptos.md
+++ b/specs/schemes/exact/scheme_exact_aptos.md
@@ -115,7 +115,8 @@ Steps to verify a payment for the `exact` scheme:
 8. Verify the transfer amount matches `requirements.amount`.
 9. Verify the transfer recipient matches `requirements.payTo`.
 10. Verify the transaction sender has sufficient balance of the `asset` to cover the required amount.
-11. Simulate the transaction using the Aptos REST API to ensure it would succeed and has not already been executed/committed to the chain. This also validates the sequence number to prevent replay attacks.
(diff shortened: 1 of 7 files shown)


phdargen:
Looks great @CarsonRoscoe, could we also a brief note to the specs like `gas_unit_price MUST be bounded to prevent fee-payer draining`?

phdargen (approved):

PR #2480, opened 2026-05-26
fix: erc-6492 factory call was server driven

CarsonRoscoe:
## Description

When deployERC4337WithEIP6492 was enabled, settle() for EIP-3009 payments extracted the factory address and factoryCalldata verbatim from the payer's ERC-6492 signature wrapper and called signer.sendTransaction from the facilitator's funded EOA with no validation. An attacker could craft a malicious ERC-6492 wrapper to make the facilitator broadcast arbitrary on-chain calls — token drains, admin role hijacks, anything.

The fix adds a factory address allowlist to the facilitator config. Settlement now rejects any factory address absent from that list. The default is an empty allowlist, which denies all factory deployment calls — fail-safe by design.

### Changes

Core fix (Go v1+v2, TypeScript v1+v2, Python v1+v2):
- New config field: EIP6492AllowedFactories []string / eip6492AllowedFactories?: string[] / eip6492_allowed_factories: list[str]
- settle() checks the factory address against the allowlist (case-insensitive) before calling sendTransaction; rejects with eip6492_factory_not_allowed if not found
- Go: exported IsFactoryAllowed helper in eip3009_helpers.go shared between v1 and v2
- ExactEvmSchemeConfig in TypeScript scheme.ts now correctly threads eip6492AllowedFactories through to settleEIP3009 (was silently dropped before)
- Python register_exact_evm_facilitator helper updated to accept and forward eip6492_allowed_factories

**Examples** (all 13 call sites across Go, TypeScript, Python): explicit eip6492AllowedFactories: [] with a comment directing operators to populate it with their trusted factory addresses.

**E2E facilitators**: deployERC4337WithEIP6492 flipped to false — the current E2E suite only covers EOA payers; counterfactual smart wallet E2E testing is tracked separately.

## Tests

Updated unit tests

## Checklist


typescript/packages/mechanisms/evm/src/exact/facilitator/scheme.ts
```diff
@@ -16,9 +16,20 @@ export interface ExactEvmSchemeConfig {
...
   deployERC4337WithEIP6492?: boolean;
+  /**
+   * Allowlist of factory contract addresses (hex strings, case-insensitive) that the facilitator
+   * will call when deploying an undeployed smart wallet via ERC-6492.
+   *
+   * An empty or omitted list denies all factory calls even when deployERC4337WithEIP6492 is true.
+   *
+   * @default []
+   */
+  eip6492AllowedFactories?: string[];
```
phdargen: can we remove the `deployERC4337WithEIP6492` config then? Should deploy when `eip6492AllowedFactories` not empty and factory address is allowed

phdargen (approved):

PR #2482, opened 2026-05-26
feat: solana cache deduplication

CarsonRoscoe:
## Description

The SVM exact facilitator uses the full base64-encoded signed transaction as the dedup cache key. In fee-payer mode, signature slot 0 (the facilitator's) is overwritten before broadcast, so an attacker can randomize those 64 bytes to produce many distinct cache keys that all decode to the same valid payer-signed message. Each variant misses the cache, passes verification, and the resource server serves N responses — but only one on-chain transfer settles.

This PR fixes the deduplication logic to handle this.

## Tests

Added unit tests

## Checklist


phdargen:
Looks great @CarsonRoscoe, but please add changelogs

phdargen (approved):

PR #2483, opened 2026-05-26
Onchain-actions extension

phdargen:
# Extension: `onchain-actions`

Adds a spec for the **`onchain-actions`** x402 extension for noncustodial onchain execution where the server prepares transactions and the client executes them.

## Problem

x402 is a poor fit for onchain actions today (swaps, token/NFT mints, vault deposits, etc.) for two reasons:

1. **Settlement timing** — the server only has access to client funds after settlement, which is the wrong model for most onchain operations.
2. **Trust** — the client must trust the server to perform the onchain action on its behalf. That is unacceptable for meaningful amounts (e.g. >$100).

## Solution

The `onchain-actions` extension enables **noncustodial** onchain actions:

- The server **prepares** the transaction(s) and returns them in a standardized format.
- The client **pays** for preparation (via normal x402 settlement).
- The client **executes** the transaction(s) onchain — the server never touches user funds.

This lets any x402-compatible client parse, verify, simulate, and execute prepared actions from any compliant server without bespoke integrations.

## Example use case

An agent asks a yield API for opportunities on Base. The server responds with something like:

> **5% USDC yield on Aave**

…plus the ordered transaction calldata (e.g. approval + deposit) for the agent to review, simulate and execute locally.

### DIFF
--- a/specs/extensions/onchain-actions.md
+++ b/specs/extensions/onchain-actions.md
@@ -1,0 +1,300 @@
+# Extension: `onchain-actions`
+
+## Summary
+
+The `onchain-actions` extension enables **noncustodial onchain execution** through x402. Servers prepare and return executable transaction data that the **client executes onchain**. The client pays for transaction preparation, not execution. The server never touches user funds.
+
+This standardizes the response format so any x402-compatible client can parse, verify, simulate and execute onchain actions from any compliant server, without bespoke integrations. Example use cases include token swaps, cross-chain bridges, portfolio rebalancing, LP management, multisend or dust cleanup.
+
+
+---
+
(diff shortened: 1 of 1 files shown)


erikreppel (commented):
I think this is well structured and a good proposal. The implication of this seems to be the client is broadcasting the tx?

I think its worth fleshing out the usecases a bit more to see if thats actually the right assumption. Some protocols like 0x API want signed data, but not necessarily an actual transaction onchain. I'd think about representing the transaction list as broadcasted, or submitted to some endpoint.

Security is of course gnarly with this but I think you cover it pretty well in the security section.

PR #2486, opened 2026-05-27
auth-capture client scheme (TypeScript)

A1igator:
At [x402r.org](https://x402r.org), we've been working on refundable payments for x402. This PR splits the client portion out of the full auth-capture implementation (#2308) so agents can detect and sign auth-capture payloads now, ahead of full server and facilitator SDK support.

## Included

- `@x402/evm/auth-capture/client`: `AuthCaptureEvmScheme` implementing `SchemeNetworkClient`. Validates the spec-mandated `extra` fields, reconstructs the canonical PaymentInfo struct, computes the payer-agnostic PaymentInfo hash, and signs an ERC-3009 `ReceiveWithAuthorization` (default) or Permit2 `PermitTransferFrom` (`assetTransferMethod: "permit2"`).
- Supporting modules: payload/extra types and guards, nonce/salt derivation and signing helpers, canonical escrow and collector constants, CAIP-2 chain parsing.
- 51 unit tests covering payload construction, nonce/salt derivation, type guards, and the ERC-3009 vs Permit2 paths.
- Client example under `examples/typescript/clients/auth-capture/`.

The client is pure signing: no on-chain reads, no contract calls. capture/void/refund are operator-side and out of client scope; authorize vs charge produce an identical client signature, so the client is agnostic to that choice.

## Spec

Implements the client half of the auth-capture scheme (spec #1425). Original proposals: #1011, #834.

## Disclosure

This PR was written with AI assistance (Claude Code). Output was reviewed before submission per the AI guidance in `CONTRIBUTING.md`.

## Tests

- Unit tests: 51/51 passing
- Package builds; client subpath export resolves
- Client example typechecks

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -13,17 +13,3 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-
----
-
-auth-capture scheme:
-
-The auth-capture payment scheme and its TypeScript implementation in
-@x402/evm were contributed by the x402r team. See proposal issue
-#1011 and specification PR #1425. The implementation was developed
-in https://github.com/BackTrackCo/x402r-scheme prior to upstreaming.
-
-The implementation originated from the "x402-escrow" proposal by
-Agentokratia in issue #834, published as @agentokratia/x402-escrow
-on npm under the MIT License, and incorporates code from that
-package.
--- a/NOTICE
+++ b/NOTICE
@@ -13,3 +13,17 @@ distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
+
+---
+
+auth-capture scheme:
+
+The auth-capture payment scheme and its TypeScript implementation in
+@x402/evm were contributed by the x402r team. See proposal issue
+#1011 and specification PR #1425. The implementation was developed
+in https://github.com/BackTrackCo/x402r-scheme prior to upstreaming.
+
+The implementation originated from the "x402-escrow" proposal by
+Agentokratia in issue #834, published as @agentokratia/x402-escrow
+on npm under the MIT License, and incorporates code from that
+package.
--- a/typescript/packages/mechanisms/evm/src/auth-capture/types.ts
+++ b/typescript/packages/mechanisms/evm/src/auth-capture/types.ts
@@ -1,0 +1,172 @@
+/**
+ * auth-capture wire-format types.
+ *
+ * Spec-level field names (captureAuthorizer, captureDeadline, refundDeadline,
+ * feeRecipient) live here at the extra/wire layer. The on-chain PaymentInfo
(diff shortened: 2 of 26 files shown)


typescript/packages/mechanisms/evm/src/auth-capture/types.ts
```diff
@@ -0,0 +1,172 @@
...
+    from: `0x${string}`;
+    to: `0x${string}`; // EIP3009_TOKEN_COLLECTOR_ADDRESS
+    value: string;
+    validAfter: string;
+    validBefore: string; // = preApprovalExpiry
+    nonce: `0x${string}`; // = payer-agnostic PaymentInfo hash
+  };
+  signature: `0x${string}`;
+  salt: `0x${string}`; // bytes32, fresh per request, used to reconstruct PaymentInfo
+}
```
phdargen, line 79: think better to introduce `payload.type = authorize, charge` already

NOTICE
```diff
@@ -13,3 +13,17 @@ distributed under the License is distributed on an "AS IS" BASIS,
...
+
+The auth-capture payment scheme and its TypeScript implementation in
+@x402/evm were contributed by the x402r team. See proposal issue
+#1011 and specification PR #1425. The implementation was developed
+in https://github.com/BackTrackCo/x402r-scheme prior to upstreaming.
+
+The implementation originated from the "x402-escrow" proposal by
+Agentokratia in issue #834, published as @agentokratia/x402-escrow
+on npm under the MIT License, and incorporates code from that
+package.
```
phdargen: Please revert this, suggest to add yourself as author of the spec file instead similar to eghttps://github.com/x402-foundation/x402/blob/51f889bc58bdbfa6456295317db1c1646812e1d9/specs/extensions/extension-offer-and-receipt.md?plain=1#L877

A1igator: Since our code started from the Agentokratia MIT implementation (it used to be open source but not anymore), I think I legally have to give them credit for the code somewhere. I think most people do it in NOTICE.md and can put it elsewhere but it's the code not just spec.

typescript/packages/mechanisms/evm/src/auth-capture/types.ts
A1igator: I might be wrong but I don't think it makes sense here? client payload depends on the token collector not where the facilitator actually calls. Client signs to the token collector not authorize/charge.

NOTICE
phdargen: Hm ok, lets move this in a new NOTICE file at top-level of the x402/evm TS sdk

typescript/packages/mechanisms/evm/src/auth-capture/types.ts
A1igator: Yeah I looked into it more and if you want the client to pick authorize vs charge, you need to block the other one in the smart contract captureAuthorizer, otherwise it's at the whims of the facilitator to honor anything. Including it in the payload is misleading since the contracts will ignore it. I can document this and include it in the future minimal reference captureAuthorizer smart contract.

I can change it in the extras for now but might want to update the spec before/during the change and it won't affect the client implementation either way. Same thing applied to the autoCapture setup. Let me know if you have other thoughts though.

NOTICE
A1igator: Done!

phdargen (approved):

PR #2412, opened 2026-05-21
fixtures: canonicalisation-substrate v0 ; 53-vector in-tree mirror

seritalien:
Per the commitment on #2326 (and the public follow-up on #2357), Vauban Pay is opening this PR to land the in-tree mirror of the four AlgoVoi-published JCS conformance vector sets that anchor the shared canonicalisation discipline section under coordination at #2326.

## What this PR adds

A single Apache 2.0 directory `fixtures/canonicalisation-substrate/v0/` containing :

| File | Vectors | Pair invariants | Source gist | Layer |
|---|---|---|---|---|
| `ap2-omh-v0.json` | 7 | 4 | `chopmob-cloud/1dca25fd6107db4b7a30bed5dbf2ded8` | application-mandate (AP2 OMH) |
| `ctef_vectors.json` | 7 | (CTEF) | `chopmob-cloud/5f35eaa527d292bf3ddc52f8725a85c9` | trust-evidence container |
| `aps_vectors.json` | 7 | (APS) | `chopmob-cloud/5f35eaa527d292bf3ddc52f8725a85c9` | attestation production |
| `privacy_class_v0.1.json` | 13 | 12 | `chopmob-cloud/30bcbc717c86493f737feb92c415ba07` | attestation (privacy_class) |
| `per_chain_envelope_v0.json` | 19 | 9 | `chopmob-cloud/e1bf4c9efde6f0e94b77c238cb33d78d` | chain-identifier + chain-native-value |

**Total : 53 vectors + 37 pair invariants.** All files are byte-for-byte mirrors of the canonical gists ; provenance verification command is in the directory README.

## Why this PR

The v3 draft of the shared canonicalisation discipline section under #2326 references these four AlgoVoi-published vector sets as the conformance anchors. Today they live in gists, which are mutable URLs and not a stable foundation for a normative reference. The retention-property clause @feedoracle added to v3 (MiCA Art. 80, AMLR Art. 56, DORA Art. 14 retention obligations) makes this even sharper : an auditor in year 5 pinning a vector revision needs a repo tag, not a gist that may have been silently edited. This PR provides that repo-pinned anchor.

## Cross-implementation status

All five reference JCS implementations validate every vector and every pair invariant byte-for-byte (re-verified 2026-05-21) :

| Library | Lang | Author |
|---|---|---|
| \`rfc8785@0.1.4\` | Python | Trail of Bits |
| \`canonicalize@3.0.0\` | JS | Erdtman + Rundgren |
[...]

PR #2199, opened 2026-05-05
feat(python/batch-settlement): add foundation (PR1)

shuhei0866:
## Summary

This is **part 1 of a planned 13-PR Python port** of `@x402/evm/batch-settlement` (TS PR #2061). It delivers the wire-level foundation that subsequent PRs will build on: constants, ABI, and 18 Pydantic wire-format models with manual discriminator/XOR helpers.

Per [our intent comment on #2061](https://github.com/x402-foundation/x402/pull/2061#issuecomment-4377639789), this opens the series.

## Stacking strategy & branch request

We would like to land this as a stacked PR series into a long-lived `feat/batch-settlement-python-sdk` branch on upstream, mirroring two existing precedents in this repo:

1. **`feat/python-v2-sdk`** — Coinbase ran a 16-PR stacked series (2025-12-29 → 2026-01-21) into this branch to land Python SDK v2. External and internal PRs were both accepted there.
2. **`feat/batch-settlement-go-sdk`** — sibling Go-language port currently in progress under the same naming convention.

If a `feat/batch-settlement-python-sdk` branch is created, we'll rebase this PR onto it and continue with PR2-PR13 stacked. If you'd prefer a different shape (sequential into main, or fewer larger PRs), happy to adapt — flagging this up front so we can align before review effort accumulates.

## What's in PR1

```
python/x402/mechanisms/evm/batch_settlement/
├ __init__.py
├ constants.py   — addresses, EIP-712 type structs, withdraw delays, scheme IDs
├ abi.py         — contract ABI as a Python dict (private-marked claim components)
└ types.py       — 18 wire-format Pydantic models, parse_payload / parse_facilitator_payload helpers, _validate_deposit_xor

python/x402/tests/unit/mechanisms/evm/batch_settlement/
├ test_constants.py — 12 tests (D2/D3/D6 sanity + L2.x preconditions)
├ test_abi.py       — 8 tests (function/event names, private marker)
└ test_types.py     — 36 tests (D1/D4/D5/D7/D8 + extra="ignore" + round-trip)

python/x402/changelog.d/+batch-settlement-foundation.feature.md
scripts/check.sh — Layer 1 universal CI gate
```

## Wire decisions affirmed

PR1 locks down 12 wire-level decisions to keep subsequent PR review focused on behavior. Statement of intent:

[...]

PR #2489, opened 2026-05-27
feat(python/batch-settlement): add cross-language byte-equivalence fixtures

shuhei0866:
## Description

Adds cross-language byte-equivalence fixtures for the `batch-settlement` EVM scheme, plus the public Python digest API needed to verify them.

After #2402 landed the Python `batch-settlement` SDK, the TS and Python sides share no `bytes`-level fixture that pins their EIP-712 digests against the same input. A future change in `viem`, in the Python `eth_account` library, or in the spec itself could silently shift one side's digest while the other stays stable — breaking `ecrecover` resolution in production with no test signal. This PR closes that gap for the four signing-time primitives (`ChannelConfig` / `Voucher` / `Refund` / `ClaimBatch`). The complementary `collectorData` ABI-encoding equivalence is tracked separately as a follow-up.

## What this PR adds

The three parts together form a single concern — **keeping the cross-language EIP-712 digest invariant verifiable over time**:

```mermaid
flowchart LR
    TS["TS SDK<br/>(viem.hashTypedData)"]
    JSON["Static JSON fixtures<br/>L2.1-L2.4"]
    PY["Python SDK<br/>compute_*_digest"]
    CI["CI: regenerate &<br/>diff committed JSON"]

    TS -->|"(1) generates"| JSON
    JSON -->|"(2) verifier re-computes<br/>and byte-compares"| PY
    JSON -.->|"(3) drift guard"| CI
```

1. **Python digest API** (`mechanisms/evm/batch_settlement/digest.py`): four pure functions returning the 32-byte EIP-712 digest of each signing-time primitive. The existing `compute_channel_id` is now a thin wrapper over `compute_channel_config_digest`, so behaviour is unchanged but external callers can now obtain raw digests without going through the signer / verifier paths (useful for client-side pre-computation, debug / observability, and any future Go / Java verifier).

2. **Static JSON fixtures** (`tests/fixtures/batch-settlement-byte-equivalence/v0/`): four vectors (`L2.1` ChannelConfig, `L2.2` Voucher, `L2.3` ClaimBatch, `L2.4` Refund) generated by `_generator.ts` using `viem.hashTypedData`. Each file records the input, the expected 32-byte digest, the EIP-712 domain + types, and generator metadata. The Python verifier test (`test_byte_equivalence_fixtures.py`) reads each JSON, re-computes the digest via the Python SDK, and asserts byte-for-byte equality.

3. **CI drift detection** (`.github/workflows/check_byte_equivalence_fixtures.yml`): on every PR, the workflow re-runs the TS generator and fails if the regenerated JSON differs from the committed JSON. This catches TS-side or spec drift; the Python verifier test catches Python-side drift. Together the two guard the invariant from both directions.

## Test plan

```bash
cd python/x402

# 1. Python digest API unit tests (20 tests)
uv run pytest tests/unit/mechanisms/evm/batch_settlement/test_digest.py

# 2. Cross-language byte-equivalence verifier (4 vectors + 2 guards = 6 tests)
uv run pytest tests/unit/mechanisms/evm/batch_settlement/test_byte_equivalence_fixtures.py

# 3. Regression: existing batch-settlement tests
uv run pytest tests/unit/mechanisms/evm/batch_settlement/
```

All green locally (20 + 6 + 57 = 83 tests, full module 372 tests). The drift-detection workflow can also be reproduced locally:

```bash
cd python/x402/tests/fixtures/batch-settlement-byte-equivalence/v0
npm install && npx tsx _generator.ts
# expected: no `git diff` against the committed JSON
```

## Acknowledged scope choices

- **Fixture path**: placed at `python/x402/tests/fixtures/...` rather than a repo-root `fixtures/` directory. The latter is being established in #2412 by another contributor; this PR avoids the top-level-precedent overhead and keeps the scope cleanly inside the python label. If `fixtures/` graduates, this directory can be promoted later in a separate PR — the naming `batch-settlement-byte-equivalence` already preserves the cross-language intent.

- **Generator is self-contained**: `_generator.ts` inlines the EIP-712 domain and type definitions and depends only on `viem`. Drift between the inlined definitions and the TS SDK constants is caught by the drift-detection job via the regenerated digest, not by structural comparison. This avoids cross-package `@x402/evm` workspace coupling in a python tree.

- **TS SDK exports unchanged**: only the digest helpers are added on the Python side. The TS SDK already exports `computeChannelId`; adding parity helpers (`computeVoucherDigest` etc) on the TS side is out of scope here and is a natural follow-up if you'd like it.

- **L2.5 / L2.6 deferred**: the ERC-3009 deposit nonce and the `collectorData` ABI encoding (the dynamic-bytes-of-tuple case where 2-stage encoding errors historically appear in TS↔Python interop) are out of scope here and tracked separately so this PR stays scoped to the fixture-format + digest-API establishment.

## Refs

- #2061 (TS batch-settlement) — introduced the wire shape
- #2402 (Python batch-settlement) — landed the Python SDK
- #2199 (closed) — earlier Python port draft that prefigured this work, including discussion of cross-language test vectors.

Happy to split this into per-component PRs if that's preferred.

### DIFF
--- a/.github/workflows/check_byte_equivalence_fixtures.yml
+++ b/.github/workflows/check_byte_equivalence_fixtures.yml
@@ -1,0 +1,35 @@
+name: Check Batch-Settlement Fixture Drift
+on: [pull_request]
+
+permissions:
+  contents: read
+  id-token: none
+
+jobs:
+  check-byte-equivalence-drift:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: "24"
+
+      - name: Install generator deps
+        working-directory: ./python/x402/tests/fixtures/batch-settlement-byte-equivalence/v0
+        run: npm install --no-audit --no-fund --ignore-scripts
+
+      - name: Regenerate fixtures
+        working-directory: ./python/x402/tests/fixtures/batch-settlement-byte-equivalence/v0
+        run: npx tsx _generator.ts
+
+      - name: Check for drift
+        run: |
+          if ! git diff --exit-code -- 'python/x402/tests/fixtures/batch-settlement-byte-equivalence/v0/L[0-9].[0-9]*.json'; then
+            echo "::error::Batch-settlement byte-equivalence fixtures are out of date."
+            echo "::error::Regenerate locally:"
+            echo "::error::  cd python/x402/tests/fixtures/batch-settlement-byte-equivalence/v0"
+            echo "::error::  npm install && npx tsx _generator.ts"
+            echo "::error::Then commit the updated JSON files."
+            exit 1
+          fi
--- a/python/x402/changelog.d/+batch-settlement-byte-equiv-fixtures.feature.md
+++ b/python/x402/changelog.d/+batch-settlement-byte-equiv-fixtures.feature.md
@@ -1,0 +1,8 @@
+Add pure EIP-712 digest helpers for the batch-settlement EVM scheme
+(`compute_channel_config_digest` / `compute_voucher_digest` /
+`compute_refund_digest` / `compute_claim_batch_digest` in
+`x402.mechanisms.evm.batch_settlement.digest`), plus cross-language
+byte-equivalence fixtures under
+`tests/fixtures/batch-settlement-byte-equivalence/v0/` and a CI
+drift-detection job. The existing `compute_channel_id` helper is now a thin
+wrapper over `compute_channel_config_digest` (behaviour unchanged).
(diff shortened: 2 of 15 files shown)


shuhei0866:
Rebased on current main (`6dba93e`). The fixtures regenerate byte-identical, and the domain and type definitions the generator inlines still match `batch-settlement/constants.ts` field for field — the only change to that file since the vectors were written adds two polling bounds, which do not enter any EIP-712 encoding.

What has changed is the stakes. #2863 promoted `computeChannelId` to a security boundary: `channelIdBindingError` rejects a payload whose claimed `channelId` does not hash back to its `channelConfig`, before any storage is read or written. That check was hand-ported into TypeScript, Python, and Go in parallel.

`L2.1-channel-config.json` pins exactly that computation — the `ChannelConfig` EIP-712 digest *is* the channel id. Each implementation is currently checked only against its own unit tests; nothing pins the three to each other. If the digest diverged between two SDKs, the binding check would reject valid payloads on one rail while accepting them on another, and it would surface as an unexplained channel-id mismatch rather than as an encoding bug.

That is the regression this fixture set closes, and the drift workflow keeps it closed as the structs evolve.

#2499 adds the deposit-side vectors (L2.5–L2.8) against the same fixture directory. Happy to fold them in here if one PR is easier to review than the current split.

PR #2491, opened 2026-05-27
spec(batch-settlement): add corrective-402 recovery contract spec and reference vectors

shuhei0866:
## Description

Adds a wire-level specification and three machine-readable reference vectors for the `batch-settlement` scheme's corrective-402 recovery contract, so any x402 SDK (TS / Python / Go / Java) can converge on the same recovery semantics.

The `batch-settlement` scheme uses cumulative vouchers; when two concurrent requests on the same channel race, or when a client's local cumulative drifts behind the server's, the server returns `402` with an authoritative state snapshot so the client can recover. The wire shape for this corrective response and the expected client behaviour (verify against onchain, update local state, retry exactly once, hard-error on a second corrective) are discussed in #2061 by @phdargen, @jithinraj and @Bortlesboat but not pinned in machine-readable form. This PR pins them.

## What this PR adds

```mermaid
flowchart LR
    Spec["Spec doc<br/>(batch-settlement-recovery.mdx)"]
    Vectors["Reference vectors<br/>(v0/*.json)"]
    SDKs["x402 SDKs<br/>(TS / Py / Go / Java)"]
    Conformance["wire-level portable<br/>recovery behaviour"]

    Spec -->|"defines contract"| Vectors
    Vectors -->|"SDKs read & implement<br/>integration tests"| SDKs
    SDKs -->|"converge by construction"| Conformance
```

Five files, no executable code:

1. **`docs/schemes/batch-settlement-recovery.mdx`** — normative spec: when a corrective 402 is returned, the wire shape of `extra.channelState`, the invariant `onchain.total_claimed ≤ server.charged_cumulative_amount ≤ voucher.max_claimable_amount`, the recovery flow (Mermaid sequence diagram), the loop-guard rationale (resource burn / bug detection / DoS surface), the telemetry labels (`stale_state` / `cas_conflict` / `persistent_stale`), and cross-references to the byte-equivalence fixtures.

2. **`docs/schemes/batch-settlement-recovery-vectors/v0/README.md`** — vector format documentation plus a quick-start for SDK integration tests (how to walk a `steps` array and assert against `expected_client_behaviour`).

3-5. **Three scenario vectors** under `v0/`:
- `happy-path.json` — stale local cumulative, single retry succeeds
- `loop-guard.json` — second corrective on retry, hard error (`PERSISTENT_STALE`)
- `cas-conflict.json` — concurrent same-channel requests, one wins, the other recovers

Each vector describes a wire-level sequence of `(request → response)` steps, an `expected_client_behaviour` block per step, and the operational `invariants` that must hold.

## Why vectors, not just spec prose

Two SDKs implementing the same prose can diverge at edge cases (is `total_claimed > charged_cumulative_amount` valid? does a retry that produces the same digest as the first attempt re-trigger the loop guard? what telemetry label applies when the corrective `charged_cumulative_amount` is unchanged?). Vectors fix the exact wire payload, state transitions, and expected behaviour. Two SDKs implementing the same vectors converge by construction.

## Acknowledged scope choices

- **Spec-only, no implementation in this PR**: this ships normative documentation + machine-readable vectors. SDK integration tests that walk each vector are out of scope here and should land as follow-ups on each language's SDK. Python's recovery handler in `client/recovery.py` (from #2402) can be tested against `happy-path.json` immediately.

- **Three scenarios, not six**: the #2061 thread also discussed dynamic-price-below-cap, HTTP-compatibility, and telemetry-privacy as recovery-adjacent. Those are operational concerns that vary by SDK and deployment; this PR is scoped to the three wire-level invariants every SDK must agree on. Additional vectors can be added under the same `v0/` schema (non-breaking), or as `v1/` if the schema needs a breaking change.

- **`signature` fields are placeholders**: EIP-712 signing semantics are independently pinned by #2489 (`tests/fixtures/batch-settlement-byte-equivalence/v0/`). The recovery contract is independent of signature validity — what matters here is the state-machine transition, not the cryptographic signing.

- **Placed under `docs/schemes/<scheme>-vectors/v<N>/` rather than a repo-root `fixtures/` directory**: same reasoning as #2489 — the repo-root `fixtures/` directory is being established by #2412 and avoiding precedent overhead keeps this PR's review surface inside the existing `docs/schemes/` area.

## Refs

- #2061 (TS batch-settlement) — original corrective-402 contract proposal and recovery-semantics discussion
- #2402 (Python batch-settlement) — landed `client/recovery.py` which already implements the contract; this PR pins it spec-side so other SDKs can converge
- #2489 (this same series) — byte-equivalence fixtures for the signing-time primitives; this PR complements that by pinning the state-machine layer

Happy to split spec / vectors / individual scenarios into separate PRs if preferred.

### DIFF
--- a/docs/schemes/batch-settlement-recovery-vectors/v0/README.md
+++ b/docs/schemes/batch-settlement-recovery-vectors/v0/README.md
@@ -1,0 +1,62 @@
+# batch-settlement recovery vectors (v0)
+
+Machine-readable test vectors for the corrective-402 recovery contract specified in [`../batch-settlement-recovery.mdx`](../batch-settlement-recovery.mdx).
+
+Each JSON file describes a single recovery scenario as a sequence of `(client request → server response)` pairs, plus the expected client behaviour and the wire-level invariants that must hold.
+
+## Files
+
+| File | Scenario | Recovery outcome |
+| --- | --- | --- |
+| `happy-path.json` | Stale local cumulative | Single retry succeeds |
+| `loop-guard.json` | Retry also returns corrective 402 | Client emits hard error (`PERSISTENT_STALE`) |
+| `cas-conflict.json` | Concurrent same-channel requests, one wins | Loser recovers via single retry, tagged separately for telemetry |
+
+## Vector schema
+
+Each file has this shape:
+
+```jsonc
+{
+  "scenario": "<id>",
+  "description": "<human-readable summary>",
+  "channel": {
+    "channel_id": "0x…",
+    "domain": { "name": "x402 Batch Settlement", "version": "1", "chainId": 84532, "verifyingContract": "0x…" }
+  },
+  "initial_state": {
+    "onchain_total_claimed": "<wei>",
+    "server_charged_cumulative": "<wei>",
+    "client_local_cumulative": "<wei>",
+    "comment": "<optional human note about the precondition>"
+  },
+  "steps": [
+    { "step": <n>, "actor": "client" | "client_a" | "client_b" | "server",
+      "action": "<text>",
+      "request": { … } /* OR */ "response": { … } }
+  ],
+  "expected_client_behaviour": { "after_step_<n>": "<text>" },
+  "invariants": [ "<text>" ]
+}
+```
+
+`signature` fields are placeholders. EIP-712 signing is independently pinned by the [byte-equivalence fixtures](../../../../python/x402/tests/fixtures/batch-settlement-byte-equivalence/v0/) and does not affect the recovery state-machine. SDK integration tests can use any deterministic signing key; the recovery contract is independent of signature validity.
+
+## How SDK integration tests use these vectors
+
+A conformant SDK should read each `steps` entry in order and:
+
+1. **`actor: "client"` with `request`** — construct the request shape from the vector, sign with a test key (any deterministic signer is fine), submit to a test server.
+2. **`actor: "server"` with `response`** — assert the server's response matches the vector's status and `extra.channelState` shape, byte-by-byte where possible.
+3. After each step, evaluate `expected_client_behaviour.after_step_<n>` as the contract the SDK's client implementation must satisfy (e.g. parse and verify `channelState`, retry exactly once, emit a specific telemetry label, escalate to hard error).
+4. At the end of the sequence, assert all `invariants` still hold for the captured state.
+
+Two SDKs implementing the same vector will produce byte-identical request bodies (modulo signatures) and state transitions, and either both pass or both fail at the same step. That is the convergence property these vectors exist to provide.
+
+## Version pinning
+
+This directory is `v0/`. Breaking changes to the schema (new top-level fields, removed fields, semantics change) will be introduced as `v1/` rather than mutating `v0/`, so downstream SDK tests can pin the version they implement against. Adding new scenarios under the same schema is non-breaking and stays in `v0/`.
+
+## Why vectors, not just spec prose
+
+Spec prose tells implementers *what* the recovery contract is; the vectors fix *exactly* what each wire payload looks like, what state transitions occur, and what the client must do after each step. Two SDKs implementing the same prose can diverge in edge-case interpretation (is `total_claimed > charged_cumulative_amount` valid? does a `cas_conflict` telemetry label apply when `charged_cumulative_amount` is unchanged?). Two SDKs implementing the same vectors converge by construction.
--- a/docs/schemes/batch-settlement-recovery-vectors/v0/README.md
+++ b/docs/schemes/batch-settlement-recovery-vectors/v0/README.md
@@ -21,21 +21,21 @@ Each file has this shape:
   "scenario": "<id>",
   "description": "<human-readable summary>",
   "channel": {
-    "channel_id": "0x…",
+    "channelId": "0x…",
     "domain": { "name": "x402 Batch Settlement", "version": "1", "chainId": 84532, "verifyingContract": "0x…" }
   },
-  "initial_state": {
-    "onchain_total_claimed": "<wei>",
-    "server_charged_cumulative": "<wei>",
-    "client_local_cumulative": "<wei>",
+  "initialState": {
+    "onchainTotalClaimed": "<wei>",
+    "serverChargedCumulative": "<wei>",
+    "clientLocalCumulative": "<wei>",
     "comment": "<optional human note about the precondition>"
   },
   "steps": [
-    { "step": <n>, "actor": "client" | "client_a" | "client_b" | "server",
+    { "step": <n>, "actor": "client" | "clientA" | "clientB" | "server",
       "action": "<text>",
       "request": { … } /* OR */ "response": { … } }
   ],
-  "expected_client_behaviour": { "after_step_<n>": "<text>" },
+  "expectedClientBehaviour": { "afterStep<n>": "<text>" },
   "invariants": [ "<text>" ]
 }
 ```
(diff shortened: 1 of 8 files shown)


sunsetpatrol:
Strong +1 on pinning the corrective-402 recovery contract as machine-readable vectors, and especially on the invariant chain (`totalClaimed ≤ chargedCumulativeAmount ≤ maxClaimableAmount`) — we built a comparable deferred-settlement rail (receiver-signed settlement over a prepaid hold) in the Stipendex project, and recovery-after-state-loss was one of the findings that reached CRITICAL in our review. One scenario we ended up needing that I don't see in the current vector set:

**Partial state restore.** The server-side replay defense really lives in *two* stores: the per-channel cumulative accounting (`chargedCumulativeAmount`, expected next increment) and the redemption/idempotency cache for already-settled work. They are often persisted by different components, so the realistic failure mode after a crash/deploy is restoring *one* without the other:

- cumulative state restored, redemption cache lost → a previously redeemed request replays as "new" work against correct counters (double service, or double charge if the cache was the only replay guard);
- redemption cache restored, cumulative state lost → the server's counters reset, corrective 402 fires, and depending on the loop-guard the client is either healed via the normal `recovery.stale_state` path or cut off with `PERSISTENT_STALE` — the contract should say which outcome is conformant.

Suggested vector, in this PR's `(client request → server response)` format: *Arrange* — server has accepted increment `n` and recorded its redemption; snapshot only one of the two stores. *Act* — restart the server from the partial snapshot; replay increment `n`. *Assert* — the conformant server refuses the replay in both partial-restore cases, which in practice requires normative language that the two stores MUST be snapshotted/restored together (or unified in one atomic store).

In our implementation this ended as an explicit paired `snapshot()/restore()` API plus a documented operator requirement to persist both stores atomically. Happy to write this out as a fourth vector JSON against this PR's schema if you'd take it.

Bortlesboat (commented):
Read through the spec and all three vectors — this pins exactly what we went back and forth on in #2061, and the vectors-over-prose argument is right.

Two gaps, both of which are the motivating examples in the README's "why vectors, not just spec prose" section but don't end up pinned anywhere:

1. **`totalClaimed > chargedCumulativeAmount`** — the README asks whether it's valid, the invariant section implies it can't happen, but nothing specifies what a client does when a corrective snapshot *violates* the invariant (e.g. a server that restarted and lost its cumulative while a batched `claim()` landed). Since the client verifies `totalClaimed` onchain anyway, it detects this case for free. I'd add a normative line: an invariant-violating snapshot is an invalid corrective → hard error, don't retry. It's the one case where following the recovery recipe (`maxClaimableAmount = chargedCumulativeAmount + requestCharge`) produces a voucher below what the contract will enforce, so the retry is guaranteed to be wrong.

2. **Unchanged cumulative on the second corrective** — the README asks "what telemetry label applies when the corrective `charged_cumulative_amount` is unchanged", but `loop-guard.json` only covers the value moving (100 → 200). Unchanged means the retry voucher was rejected at the *same* cumulative it was constructed against, which points at client-side voucher reconstruction rather than drift. Worth either a fourth vector or a sentence in the loop-guard section stating `PERSISTENT_STALE` applies whether or not the value moved.

Smaller one: `cas-conflict.json` says the label MUST be `recovery.cas_conflict`, but the spec never defines how a client *decides* between `stale_state` and `cas_conflict` — the two are wire-identical by design. The implementable rule implied by the vectors seems to be: `cas_conflict` iff the signing-time local cumulative was previously confirmed by a server response on this channel, `stale_state` otherwise. If that's the intent, pinning it in the telemetry table would stop SDKs from inventing their own heuristics — which is the exact divergence this PR exists to prevent.

None of these block the direction. (1) is the one I'd want before merge, since it's the only case where the recovery recipe itself emits a wrong voucher.

shuhei0866:
@Bortlesboat 

First, thank you for such a thoughtful and high-quality review. Each point carried meaningful design implications and deserved deeper consideration. In particular, your comments pushed me to think more carefully about what the client can actually know, which states are recoverable, and where normative protocol behavior should end and diagnostic inference should begin. The review genuinely broadened and sharpened my thinking here.

I walked through all three points and agree with the direction:

1. A corrective snapshot where the verified onchain `totalClaimed` exceeds the server’s `chargedCumulativeAmount` is not recoverable client-side. I’ve made this an `INVALID_CORRECTIVE_STATE`: no local mutation, replacement signature, or retry, with a negative vector covering it.

2. The loop guard is now attempt-based. Any second corrective 402 for the same logical request results in `PERSISTENT_STALE`, regardless of whether the cumulative advanced or stayed unchanged. Value movement is diagnostic only. I added coverage for the unchanged case.

3. I don’t think the client can normatively distinguish a CAS conflict from ordinary stale state using the corrective response alone; the cases are wire-identical. The spec now uses `stale_state` as the protocol outcome and allows `cas_conflict` only as an optional suspected diagnostic cause when supporting evidence exists.

I’ve updated the spec and vectors accordingly. Thanks again for pushing on these edge cases—the resulting recovery contract is substantially clearer and more robust because of it.

shuhei0866:
Update on where this stands.

The three gaps raised in review are addressed in `01a4130`:

- An invariant-violating corrective is terminal (`INVALID_CORRECTIVE_STATE`): no local mutation, no replacement signature, no retry. Covered by `invalid-corrective-state.json`.
- The loop guard is attempt-based, so a second corrective for the same logical request is `PERSISTENT_STALE` whether or not the cumulative moved. Covered by `loop-guard-unchanged.json`.
- `cas_conflict` is demoted to an optional suspected cause, with `stale_state` as the protocol outcome, since the two are wire-identical from where the client stands.

One further outcome was added in `447ff5f`, and it came from implementer feedback rather than from me. While scoping a Go implementation against these vectors in #2908, @256dino found that the document requires the client to verify `totalClaimed` onchain before trusting a corrective snapshot, and then says nothing about that read itself failing. `INVALID_CORRECTIVE_STATE` would blame the server for something the client never established, and reporting nothing strands the request. `TRANSIENT_STATE_UNAVAILABLE` now covers it as a non-terminal outcome that preserves the underlying cause, mutates nothing, and consumes no loop-guard attempt, since no replacement voucher is signed. The same commit pins what the second corrective may do to local state, which the document had left open.

That shape is the client-side counterpart to what #3083 landed for settle last week: a broadcast whose confirmation cannot be established is reported as non-terminal with its evidence attached, rather than collapsed into either success or failure.

The vector set is six files. CI is green and the commits are verified.

If this is more surface than the scheme wants to take on right now, I am happy to trim it — the invariant and the single-retry rule stand on their own, and the vectors could land separately or not at all.

Bortlesboat (commented):
Re-read `01a4130..447ff5f` against the prose and all six vectors. The three points from my July review are now pinned: invalid corrective state fails closed, the loop guard is attempt-based for both moved and unchanged cumulatives, and `cas_conflict` is diagnostic rather than a wire-visible protocol outcome.

One precedence case is now ambiguous because of the added transient outcome:

- the loop-guard section says a second corrective 402 **MUST** produce `PERSISTENT_STALE` and no further retry;
- the transient section/table says a corrective whose onchain verification is unavailable **MUST** produce non-terminal `TRANSIENT_STATE_UNAVAILABLE`, consumes no loop-guard attempt, and may start later with a fresh budget.

Those rules conflict if the onchain read fails while handling the *second* corrective. Since a replacement voucher was already signed and the single retry was already sent, I think `PERSISTENT_STALE` should take precedence at that point: stop without trusting or persisting the second snapshot, and do not reset the logical request's retry budget. `TRANSIENT_STATE_UNAVAILABLE` can then be scoped explicitly to verification of the first corrective. A normative sentence is enough; a seventh vector only seems necessary if you want SDK conformance to pin this composition case too.

All required GitHub checks on `447ff5f` pass; the lone Vercel status is the existing team-authorization failure, not a content check.

PR #2498, opened 2026-05-27
spec: capabilities field on SupportedKind for explicit facilitator variant support

CarsonRoscoe:
## Problem

Facilitators' GET /supported responses have no way to declare which variants of a scheme they support — e.g. which assetTransferMethod values (eip3009, permit2, erc7710) an EVM facilitator handles. Servers are expected to know this out-of-band, which breaks for third-party facilitators and makes startup validation impossible.

## Solution

Adds an optional capabilities object to SupportedKind entries in GET /supported. Each key is a discriminator field name from  PaymentRequirements.extra for that scheme; each value is the array of accepted string values for that discriminator.

**This keying is intentional**: capabilities keys mirror the exact field names already present in PaymentRequirements.extra, enabling generic validation logic with no scheme-specific code — iterate capabilities entries, check each against the matching extra field of the same name.

**Backwards compatibility**: absence of capabilities means full support. Existing facilitators require no changes.

**Scheme obligation**: scheme specs that introduce discriminator fields in extra must document the corresponding capabilities key and enumerate valid values. Schemes with no discriminators omit capabilities entirely.

##  Changes

`specs/x402-specification-v2.md`
- SupportedKind field table: adds capabilities (optional)
- §7.3 example: one EVM kind now shows capabilities: { assetTransferMethod: ["eip3009", "permit2"] }
- §7.3.2 (new): defines semantics, default behavior, and scheme documentation obligation

`specs/schemes/exact/scheme_exact_evm.md`
- New "Facilitator Capability Declaration" section: names assetTransferMethod as the discriminator, enumerates valid values ("eip3009", "permit2", "erc7710"), shows an example kind entry, and states that servers must not advertise a method absent from the matched kind's capabilities

### DIFF
--- a/specs/schemes/exact/scheme_exact_evm.md
+++ b/specs/schemes/exact/scheme_exact_evm.md
@@ -16,6 +16,25 @@ If no `assetTransferMethod` is specified in `PaymentRequired.extra`, clients sho
 
 In all cases, the Facilitator cannot modify the amount or destination. They serve only as the transaction broadcaster.
 
+### Facilitator Capability Declaration
+
+`assetTransferMethod` is the discriminator field for this scheme. Facilitators declare which values they support via `capabilities.assetTransferMethod` in their `/supported` response (see core spec §7.3.2). Valid values are `"eip3009"`, `"permit2"`, and `"erc7710"`. A facilitator omitting `capabilities` is assumed to support all three.
+
+Example kind entry for a facilitator that supports `eip3009` and `permit2` but not `erc7710`:
+
+```json
+{
+  "x402Version": 2,
+  "scheme": "exact",
+  "network": "eip155:84532",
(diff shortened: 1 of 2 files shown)


IkerAlus:
+1 to reviving this as it now covers a second reserved key: since #3053 made `paymentFlow` a protocol-reserved key in `extra` and keys flow support per `assetTransferMethod`, a facilitator's supported flows need declaring too. 

This PR closes a concrete gap: authorization-based method and a client-prepaid one can both be `exact` on the same CAIP-2 network while requiring completely different settle paths, so today a resource server cannot tell from `GET /supported` whether its facilitator implements the one it needs. #3053 validates declared routes against locally registered schemes, but nothing validates remote facilitator capability.

PR #2520, opened 2026-05-29
Upfront specs

phdargen:
## Introducing the `upfront` scheme

`upfront` settles a fixed payment **before** the resource server executes the protected request. Like [`exact`](../exact/scheme_exact.md), it transfers a server-specified amount, but the payment is final before the route handler runs.

```
exact:    verify → route handler → settle → resource delivery
upfront:  verify + settle → route handler → resource delivery
```

## Example Use Cases

`upfront` is intended for cases where the server needs payment finality **before** doing work:

- Resources that perform expensive compute (e.g. long-running inference).
- Resources that trigger irreversible or hard-to-reverse actions (e.g. sending a message, minting an NFT).
- Networks that have no smart-contract pull-settlement and can only verify a payment the client has already made (see [Asset Transfer Methods](#asset-transfer-methods)).

## Asset Transfer Methods

`upfront` is defined independently of how the asset moves. The `extra.assetTransferMethod` field selects a method, and methods fall into two families. Both satisfy the defining property: **payment is final before the route handler executes.**

### 1. Authorization-based (facilitator-settled)

The client signs an off-chain payment authorization; the facilitator submits the settlement transaction **before** the route handler runs. This gives the client a gasless UX while giving the server finality first.

- Requires a network primitive that lets a third party pull funds against a client signature (e.g. EIP-3009 / Permit2 on EVM).
- Verification confirms the signed authorization is valid and will settle; settlement is then performed by the facilitator ahead of execution.
- EVM upfront spec for `"assetTransferMethod": "eip3009,permit2"` included in this PR

### 2. Payment-proof (client-settled)

The client **pays first** (broadcasting the transfer and paying any network fee) and presents a **payment proof** with the request. The server/facilitator verifies the proof before executing the route handler.

This would cover eg:
- EVM TxID: https://github.com/x402-foundation/x402/pull/1004
- ZCash: https://github.com/x402-foundation/x402/pull/1634
- BTC lightning: https://github.com/x402-foundation/x402/pull/1311
- BSV: https://github.com/x402-foundation/x402/pull/1844 
- Near intents: https://github.com/x402-foundation/x402/pull/2102

### DIFF
--- a/specs/schemes/upfront/scheme_upfront.md
+++ b/specs/schemes/upfront/scheme_upfront.md
@@ -1,0 +1,88 @@
+# Scheme: `upfront`
+
+## Summary
+
+`upfront` settles a fixed payment **before** the resource server executes the protected request. Like [`exact`](../exact/scheme_exact.md), it transfers a server-specified amount, but the payment is final before the route handler runs.
+
+```
+exact:    verify → route handler → settle → resource delivery
+upfront:  verify + settle → route handler → resource delivery
+```
+
+## Example Use Cases
+
+`upfront` is intended for cases where the server needs payment finality **before** doing work:
+
+- Resources that perform expensive compute (e.g. long-running inference).
+- Resources that trigger irreversible or hard-to-reverse actions (e.g. sending a message, minting an NFT).
+- Networks that have no smart-contract pull-settlement and can only verify a payment the client has already made (see [Asset Transfer Methods](#asset-transfer-methods)).
+
+## `upfront` vs `exact`
+
+The two schemes differ only in *when* settlement happens relative to the route handler, but that ordering changes who carries the residual risk.
+
+### `exact`: settle after a successful response
+
+Under `exact`, the server settles only if its route handler returns success. If the handler returns a status `>= 400`, the server does not settle and **the client is never charged**, so refunds are generally unnecessary.
+
+`exact` guarantees that **no resource is delivered if settlement fails**. The residual risk falls on the *server*, and is small:
+
+- The payment authorization can become invalid between `verify` and `settle` (for example it expires, or the client double-spends the nonce, while the route handler runs).
+- The settlement transaction can fail to land (e.g. network congestion).
+
+In either case the server may have already spent resources producing a response it cannot get paid for.
+
+### `upfront`: settle before the route handler
+
+Under `upfront`, payment is final before the route handler runs. This removes the server's settlement-failure risk (funds are confirmed before any work begins), at the cost of shifting **delivery risk to the client**: the client has paid before knowing whether the resource will be delivered successfully.
+
+- Clients **opt in** to `upfront` acknowledging this risk, and SHOULD prefer `exact` when it is offered for the same resource.
+- Servers **MAY** offer refunds for failed or undelivered resources, but refunds are **out of protocol**: whether, when, and how a refund is issued depends entirely on the server implementation. The x402 protocol provides no refund mechanism for `upfront`.
+
+### When to choose `upfront`
+
+Choose `upfront` when the server needs a payment-finality guarantee before route execution. This is most relevant for expensive compute or irreversible actions, where the server cannot safely begin work on the promise of a later settlement. It also applies when the underlying network cannot pull funds and only supports client-initiated payment with a proof.
+
+## Asset Transfer Methods
+
+`upfront` is defined independently of how the asset moves. The `extra.assetTransferMethod` field selects a method, and methods fall into two families. Both satisfy the defining property: **payment is final before the route handler executes.**
+
+### 1. Authorization-based (facilitator-settled)
+
+The client signs an off-chain payment authorization; the facilitator submits the settlement transaction **before** the route handler runs. This gives the client a gasless UX while giving the server finality first.
+
+- Requires a network primitive that lets a third party pull funds against a client signature (e.g. EIP-3009 / Permit2 on EVM).
+- Verification confirms the signed authorization is valid and will settle; settlement is then performed by the facilitator ahead of execution.
+
+### 2. Payment-proof (client-settled)
+
+The client **pays first** (broadcasting the transfer and paying any network fee) and presents a **payment proof** with the request. The server/facilitator verifies the proof before executing the route handler.
+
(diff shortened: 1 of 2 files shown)


chopmob-cloud:
The authorization-based (facilitator-settled) path maps cleanly to how several non-EVM chains work in practice.

On ALGO/VOI, a signed authorization is submitted by the facilitator against the client's signed delegation -- the settlement transaction is on-chain before the route handler runs. The client never submits a transaction; they sign a typed authorization that the facilitator executes. That is method 1 exactly, without EIP-3009.

On Solana, Stellar, and Base we currently use client-settled (method 2): the client broadcasts and presents proof. The `assetTransferMethod` discriminator is the right field to carry this distinction -- a server accepting ALGO authorization-based and Solana payment-proof under the same `upfront` scheme needs to know which verification path to run before it can declare finality.

One edge case worth specifying: on chains where settlement finality is probabilistic (e.g., Solana 32-confirmation threshold), what does "payment is final before route handler executes" mean in practice for method 2? We currently gate on confirmed status plus a configurable confirmation depth. If the spec stays chain-agnostic on finality definition, implementations will diverge on what "final" means at the boundary.

AlgoVoi runs upfront-style settlement live across 8 chains. The confirmation-depth question is the one operational gap we hit when normalising the verify-then-execute boundary across chains with different finality models.

*AlgoVoi (chopmob-cloud) -- Acquisition enquiries: https://docs.algovoi.co.uk/acquisition*

TheBlueMatt:
Wouldn't this be required for a substantial number of cryptocurrencies, including on-chain Bitcoin (in the same way its required for LN)?

chopmob-cloud:
Yes. The payment-proof (client-settled) family is not just an option here, it is the only upfront settlement model available on a network that has no third-party pull primitive. On-chain Bitcoin is exactly that case: with no equivalent of EIP-3009 or Permit2, the client has to pay first and present a proof, which is the same reason Lightning needs it. The pay-first PRs already linked in the description (EVM TxID, ZCash, Lightning, BSV, Near intents) are all members of that one family.

One distinction worth pinning in the base doc for that family: Core Property 1 says settlement MUST be "confirmed final" before the route handler runs, but "final" is deterministic for Lightning, where the preimage is the settlement, and probabilistic for on-chain Bitcoin, where it is a function of confirmation depth. If the payment-proof family stays silent on finality, implementations will diverge on what "final" means at the verify-then-execute boundary. A per-network finality parameter for that family, a confirmation depth or an explicit finality rule the way the EVM doc pins its verification, would keep on-chain BTC, Lightning, and the other pay-first networks interoperable under one scheme.

javierpmateos:
Strong +1 to the payment-proof (client-settled) family as defined here ... Lightning is arguably its canonical instance (no pull-settlement primitive; the "cryptographic settlement secret" the spec anticipates is exactly the payment preimage).

On the open question of what "final" means in this family: I'd suggest the abstract spec distinguish two finality classes, and have each per-network document declare which one applies.

- **Probabilistic finality** — proof is an on-chain transaction hash/id. Finality is a function of confirmation depth; the verifier needs chain access and a per-network confirmation policy (N confs on Bitcoin L1, reorg assumptions on EVM, etc.). "Settle-before-execute" here means "confirmed to the network's declared policy before execute."
- **Deterministic finality** — proof is a settlement secret. A Lightning preimage cannot exist until the HTLC has been accepted, so the proof *is* the settlement: validity is a pure function of the request (`SHA-256(preimage) == payment_hash` plus invoice validation), with no node query, no confirmation policy, and no chain access at all. Even a hold invoice can't produce a false proof — the preimage only exists once payment is actually accepted — so holds are a liveness concern, not a safety one.

Naming the classes in the abstract doc would let Core Property 1 stay method-agnostic ("payment MUST be confirmed final per the method's declared finality class") while making verifier requirements explicit per network.

A second point that may deserve a line in the abstract doc: in this family the payment happens outside the facilitator's view, so a verifier can only legitimately attest to what the authenticated path commits to. In v2 `resource` sits in the `PaymentRequired` envelope and reaches the facilitator only through the client echo — which means a facilitator attesting to *what* was paid for is attesting to a client claim, unless the proof artifact itself commits to it. The resource server can of course check the echo against its own route before calling out; the gap is specific to what a facilitator can put its name on, which matters as soon as attestations are meant to be portable. Worth stating explicitly what verifiers in this family can and cannot vouch for.

To reduce the consolidation load once this merges: I've restructured my earlier Lightning draft (#2938, originally filed under `exact`) as `upfront` + `extra.assetTransferMethod: "bolt11"` following this PR's conventions — deterministic-finality profile, preimage as proof, single-use tracking per Core Property 4, and a mandatory recipient-signed binding via BOLT11 `description_hash` that commits the invoice to both the payment requirements and the resource, which closes the substitution vector cryptographically. Draft + reference implementation with passing tests: https://github.com/javierpmateos/x402-lightning — happy to consolidate with #2861, which shares the preimage/stateless approach, into a single `scheme_upfront_lightning.md` once this PR lands.

zjzJoez:
A placement question this PR raises, from someone with a proposal that runs into it — asking rather than arguing, and happy to be told it is obvious.

`upfront` is already a `paymentFlow` value in §6.1 (`settle → resource → respond`), and this PR introduces it again as a scheme. So the same name now describes both when settlement happens relative to the handler, and how a particular payment is formed. Is the intended model that a scheme exists per ordering, with the §6.1 value naming it — or that the flow field is the general mechanism and this scheme is the EVM-and-friends realisation of one value?

I ask because #3182 proposes a fourth ordering (respond, then settle at a signed future `validAfter`, cancellable by the payer until then) and the answer decides whether that is a §6.1 amendment, a scheme, or neither. I would rather build it the way you would want it than guess and submit the wrong shape.

Concrete on the "is it real" question: two live public facilitators accept an identical `exact`/`eip3009` payload with a backdated `validAfter` and reject it with `invalid_exact_evm_payload_authorization_valid_after` when it is future-dated — reproducible, read-only, in the linked repo. Not asking for a decision on the proposal here, only on where such a thing belongs.

phdargen:
Replaced by payment flow mechanism, see https://github.com/x402-foundation/x402/pull/3053 and https://github.com/x402-foundation/x402/pull/3240

PR #2530, opened 2026-05-31
Reject partial money strings

samsamtrum:
Several server-side money parsers use `parseFloat` after trimming a leading `$`. `parseFloat` accepts numeric prefixes, so values like `1abc` are parsed as `1` instead of rejected.

This changes the parsers to require the full string to be a plain decimal amount before converting it, and adds regression coverage for SVM and Stellar server parsing.

### DIFF
--- a/examples/typescript/facilitator/advanced/bazaar.ts
+++ b/examples/typescript/facilitator/advanced/bazaar.ts
@@ -19,7 +19,10 @@ import { ExactEvmScheme } from "@x402/evm/exact/facilitator";
 import { UptoEvmScheme } from "@x402/evm/upto/facilitator";
 import { toFacilitatorSvmSigner } from "@x402/svm";
 import { ExactSvmScheme } from "@x402/svm/exact/facilitator";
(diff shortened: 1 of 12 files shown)


phdargen:
Thanks for the contribution @samsamtrum 

Note that we require commit signing, please rebase and verify your commits.

phdargen:
I'd also suggest to add a shared helper into `@x402/core/utils` and reusing it everywhere, sth like:
```
export function parseMoneyString(money: string): number {
  const cleaned = money.replace(/^\$/, "").trim();
  if (!/^-?\d+(?:\.\d+)?$/.test(cleaned) || /[eE]/.test(cleaned)) {
    throw new Error(`Invalid money format: ${money}`);
  }
  const amount = Number(cleaned);
  if (!Number.isFinite(amount) || amount < 0) {
    throw new Error(`Invalid money format: ${money}`);
  }
  return amount;
}
```

phdargen:
Thanks for the update @samsamtrum, could you please fix the formatting/lint?

phdargen (approved):

PR #2537, opened 2026-06-01
feat: add Cardano implementation for Typescript package

fabianbormann:
## Checklist


phdargen:
Thanks for the contribution @fabianbormann!

A few initial comments:
- the java sdk is legacy v1 only and effectively frozen, please remove from the PR. You could release it as a 3rd party sdk if you'd like
- I'd suggest we nail the implementation of one reference sdk first and to do the others as follow-ups. At first glance, your TS implementation looks most complete (Go seems to be server only), so this should be a good start. Please remove the others from the PR for now
- please add full e2e and integration tests for the reference sdk and add cardano to the */all_networks examples, see also https://github.com/x402-foundation/x402/blob/main/CONTRIBUTING.md#pr-2-reference-implementation
- please point me to a testnet faucet and also add this info to the READMEs

Kammerlo:
@phdargen Thanks for the feedback. We will incorporate that. We'll move the PR back to draft until we are done. I'm expecting it in the next few days.

I removed the other implementations already and will add an e2e example soon. 
A testnet faucet can be found here: https://docs.cardano.org/cardano-testnets/tools/faucet 

But I'll add it into the documentation of the e2e test as well, such that others can run the tests themselves easily.

Kammerlo:
@phdargen Update from our side :) 
- I added e2e tests for cardano
- removed the other languages and only stick with typescript
- Added all of that into the readmes - I hope it's understandable, if somethings still unclear for non-cardano let me know then I'm happy to add more docs, but I tried to keep it short
- Needed to force push to add signatures to all commits

phdargen:
Building the examples fails with:
```
cd examples/typescript
pnpm install && pnpm build
[WARN] Issues with peer dependencies found. Run "pnpm peers check" to list them.
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @parcel/watcher@2.5.6, msgpackr-extract@3.0.4

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
```

These new transitive dep of the cardano sdk need to be added to `allowBuilds` in:
https://github.com/x402-foundation/x402/blob/main/examples/typescript/pnpm-workspace.yaml

Likely similar issue in e2e

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -49,6 +51,10 @@ const PORT = process.env.PORT || "4022";
 
 // Configuration - optional per network (alphabetic order)
 const avmPrivateKey = process.env.AVM_PRIVATE_KEY as string | undefined;
+const cardanoMnemonic = process.env.CARDANO_MNEMONIC as string | undefined;
+const cardanoNetwork = process.env.CARDANO_NETWORK || "cardano:preprod";
```
phdargen: Build fails with:
```
@x402/advanced-facilitator-example:build: all_networks.ts:129:24 - error TS2345: Argument of type 'string' is not assignable to parameter of type '`${string}:${string}` | `${string}:${string}`[]'.
@x402/advanced-facilitator-example:build:
@x402/advanced-facilitator-example:build: 129   facilitator.register(CARDANO_NETWORK, new ExactCardanoScheme(cardanoSigner));
@x402/advanced-facilitator-example:build:                            ~~~~~~~~~~~~~~~
@x402/advanced-facilitator-example:build:
@x402/advanced-facilitator-example:build:
@x402/advanced-facilitator-example:build: Found 1 error in all_networks.ts:129
```

Kammerlo: Damn I tested it back and forth. I'll fix it asap. Sorry!

typescript/packages/mechanisms/cardano/src/signer.ts
```diff
@@ -0,0 +1,521 @@
...
+import { parseAssetUnit, parseUtxoRef } from "./utils";
+
+/**
+ * Provider connection used by the reference signers. Exactly one of
+ * `blockfrost` or `koios` must be supplied. These map directly onto the
+ * Evolution SDK provider configs.
+ */
+export type CardanoProviderConfig =
+  | { blockfrost: { baseUrl: string; projectId?: string }; koios?: never }
+  | { koios: { baseUrl: string; token?: string }; blockfrost?: never };
```
phdargen: Could this be framed more vendor neutral? For example just a single `rpcUrl` + `rpcApiKey` ? Is there any public rpc we could use as fallback?

examples/typescript/facilitator/advanced/all_networks.ts
phdargen: all good, no worries

phdargen:
Got some Ada from the faucet, but how to get USDM @Kammerlo ?

typescript/packages/mechanisms/cardano/src/signer.ts
Kammerlo: These are two different API providers, with different APIs. When building the client, we need to know which one it is. The sdk in the background is then handling the connection. In Cardano blockfrost and koios are usally the standard.

typescript/packages/mechanisms/cardano/src/constants.ts
```diff
@@ -0,0 +1,214 @@
+/**
+ * Network identifier for Cardano Mainnet.
+ *
+ * The x402 Cardano spec uses human-readable names ("cardano:mainnet") rather
+ * than canonical CAIP-2 form ("cardano:1" or genesis-hash CAIP-2). Treat these
+ * as x402-local network identifiers; do not "fix" them to canonical CAIP-2.
```
phdargen: x402 does use CAIP-2 throughout, if `cardano:mainnet` is not the canonical CAIP-2 form then please change to "cardano:1" or whatever the canonical form is

typescript/packages/mechanisms/cardano/src/constants.ts
```diff
@@ -0,0 +1,214 @@
...
+ */
+export const USDM_DEFAULT_DECIMALS = 6;
+
+/**
+ * Asset identifier for native ADA. The Cardano facilitator special-cases
+ * this value: lovelace lives in an output's `coin` field, not in its
+ * multi-asset map, so the verifier compares against `output.coin` when the
+ * asset string is exactly `"lovelace"`.
+ */
+export const LOVELACE_ASSET = "lovelace";
```
phdargen, line 119: Can we define this convention is the specs: https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_cardano.md

Why `lovelace` and not `ADA`?

typescript/packages/mechanisms/cardano/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  /**
+   * Parses a Money value (string with optional `$`/USD suffix or number) into a
+   * decimal number.
+   *
+   * @param money - The Money value to parse.
+   * @returns The decimal value as a number.
+   */
+  private parseMoneyToDecimal(money: Money): number {
```
phdargen: Please use core utils where appropriate:
`import { convertToTokenAmount, numberToDecimalString, parseMoneyString } from "@x402/core/utils";`

parseMoneyToDecimal + toAtomic duplicates this logic

typescript/packages/mechanisms/cardano/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,503 @@
...
+      // We keep the claim in place even on rejection so retries cannot
+      // rebroadcast the same transaction repeatedly.
+      if (submission.status !== "confirmed" && !this.acceptMempool) {
+        return {
+          success: false,
+          errorReason: ERR_SETTLEMENT_NOT_CONFIRMED,
+          transaction: submission.txHash,
+          network: payload.accepted.network,
+          payer: verifyResult.payer,
+          extensions: { status: submission.status },
```
phdargen: scheme specific info should go in `extra`, not `extensions`
```suggestion
          extra: { status: submission.status },
```

e2e/servers/express/index.ts
```diff
@@ -543,6 +563,38 @@ app.use(
...
         : {}),
+      ...(CARDANO_PAYEE_ADDRESS
+        ? {
+          "GET /exact/cardano": {
+            accepts: {
+              payTo: CARDANO_PAYEE_ADDRESS!,
+              scheme: "exact",
+              // Pay in lovelace (native tADA) so the e2e is fundable directly
+              // from the Cardano testnet faucet; preprod USDM is not faucet-
+              // available. 2 ADA is comfortably above the min-UTXO threshold.
```
phdargen: whats the min-UTXO threshold exactly? Would also be good to put this info in the specs

typescript/packages/mechanisms/cardano/src/constants.ts
Kammerlo: We use lovelace because it is the canonical on-chain unit of Cardano. Similar to how Ethereum specifications typically use wei rather than ETH, using the smallest indivisible unit avoids ambiguity around decimal conversions and makes the amount representation deterministic.

I'll add a note to the spec explicitly stating that amounts are denominated in lovelace (1 ADA = 1,000,000 lovelace).

e2e/servers/express/test.config.json
```diff
@@ -179,6 +179,13 @@
       "requiresPayment": true,
       "protocolFamily": "stellar"
     },
+    {
+      "path": "/exact/cardano",
```
phdargen: we will need tests for all 3 `assetTransfer` here, similar to `/exact/evm/eip3009` + `/exact/evm/permit2`:
```
/exact/cardano/default
/exact/cardano/masumi
/exact/cardano/script
```

typescript/packages/mechanisms/cardano/src/constants.ts
phdargen: OK more broadly, the spec doesnt mention native transfers at all. Does this not require its own payload creation and verification rules? Does ADA work identical to any other token like USDM?

e2e/servers/express/index.ts
Kammerlo: `minLovelace = utxoSize x coinsPerUTxOByte` 
I'll add a comment to the spec.

typescript/packages/mechanisms/cardano/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,503 @@
...
+
+  /**
+   * Verifies a Cardano payment against the supplied requirements following
+   * the spec's six rules.
+   *
+   * @param payload - The Cardano payment payload.
+   * @param requirements - The payment requirements being fulfilled.
+   * @returns A verify response describing success or failure.
+   */
+  async verify(
```
phdargen, line 295: Verification checks are so far only structural payload checks but should also check if the tx would actually succeed onchain, ideally with a full tx simulation. Or if thats not possible, at least some targeted checks like sufficient client token balance etc

typescript/packages/mechanisms/cardano/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,503 @@
...
+      // return a false-positive that would let callers grant access on an
+      // unpaid request. This witness-count check is structural only: it proves
+      // witness material is PRESENT, not that the signatures are valid. vkey
+      // signatures are not cryptographically validated here — that gate is the
+      // node's acceptance at submit time (settle). Per the eUTXO model,
+      // authorization is confirmed by settlement, so callers MUST grant access
+      // on confirmed settlement, never on verify() alone.
+      if (decoded.vkeyWitnessCount === 0 && decoded.scriptWitnessCount === 0) {
+        return { isValid: false, invalidReason: ERR_TRANSACTION_UNSIGNED, payer: "" };
+      }
```
phdargen, line 758: This should validate that the signature is actually valid. Otherwise the server might waste resources and do unnecessary work for an invalid authorization that is impossible to settle

typescript/packages/mechanisms/cardano/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,503 @@
...
+ *
+ * The duplicate-settlement cache is in-process only; across multiple
+ * facilitator instances the authoritative replay guard is the on-chain UTXO
+ * spend (rule 5), which makes the consumed nonce UTXO fail re-verification.
+ */
+export class ExactCardanoScheme implements SchemeNetworkFacilitator {
+  readonly scheme = SCHEME_EXACT;
+  readonly caipFamily = "cardano:*";
+
+  private readonly settlementCache = new Map<string, number>();
```
phdargen: The settlement cache is also missing in the specs. Could you please add a dedicated section to the spec justifying why this is needed? Similar to SVM: https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_svm.md#duplicate-settlement-mitigation-recommended

examples/typescript/facilitator/advanced/README.md
```diff
@@ -21,6 +22,9 @@ cp .env-local .env
 
 and fill required environment variables:
 
+- `CARDANO_MNEMONIC` - Cardano wallet seed phrase for the fee-paying facilitator wallet (optional)
```
phdargen: Hm reading through the specs and code, it is my understanding that the client builds a fully signed transaction and the facilitator only broadcasts it, implying the client pays for gas? Could you please clarify who pays the tx fee (client vs facilitator) @Kammerlo @fabianbormann?

If the client indeed pays it, statements like `fee-paying facilitator wallet` are misleading and why does the facilitator then need a wallet signer at all? All other network schemes support fee sponsorship, what we consider very important for user experience. Any options to add this or is this a fundamental network-level limitation?

typescript/packages/mechanisms/cardano/src/constants.ts
Kammerlo: Yes in Cardano any token is treated like a native token.

typescript/packages/mechanisms/cardano/src/exact/facilitator/scheme.ts
Kammerlo: We can't do a full tx simulation. What is possible is to check if all inputs are unspent. It's added.

examples/typescript/facilitator/advanced/README.md
Kammerlo: Made the facilitator wallet optional. It was only used for returning an address. If not set it's returning an empty list on the `/supported` endpoint. 
We could add in a later iteration also add fee sponsorship. This can be done through co-signing a transaction, but that would require a more interactive flow and more changes in the spec.

typescript/packages/mechanisms/cardano/src/constants.ts
Kammerlo: There's actually no strictly canonical CAIP-2 form for Cardano. There's no registered CASA cardano namespace, so `cardano:mainnet` and `cardano:1` are equally non-canonical, and `cardano:1` is in fact ambiguous since mainnet is network id 1 but both preprod and preview are id 0. The only standardized identifier is CIP-34's `cip34:NetworkId-NetworkMagic` (e.g. `cip34:1-764824073`), which disambiguates correctly but has poor DX, so we keep the readable `cardano:mainnet/preprod/preview` as the canonical advertised ids. I've added normalization so clients/facilitators also accept the CIP-34 forms as input aliases and fold them to the canonical id — same approach the SVM mechanism already uses, where normalizeNetwork (in mechanisms/svm/src/utils.ts) maps legacy names like solana/solana-devnet onto their canonical solana:<genesisHash> CAIP-2 ids. I've also added a "Network Identifiers" section to the spec documenting the rationale.

Kammerlo:
@phdargen I implemented all of your comments. Thanks first of all! 
if you need anything else changed, let me know.

phdargen:
> Building the examples fails with:
> 
> ```
> cd examples/typescript
> pnpm install && pnpm build
> [WARN] Issues with peer dependencies found. Run "pnpm peers check" to list them.
> [ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: @parcel/watcher@2.5.6, msgpackr-extract@3.0.4
> 
> Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
> ```
> 
> These new transitive dep of the cardano sdk need to be added to `allowBuilds` in: https://github.com/x402-foundation/x402/blob/main/examples/typescript/pnpm-workspace.yaml
> 
> Likely similar issue in e2e

Still getting this issue in both examples and e2e, please add '@parcel/watcher' and msgpackr-extract to `allowBuilds` in:
https://github.com/x402-foundation/x402/blob/main/examples/typescript/pnpm-workspace.yaml
and 
https://github.com/x402-foundation/x402/blob/main/e2e/pnpm-workspace.yaml

examples/typescript/servers/advanced/all_networks.ts
```diff
@@ -81,6 +85,14 @@ if (avmAddress) {
     payTo: avmAddress,
   });
 }
+if (cardanoAddress) {
+  accepts.push({
+    scheme: "exact",
+    price: "$0.001",
+    network: CARDANO_NETWORK,
+    payTo: cardanoAddress,
```
phdargen, line 122: It appears no USDM faucet is available, so the example should also charge ADA like e2e or it is unusable

phdargen: would it be possible to send me some testnet USDM, so I can at least test non ADA payments once locally? 
addr_test1qplc7wx60kzxe7vsshrmrr3dx6t9ldfeth5dnl3h5s0gn9ac3hs3rekjj4ssz5gc236h60uc7w083l5rwsknp2rc6pps6th6c8

phdargen:
e2e run result: 

`pnpm test --testnet --min -v --families=cardano`
```
📋 Detailed Test Results
========================

✅ PASSED TESTS:

  # 1: axios → express → /exact/cardano/default
      Facilitator: typescript
      Network: cardano:preprod
      Tx: 6755eb08d1991e47bdb8bfb872747d1c4801b88c12266d4f1defff818702e9f7
  # 2: axios → express → /exact/cardano/masumi
      Facilitator: typescript
      Network: cardano:preprod
      Tx: 8da3f6b41fc621c45f9ac901ea7a6c3d8f65d8d5191cafed82add03f4bdaf33b
  # 3: axios → express → /exact/cardano/script
      Facilitator: typescript
      Network: cardano:preprod
      Tx: 3700403c36c1622ecabf4629ea1e7b2b1aff752858ab34d4c6e3aebd7c080f5c
  # 4: fetch → express → /exact/cardano/default
      Facilitator: typescript
      Network: cardano:preprod
      Tx: 541ead733127e5199c3c71e6715f18c256faff196e8d6267410cdd2fa36f4454

❌ FAILED TESTS:

  # 5: axios → fastify → /exact/cardano
      Facilitator: typescript
      Error: $ tsx index.ts
{"success":false,"error":"Failed to create payment payload: Coin selection failed for lovelace: 0, lovelace...: 0, 16a55b2a349361ff...: 1000","status_code":500}

  # 6: axios → hono → /exact/cardano
      Facilitator: typescript
      Error: $ tsx index.ts
{"success":false,"error":"Failed to create payment payload: Coin selection failed for lovelace: 0, lovelace...: 0, 16a55b2a349361ff...: 1000","status_code":500}

  # 7: axios → next → /api/exact/cardano
      Facilitator: typescript
      Error: $ tsx index.ts
{"success":false,"error":"Failed to create payment payload: Coin selection failed for lovelace: 0, lovelace...: 0, 16a55b2a349361ff...: 1000","status_code":500}

  # 8: axios → next → /api/exact/cardano/withx402
      Facilitator: typescript
      Error: $ tsx index.ts
{"success":false,"error":"Failed to create payment payload: Coin selection failed for lovelace: 0, lovelace...: 0, 16a55b2a349361ff...: 1000","status_code":500}


📊 Breakdown by Facilitator:
 typescript      ✅ 4 / ❌ 4 (50%)

📊 Breakdown by Server:
 express              ✅ 4 / ❌ 0 (100%)
 fastify              ✅ 0 / ❌ 1 (0%)
 hono                 ✅ 0 / ❌ 1 (0%)
 next                 ✅ 0 / ❌ 2 (0%)

📊 Breakdown by Client:
   axios                ✅ 3 / ❌ 4 (43%)
   fetch                ✅ 1 / ❌ 0 (100%)
```

Seems the issue is that all servers but express still use USDM instead of ADA

specs/schemes/exact/scheme_exact_cardano.md
```diff
@@ -354,10 +365,59 @@ Schema:
...
+```
+
+- `coinsPerUtxoByte` — a protocol parameter, currently **4310 lovelace/byte** on mainnet and the public testnets. It is governance-settable, so implementations MUST read it from live protocol parameters rather than hardcode it.
+- `160` — a constant overhead (20 words x 8 bytes) covering the transaction input and its entry in the UTXO map.
+- `|serialized_output|` — the CBOR byte length of the output (address, value, and any datum or script reference).
+
+Because the threshold is a function of output size, **there is no single fixed value**:
+
+- **Pure-lovelace payment** (`asset: "lovelace"`): the requested `amount` is itself the output coin and must satisfy the formula — in practice ~1 ADA (≈ 1,000,000 lovelace) for a minimal output. An `amount` below the minimum yields an unbalanceable transaction and MUST be rejected at build time.
+- **Native-asset payment** (e.g. USDM): the multi-asset entry (28-byte policy id + asset name + quantity) enlarges the output, raising the minimum to roughly **1.2–1.5 ADA**. This ADA is sent **in addition to** the token amount, is drawn from the client's inputs, and travels with the token until the recipient spends the output.
```
phdargen, line 909: Do I understand correctly that for a USDM payment, the total client wallet spend = requested USDM + tx fee + min ADA? 

The tx fee is consumed by the network, while the server gets 1.2–1.5 ADA on top of the requested USDM? So they should consider this in their pricing that they get an extra ~0.2$ in ADA (at todays prices)

examples/typescript/facilitator/advanced/README.md
phdargen: OK lets add a `extra.areFeesSponsored=false` then, aligning with Stellar:
https://github.com/x402-foundation/x402/blob/b1c4734bc2bdfb8719770c7939fa1a69dcbee97e/specs/schemes/exact/scheme_exact_stellar.md?plain=1#L54-L56

specs/schemes/exact/scheme_exact_cardano.md
Kammerlo: Yes exactly. Every Utxo must hold the minAda.

e2e/servers/express/index.ts
```diff
@@ -643,6 +685,60 @@ app.use(
...
+                  identifierFromPurchaser: "aabbaabb11221122aabb",
+                  sellerVkey: "deadbeef",
+                  paymentType: "Web3CardanoV1",
+                  blockchainIdentifier: "blockchain_identifier",
+                  payByTime: "1713626260",
+                  submitResultTime: "1713636260",
+                  unlockTime: "1713636260",
+                  externalDisputeUnlockTime: "1713636260",
+                  agentIdentifier: "agent_identifier",
+                  inputHash: "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
```
phdargen: The implementation of the "masumi" assetTransferMethod is almost entirely missing. The server just declares some dummy constants, the client ignores them and signs a regular 'default' tx and the facilitator also treats it like a regular 'default' tx without any "masumi" specific logic

e2e/servers/express/index.ts
```diff
@@ -643,6 +685,60 @@ app.use(
...
+              accepts: {
+                // payTo is the script address the facilitator reconstructs from
+                // the script descriptor below and verifies it matches.
+                payTo: CARDANO_SCRIPT_ADDRESS,
+                scheme: "exact",
+                price: { amount: "2000000", asset: "lovelace" },
+                network: CARDANO_NETWORK,
+                extra: {
+                  assetTransferMethod: "script",
+                  script: { type: "plutusV3", code: CARDANO_SCRIPT_CODE },
```
phdargen: The facilitator has some minimal additional verify logic for 'script' but seems like the client also just ignores extra and builds a regular tx

phdargen:
Hi @Kammerlo @fabianbormann, the implementation of the `default` method looks good, just a couple of minor fixes before we could get this in. 

The other two methods are neither fully implemented nor sufficiently specified, though. The client and (for the most part) the facilitator implementations ignore the server-declared extra fields, so effectively there's no difference from a regular `default` tx.

More fundamentally: after doing some research, my understanding is that the intent of `script` and `masumi` is to use the x402 payment only to lock funds into a contract, with the real "delivery" happening in later, separate transactions governed by the validator (script) or by an escrow protocol and coordinator (masumi). x402's /verify + /settle only ever cover the lock step, and what comes after isn't detailed in the spec at all. How and when are funds released from escrow? And what's the use case or benefit relative to a regular default tx that just settles in one go? 
For `masumi` in particular, it seems the client would also need a pre-payment negotiation/handshake before even making the resource request. All of these additional lifecycle operations would need to be specified.

Given that the payment flow for these appears to be significantly different from a regular `exact` tx, I'm not sure `script` and `masumi` belong in this scheme at all. Putting them behind an `assetTransferMethod` under `exact` hides a lot of this complexity and nuance

My suggestion would be to land the `default` implementation now and return to a spec-level discussion for `masumi` and `script` before following up with their implementations

examples/typescript/servers/advanced/all_networks.ts
Kammerlo: I found a faucet: https://tusdm.moneta.global/
Additionally I send your address a few hundred tUSDM

Kammerlo:
Hello @phdargen, thanks for your patience and for the thorough review — the deep dive into the Cardano side is much appreciated.
I gathered feedback from a few stakeholders and have since extended both methods and the specification. Let me respond to the main points directly.
The intent behind having three methods is separation of concerns:

- `default` - address-to-address, the simplest transfer.
- `script` - generic smart-contract access. It lets the client attach an arbitrary datum and lock the payment into any contract, to be retrieved later according to that contract's logic. The datum itself is intentionally not verified by the facilitator: this is generic access, so the facilitator doesn't know the target contract's rules. It's the payment path for programmable contract payments that need a datum.
- `masumi` - a specific agent-to-agent payment protocol built on top of that idea. Here we added dedicated datum fields, and the facilitator does verify them — deadlines, collateral and cooldowns — precisely to prevent funds from being locked indefinitely. Masumi is successful AI agent protocol on cardano.

I hope the added specification and the extra verification make our intent clearer. Thanks for the extensive feedback and the deep dive into Cardano — happy to discuss further, but keeping these three methods is important for us.

examples/typescript/pnpm-workspace.yaml
```diff
@@ -26,3 +26,5 @@ allowBuilds:
   sharp: true
   unrs-resolver: true
   utf-8-validate: true
+  parcel/watcher: true
```
phdargen:
```suggestion
  '@parcel/watcher': true
```

e2e/pnpm-workspace.yaml
```diff
@@ -29,3 +29,5 @@ allowBuilds:
   sharp: true
   unrs-resolver: true
   utf-8-validate: true
+  parcel/watcher: true
```
phdargen:
```suggestion
  '@parcel/watcher': true
```

examples/typescript/servers/advanced/all_networks.ts
phdargen: Please add this link to the docs

examples/typescript/facilitator/advanced/README.md
phdargen: Please add the `extra.areFeesSponsored=false` flag @Kammerlo 

Facilitator announces it in /supported as kinds[].extra.areFeesSponsored.
Resource server loads /supported on initialize(), and when building payment requirements calls enhancePaymentRequirements(…, supportedKind, …), which copies the flag into PaymentRequirements.extra

typescript/packages/mechanisms/cardano/src/constants.ts
```diff
@@ -0,0 +1,282 @@
...
+export const SCHEME_EXACT = "exact";
+
+/**
+ * USDM policy id on Cardano Mainnet.
+ */
+export const USDM_MAINNET_POLICY_ID = "c48cbb3d5e57ed56e276bc45f99ab39abe94e6cd7ac39fb402da47ad";
+/**
+ * USDM policy id on Cardano Preprod.
+ */
+export const USDM_PREPROD_POLICY_ID = "e675b46e4d2242c991a8932a99db3044e80515ae14b4c4ccf6b3f4c9";
```
phdargen, line 88: spec file uses `16a55b2a349361ff88c03788f93e1e966e5d689605d044fef722ddde`, please align

Kammerlo: Will do it right away. I'm currently resolving the merge conflicts as well. Thanks!

e2e/servers/express/test.config.json
```diff
@@ -183,6 +183,34 @@
...
+    },
+    {
+      "path": "/exact/cardano/masumi",
+      "method": "GET",
+      "description": "Protected Cardano endpoint using the masumi assetTransferMethod",
+      "requiresPayment": true,
+      "protocolFamily": "cardano"
+    },
+    {
+      "path": "/exact/cardano/masumi-usdm",
```
phdargen: Non-`express` routes charge $0.001 (defaults to USDM), while `express` uses lovelace on the base route and has a seperate /masumi-usdm path, so the test matrix is inconsistent. 

Suggest to follow Hedera/NEAR:

- Keep /exact/cardano/{default,masumi,script} (method only)
- Drop /masumi-usdm
- Drive price with something like CARDANO_ASSET (default lovelace) + CARDANO_AMOUNT for all servers: express/hono/....
- USDM runs: set env to CARDANO_ASSET = e675b46e4d2242c991a8932a99db3044e80515ae14b4c4ccf6b3f4c9.....

phdargen:
> Hello @phdargen, thanks for your patience and for the thorough review — the deep dive into the Cardano side is much appreciated. I gathered feedback from a few stakeholders and have since extended both methods and the specification. Let me respond to the main points directly. The intent behind having three methods is separation of concerns:
> 
> * `default` - address-to-address, the simplest transfer.
> * `script` - generic smart-contract access. It lets the client attach an arbitrary datum and lock the payment into any contract, to be retrieved later according to that contract's logic. The datum itself is intentionally not verified by the facilitator: this is generic access, so the facilitator doesn't know the target contract's rules. It's the payment path for programmable contract payments that need a datum.
> * `masumi` - a specific agent-to-agent payment protocol built on top of that idea. Here we added dedicated datum fields, and the facilitator does verify them — deadlines, collateral and cooldowns — precisely to prevent funds from being locked indefinitely. Masumi is successful AI agent protocol on cardano.
> 
> I hope the added specification and the extra verification make our intent clearer. Thanks for the extensive feedback and the deep dive into Cardano — happy to discuss further, but keeping these three methods is important for us.

Thanks for the expanded spec and verification! 

I can get behind `script` as it apprears analogous to an EVM server using a smart-contract address as payTo, with the Cardano-specific need to attach a datum.

I’m still concerned that `masumi` leaves too much of the real payment flow out of band (purchase creation, how binding fields reach the 402, and everything after lock). Could you clarify the complete end-to-end flow: who creates the purchase and how, how those fields get into PaymentRequirements, and what happens after /settle?

I also think there are some gaps in `masumi` 402 creation: timestamps like `payByTime` need to be dynamic per request, but currently they are hardcoded. `extra` also includes client-specific fields like `identifierFromPurchaser`, `inputHash`, `buyerReturnAddress`. How does the server know those? Even if they were stored somewhere, the client’s initial request is unauthenticated HTTP, so the server doesn’t know which client is calling and can’t fill the 402 accordingly. If these are client-provided inputs, do they need to be in the 402 at all?

Kammerlo:
@phdargen Thanks for staying on this — `payByTime` was a real bug, good catch. We've also added a clearer Masumi flow.

As the requirements grew we added more options for configuring settlement, e.g. the submission and confirmation policy. After discussions with our stakeholders we've extended x402 Masumi further to make it clearer what it does: the spec now describes the full flow end to end, and we've added request commitments and signatures so buyer and seller provably agree on the same material. This is already used in a real product. 

On the fields you flagged: `identifierFromPurchaser` and `buyerReturnAddress` aren't in `extra` at all; the buyer supplies those at lock time, and `inputHash` is just the digest of the request the server already received. Everything in the 402 is issuer-derived, so the server never needs to know who's calling.

On Hydra: it stays in the spec on purpose `settlementPolicy`, `settlementLayer` and `headId` are specified so the wire format doesn't have to change later — but this implementation refuses it explicitly rather than half-supporting it. Head suitability and snapshot evidence deserve their own discussion, so we'll follow up with a separate PR for the implementation.

Hope that's clearer now. If not, let me know and we'll realign.

phdargen:
Hi @Kammerlo, thanks for the update! We made a couple of high-priority sdk changes in the last 2 weeks that unblock several work streams. Could you please rebase to upstream/main and make sure to account for the following changes: 
- payment flows: https://github.com/x402-foundation/x402/pull/3053
- spend controls and unified default asset declaration + money parsing: https://github.com/x402-foundation/x402/pull/3124
- e2e refactor: https://github.com/x402-foundation/x402/pull/2976

Please let me know if you have questions about any of the changes

phdargen:
Thanks for expanding the Masumi spec, that clears things up and I agree on the direction.
As long as a standard x402 SDK client can make a masumi payment fully in protocol thats fine. 
The core x402 flow then ends with the http request cycle, everything after is handled by 3rd party SDKs.
 
My current understanding of the flow:

In protocol (standard @x402/cardano):
- Buyer hits the protected resource (this request is the purchase order)
- Server issues a one-shot 402 (commitment, deadlines, termsDigest) and must reuse that same object on the paid retry
- Client verifies the seller signature, builds the escrow lock, retries with PAYMENT-SIGNATURE
- Facilitator /verify + /settle lands FundsLocked
- Server returns HTTP 200. That is where x402 stops

Out of band (server that opted into Masumi, 3rd party SDK):
Result submission, refund, dispute withdrawal etc

phdargen:
This PR has significant changes in @x402/core and every HTTP adapter to implement a new HTTP replay/idempotency mechanism. In general, new network SDK additions should not modify the core packages unless absolutely needed, and should implement everything in the mechanism using scheme-level lifecycle hooks if the core flow needs to be modified.

For the default and script assetTransferMethods, no core package changes should be required.

For assetTransferMethod=masumi, I would argue it is also not needed, and the approach goes in the wrong direction.

Masumi needs to persist the issued 402 and, on the paid retry, confirm payload.accepted is that same quote. The current design assumes a single in-flight 402. It does not survive concurrent clients: A gets a 402, B gets a 402, A retries. Anonymous requests to the same path share a fingerprint, and the Masumi path keeps one in-memory “current offer” that B overwrites. A’s retry is then matched against B’s quote. In-memory / same-adapter WeakMap state has the same shape of bug across processes.

That handshake belongs in the mechanism, keyed off the payment payload, not the HTTP request. Persist the issued requirements under termsDigest and on the paid retry compare payload.accepted to that record.

This should work similarly to the server's ChannelStorage in the batch-settlement scheme, which defines a general storage interface with a default in-memory implementation; production services can implement their own backends (Redis etc.): see [storage.ts](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/evm/src/batch-settlement/server/storage.ts) and the [batch-settlement server example](https://github.com/x402-foundation/x402/tree/main/examples/typescript/servers/batch-settlement#storage).

phdargen:
Copying the result of some quick brainstorming with my AI below. Should be a good start, feel free to diverge as you see fit

> # Revert HTTP replay; Masumi terms store in-mechanism
> 
> Cardano currently patches `@x402/core` / `@x402/http` with generic HTTP replay plumbing (`getRawBody`, request fingerprints, skipHandler status/raw replay, abort HTTP statuses). The `cardanoReplayProtection` extension itself lives in the mechanism; it is not a core API, but it is what that plumbing was added for. Default/script should work like EVM exact. Masumi only needs to persist the issued `PaymentRequirements` and compare the paid retry’s `accepted` object, entirely in `@x402/cardano`.
> 
> The HTTP/in-memory design fails first under concurrent clients, not a lonely retry: Client A 402, Client B 402, Client A paid retry. That is quote/handshake lookup, not HTTP replay: A must confirm against the 402 A was issued. Anonymous requests to the same path share a fingerprint; Masumi e2e keeps one mutable `CARDANO_MASUMI_REQUIREMENTS` that B overwrites. A’s retry then matches against B’s quote. Keying the store by `termsDigest` from `payload.accepted` keeps A’s offer even if B is issued in between.
> 
> 
> ```mermaid
> sequenceDiagram
>   participant Client
>   participant Server
>   participant Store as MasumiTermsStorage
>   participant Facilitator
>   Client->>Server: unpaid request
>   Server->>Store: put termsDigest, issued requirements
>   Server-->>Client: 402 with extra
>   Client->>Server: paid retry with accepted plus tx
>   Server->>Store: load by termsDigest, deepEqual accepted, CAS first txHash
>   alt mismatch or other tx
>     Server-->>Client: 402 invalid
>   else match
>     Server->>Facilitator: verify or settle
>   end
> ```
> 
> ## 1. Revert core and HTTP adapters
> 
> Restore these files from `upstream/main` (they exist only because of Cardano HTTP replay):
> 
> - [`typescript/packages/core/src/client/x402Client.ts`](typescript/packages/core/src/client/x402Client.ts) — stop passing `resource` into `createPaymentPayload` context
> - [`typescript/packages/core/src/types/mechanisms.ts`](typescript/packages/core/src/types/mechanisms.ts) — drop `PaymentPayloadContext.resource`, `SchemePaymentRequiredContext.requirement`, `requireMatchingPayloadResource`
> - [`typescript/packages/core/src/http/x402HTTPResourceServer.ts`](typescript/packages/core/src/http/x402HTTPResourceServer.ts) — drop `getRawBody`, resource-matching, `isRaw`, `httpStatus`, `responseStatus`
> - [`typescript/packages/core/src/server/x402ResourceServer.ts`](typescript/packages/core/src/server/x402ResourceServer.ts) — drop abort `status`, skipHandler `status`/`headers`/`isRaw`
> - Matching core tests/mocks and [`typescript/packages/http/express|fastify|hono|next`](typescript/packages/http) adapter changes
> 
> ## 2. Replace the resource-server operation store with Masumi terms storage
> 
> Delete `CardanoOperationStore`, `InMemoryCardanoOperationStore`, HTTP fingerprinting, replay challenges, `skipHandler` response capture, `WeakMap` adapter ownership, `requestBinding`, and `requireMatchingPayloadResource`.
> 
> Keep [`CardanoSettlementStore`](typescript/packages/mechanisms/cardano/src/idempotency.ts) (facilitator tx-hash CAS). That stays mechanism-local and is not an HTTP protocol.
> 
> Add [`typescript/packages/mechanisms/cardano/src/exact/masumi/storage.ts`](typescript/packages/mechanisms/cardano/src/exact/masumi/storage.ts), modeled on [`ChannelStorage.updateChannel`](typescript/packages/mechanisms/evm/src/batch-settlement/server/storage.ts):
> 
> - Key: `termsDigest`
> - Value: `{ termsDigest, requirements: PaymentRequirements, claimedTxHash?: string }`
> - Atomic `updateTerms(digest, current => next)` with the same concurrency contract as `updateChannel` (in-memory lock; production backends must CAS)
> - `InMemoryMasumiTermsStorage` for tests/dev
> - No file/Redis backend in this change (integrators implement the interface)
> 
> Wire it on [`ExactCardanoScheme`](typescript/packages/mechanisms/cardano/src/exact/server/scheme.ts):
> 
> - Constructor: `new ExactCardanoScheme()` with no required store (default/script like EVM). Optional `masumiStorage`; default `new InMemoryMasumiTermsStorage()`.
> - `enrichPaymentRequiredResponse`: for each masumi accept, `updateTerms(digest, current => current ?? { termsDigest, requirements: clone(accept) })` so the first 402 wins and a paid-retry 402 cannot rotate terms.
> - `onAfterVerify`: **masumi only**. Compute digest from `payload.accepted`, abort if unknown, if `!deepEqual(accepted, stored.requirements)`, or if `claimedTxHash` is a different tx. CAS the first `txHash`. Import `deepEqual` from `@x402/core/utils`.
> - Abort with `{ abort: true, reason, message }` only (no custom HTTP status after the core revert). Do not `skipHandler`.
> 
> Default/script paths must not touch the store.
> 
> Client: drop `context?.resource` in [`exact/client/scheme.ts`](typescript/packages/mechanisms/cardano/src/exact/client/scheme.ts) (core will no longer pass it). Facilitator registry checks keep using `PaymentPayload.resource`, which core already copies after payload creation.
> 
> ## 3. Callers, tests, docs, spec
> 
> - e2e/examples: `new ExactCardanoScheme({ inMemoryStore: {} })` → `new ExactCardanoScheme()`. Express masumi extra getter can stay; enrich will persist each issued offer, and the retry compares `accepted` to the store (not the live getter).
> - Replace [`test/unit/idempotency.test.ts`](typescript/packages/mechanisms/cardano/test/unit/idempotency.test.ts) operation-store cases with `InMemoryMasumiTermsStorage` tests; keep settlement-store tests.
> - Rewrite HTTP-replay cases in [`test/unit/scheme.test.ts`](typescript/packages/mechanisms/cardano/test/unit/scheme.test.ts): store-on-issue, retry match, mutated `accepted` rejected, second tx for same digest rejected, default/script does not require storage.
> - README: remove `cardanoReplayProtection` / `CardanoOperationStore` / `requestBinding`. Document `masumiStorage` like batch-settlement channel storage.
> - Spec [`specs/schemes/exact/scheme_exact_cardano.md`](specs/schemes/exact/scheme_exact_cardano.md) (requested):
>   - Delete **Replay challenge** (`cardanoReplayProtection`)
>   - Drop handler-response replay and request-fingerprint from pending confirmation (handler waits for settlement like other exact schemes)
>   - Keep issuer “store complete requirements and compare `accepted`” (~line 197); state the resource-server key is `termsDigest`, first `txHash` is CAS-bound, comparison is the issued object vs retry `accepted`
>   - Keep facilitator **Duplicate Settlement Mitigation** as recommended (tx-id cache), not an HTTP extension
> 
> Exports: add masumi storage from server/`index.ts`; stop exporting `CardanoOperationStore`.
>

Kammerlo:
@phdargen Hey, I reverted all of it. Core and http adapters are with no changes left and I moved the code as you suggested into the mechanism. Thanks again for that recommendation! 
Additionally rebased on upstream and accounted for the three PRs you mentioned. 
One thing to mention: Cardano runs against the axios e2e client only, because hone and next reply with the unread branch of the `repsonse.clone()`. That branch doesn't survive reliably Cardano's block time. 
Happy to open a separate PR with a two-line fix and a regression test for each adapter if you think buffering the read branch instead is the right call there or adding it right into that PR. 

And fyi (if interesting for your :) ): For production we will probably use a facilitator written in java complying with the typescript library and the schema. https://github.com/cardano-foundation/cardano-x402-facilitator 
I tested the combination with a typescript client & server based on this library and then use it with the java facilitator. The demo library excercising the whole flow is here: https://github.com/Kammerlo/x402-cardano-demo

phdargen:
Thanks for addressing my previous comments @Kammerlo, the core/http revert looks clean.

The hono/next buffering bug is a good catch, I prepared a patch in 
https://github.com/x402-foundation/x402/pull/3392 which also increases the facilitator timeout to 90s.

Remaining issues below. 

## Blockers

- **Full e2e coverage:** We will need full e2e coverage of all server/client frameworks and transports and they should exercise the defaults that real users get. Please add a env config to choose `mempool` confirmations but default to not using it. Once my PR above is merged and you rebased, this should unblock the remaining client/server frameworks, please remove all test suite restrictions. This also includes MCP transport tests. Since the mausmi logic moved out of core/http into the mechanism, the implementation should be transport agnostic and it should no longer be needed to exclude MCP.  The actual gap is that Masumi quote issuance lives in the e2e harness as an HTTP-only dynamic accepts builder (`cardanoMasumiAccepts` in `e2e/servers/typescript/config.ts`). Please move per-request issuance + paid-retry lookup into the scheme's `enrichPaymentRequiredResponse` (you already export `issueMasumiRequirements`; the harness glue is exactly what every real Masumi server would otherwise reimplement)

- **Descope client submission mode (and Hydra) from this PR**:
The `submissionPolicy`/`submissionMode` design is new and conflicts with the family rules in `specs/schemes/exact/scheme_exact.md` (also see https://github.com/x402-foundation/x402/pull/3145) : client-submitted (payment-proof) methods MUST use the `upfront` flow, while this scheme declares every method uses `authorization`. This is a big increase in scope, lets please defer client submission + Hydra to a follow-up spec PR and keep this PR facilitator-broadcast only

- **Server submission must work out of the box:**
`verify()` rejects server-mode payloads unless the signer provides a complete phase-1 ledger validator, which no standard provider setup can supply (Cardano has no phase-1 dry-run API). I had to stub `validatePhase1Transaction: async () => undefined` in the advanced facilitator example to get a payment through. Please make the hook optional: by default `verify()` should validate as well as possible that settle will succeed. It already checks inputs-unspent, TTL bounds and min-UTXO; add value balance and fee floor, both computable from the input values you already fetch. Keep `validatePhase1Transaction` as an opt-in for facilitators that can provide stronger guarantees (e.g. running a full node)

- **Settlement store must have a working default:**
`ExactCardanoScheme` (facilitator) throws unless a `settlementStore` or `inMemorySettlementStoreMaxEntries` is passed explicitly. Compare SVM's duplicate-settlement cache: bounded in-memory by default, durable store as a documented production option. Please default to `new InMemoryCardanoSettlementStore()` and document the durable-store recommendation instead of throwing.

- `src/idempotency.ts` is misleadingly named: it's the duplicate-settlement guard, not HTTP idempotency. Please rename (e.g. `settlementStore.ts`) and drop "idempotency" wording from README/spec

- **Working examples:** examples are currently broken due to several of the issues above. Please make sure both e2e tests and examples run successfully before you commit 

## Optional / nice to have

- **`settlement_pending` instead of holding the connection**: core ships a non-terminal retry protocol (`pendingSettlementStore`, resource-server retry) used by the EVM path. The Cardano facilitator holds the settle HTTP call through confirmation (up to 90s) instead. Broadcasting, returning `settlement_pending` + tx hash, and letting the server retry would remove the long-held connection entirely and make timeouts a non-issue regardless of client defaults
- Please add the tUSDM faucet link (https://tusdm.moneta.global) to **both** the package README (`typescript/packages/mechanisms/cardano/README.md`, which currently says preprod USDM "must be sourced separately") and the Cardano funding section of `e2e/README.md`

phdargen:
https://github.com/x402-foundation/x402/pull/3392 has been merged @Kammerlo

Kammerlo:
@phdargen Thanks for the input and the fix. I'll add all of that right now. PR update should be there soon.

Kammerlo:
@phdargen Thanks for the review — all points addressed, rebased on `main` including #3392.

- **e2e coverage:** every TS client, server and transport now runs Cardano at the spec default (one confirmation), with no `excludeNetworks` overlays. `CARDANO_L1_CONFIRMATIONS=-1` opts into mempool settlement; unset is the default. Masumi quote issuance and the paid-retry lookup moved from the harness into `ExactCardanoScheme.enrichPaymentRequiredResponse`, so `cardanoMasumiAccepts` is gone and MCP works unchanged. Full matrix on preprod passes.
- **Client submission / Hydra:** descoped. Every method is facilitator-broadcast under the `authorization` flow; `submissionPolicy`, `submissionMode`, `settlementLayer`, `headId` and `terms.settlementPolicy` are removed from code, spec and README.
- **`verify()` without a phase-1 validator:** value conservation and the fee floor are computed from the fetched input values and live protocol parameters, on top of inputs-unspent, TTL and min-UTXO. `validatePhase1Transaction` stays an opt-in and is the only way to accept balance-changing operations.
- **Settlement store:** defaults to a bounded `InMemoryCardanoSettlementStore`; the durable-store recommendation is documented. `idempotency.ts` is now `settlementStore.ts` and the wording is dropped from README and spec.
- **`settlement_pending`:** `settle()` broadcasts, waits a bounded 75 s, and returns the non-terminal result with the tx hash; core's retry resumes the same transaction without rebroadcasting. Two harness notes: the e2e facilitator keeps its verified-payment record across `settlement_pending` so the retry passes hook 3, and the MCP e2e client passes a 180 s tool-call timeout because the MCP SDK's 60 s default is shorter than one confirmation plus the retry.
- **Examples:** run on defaults, no phase-1 stub. tUSDM faucet link added to both READMEs.

phdargen (approved):

phdargen:
Alright guys @Kammerlo @fabianbormann, great work! Let's get this in 🚀

PR #2540, opened 2026-06-01
fix(python): run failure hooks after after-hook errors

### LINKED ISSUE #2470: Python x402Client does not run failure hooks when after-payment hook throws
Summary: In Python, `_create_payment_payload_v2` drives hook execution from outside the generator. If an `on_after_payment_creation` hook throws, the exception escapes the dispatcher and never re-enters the generator’s `try`/`except`, so `on_payment_creation_failure` hooks do not run. This differs from TypeScript behavior and prevents guardrail-style integrations from rolling back provisional state after after-hook failures.

Expected: after-hook exceptions should trigger payment creation failure hooks, ideally with enough context to clean up any created payload/provisional state.

skyc1e:
Routes Python payment creation hook errors back through the shared generator driver, so `on_payment_creation_failure` hooks run when an `on_after_payment_creation` hook raises. This matches the existing failure path for scheme errors and keeps async/sync clients consistent.

Closes #2470

Checked with:
- `python -m pytest tests\unit\core\test_client.py -q`
- `python -m pytest tests\integrations\test_core.py -q`
- `python -m pytest tests\integrations\test_async_behavior.py::TestAsyncClientHooks -q`
- `python -m ruff check client.py tests\unit\core\test_client.py`

### DIFF
--- a/python/x402/changelog.d/2540.bugfix.md
+++ b/python/x402/changelog.d/2540.bugfix.md
@@ -1,0 +1,1 @@
+Run Python payment creation failure hooks when after-payment hooks raise.
--- a/python/x402/client.py
+++ b/python/x402/client.py
@@ -7,13 +7,15 @@ Async is the default with full async hook support.
 from __future__ import annotations
 
 import asyncio
-from typing import Any
+from collections.abc import Generator
+from typing import Any, TypeVar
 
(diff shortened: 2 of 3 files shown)


phdargen (approved):
Thanks for the fix @skyc1e

PR #984, opened 2026-01-17
feat: add exact payment scheme spec for XRPL

aristotle-satoshi:
Add specification for the exact payment scheme on XRP Ledger, supporting both native XRP and IOU (issued currency) payments.

## Description

This PR adds the `exact` payment scheme specification for the XRP Ledger (XRPL), following the pattern established by existing chain-specific specs (EVM, SVM, Sui, Aptos).

Key features:
- Support for native XRP payments (amounts in drops)
- Support for IOU (Issued Currency) payments with issuer validation
- CAIP-2 network identifiers: `xrpl:0` (Mainnet), `xrpl:1` (Testnet), `xrpl:2` (Devnet)
- Invoice binding via Memos or InvoiceID field
- Comprehensive facilitator verification rules
- Security considerations for replay protection and partial payment prevention

## Tests

This is a specification document only - no code changes.

- [x] Spec structure matches existing scheme specs (`scheme_exact_evm.md`, etc.)
- [x] Uses MUST/SHOULD/MAY language per specs/CONTRIBUTING.md guidelines
- [x] Includes JSON examples for PaymentRequirements and PaymentPayload
- [x] Documents Security Considerations section

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge)

PR #777, opened 2025-12-13
specs: add XRPL exact scheme (doc-only proposal)

TusharPardhe:
Adds an initial scheme document for supporting x402 v2 `exact` payments on XRPL (XRP Ledger).

Scope (intentional, to keep the first review deterministic):
- XRPL public networks only: mainnet/testnet/devnet (CAIP-2 proposal: `xrpl:0`, `xrpl:1`, `xrpl:2`)
- Native XRP payments only (`asset: "XRP"`, `amount` in drops)
- Payer-signed `Payment` tx blob as the `PaymentPayload.payload` (`txBlob`)
- Verifier rejects partial payments and any path / cross-currency fields
- Settlement requires a validated-ledger success result

Open question for maintainers:
- CAIP-2 network id convention: this doc proposes `xrpl:0`/`xrpl:1`/`xrpl:2`; happy to align with preferred convention.

Doc added:
- specs/schemes/exact/scheme_exact_xrpl.md

PR #2547, opened 2026-06-02
spec(xrpl): add exact scheme

aristotle-satoshi:
## Description

Add the XRPL `exact` payment scheme spec for x402 v2. This is now a spec-only PR under the new-chain review flow and does not add SDK implementation, package wiring, docs updates, or facilitator changes.

This carries forward the XRPL exact spec work from #984 for review.

## Scope

- Defines CAIP-2 network identifiers for XRPL as `xrpl:{network_id}`.
- Defines payer-signed XRPL `Payment` settlement for native XRP and issued currencies.
- Specifies verification, settlement, invoice binding, NetworkID handling, fee responsibility, and Sequence/TicketSequence requirements.

## Tests

- `git diff --check $(git merge-base HEAD origin/main)..HEAD`
- `git diff $(git merge-base HEAD origin/main)..HEAD | rg -n "xrpl-x402-facilitator|t54.ai|private repo|ported from|PR777|#777|pull/777|typescript/packages/mechanisms/xrpl|pnpm-lock|changeset|docs/core-concepts|docs/schemes" || true`
- `git merge-tree $(git merge-base HEAD origin/main) HEAD origin/main | rg -n "CONFLICT|<<<<<<<|>>>>>>>" || true`

### DIFF
--- a/specs/schemes/exact/scheme_exact_xrpl.md
+++ b/specs/schemes/exact/scheme_exact_xrpl.md
@@ -1,0 +1,437 @@
+# Exact Payment Scheme for XRP Ledger (XRPL) (`exact`)
+
+This document specifies the `exact` payment scheme for the x402 protocol v2 on the XRP Ledger.
+
+This scheme facilitates payments of a specific amount of XRP or an issued currency (IOU) on the XRP Ledger using a payer-signed `Payment` transaction.
+
+## Scheme Name
+
+`exact`
+
+## Payment Model
+
+| Aspect                    | Description                                                                  |
+| ------------------------- | ---------------------------------------------------------------------------- |
+| **Payment authorization** | The payer signs a standard XRPL `Payment` transaction                        |
+| **Settlement**            | The facilitator submits the signed transaction to XRPL                       |
+| **Fee payer**             | The payer pays the XRPL transaction fee embedded in the signed transaction   |
+
+XRPL charges the transaction fee to the transaction `Account`. This exact scheme therefore does not support facilitator-sponsored network fees for the signed `Payment` transaction. Supporting fee sponsorship would require a different payment model, not only a facilitator implementation change.
+
+## Network Identifier (CAIP-2)
+
+x402 v2 requires CAIP-2 network identifiers. For XRPL, the format is:
+
+```text
+xrpl:{network_id}
+```
+
+Where `network_id` is the XRPL numeric NetworkID (`uint32`).
+
+Common XRPL network identifiers:
+
+| Network | Identifier |
+| ------- | ---------- |
+| Mainnet | `xrpl:0`   |
+| Testnet | `xrpl:1`   |
+| Devnet  | `xrpl:2`   |
+
+## Protocol Flow
+
+The protocol flow for `exact` on XRPL is client-driven.
+
+1. **Client** makes a request to a **Resource Server**.
+2. **Resource Server** responds with a payment required signal containing `PaymentRequired` in the `PAYMENT-REQUIRED` header (base64-encoded JSON).
+3. **Client** creates a `Payment` transaction to the resource server's XRPL address for the specified amount.
+4. **Client** signs the transaction with their wallet, producing a fully signed transaction blob.
+5. **Client** encodes the signed transaction as a hex string.
+6. **Client** sends a new request to the resource server with the `PAYMENT-SIGNATURE` header containing the base64-encoded `PaymentPayload`.
+7. **Resource Server** forwards the `PaymentPayload` and `PaymentRequirements` to a **Facilitator Server's** `/verify` endpoint.
+8. **Facilitator** decodes the `signedTxBlob`, deserializes the proposed transaction, and validates it against the expected payment parameters.
+9. **Facilitator** returns a `VerifyResponse` to the **Resource Server**.
+10. **Resource Server**, upon successful verification, forwards the payload to the facilitator's `/settle` endpoint.
+11. **Facilitator Server** re-runs verification and submits the signed transaction to the XRPL network identified by `paymentRequirements.network`.
+12. Upon successful validated settlement, the **Facilitator Server** responds with a `SettlementResponse` to the **Resource Server**.
+13. **Resource Server** grants the **Client** access to the resource via the `PAYMENT-RESPONSE` header.
+
+## x402 v2 Headers
+
+| Direction                   | Header              | Content                                 |
+| --------------------------- | ------------------- | --------------------------------------- |
+| Server -> Client (challenge) | `PAYMENT-REQUIRED`  | Base64-encoded JSON `PaymentRequired`   |
+| Client -> Server (payment)   | `PAYMENT-SIGNATURE` | Base64-encoded JSON `PaymentPayload`    |
+| Server -> Client (result)    | `PAYMENT-RESPONSE`  | Base64-encoded JSON settlement response |
+
+Legacy header names (`X-PAYMENT`, `X-PAYMENT-RESPONSE`) are deprecated and SHOULD NOT be used for new integrations.
+
+## `PaymentRequirements` for `exact`
+
+The resource server advertises payment requirements in the `accepts` array.
+
+### XRP (Native) Example
+
+```json
+{
+  "scheme": "exact",
+  "network": "xrpl:0",
+  "asset": "XRP",
+  "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+  "amount": "1000000",
+  "maxTimeoutSeconds": 600,
+  "extra": {
+    "invoiceId": "INV-2025-001"
+  }
+}
+```
+
+### IOU (Issued Currency) Example
+
+```json
+{
+  "scheme": "exact",
+  "network": "xrpl:0",
+  "asset": "524C555344000000000000000000000000000000",
+  "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+  "amount": "10.5",
+  "maxTimeoutSeconds": 600,
+  "extra": {
+    "issuer": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q",
+    "invoiceId": "INV-2025-002"
+  }
+}
+```
+
+### Field Definitions
+
+| Field                  | Type    | Required | Description                                      |
+| ---------------------- | ------- | -------- | ------------------------------------------------ |
+| `scheme`               | string  | Yes      | Must be `"exact"`                                |
+| `network`              | string  | Yes      | CAIP-2 identifier (for example, `"xrpl:0"`)      |
+| `asset`                | string  | Yes      | `"XRP"` for native XRP, or currency code for IOU |
+| `payTo`                | string  | Yes      | XRPL classic address receiving the payment       |
+| `amount`               | string  | Yes      | XRP drops string or IOU issued-currency value    |
+| `maxTimeoutSeconds`    | integer | Yes      | Maximum validity window for payment attempt      |
+| `extra.invoiceId`      | string  | No       | Unique invoice identifier for binding            |
+| `extra.destinationTag` | integer | No       | DestinationTag for hosted accounts               |
+| `extra.issuer`         | string  | IOU only | Classic address of the IOU issuer                |
+
+`extra.destinationTag` applies to both native XRP and IOU payments. It is used when the receiver is a hosted account or otherwise requires a destination tag for attribution.
+
+No `extra.decimals` field is defined for XRPL exact payments. Implementations MUST NOT derive the signed transfer amount from server-provided decimal precision metadata.
+
+### Asset Field Values
+
+| Asset Type  | Format            | Example                                      |
+| ----------- | ----------------- | -------------------------------------------- |
+| Native XRP  | `"XRP"`           | `"XRP"`                                      |
+| 3-char IOU  | 3-character code  | `"USD"`                                      |
+| 160-bit IOU | 40 hex characters | `"524C555344000000000000000000000000000000"` |
+
+## Amount Formatting
+
+### XRP (Native)
+
+For native XRP, `PaymentRequirements.amount` is a string containing integer drops. One XRP equals 1,000,000 drops.
+
+| Human Amount | `amount` Value |
+| ------------ | -------------- |
+| 1 XRP        | `"1000000"`    |
+| 0.1 XRP      | `"100000"`     |
+| 0.000001 XRP | `"1"`          |
+
+### IOU (Issued Currency)
+
+For XRPL issued currencies, `PaymentRequirements.amount` is the exact XRPL issued-currency `value` string to be encoded in the destination amount object.
+
+XRPL issued currencies are identified by `(currency, issuer)` and the ledger `Payment` amount uses a decimal `value` string. XRPL does not define a universal token-decimals field for arbitrary issued currencies, so this scheme does not accept server-declared decimal precision.
+
+| Human Amount | `amount` Value | XRPL destination amount `value` |
+| ------------ | -------------- | ------------------------------- |
+| 10.50 USD    | `"10.5"`       | `"10.5"`                        |
+| 0.01 RLUSD   | `"0.01"`       | `"0.01"`                        |
+
+The facilitator MUST compare IOU amounts using exact decimal arithmetic suitable for XRPL issued-currency values, not binary floating point.
+
+## `PaymentPayload` for `exact`
+
+The `PAYMENT-SIGNATURE` header contains a base64-encoded `PaymentPayload`.
+
+### XRP Example
+
+```json
+{
+  "x402Version": 2,
+  "accepted": {
+    "scheme": "exact",
+    "network": "xrpl:0",
+    "asset": "XRP",
+    "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+    "amount": "1000000",
+    "maxTimeoutSeconds": 600,
+    "extra": {
+      "invoiceId": "INV-2025-001"
+    }
+  },
+  "payload": {
+    "signedTxBlob": "120000228000000024000000036840000000000000C732103AB40A0490F9B7ED8DF29D246BF2D6269820A0EE7742ACDD457BEA7C7D0931EDB74473045022100..."
+  }
+}
+```
+
+### IOU Example
+
+```json
+{
+  "x402Version": 2,
+  "accepted": {
+    "scheme": "exact",
+    "network": "xrpl:0",
+    "asset": "524C555344000000000000000000000000000000",
+    "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+    "amount": "10.5",
+    "maxTimeoutSeconds": 600,
+    "extra": {
+      "issuer": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q",
+      "invoiceId": "INV-2025-002"
+    }
+  },
+  "payload": {
+    "signedTxBlob": "120000228000000024000000036840000000000000C732103AB40A0490F9B7ED8DF29D246BF2D6269820A0EE7742ACDD457BEA7C7D0931EDB74473045022100..."
+  }
+}
+```
+
+### Payload Fields
+
+| Field          | Type   | Required | Description                              |
+| -------------- | ------ | -------- | ---------------------------------------- |
+| `signedTxBlob` | string | Yes      | Hex-encoded signed XRPL transaction blob |
+
+## Facilitator Verification Rules (MUST)
+
+A facilitator verifying an `exact`-scheme XRPL payment MUST enforce all of the following checks.
+
+### 1. Envelope Checks (x402 v2)
+
+The facilitator MUST reject if:
+
+- `paymentPayload.x402Version != 2`
+- `paymentPayload.accepted.scheme != "exact"`
+- `paymentPayload.accepted.network` is unsupported
+- `paymentPayload.accepted` does not match `paymentRequirements` on `scheme`, `network`, `asset`, `payTo`, `amount`, or `maxTimeoutSeconds`
+- Required `extra` keys are missing or mismatched:
+  - `issuer` for IOU payments
+  - `invoiceId` when invoice binding is required
+  - `destinationTag` when destination tag binding is required
+
+### 2. Transaction Decoding
+
+- Decode `signedTxBlob` (hex) into bytes.
+- Decode bytes using the XRPL binary codec to obtain `tx_json`.
+- If decoding fails, verification MUST fail.
+
+### 3. Transaction Type
+
+- `tx_json.TransactionType` MUST equal `"Payment"`.
+
+### 4. Destination Validation
+
+- `tx_json.Destination` MUST equal `paymentRequirements.payTo`.
+- If `paymentRequirements.extra.destinationTag` is present, `tx_json.DestinationTag` MUST be present and equal.
+
+### 5. Network Binding
+
+Let `networkId` be the integer parsed from `paymentRequirements.network` (for example, `"xrpl:1"` -> `1`).
+
+| Condition           | Requirement                                |
+| ------------------- | ------------------------------------------ |
+| `networkId <= 1024` | `tx_json.NetworkID` MUST be omitted        |
+| `networkId > 1024`  | `tx_json.NetworkID` MUST equal `networkId` |
+
+For XRPL mainnet, testnet, devnet, and other standard networks with `networkId <= 1024`, `NetworkID` is omitted by XRPL protocol rules. This omission is a transaction-format requirement, not a standalone cryptographic replay guarantee between standard XRPL networks.
+
+For `networkId <= 1024`, the facilitator MUST submit the transaction only to the XRPL network identified by `paymentRequirements.network`. For custom XRPL networks with `networkId > 1024`, the signed `NetworkID` field provides explicit network binding.
+
+### 6. Amount Validation
+
+XRPL API v2 uses `DeliverMax`; API v1 uses `Amount`. The facilitator MUST determine the destination amount field:
+
+- If `tx_json.DeliverMax` is present, use it.
+- Else use `tx_json.Amount`.
+- If neither is present, reject.
+- If both are present, reject.
+
+#### XRP Amount Rules
+
+If `paymentRequirements.asset == "XRP"`:
+
+- Destination amount field MUST be a string of digits representing drops.
+- `int(destinationAmount) == int(paymentRequirements.amount)`.
+- `tx_json.SendMax` MUST be omitted.
+- `tx_json.Paths` MUST be omitted.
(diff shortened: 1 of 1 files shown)


specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+
+### IOU (Issued Currency) Example
+
+```json
+{
+  "scheme": "exact",
+  "network": "xrpl:0",
+  "asset": "524C555344000000000000000000000000000000",
+  "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+  "amount": "10.50",
```
phdargen: should use atomic units

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+
+| Human Amount | `amount` Value |
+| ------------ | -------------- |
+| 1 XRP        | `"1000000"`    |
+| 0.1 XRP      | `"100000"`     |
+| 0.000001 XRP | `"1"`          |
+
+### IOU (Issued Currency)
+
+IOU amounts are specified as **decimal strings**:
```
phdargen: please use atomic units throughout

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+| `extra.destinationTag` | integer | No       | DestinationTag for hosted accounts            |
+| `extra.issuer`         | string  | IOU only | Classic address of the IOU issuer             |
+
+### Asset Field Values
+
+| Asset Type  | Format            | Example                                      |
+| ----------- | ----------------- | -------------------------------------------- |
+| Native XRP  | `"XRP"`           | `"XRP"`                                      |
+| 3-char IOU  | 3-character code  | `"USD"`                                      |
+| 160-bit IOU | 40 hex characters | `"524C555344000000000000000000000000000000"` |
```
phdargen, line 173: better to remove ambiguity
```suggestion
| 160-bit IOU | 40 hex characters | `"524C555344000000000000000000000000000000"` |
```

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+### 5. Network Binding
+
+Let `networkId` be the integer parsed from `paymentRequirements.network` (e.g., `"xrpl:1"` → `1`).
+
+| Condition           | Requirement                                |
+| ------------------- | ------------------------------------------ |
+| `networkId <= 1024` | `tx_json.NetworkID` MUST be **omitted**    |
+| `networkId > 1024`  | `tx_json.NetworkID` MUST equal `networkId` |
+
+> This prevents cross-network replay attacks and guaranteed on-chain failures.
```
phdargen: Per the spec above only network_id=0,1,2 are allowed. For all of those `tx_json.NetworkID` would be omitted, how exactly is this preventing cross-network replay? Naively this seems to do rather the opposite

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+### 9. Safety Checks (Recommended)
+
+The facilitator SHOULD reject transactions with:
+
+- Excessive `Fee` (define a max fee policy, e.g., 1 XRP)
+- Unsupported transaction features
+
+### 10. Signature Validation
+
+- `/verify` MAY validate the signature offline if the implementation supports XRPL signature verification
```
phdargen:
```suggestion
- `/verify` MUST validate the signature offline
```

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+}
+```
+
+### Payload Fields
+
+| Field          | Type   | Required | Description                              |
+| -------------- | ------ | -------- | ---------------------------------------- |
+| `signedTxBlob` | string | Yes      | Hex-encoded signed XRPL transaction blob |
+
+## Facilitator Verification Rules (MUST)
```
phdargen, line 263: Verification checks are so far only structural payload checks but should also check if the tx would actually succeed onchain, ideally with a full tx simulation. Or if thats not possible, at least some targeted checks like sufficient client token balance etc

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+  "scheme": "exact",
+  "network": "xrpl:0",
+  "asset": "524C555344000000000000000000000000000000",
+  "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+  "amount": "10.50",
+  "maxTimeoutSeconds": 600,
+  "extra": {
+    "issuer": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q",
+    "invoiceId": "INV-2025-002",
+    "destinationTag": 12345
```
phdargen: Whats the use case for `destinationTag`? And why is it only included for IOU but not native?

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+- `tx_json.LastLedgerSequence` MUST be present
+
+Recommended policy:
+
+- Convert `maxTimeoutSeconds` to ledgers: `maxLedgerDelta = ceil(maxTimeoutSeconds / 5) + 2`
+- Require: `LastLedgerSequence <= currentValidatedLedgerIndex + maxLedgerDelta`
+
+### 8. Invoice Binding (Required)
+
+The signed transaction MUST commit to the invoice via one of:
```
phdargen: I think this should be optional, similar to SVM memo: https://github.com/x402-foundation/x402/blob/17bd7e59f95b15f45f93f363bcd99c2b20f0c043/specs/schemes/exact/scheme_exact_svm.md?plain=1#L71

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+- Comparison: case-insensitive
+
+#### Option B: InvoiceID Field
+
+The transaction includes:
+
+- `InvoiceID = SHA-256(invoiceId)` as 32-byte hex (64 hex characters)
+- Comparison: case-insensitive
+
+The facilitator MUST reject if neither binding is present or if any present binding is mismatched.
```
phdargen: I'd suggest to define one canonical way, reject the other

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+
+### 9. Safety Checks (Recommended)
+
+The facilitator SHOULD reject transactions with:
+
+- Excessive `Fee` (define a max fee policy, e.g., 1 XRP)
+- Unsupported transaction features
```
phdargen: what are those unsupported transaction features and how does the facilitator checks this?

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+
+`exact`
+
+## Payment Model
+
+| Aspect                    | Description                                                                  |
+| ------------------------- | ---------------------------------------------------------------------------- |
+| **Payment authorization** | The payer signs a standard XRPL `Payment` transaction                        |
+| **Settlement**            | The facilitator submits the signed transaction to XRPL                       |
+| **Fee payer**             | The payer pays the XRPL transaction fee (embedded in the signed transaction) |
```
phdargen: This would be the first network scheme that doesnt support fee sponsorship, what we consider very important for user experience. Any options to add this (at least for IOU) or is this a fundamental network level limitation?

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,398 @@
...
+This scheme is designed with trust minimization:
+
+- The facilitator **cannot redirect funds** because any mutation of the signed transaction invalidates the payer's signature
+- The resource server **cannot collect more** than the amount the payer signed for
+- Invoice binding ensures the payer's intent is cryptographically committed
+
+### Replay Protection
+
+- `LastLedgerSequence` ensures transactions expire
+- `Sequence` number prevents double-submission by the same account
```
phdargen: I take it `Sequence` is a monotonically increasing nonce. This would be problematic. Client account would be effectively locked until server settles (could be many mins, in principle unbounded) as the flow is verify -> server route handler runs -> settle (if api success). And if client does another tx in-between verify and settle, settle fails and the server wasted work

phdargen:
Thanks @aristotle-satoshi, lets pick this up again! 

Made a first pass through the spec only. Once we are aligned there, will check the implementation. Generally preferred to separate this in 2 PRs, see https://github.com/x402-foundation/x402/blob/main/CONTRIBUTING.md#new-chains

Please also run the implementation through our new skill here: https://github.com/x402-foundation/x402/blob/main/.agents/skills/contributing/references/new-mechanism.md

phdargen:
Please also verify your commits

aristotle-satoshi:
Updated the PR description to clarify the spec source and review flow.

For this work, #984 is the canonical XRPL exact spec input/source. It was closed without being merged, so I am not assuming XRPL is already part of `main`; this PR carries that spec work forward and pairs it with the TypeScript implementation.

On the new-chain guidance: I can split this into a spec-only PR first and a separate TypeScript implementation PR after the spec is aligned if that is the preferred review path. I will also run the implementation against `.agents/skills/contributing/references/new-mechanism.md` before requesting implementation review. The commit verification check is now passing.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Applied this suggestion: the 160-bit IOU row now uses the 40-character hex example to remove the ambiguity.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Clarified this in the payment model. XRPL charges the transaction fee to the transaction `Account`, so in this payer-signed `Payment` model the payer pays the network fee embedded in the signed transaction. That applies to both XRP and IOU payments; facilitator-sponsored fees for the same signed transaction are a network/model limitation unless the payment model changes.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Added this to the spec. `/verify` now MUST check that the signed transaction would currently succeed on XRPL, SHOULD use XRPL transaction simulation, and MUST fall back to targeted checks covering account existence, sequence or ticket availability, XRP fee balance, and IOU balance or trust line sufficiency if simulation is unavailable.

phdargen:
> Updated the PR description to clarify the spec source and review flow.
> 
> For this work, #984 is the canonical XRPL exact spec input/source. It was closed without being merged, so I am not assuming XRPL is already part of `main`; this PR carries that spec work forward and pairs it with the TypeScript implementation.
> 
> On the new-chain guidance: I can split this into a spec-only PR first and a separate TypeScript implementation PR after the spec is aligned if that is the preferred review path. I will also run the implementation against `.agents/skills/contributing/references/new-mechanism.md` before requesting implementation review. The commit verification check is now passing.

Yes please make this PR spec only and follow up with the implementation in a separate PR

BTW, there is no need to edit docs/ as this is automatically updated by mintlify  (but you may, if you like)

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,407 @@
...
+| `scheme`               | string  | Yes      | Must be `"exact"`                             |
+| `network`              | string  | Yes      | CAIP-2 identifier (e.g., `"xrpl:0"`)          |
+| `asset`                | string  | Yes      | `"XRP"` for native, or currency code for IOUs |
+| `payTo`                | string  | Yes      | XRPL classic address (destination)            |
+| `amount`               | string  | Yes      | Atomic-unit payment amount                    |
+| `maxTimeoutSeconds`    | integer | Yes      | Maximum validity window for payment attempt   |
+| `extra.invoiceId`      | string  | No       | Unique invoice identifier for binding         |
+| `extra.destinationTag` | integer | No       | DestinationTag for hosted accounts            |
+| `extra.issuer`         | string  | IOU only | Classic address of the IOU issuer             |
+| `extra.decimals`       | integer | IOU only | Decimal precision for IOU atomic units        |
```
phdargen: there should be no server provided decimals in `extra`

If client trusts a server-declared `extra.decimals`, a malicious/misconfigured server can understate decimals so the same small-looking atomic amount signs a far larger numeric transfer

Atomic units should be used throughout. Server declares price in atomic unit, client creates payload in atomic units. Facilitator verifies amount in atomic units

If for some reason decimals is needed somewhere in the flow (this has to be well justified), then this should be retrieved from onchain state via RPC

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,407 @@
...
+
+If `paymentRequirements.asset != "XRP"`:
+
+- Destination amount field MUST be an issued-currency object:
+  ```json
+  { "currency": "...", "issuer": "...", "value": "..." }
+  ```
+- `currency` MUST match `paymentRequirements.asset` (3-char or 160-bit hex)
+- `issuer` MUST match `paymentRequirements.extra.issuer`
+- `Decimal(value) == Decimal(paymentRequirements.amount / 10^paymentRequirements.extra.decimals)`
```
phdargen: should compare atomic units

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,407 @@
...
+### 5. Network Binding
+
+Let `networkId` be the integer parsed from `paymentRequirements.network` (e.g., `"xrpl:1"` → `1`).
+
+| Condition           | Requirement                                |
+| ------------------- | ------------------------------------------ |
+| `networkId <= 1024` | `tx_json.NetworkID` MUST be **omitted**    |
+| `networkId > 1024`  | `tx_json.NetworkID` MUST equal `networkId` |
+
+For XRPL mainnet, testnet, and devnet, `NetworkID` is omitted by XRPL convention. For those networks this check prevents invalid `NetworkID` fields but does not by itself cryptographically prevent replay across standard XRPL networks. Facilitators MUST submit only to `paymentRequirements.network`; custom XRPL networks with `networkId > 1024` are bound by the signed `NetworkID` field.
```
phdargen: Still don't see the benefit/reason of omitting `networkId`. Does it mean a malicious/misconfigured facilitator could submit a tx on mainnet that is intended to be on testnet? We should avoid any trust assumptions

specs/schemes/exact/scheme_exact_xrpl.md
phdargen: Could you please comment on this issue @aristotle-satoshi ?

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Updated the spec to remove server-provided `extra.decimals`. XRPL IOU payments now use the ledger-native issued-currency `value` string, and the spec says implementations MUST NOT derive transfer amounts from server-declared decimal precision.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Adjusted the amount validation section. XRP compares drops, while IOU payments compare the XRPL issued-currency `value` directly against `paymentRequirements.amount` using exact decimal arithmetic; no `amount / 10^extra.decimals` conversion remains.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Clarified the NetworkID limitation. For `networkId <= 1024`, omission is an XRPL transaction-format rule, not a standalone cryptographic replay guarantee. The spec now states that facilitators MUST submit only to `paymentRequirements.network`; explicit signed NetworkID binding is only available for custom networks with `networkId > 1024`.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Added an Expiry and Account Sequencing section. `TicketSequence` is now SHOULD for flows where `/verify` and `/settle` are separated by resource-handler execution. If normal `Sequence` is used, `/verify` must check current sequence availability and resource servers should settle promptly to minimize sequence-race risk.

aristotle-satoshi:
Done. This PR is now spec-only: the current diff only adds `specs/schemes/exact/scheme_exact_xrpl.md`; docs, SDK implementation, package, and lockfile changes have been removed. Commit verification and GitHub Actions are passing; Vercel still requires maintainer authorization. I will follow up with the implementation in a separate PR after the spec is aligned.

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,437 @@
...
+
+## Payment Model
+
+| Aspect                    | Description                                                                  |
+| ------------------------- | ---------------------------------------------------------------------------- |
+| **Payment authorization** | The payer signs a standard XRPL `Payment` transaction                        |
+| **Settlement**            | The facilitator submits the signed transaction to XRPL                       |
+| **Fee payer**             | The payer pays the XRPL transaction fee embedded in the signed transaction   |
+
+XRPL charges the transaction fee to the transaction `Account`. This exact scheme therefore does not support facilitator-sponsored network fees for the signed `Payment` transaction. Supporting fee sponsorship would require a different payment model, not only a facilitator implementation change.
```
phdargen, line 19: OK lets add a `extra.areFeesSponsored=false` then, aligning with Stellar:
https://github.com/x402-foundation/x402/blob/b1c4734bc2bdfb8719770c7939fa1a69dcbee97e/specs/schemes/exact/scheme_exact_stellar.md?plain=1#L54-L56

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,437 @@
...
+Let `networkId` be the integer parsed from `paymentRequirements.network` (for example, `"xrpl:1"` -> `1`).
+
+| Condition           | Requirement                                |
+| ------------------- | ------------------------------------------ |
+| `networkId <= 1024` | `tx_json.NetworkID` MUST be omitted        |
+| `networkId > 1024`  | `tx_json.NetworkID` MUST equal `networkId` |
+
+For XRPL mainnet, testnet, devnet, and other standard networks with `networkId <= 1024`, `NetworkID` is omitted by XRPL protocol rules. This omission is a transaction-format requirement, not a standalone cryptographic replay guarantee between standard XRPL networks.
+
+For `networkId <= 1024`, the facilitator MUST submit the transaction only to the XRPL network identified by `paymentRequirements.network`. For custom XRPL networks with `networkId > 1024`, the signed `NetworkID` field provides explicit network binding.
```
phdargen, line 308: OK lets add a warning then that one should use different testnet and mainnet wallets. If one uses an account on testnet that also holds mainnet funds, a malicious/misconfigured facilitator could replay a testnet tx on mainnet

specs/schemes/exact/scheme_exact_xrpl.md
```diff
@@ -0,0 +1,437 @@
...
+If `TicketSequence` is present:
+
+- `tx_json.Sequence` MUST be `0`.
+- `tx_json.TicketSequence` MUST refer to an available ticket for `tx_json.Account`.
+
+If `TicketSequence` is absent:
+
+- `tx_json.Sequence` MUST be present.
+- `/verify` MUST check that `tx_json.Sequence` is currently available for `tx_json.Account`.
+- Resource servers SHOULD settle promptly after successful verification to minimize sequence-race risk.
```
phdargen: The only `exact` spec compliant flow is where `/verify` and `/settle` are separated by resource-handler execution. Servers do not have a choice when to settle, they do it after the resource handler finished successfully. Thus they can't "settle promptly after successful verification". The decision when to settle can also not depend on what kind of transaction the client signed as the server never inspects it, the facilitator does. 

It seems the `TicketSequence` is solving the locked nonce issue, so why not remove this ambiguity and define that clients MUST use `TicketSequence`? Any disadvantages wrt normal `Sequence`?

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Added `PaymentRequirements.extra.areFeesSponsored=false` as a required field, included it in the XRP/IOU examples and payload examples, and kept XRPL exact explicitly payer-fee-only. I also reject `Delegate` so the fee model stays unambiguous.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Updated this to remove the normal `Sequence` fallback. XRPL x402 exact payments now MUST use `TicketSequence`; `Sequence` must be `0`, the ticket must be available, and clients must create a ticket before signing if none is available. This avoids relying on resource servers to settle promptly, which is not part of the exact flow.

specs/schemes/exact/scheme_exact_xrpl.md
aristotle-satoshi: Added the warning. The spec now says standard XRPL networks omit signed `NetworkID`, so clients/wallets SHOULD use different XRPL accounts for mainnet, testnet, devnet, and other standard networks to reduce replay risk when the same account has funds and compatible ticket state on multiple networks.

specs/schemes/exact/scheme_exact_xrpl.md
phdargen: Thanks for the update! 

The `TicketSequence` lifecycle and implications still feel underspecified

After doing some digging, it does seem there are some disadvantages wrt normal Sequence as it requires a second tx in the flow for creating the ticket before even doing the server request increasing per-request latency and incurring an additional tx fee plus a small XRP deposit that gets locked. Could you please clarify if my understanding is accurate?

aristotle-satoshi: Your understanding is accurate. `TicketSequence` avoids the normal account-sequence race, but it has a real cost profile that is especially awkward for micropayments.

If the payer has no available ticket, they need a `TicketCreate` transaction first. That adds latency and fee, and each unused ticket consumes owner reserve until it is used or removed. For low-value payments, the reserve requirement can be much larger than the payment amount itself. For example, a payment may be only a few drops or `0.01 XRP`, while maintaining an unused ticket currently requires materially more XRP locked as reserve. This also makes low-balance wallets harder to support.

Because of that, I think normal `Sequence` may be worth considering for this spec, with the tradeoff clearly documented.

Normal `Sequence` would avoid the ticket reserve and TicketCreate preflight entirely, which is much better for low-balance micropayment wallets. The downside is that it cannot reserve the payer account sequence at the XRPL protocol level across the x402 `/verify -> resource handler -> /settle` gap. A facilitator can mitigate this by checking account sequence at `/verify`, enforcing `LastLedgerSequence`, re-checking immediately before `/settle`, and clients/wallets can locally lock the next sequence while a payment is pending. But this remains a cooperative-client mitigation rather than a protocol-level guarantee: the payer could still consume the same sequence elsewhere before settlement.

So I see the tradeoff as:

- `TicketSequence`: stronger delayed-settlement safety, but requires ticket inventory, reserve, and possible TicketCreate latency.
- normal `Sequence`: better UX and lower balance requirements for micropayments, but leaves a settlement invalidation race that facilitators/resource servers must explicitly accept and mitigate.

Would you prefer the XRPL exact spec to use normal `Sequence` with explicit facilitator/client mitigation requirements, or keep `TicketSequence` as the stricter path and document the reserve limitation?

aristotle-satoshi: To make the cost tradeoff more concrete:

- For a `0.01 XRP` payment, one unused ticket currently locks `0.2 XRP` of owner reserve while it is outstanding. That is about **20x the payment amount** in temporary locked liquidity.
- For `10` concurrent pending payments, a wallet would need `10` available tickets, temporarily locking about `2 XRP` in reserve.
- For `50` concurrent pending payments, that becomes about `10 XRP` locked.
- At the XRPL ticket limit of `250` outstanding tickets, the reserve footprint is about `50 XRP`.

This reserve is released when tickets are used or removed, so it is not a permanent per-payment fee. But for low-balance wallets or high-concurrency micropayment agents, it can still be a meaningful bottleneck: the wallet may have enough XRP to pay the merchant and the network fee, but not enough spendable balance above reserve to maintain the ticket inventory.

phdargen: Thanks for the additional details @aristotle-satoshi, there are certainly pros/cons to either approach and doesnt seem one would be strictly better than the other in every scenario. 

So I suggest we give servers and clients the choice by having 
`assetTransferMethod="sequence" `or `assetTransferMethod="ticketSequence"`, similar to EVM:
https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md

For example, it should be fine to use `Sequence` if the server just does a quick database lookup, so settles ~immediately. They may also offer `TicketSequence` in that case for clients that want to do concurrent requests. Servers doing longer running tasks, might instead restrict to `TicketSequence` only or opt-in to accept the increased risk for `Sequence`

phdargen: If that sounds good to you, please add `extra.assetTransferMethod` to the spec and document the tradeoffs similar to what we discussed above

aristotle-satoshi: Sounds good. Added `extra.assetTransferMethod` with values `"sequence"` and `"ticketSequence"`, mirroring the EVM spec's negotiation pattern: if the server specifies a method the client MUST use it; if omitted, clients default to `"sequence"` and MUST echo a non-default selection in `accepted.extra`; servers can offer both via multiple `accepts` entries. The new Asset Transfer Methods section documents the tradeoffs we discussed (ticket reserve of 0.2 XRP per outstanding ticket, TicketCreate preflight latency/fee, 250-ticket cap vs. the sequence-race risk for delayed settlement) plus guidance on choosing per settle latency. Verification rules are now split per method: the `sequence` path requires the tx sequence to be current at `/verify` and re-checked at `/settle`, and the `ticketSequence` path keeps the `Sequence=0` + available-ticket checks.

phdargen (approved):
Thanks @aristotle-satoshi, LGTM 🚀

aristotle-satoshi:
Thanks @phdargen for the review and for merging the spec!

As discussed, the follow-up TypeScript implementation is now ready for review in #2801. It implements the merged spec (including `assetTransferMethod`) following `.agents/skills/contributing/references/new-mechanism.md`, with unit/integration tests and the full 10-scenario e2e matrix verified on XRPL Testnet.

Issue #2551, opened 2026-06-04
Proposal: exact scheme on Canton Network


Denend:
## Why Canton

- Privacy-preserving institutional settlement network using the Global Synchronizer
- Daml contracts as settlement primitives, not EVM/SVM transaction semantics
- Stakeholder-scoped contract visibility rather than globally visible account/contract state
- Participant-mediated submission, not EVM-style gas

## Proposed approach

- Client exercises CIP-56 `TransferFactory_Transfer` on its own participant → receives `updateId`
- Facilitator reads the completed update through SV Scan, validates `sender`, `receiver`, `amount`, `instrumentId` from the exercised choice argument directly
- Facilitator does not submit on-ledger settlement transactions

## Key design decisions

1. Agent pays directly — `TransferFactory_Transfer actAs: [agentParty]`
2. Facilitator verification is read-only via SV Scan; on RBAC-gated
   deployments (MainNet) facilitator authenticates to Scan
3. Atomic settlement via merchant `TransferPreapproval`; pending transfers
   are rejected, not queued
4. HTTP replay protection via consumed-payment store; ledger double-spend
   impossible — completed transfer is an immutable on-ledger record
5. CIP-56-only path covers Canton Coin (Amulet) and any CIP-56 instrument

## v2 alignment

- `PAYMENT-REQUIRED` / `PAYMENT-SIGNATURE` / `PAYMENT-RESPONSE`
- `x402Version: 2`, amount as atomic string
- `/settle` re-verifies independently

## Implementation status

Reference implementation is in progress and the CIP-56 flow has been validated
against Canton environments. Spec-only PR follows if the direction is accepted

PR #2552, opened 2026-06-04
Specs(exact): propose Canton exact scheme for x402 (spec-only)

Denend:
## Description

Adds an `exact` scheme implementation spec for Canton Network under
`specs/schemes/exact/`.

The spec defines Canton payment requirements, payment payload shape,
read-only SV Scan verification, settlement semantics, replay protection, and
Canton-specific error reasons for the CIP-56 `TransferFactory_Transfer` path.

Related discussion: #2551

## Tests

Docs/spec-only change. No SDK code changed.

## Checklist

Docs-only/spec-only change; changelog fragment skipped per template guidance.

### DIFF
--- a/specs/schemes/exact/scheme_exact_canton.md
+++ b/specs/schemes/exact/scheme_exact_canton.md
@@ -1,68 +1,56 @@
 # Exact Payment Scheme for Canton Network (`exact`)
 
 This document specifies the `exact` payment scheme for the x402 protocol on
-Canton Network. Payments use the CIP-56 `TransferFactory_Transfer` primitive;
-the facilitator verifies read-only via SV Scan and does not submit on-ledger
-settlement transactions.
+Canton Network. The client creates a `Splice.ExternalPartyAmuletRules.TransferCommand`
+via Ed25519 external-party signing; the facilitator, acting as `delegate`,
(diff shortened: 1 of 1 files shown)


phdargen:
Thanks a lot for the proposal @Denend!

Client prepayments are a significantly different payment flow with a different security/trust model than existing `exact` schemes.

We are currently discussing if and how we want to support such a flow in the x402 protocol, possibly through a new distinct scheme, see https://github.com/x402-foundation/x402/pull/2520

In any case, it would be strongly preferred if your proposal could be reframed into a true 'exact' flow, where the client only signs an authorization and then the facilitator submits the tx and sponsors gas. If Canton supports such a flow, thats the best path forward

Denend:
@phdargen Thanks for the review!

We actually had an initial v1 implementation that followed the "exact" flow, so rolling back to that approach is not a problem for us

We've updated the scheme according to your recommendations

Denend:
Hey @erikreppel-cb @CarsonRoscoe @ryanRfox , could you please check out this PR when you have time? Thanks!

Denend:
new updated PR is https://github.com/x402-foundation/x402/pull/2634

PR #2554, opened 2026-06-04
feat: reject eoas as assets

CarsonRoscoe:
## Description

Fixes a silent payment bypass: the EVM facilitator's verify() accepted payments where the asset field pointed to a wallet address (EOA) rather than a token contract. The root cause is an EVM property — calling any function on an EOA via eth_call returns empty data without reverting, so the on-chain simulation (transferWithAuthorization or Permit2 settle) incorrectly passes. The subsequent settle transaction lands on-chain and succeeds, but no Transfer event is ever  emitted, producing a silent no-op.

**Fix**: Add a eth_getCode pre-flight check early in each EVM verify path. If the asset address has no bytecode it is rejected with asset_not_deployed_contract before signature verification or simulation are attempted. The check applies to all three EVM payment types — EIP-3009, Permit2 exact, and Permit2 upto — since no valid payment scheme targets an EOA as its token contract. SVM is unaffected.

_Why eth_getCode and not supportsInterface?_ Calling supportsInterface on an EOA returns empty bytes (not a revert), which  would be treated as "uncertain, allow through" — the same failure mode as the bug itself. eth_getCode is the definitive check: an EOA has no bytecode by definition. For contracts that have bytecode but don't implement the expected transfer interface, the EVM does revert on the function call (unlike EOAs), so simulation correctly catches those cases.

This is implemented at the SDK level (not via facilitator hooks), which allows the check to be placed directly in the verify logic, before signature verification and simulation.

### Changes

Go (go/mechanisms/evm/)
- permit2_errors.go — new shared constant ErrAssetNotDeployedContract = "asset_not_deployed_contract"
- utils.go — new helper ValidateAssetIsContract(ctx, signer, asset) that calls signer.GetCode and returns the error reason or propagates RPC failures
- exact/facilitator/eip3009.go — call ValidateAssetIsContract early in verifyEIP3009
- exact/facilitator/permit2.go — call ValidateAssetIsContract early in VerifyPermit2
- upto/facilitator/permit2.go — call ValidateAssetIsContract early in VerifyUptoPermit2

TypeScript (typescript/packages/mechanisms/evm/src/)
- exact/facilitator/errors.ts — new export ErrAssetNotDeployedContract
- upto/facilitator/errors.ts — re-exports ErrAssetNotDeployedContract from exact errors
- exact/facilitator/eip3009.ts, exact/facilitator/permit2.ts, upto/facilitator/permit2.ts — await signer.getCode({ address: tokenAddress }) check added early in each verify function
  
Python (python/x402/mechanisms/evm/)
- constants.py — new constant ERR_ASSET_NOT_DEPLOYED_CONTRACT
- exact/facilitator.py, exact/permit2_utils.py, upto/permit2_utils.py — signer.get_code(token_address) check added early in each verify path

## Tests

New unit tests added across all three SDKs, each covering:
- EOA asset rejected — GetCode/get_code/getCode returns empty → verify returns asset_not_deployed_contract before any signature check fires
- Deployed contract passes — non-empty bytecode → verify proceeds normally
- RPC error propagates — GetCode throwing propagates as an internal error (500), not a 400, consistent with how other RPC failures (KYT, balance checks) are handled

- Manual tested Go to confirm the report and confirm the fix

## Checklist


typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts
```diff
@@ -97,6 +97,14 @@ export async function verifyEIP3009(
     };
   }
 
+  // Reject payments whose asset is an EOA — eth_call on an EOA silently returns
+  // empty data without reverting, so simulation would pass but no Transfer event
+  // would be emitted, producing a silent no-op settlement.
+  const assetBytecode = await signer.getCode({ address: erc20Address });
+  if (!assetBytecode || assetBytecode === "0x") {
+    return { isValid: false, invalidReason: Errors.ErrAssetNotDeployedContract, payer };
+  }
```
phdargen: Thats a good fix, but would consider moving after the cheap payload checks just before simulation as it requires a RPC call

phdargen:
Possible optimization to avoid an additional RPC call on the happy path: change the transfer simulation from a direct readContract call to a multicall that batches balanceOf() with the raw transferWithAuthorization simulation. Then use the decoded balanceOf result to prove the asset behaves like an ERC-20 
```
const results = await multicall(signer.readContract.bind(signer), [
  {
    address: erc20Address,
    abi: eip3009ABI,
    functionName: "balanceOf",
    args: [auth.from],
  },
  {
    address: erc20Address,
    callData: transferCalldata,
  },
]);

if (results[0]?.status !== "success") {
  return false;
}
```
and do the getCode check only in `diagnoseEip3009SimulationFailure` to return specific  `asset_not_deployed_contract` error code


Might be more tricky for permit2 flows though

phdargen (approved):

PR #2556, opened 2026-06-05
Fix settlementOverrides for next

phdargen:
## Description

- Fixes a bug in @x402/next's `handleSettlement` which never forwarded response headers to processSettlement, so setSettlementOverrides was ignored breaking dynamic pricing for `upto` and `batch-settlement`
- strips internal settlement-overrides header in all TS middleware packages after settlement reads it, so its not exposed to the client. Go/py middlewares already do this

## Tests

Added dynamic pricing to `next-batch-settlement-redis` and tested manually 

## Checklist

### DIFF
--- a/examples/typescript/fullstack/next-batch-settlement-redis/next-env.d.ts
+++ b/examples/typescript/fullstack/next-batch-settlement-redis/next-env.d.ts
@@ -1,6 +1,0 @@
-/// <reference types="next" />
-/// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
-
-// NOTE: This file should not be edited
-// see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/examples/typescript/fullstack/next-batch-settlement-redis/next-env.d.ts
+++ b/examples/typescript/fullstack/next-batch-settlement-redis/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
+import "./.next/dev/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/typescript/packages/http/express/src/index.ts
+++ b/typescript/packages/http/express/src/index.ts
@@ -288,6 +288,7 @@ export function paymentMiddlewareFromHTTPServer(
             reason: "handler_failed",
             responseStatus: res.statusCode,
           });
+          res.removeHeader(SETTLEMENT_OVERRIDES_HEADER);
           restoreResponseMethods();
           // Replay all buffered calls in order
           for (const [method, args] of bufferedCalls) {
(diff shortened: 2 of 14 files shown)


examples/typescript/fullstack/next-batch-settlement-redis/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
+import "./.next/dev/types/routes.d.ts";
```
CarsonRoscoe: I think this file should be removed and gitignored

typescript/packages/http/hono/src/index.ts
```diff
@@ -221,6 +221,10 @@ export function paymentMiddlewareFromHTTPServer(
             reason: "handler_failed",
             responseStatus: res.status,
           });
+          const headers = new Headers(res.headers);
+          headers.delete(SETTLEMENT_OVERRIDES_HEADER);
+          const body = await res.clone().arrayBuffer();
```
CarsonRoscoe: The success path just calls `res.headers.delete(SETTLEMENT_OVERRIDES_HEADER)` directly (a few lines below in this same file). The same works here, Response.headers is mutable. The body clone to construct a whole new Response is unnecessary and inconsistent. Suggestion:
```
res.headers.delete(SETTLEMENT_OVERRIDES_HEADER); // c.res already points to res — no reassignment needed
```

typescript/packages/http/express/src/index.ts
```diff
@@ -346,17 +348,21 @@ export function paymentMiddlewareFromHTTPServer(
         } catch (error) {
           if (error instanceof FacilitatorResponseError) {
             bufferedCalls = [];
+            res.removeHeader(SETTLEMENT_OVERRIDES_HEADER);
```
CarsonRoscoe: extraneous as `res.removeHeader(SETTLEMENT_OVERRIDES_HEADER);` is called in finally

typescript/packages/http/express/src/index.ts
```diff
@@ -346,17 +348,21 @@ export function paymentMiddlewareFromHTTPServer(
...
             sendFacilitatorError(res, error);
             return;
           }
           console.error(error);
           // If settlement fails, don't send the buffered response
           bufferedCalls = [];
+          res.removeHeader(SETTLEMENT_OVERRIDES_HEADER);
```
CarsonRoscoe: extraneous as `res.removeHeader(SETTLEMENT_OVERRIDES_HEADER);` is called in finally

typescript/packages/http/next/src/utils.ts
```diff
@@ -172,18 +173,24 @@ export async function handleSettlement(
...
+    response.headers.forEach((value, key) => {
+      responseHeaders[key] = value;
+    });
+
     const result = await httpServer.processSettlement(
       paymentPayload,
       paymentRequirements,
       declaredExtensions,
-      httpContext ? { request: httpContext, responseBody } : undefined,
+      httpContext ? { request: httpContext, responseBody, responseHeaders } : undefined,
```
CarsonRoscoe: When `httpContext` is undefined, `responseHeaders` is dropped entirely from the transport context, so a `setSettlementOverrides` call on the response would be silently ignored. This matches the pre-existing behavior for the `undefined` case, so it's not a regression. But it's worth noting that the `?` guard means overrides are only respected when the request context is present

phdargen: good point, think there was no good reason to leave httpContext optional here

CarsonRoscoe (approved):

PR #2558, opened 2026-06-05
improve client error surfacing

phdargen:
## Description

Adds transport-agnostic `parsePaymentResult` and simplify the parsed result to `HTTPResourceResponse` (`{ status, body, header }`), where `header` is the decoded `SettleResponse` (from `PAYMENT-RESPONSE`) or `PaymentRequired` (from `PAYMENT-REQUIRED`, whose `error` carries the server's failure reason). This lets clients surface server-delivered payment errors without branching.

For failed payments, client examples currently just show a generic error message:
```
Response body: {}
Error: Payment response header not found
```

After fix:
```
{
  status: 402,
  body: {},
  header: {
    x402Version: 2,
    error: '....',
    resource: {
      url: 'http://localhost:4021/api/generate',
      description: 'AI text generation — billed by token usage',
      mimeType: 'application/json'
    },
    accepts:  ...,
    extensions: ...,
  }
}
```

## Checklist

### DIFF
--- a/examples/typescript/clients/batch-settlement/index.ts
+++ b/examples/typescript/clients/batch-settlement/index.ts
@@ -70,7 +70,7 @@ async function main(): Promise<void> {
     const response = await fetchWithPayment(url, { method: "GET" });
     const result = await httpClient.processResponse(response);
 
-    if (result.status === 200) {
+    if (result.paymentStatus === "settled") {
       console.log(`Request ${i + 1} — RESPONSE`);
       console.log(result.body);
       console.log(JSON.stringify(result.header, null, 2));
--- a/examples/typescript/clients/batch-settlement/index.ts
+++ b/examples/typescript/clients/batch-settlement/index.ts
@@ -70,12 +70,12 @@ async function main(): Promise<void> {
     const response = await fetchWithPayment(url, { method: "GET" });
     const result = await httpClient.processResponse(response);
 
-    if (result.kind === "success") {
+    if (result.status === 200) {
       console.log(`Request ${i + 1} — RESPONSE`);
       console.log(result.body);
-      console.log(JSON.stringify(result.settleResponse, null, 2));
+      console.log(JSON.stringify(result.header, null, 2));
     } else {
-      console.log(`Request ${i + 1} — ${result.kind}`);
+      console.log(`Request ${i + 1} — no settlement`);
       console.log(JSON.stringify(result, null, 2));
     }
     console.log(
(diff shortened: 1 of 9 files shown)


typescript/packages/core/src/http/x402HTTPClient.ts
```diff
@@ -220,48 +220,53 @@ export class x402HTTPClient {
...
+    const { status, getHeader, body } = args;
+
+    let header: SettleResponse | PaymentRequired | undefined;
     try {
-      settleResponse = this.getPaymentSettleResponse(getHeader);
+      header = this.getPaymentSettleResponse(getHeader);
     } catch {
-      /* no header */
+      try {
+        header = this.getPaymentRequiredResponse(getHeader, body);
```
CarsonRoscoe: The original code guarded this behind `response.status === 402`. We now run this regardless of status, so the code runs against any HTTP status

examples/typescript/clients/batch-settlement/index.ts
```diff
@@ -70,12 +70,12 @@ async function main(): Promise<void> {
     const response = await fetchWithPayment(url, { method: "GET" });
     const result = await httpClient.processResponse(response);
 
-    if (result.kind === "success") {
+    if (result.status === 200) {
```
CarsonRoscoe: nit: If we can still check for explicit success here, we should,

Success = Payment was successful
200 = Call was successful

Distinction is 200 will be returned for passthroughs that don't gate 402's, while success ensures 402 was attempted and paid

typescript/packages/core/src/http/x402HTTPClient.ts
```diff
@@ -220,48 +220,53 @@ export class x402HTTPClient {
...
+   * @param args.status - HTTP response status code
+   * @param args.getHeader - Callback to read response headers by name
+   * @param args.body - Response body payload
+   * @returns The parsed status, body, and decoded payment header
    */
-  async processResponse(response: Response): Promise<x402PaymentResult> {
-    const getHeader = (name: string) => response.headers.get(name);
-
-    let settleResponse: SettleResponse | undefined;
+  parsePaymentResult(args: {
```
CarsonRoscoe, line 235: Please add a unit test for parsePaymentResult

examples/typescript/clients/batch-settlement/index.ts
phdargen: good point, added new `paymentStatus` field

CarsonRoscoe (approved):

PR #2575, opened 2026-06-08
Builder-code extension for go SDK

phdargen:
## Description

Implements the builder-code extension for the go SDK following the TS reference https://github.com/x402-foundation/x402/pull/2329 and follow-up https://github.com/x402-foundation/x402/pull/2561

## Tests

Cross sdk tests with TS builder-code examples

## Checklist


go/mechanisms/evm/exact/v1/facilitator/scheme.go
```diff
@@ -305,7 +305,12 @@ func (f *ExactEvmSchemeV1) Settle(
 		return nil, x402.NewSettleError(ErrInvalidPayload, verifyResp.Payer, network, "", err.Error())
 	}
 
-	txHash, err := exactfacilitator.ExecuteTransferWithAuthorization(ctx, f.signer, tokenAddress, parsedAuthorization, sigData)
+	dataSuffix, err := evm.ResolveDataSuffix(fctx, evm.DataSuffixContext{})
```
CarsonRoscoe, line 311: Is there a reason in v1 we're passing n the default DataSuffixContext here rather than passing in the derived one we do in the other schemes?

I imagine this should be `evm.ResolveDataSuffix(fctx, evm.DataSuffixContext{Payload: payload, Requirements: requirements})`, as with a zero-value context, `extractClientExtension` will read from an empty `Extensions` map, so a and s are always empty. Only w (if this facilitator has one) would ever be written to the CBOR suffix for v1 payments

I understand v1 uses types.PaymentPayloadV1, which can't be assigned to the v2 types.PaymentPayload field on DataSuffixContext, so this is an architectural constraint. But it needs to be explicit, not silent. 

Imo, it's "fine" since we aren't supposed to be focusing our efforts too hard on building up v1. In fact, it looks like Typescript does nothing in the v1 case. So perhaps just document this distinction in a docstring or comments and leave it be?

CarsonRoscoe (dismissed):

go/mechanisms/evm/exact/v1/facilitator/scheme.go
phdargen: This is just an artifact of Go's stricter typing, the behavior of TS is aligned and appends the facilitator code for v1

Added some comments

CarsonRoscoe (approved):

PR #2579, opened 2026-06-09
feat(keeta): add @x402/keeta TypeScript package

sc4l3r:
## Description

This PR adds the initial TypeScript implementation for Keeta following the specification merged in https://github.com/x402-foundation/x402/pull/1575 . It's been reviewed by @ezraripps in my fork https://github.com/sc4l3r/coinbase-x402/pull/2 .

Commits can be squashed before merge if desired.

## Tests

Unit and integration tests have been added in the package's `test` directory with a description on how to run them in the README. Additionally, this implementation is used successfully in this example https://github.com/sc4l3r/keeta-x402 .

## Checklist


.github/workflows/publish_npm_scoped_x402_keeta.yml
```diff
@@ -0,0 +1,56 @@
+name: Publish @x402/keeta package to NPM
```
phdargen, line 1: Please align with latest version of publishing actions, see eg https://github.com/x402-foundation/x402/blob/main/.github/workflows/publish_npm_scoped_x402_evm.yml

and also add to 
https://github.com/x402-foundation/x402/blob/main/.github/workflows/publish_npm_scoped_x402_all.yml

examples/typescript/clients/advanced/README.md
```diff
@@ -44,6 +44,7 @@ and fill required environment variables:
 - `HEDERA_ACCOUNT_ID` - Hedera account id for Hedera payments (optional)
 - `HEDERA_PRIVATE_KEY` - Hedera **ECDSA** private key (0x-prefixed or DER-encoded) for Hedera payments (optional)
 - `HEDERA_NETWORK` - Hedera network (optional, defaults to `hedera:testnet`)
+- `KEETA_PRIVATE_KEY` - Keeta passphrase for Keeta payments
```
phdargen: Seems pk isnt supported by the sdk, only mnemonic

examples/typescript/facilitator/advanced/all_networks.ts
```diff
@@ -50,6 +53,7 @@ const PORT = process.env.PORT || "4022";
 // Configuration - optional per network (alphabetic order)
 const avmPrivateKey = process.env.AVM_PRIVATE_KEY as string | undefined;
 const evmPrivateKey = process.env.EVM_PRIVATE_KEY as `0x${string}` | undefined;
+const keetaPassphrase = process.env.KEETA_PASSPHRASE as string | undefined;
```
phdargen: Suggest to rename here and elsewhere:
```suggestion
const keetaMnemonic = process.env.KEETA_MNEMONIC as string | undefined;
```

examples/typescript/clients/advanced/README.md
```diff
@@ -69,6 +70,14 @@ Stellar accounts need to be created and funded with both XLM and USDC. Instructi
 2. Add USDC trustline (required to transact USDC): go to [Fund Account](https://lab.stellar.org/account/fund) ➡️ Paste your `Public Key` ➡️ Add USDC Trustline ➡️ paste your `Secret key` ➡️ Sign transaction ➡️ Add Trustline.
 3. Get testnet USDC from [Circle Faucet](https://faucet.circle.com/) (select Stellar network).
 
+#### Keeta Testnet
+
+To create a Keeta Testnet wallet:
+
+1. Go to [Keeta Testnet Wallet](https://wallet.test.keeta.com/) and follow the steps to create your wallet. Make sure to save your passphrase to keep access to your wallet. To get your Keeta address, click on "Receive" and copy the deposit address (starting with `keeta_`).
+2. Use the [Keeta Testnet Faucet](https://faucet.test.keeta.com/) to send Testnet KTA to your wallet.
+3. To get Testnet USDC on Keeta, go to the "Receive" page in the wallet, click on "Any token from Keeta Testnet", select "USDC from Base (Sepolia) Testnet" and copy the deposit address (starting with `0x`). Then go the [Circle Faucet](https://faucet.circle.com/), select Base network and enter your Base deposit address.
```
phdargen: Thanks for providing this info, I managed to get some testnet KTA. 

But after clicking "Deposit USDC from Base (Sepolia) Testnet", it made me fill out a lot of forms, where I reluctantly put in some random stuff only to get asked to download an app to upload some ID document, at which point I gave up on the process. Is this simulated KYC process really needed on testnet, can I skip this somehow?

Image
Image

e2e/servers/express/test.config.json
```diff
@@ -172,6 +172,13 @@
       "requiresPayment": true,
       "protocolFamily": "hedera"
     },
+    {
+      "path": "/exact/keeta",
+      "method": "GET",
+      "description": "Protected endpoint requiring payment on Keeta network",
+      "requiresPayment": true,
+      "protocolFamily": "keeta"
+    }
```
phdargen:
```suggestion
    },
```

e2e/clients/axios/index.ts
```diff
@@ -122,6 +133,8 @@ if (aptosAccount) {
 }
 if (hederaClientSigner) {
   client.register("hedera:*", new ExactHederaScheme(hederaClientSigner));
+if (keetaSigner) {
+  client.register(KEETA_TESTNET_CAIP2, new ExactKeetaScheme(keetaSigner));
 }
```
phdargen, line 183: There are similar bugs in all other e2e scripts that would break the e2e tests at startup or runtime, please fix and run them to confirm they succeed before committing 
```suggestion
if (hederaClientSigner) {
   client.register("hedera:*", new ExactHederaScheme(hederaClientSigner));
}
if (keetaSigner) {
  client.register(KEETA_TESTNET_CAIP2, new ExactKeetaScheme(keetaSigner));
}
```

phdargen:
Please also add e2e tests for the remaining server packages: hono, next, fastify

typescript/packages/mechanisms/keeta/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,367 @@
...
+
+      return {
+        isValid: false,
+        invalidReason: "invalid_exact_keeta_payload_block_could_not_be_decoded",
+        payer: "",
+      };
+    }
+
+    const caip = requirements.network.split(":");
+    const networkId = BigInt(caip[1]);
```
phdargen: might throw on a malformed `requirements.network`, facilitator should never return 500 but isValid: false

please wrap code paths that could throw in a try/catch

typescript/packages/mechanisms/keeta/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,367 @@
...
+      };
+    }
+
+    // 4.4 Verify the payment operation
+    const [payOperation] = block.operations;
+    const payOperationVerificationResult = this.verifyPaymentOperation(payOperation, requirements);
+    if (payOperationVerificationResult !== null) {
+      return payOperationVerificationResult;
+    }
+
```
phdargen, line 144: should add explicit check that the sponsor cannot be tricked into moving its own funds like:
```
const facilitatorAddresses = this.signer.getAddresses();
if (facilitatorAddresses.includes(block.account.publicKeyString.toString())) {
  return {
    isValid: false,
    invalidReason: "invalid_exact_keeta_payload_payer_is_facilitator",
    payer: "",
  };
}
```

typescript/packages/mechanisms/keeta/src/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/**
+ * Convert a decimal amount to token smallest units
+ *
+ * @param decimalAmount - The decimal amount (e.g., "0.10")
+ * @param decimals - The number of decimals for the token (e.g., 6 for USDC)
+ * @returns The amount in smallest units as a string
+ */
+export function convertToTokenAmount(decimalAmount: string, decimals: number): string {
```
phdargen: replace by `import { convertToTokenAmount } from "@x402/core/utils";`

typescript/packages/mechanisms/keeta/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Parse Money (string | number) to a decimal number.
+   * Handles formats like "$1.50", "1.50", 1.50, etc.
+   *
+   * @param money - The money value to parse
+   * @returns Decimal number
+   */
+  private parseMoneyToDecimal(money: string | number): number {
```
phdargen, line 100: Please use core utils where possible:
` import { convertToTokenAmount, numberToDecimalString, parseMoneyString } from "@x402/core/utils";`

examples/typescript/facilitator/advanced/all_networks.ts
sc4l3r: The SDK explicitly calls it passphrase in [seedFromPassphrase](https://static.network.keeta.com/docs/classes/KeetaNetSDK.Referenced.Account.html#seedfrompassphrase) which is why I stuck to using passphrase everywhere.

Do you think that mnemonic is better to read / understand for other users and thus outweighs the benefit of sticking to the SDK term?

examples/typescript/clients/advanced/README.md
sc4l3r: The web wallet got updated recently and it seems that on testnet it defaults to the bridge that requires identity verification (even though there's a simple base -> keeta bridge available as well). I've brought that up with the team and I'm sure they'll fix it soon.

In the meantime, simply select cbBTC instead of USDC and use the deposit address it shows there, that's the same.

Alternatively, you can use the v1 web wallet interface (which this guide was originally written for) at https://v1.wallet.test.keeta.com/ . Once you've selected "USDC from Base (Sepolia) testnet" there, it should show a provider selection dialog with "BRIDGE1" and "DEV2" as the options. Select "DEV2" for the normal Base -> Keeta bridge. Then you should see your Base deposit address.

image

sc4l3r:
@phdargen Thanks for the quick review! I've addressed all the feedback above. I've added E2E tests for hono, next, and fastify and all E2E tests are working now (faulty rebase broke them initially...); tested with:

```
📊 Breakdown by Facilitator:
 typescript      ✅ 10 / ❌ 0 (100%)

📊 Breakdown by Server:
 express              ✅ 2 / ❌ 0 (100%)
 fastify              ✅ 2 / ❌ 0 (100%)
 hono                 ✅ 2 / ❌ 0 (100%)
 next                 ✅ 4 / ❌ 0 (100%)

📊 Breakdown by Client:
   axios                ✅ 5 / ❌ 0 (100%)
   fetch                ✅ 5 / ❌ 0 (100%)
```

typescript/site/app/facilitator/index.ts
```diff
@@ -161,6 +164,21 @@ async function createFacilitator(): Promise<x402Facilitator> {
...
+  if (process.env.FACILITATOR_KEETA_PASSPHRASE) {
+    const keetaAccounts = await Promise.all(
+      process.env.FACILITATOR_KEETA_PASSPHRASE.split(",")
+        .map(p => p.trim())
+        .filter(p => p.length > 0)
+        .map(async p => {
+          const seed = await KeetaNet.lib.Account.seedFromPassphrase(p);
+          return KeetaNet.lib.Account.fromSeed(seed, 0);
+        }),
+    );
```
sc4l3r: @phdargen Do you prefer multiple passphrases or using a single passphrase and deriving multiple accounts from that via the index?

We could add an env `FACILITATOR_KEETA_SIGNER_AMOUNT` and then do this to generate a set of signers from a single passphrase:

```typescript
const amountSigners = parseInt(
  process.env.FACILITATOR_KEETA_SIGNER_AMOUNT ?? "1",
);
const keetaAccounts = [];
for (let i = 0; i < amountSigners; i++) {
  const account = KeetaNet.lib.Account.fromSeed(
    await KeetaNet.lib.Account.seedFromPassphrase(
      process.env.FACILITATOR_KEETA_PASSPHRASE,
    ),
    i,
  );
  keetaAccounts.push(account);
}
```

Only downside is that you currently can't use the accounts derived from a non-zero index in the web wallet (since it takes a passphrase and defaults to index 0). But if you simply fund these accounts via the faucet you wouldn't have to use the web wallet anyway so that shouldn't be a problem really.

Both is fine and only depends on how you'd want to deploy this (if you even want to deploy multiple signers at all).

typescript/packages/mechanisms/keeta/src/exact/facilitator/scheme.ts
sc4l3r: I added that since it's a cheap check. However, as described in https://github.com/x402-foundation/x402/pull/1575#discussion_r2987947829 it's not really a concern since either `block.signer == block.account` in which case the facilitator essentially created the block or `blocker.signer` has the permission to act on behalf of `block.account` (or else the network would reject the block). In both cases such block could be submitted to the network directly without going through the facilitator here.

examples/typescript/clients/advanced/all_networks.ts
```diff
@@ -75,6 +80,17 @@ async function main(): Promise<void> {
     console.log(`Initialized EVM account: ${evmSigner.address}`);
   }
 
+  // Register Keeta scheme if passphrase is provided
+  if (keetaPassphrase) {
+    const keetaAccount = KeetaNet.lib.Account.fromSeed(
+      await KeetaNet.lib.Account.seedFromPassphrase(keetaPassphrase),
+      0,
+    );
+    const keetaSigner = toClientKeetaSigner(keetaAccount);
```
phdargen: A test payment succeeded but the process never exited because the SDK's `UserClient` keeps network connections open. `toClientKeetaSigner() `caches those clients and never tore them down, so Node stayed alive after `main()` finished. Please add some auto-cleanup

examples/typescript/facilitator/advanced/all_networks.ts
phdargen: Yes I think `mnemonic` would be clearer to most users/devs. At least to me, it wasnt immediately obvious what a `passphrase` is supposed to be

https://wallet.test.keeta.com/  doesnt use the term `passphrase` either, but `seed phrase`

examples/typescript/clients/advanced/README.md
phdargen: OK thanks, that worked

typescript/site/app/facilitator/index.ts
phdargen: a single passphrase seems more convenient

typescript/packages/mechanisms/keeta/src/exact/facilitator/queue.ts
```diff
@@ -0,0 +1,234 @@
...
+   * @param network - The network identifier
+   * @returns The block hash from the submitted transaction
+   */
+  async enqueue(feePayer: string, encodedBlock: string, network: Network): Promise<string> {
+    const blockId = new KeetaNet.lib.Block(
+      encodedBlock,
+    ).hash.toString() as unknown as KeetaAnchorQueueRequestID;
+
+    if (this.pendingPromises.has(blockId)) {
+      throw new Error("duplicate_block");
```
phdargen: this error code should be propagated to the client, currently it only surfaces as generic `errorReason: 'transaction_failed'`

typescript/packages/mechanisms/keeta/src/exact/facilitator/queue.ts
```diff
@@ -0,0 +1,234 @@
...
+
+    let resolve!: (blockHash: string) => void;
+    let reject!: (error: Error) => void;
+    const promise = new Promise<string>((res, rej) => {
+      resolve = res;
+      reject = rej;
+    });
+
+    await runnerEntry.runner.add({ feePayer, encodedBlock, network }, { id: blockId });
+    this.pendingPromises.set(blockId, { resolve, reject });
```
phdargen: `pendingPromises.set` runs after `await runner.add`, but a drain loop from a prior enqueue may already be running and pick up the entry in that window. The processor then hits `if (!pending)` and throws `"not pending anymore"` (queue.ts:74-77); since the entry is already consumed and the promise was never registered, enqueue's promise never settles and the settle HTTP request may hang indefinitely 

Register the pending promise before adding:
```suggestion
    this.pendingPromises.set(blockId, { resolve, reject });
    try {
      await runnerEntry.runner.add({ feePayer, encodedBlock, network }, { id: blockId });
    } catch (error) {
      this.pendingPromises.delete(blockId);
      throw error;
    }
```

examples/typescript/facilitator/advanced/all_networks.ts
sc4l3r: Alright, renamed all usages of `passphrase` to `mnemonic`. Make sure to adjust your `.env` files as well (e.g. `KEETA_PASSPHRASE` is now `KEETA_MNEMONIC`).

examples/typescript/clients/advanced/all_networks.ts
sc4l3r: I've added a `destroy` method as well as a `Symbol.asyncDispose` implementation (which calls `destroy`) on both client and facilitator signers now and adjusted the examples accordingly. Locally this client now terminates correctly.

typescript/packages/mechanisms/keeta/src/exact/facilitator/queue.ts
sc4l3r: Good catch! Fixed as suggested.

phdargen (approved):
Thanks for the update @sc4l3r, LGTM 🚀

PR #2595, opened 2026-06-11
feat(extensions): add code and cause to SIWx failure results

### LINKED ISSUE #2586: Expose structured error codes from validateSIWxMessage and preserve verifier errors in verifySIWxSignature
Hey, I was working on a merchant server doing per-request SIWX auth and ran into the following while using the sign-in-with-x extension from `@x402/extensions`. It feels like the same class of issue as #2394, just scoped to the SIWX extension. There are two error signaling gaps that force consumers to string match prose.

## 1. `validateSIWxMessage` returns prose-only errors ([validate.ts](https://github.com/x402-foundation/x402/blob/main/typescript/packages/extensions/src/sign-in-with-x/validate.ts))

`SIWxValidationResult` is `{ valid: boolean; error?: string }` ([types.ts](https://github.com/x402-foundation/x402/blob/main/typescript/packages/extensions/src/sign-in-with-x/types.ts)), and the error values are prose: `"Message expired"`, `` `Message too old: ${n}s exceeds ${m}s limit` ``, `"Nonce validation failed (possible replay attack)"`, `"Domain mismatch: ..."`, etc.

My server exposes machine-readable auth errors to clients (`WWW-Authenticate: ... error="expired"`, so an HTTP client knows whether to re-sign or give up). The only way to produce those today is to regex match the library's strings, which means exact-pinning the package and pinning the strings in tests, and it breaks silently on any wording change.

It would be nice to have a stable code next to the existing `error`, one value per existing check.  For example,
```ts
type SIWxValidationCode =
  | 'domain_mismatch' | 'uri_mismatch'
  | 'invalid_issued_at' | 'too_old' | 'issued_at_in_future'
  | 'invalid_expiration_time' | 'expired'
  | 'invalid_not_before' | 'not_yet_valid'
  | 'nonce_invalid';

interface SIWxValidationResult { valid: boolean; code?: SIWxValidationCode; error?: string }
```
`@signinwithethereum/siwe` (used by [evm.ts](https://github.com/x402-foundation/x402/blob/main/typescript/packages/extensions/src/sign-in-with-x/evm.ts)) exports `SiweError.type: SiweErrorType` for exactly this, but the extension doesn't surface any equivalent structure.

## 2. `verifySIWxSignature` swallows `evmVerifier` errors ([verify.ts](https://github.com/x402-foundation/x402/blob/main/typescript/packages/extensions/src/sign-in-with-x/verify.ts))

`verifyEVMPayload` (and the outer catch in `verifySIWxSignature`) does roughly:

```ts
} catch (error) {
  return {
    valid: false,
    error: error instanceof Error ? error.message : "Signature verification failed",
  };
}
```

My `evmVerifier` does an EIP-1271/6492 RPC lookup, and from the return value I can't tell  if it is "signature did not verify" (401 error )or  "RPC is down, could not check" (503 Error). The workarounds I've found are mutable flags set inside the callback as a side channel, or string matching `error`.

It would be nice if the thrown error was preserved instead of collapsed, e.g. a `cause` field on the failure result, where its presence means the verifier threw rather than returned false:

```ts
interface SIWxVerifyResult { valid: boolean; address?: string; error?: string; cause?: unknown }
```

The function keeps its never-throws contract, and a consumer can stay fail-closed by default (unrecognized errors still map to 401) and only upgrade recognized infrastructure errors to 503.

One small related thing: `validateSIWxMessage` propagates `checkNonce` throws uncaught, which is useful (I rely on it to fail closed when the nonce store errors), but it's undocumented. A JSDoc note on `checkNonce` guaranteeing that behavior, like the `@throws` tags elsewhere in the module, would make it safe to rely on.

Both proposals are purely additive and existing consumers see no behavior change.  

I would be to happy to open up PR if above suggestions works!

vraspar:
## Description

Closes #2586. 

- `validateSIWxMessage` failure results now include a `code` field (new exported `SIWxValidationCode` union) next to the existing prose `error`. 
- `verifySIWxSignature` failures now include `cause` with the original thrown error when verification threw, so callers can distinguish an invalid signature (401-class) from a verifier failure such as an RPC outage (503-class) without string matching. 
- The `checkNonce` JSDoc now states that errors it throws propagate to the `validateSIWxMessage` caller. All changes are additive and backward compatible.
- This also fixes a missing `await` in the EVM branch of `verifySIWxSignature`: a malformed CAIP-2 chainId such as `eip155:abc` previously made the function reject, and it now resolves `{ valid: false }` with `error` and `cause` as documented.

## Disclosure
This PR was prepared with the assistance of a coding agent (Claude Code) and reviewed by the author before submission.

## Tests

- `code` assertions on existing validate tests, plus an `it.each` table covering the remaining 9 codes (`domain_mismatch` is asserted on the existing domain-mismatch test)
- `checkNonce` throw propagation to the `validateSIWxMessage` caller
- `cause` is the thrown error when `evmVerifier` throws, and absent when the verifier returns `false`
- malformed chainId resolves `{ valid: false }` instead of rejecting

Verified locally from `typescript/`:
- `pnpm --filter @x402/extensions test` → 476 passed (95 in sign-in-with-x)
- `pnpm --filter @x402/extensions format:check` and `lint:check` → clean

## Checklist

### DIFF
--- a/typescript/packages/extensions/src/sign-in-with-x/types.ts
+++ b/typescript/packages/extensions/src/sign-in-with-x/types.ts
@@ -171,10 +171,27 @@ export interface DeclareSIWxOptions {
 }
 
 /**
+ * Machine-readable code identifying which validation check failed
+ */
+export type SIWxValidationCode =
+  | "domain_mismatch"
+  | "uri_mismatch"
+  | "invalid_issued_at"
+  | "too_old"
+  | "issued_at_in_future"
+  | "invalid_expiration_time"
+  | "expired"
+  | "invalid_not_before"
+  | "not_yet_valid"
+  | "nonce_invalid";
+
+/**
  * Validation result from validateSIWxMessage
  */
 export interface SIWxValidationResult {
   valid: boolean;
+  /** Code for the failed check (present when valid is false) */
+  code?: SIWxValidationCode;
   error?: string;
 }
 
@@ -184,7 +201,12 @@ export interface SIWxValidationResult {
 export interface SIWxValidationOptions {
   /** Maximum age for issuedAt in milliseconds (default: 5 minutes) */
   maxAge?: number;
-  /** Custom nonce validation function */
+  /**
+   * Custom nonce validation function.
+   *
+   * Errors thrown here are not caught: they propagate to the
+   * validateSIWxMessage caller, so a failing nonce backend fails closed.
+   */
   checkNonce?: (nonce: string) => boolean | Promise<boolean>;
 }
 
@@ -196,6 +218,8 @@ export interface SIWxVerifyResult {
   /** Recovered/verified address (checksummed) */
   address?: string;
   error?: string;
+  /** Error thrown during verification (absent when a signature check simply failed) */
+  cause?: unknown;
 }
 
 /**
--- a/typescript/packages/extensions/src/sign-in-with-x/types.ts
+++ b/typescript/packages/extensions/src/sign-in-with-x/types.ts
@@ -174,26 +174,26 @@ export interface DeclareSIWxOptions {
  * Machine-readable code identifying which validation check failed
  */
 export type SIWxValidationCode =
-  | "domain_mismatch"
-  | "uri_mismatch"
-  | "invalid_issued_at"
-  | "too_old"
-  | "issued_at_in_future"
-  | "invalid_expiration_time"
-  | "expired"
-  | "invalid_not_before"
-  | "not_yet_valid"
-  | "nonce_invalid";
+  | "invalid_siwx_domain_mismatch"
+  | "invalid_siwx_uri_mismatch"
+  | "invalid_siwx_issued_at"
+  | "invalid_siwx_issued_at_too_old"
+  | "invalid_siwx_issued_at_in_future"
+  | "invalid_siwx_expiration_time"
+  | "invalid_siwx_expired"
+  | "invalid_siwx_not_before"
+  | "invalid_siwx_not_yet_valid"
+  | "invalid_siwx_nonce";
 
 /**
- * Validation result from validateSIWxMessage
+ * Validation result from validateSIWxMessage.
+ *
+ * Failure fields follow the facilitator VerifyResponse naming
+ * (isValid/invalidReason/invalidMessage).
  */
-export interface SIWxValidationResult {
-  valid: boolean;
-  /** Code for the failed check (present when valid is false) */
-  code?: SIWxValidationCode;
-  error?: string;
-}
+export type SIWxValidationResult =
+  | { isValid: true }
+  | { isValid: false; invalidReason: SIWxValidationCode; invalidMessage: string };
 
 /**
  * Options for message validation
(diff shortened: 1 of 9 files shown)


typescript/packages/extensions/src/sign-in-with-x/types.ts
```diff
@@ -170,11 +170,28 @@ export interface DeclareSIWxOptions {
...
+  | "domain_mismatch"
+  | "uri_mismatch"
+  | "invalid_issued_at"
+  | "too_old"
+  | "issued_at_in_future"
+  | "invalid_expiration_time"
+  | "expired"
+  | "invalid_not_before"
+  | "not_yet_valid"
+  | "nonce_invalid";
```
phdargen: Suggest adding a `invalid_siwx_*` prefix similar to scheme specific errors like `invalid_exact_evm_payload_signature`

typescript/packages/extensions/src/sign-in-with-x/types.ts
```diff
@@ -170,11 +170,28 @@ export interface DeclareSIWxOptions {
...
+
 /**
  * Validation result from validateSIWxMessage
  */
 export interface SIWxValidationResult {
   valid: boolean;
+  /** Code for the failed check (present when valid is false) */
+  code?: SIWxValidationCode;
   error?: string;
 }
```
phdargen: Consider discriminated union for better typing
```suggestion
type SIWxValidationResult =
  | { valid: true }
  | { valid: false; code: SIWxValidationCode; error: string };
```

phdargen: Might be worth aligning the shape with a regular verify response:
```
valid -> isValid
code -> invalidReason
error -> invalidMessage 
```
https://github.com/x402-foundation/x402/blob/530400528bbd721c719645fc9018d0f34260c395/typescript/packages/core/src/types/facilitator.ts#L10-L17

typescript/packages/extensions/src/sign-in-with-x/types.ts
phdargen: Could you please also add the error codes to the siwx spec?

vraspar: Done, All codes now use the `invalid_siwx_` prefix and also added error codes table to `specs/extensions/sign-in-with-x.md` under Verification Logic, Validate Message Fields.

typescript/packages/extensions/src/sign-in-with-x/types.ts
vraspar: Done, combined with the naming suggestion

vraspar: Done

phdargen:
Thanks for the update @vraspar, I took this over with a few more fixes in https://github.com/x402-foundation/x402/pull/2888 as I couldnt push to your branch. Want to get this out in today's releases along with some other SIWX changes made in https://github.com/x402-foundation/x402/pull/2859. Feel free to sync this PR instead if you prefer

Will also follow up with the equivalent go/py changes shortly

phdargen:
Continued in https://github.com/x402-foundation/x402/pull/2888

Issue #2452, opened 2026-05-23
Composition with AP2 mandate verification — reference implementation


EfeDurmaz16:
x402 standardizes the rail (HTTP 402 → signed payment payload → settlement). It deliberately does not standardize *authorization* — whether the agent making the call was permitted by its principal to spend on this resource, for this amount, at this time.

In practice this leaves a gap for any agent operator deploying x402 in production: a stolen key, a prompt-injected agent, or a buggy planner can produce a perfectly valid x402 payment that the principal never authorized. The operator carries strict liability for those settlements, and the spec gives them no in-band hook to prevent it.

Google's Agent Payments Protocol (AP2) fills exactly that gap on the policy side: Intent → Cart → Payment mandates, each a VC-signed object that proves the principal authorized this specific spend. AP2 says nothing about rails. The two protocols are mechanically complementary.

Sardis (MIT, `sardis@2.0.0a0`) composes them:

```
AP2 mandate verify → policy.check → x402 challenge/payload → audit hash chain
```

Reference implementation:
- AP2 ↔ internal mandate adapter: [`packages/sardis/src/sardis/ucp/adapters/ap2.py`](https://github.com/EfeDurmaz16/sardis/blob/main/packages/sardis/src/sardis/ucp/adapters/ap2.py)
- x402 challenge / payload / verifier: [`packages/sardis/src/sardis/protocol/x402.py`](https://github.com/EfeDurmaz16/sardis/blob/main/packages/sardis/src/sardis/protocol/x402.py)
- Mandate cache (replay/expiry): [`packages/sardis/src/sardis/protocol/mandate_cache.py`](https://github.com/EfeDurmaz16/sardis/blob/main/packages/sardis/src/sardis/protocol/mandate_cache.py)

The question I'd like to put to the maintainers: **is AP2-style mandate verification something the x402 spec should reference as a recommended pre-flight verifier interface (non-normative appendix), or should this composition stay entirely external to the spec?**

Either answer is useful — but if x402 stays silent on authorization, every agent-economy deployment will reinvent this layer, mostly badly.

Happy to draft a spec appendix if you'd find it useful. — Efe

TateLyman:
I would keep AP2 mandate verification non-normative for x402 itself, but add an explicit pre-flight verifier hook/appendix.

Reasoning:

- x402 should stay rail-focused: challenge shape, payment payload, settlement/facilitator behavior, and receipts.
[...]

Issue #2273, opened 2026-05-12
[Extension Proposal] metered-session — prepaid metered sessions with auto-refund for streaming x402 transports


opcastil11:
## Summary

This extension defines a **session-level payment model** for x402 transports where consumption is continuous and non-deterministic (WebSocket, gRPC streaming, SSE). The client prepays for an estimated budget; the server meters actual consumption in real time over the same transport; on session close, the server settles the consumed amount and refunds the remaining balance on-chain.

Strictly complementary to the per-message WS transport in #1231: that transport handles RPC-style paid calls, this extension handles stateful streams where per-call settlement is impractical (LLM token streams, live video, metered RPC firehose, real-time data feeds, AI inference, metered compute, paid VPN).

## Motivation

x402's `402 → pay → response` cycle is inherently per-request. Streaming workloads break this in two ways:

1. **Settlement granularity** — paying per chunk/frame adds latency and on-chain overhead that defeats WS use cases.
2. **Unknown total cost** — the client cannot pay upfront for "what the stream will cost" because it doesn't know how long it will subscribe or how much it will consume.

The fix is the prepay-with-refund pattern, standard in metered SaaS billing and already shipping as vendor SDKs in the x402 ecosystem (Vybe's `@vybenetwork/x402-client`, Atomic Stream, QuickNode's WS RPC tier). Without a canonical extension, every provider implements this differently and clients can't reuse a single SDK across services.

### Schema additions (`specs/extensions/metered-session.md`)

- **`SessionRequirementsResponse`** — extends `PaymentRequirementsResponse`:
  - `pricePerUnit`, `unit` (`second` | `byte` | `token` | custom)
  - `estimatedDuration`, `maxSessionDuration`
  - `refundPolicy` (`on-close` | `signed-receipt` — see open questions)

- **`SessionUpdate`** frame (server → client, periodic):
  - `elapsed`, `consumedAmount`, `remainingBalance`
  - optional `bytesTransferred`, `messageCount`, `unitsConsumed`

- **`SessionSettlement`** frame (server → client on close):
  - `consumedAmount`, `refundAmount`, `refundTxHash`, `cause`

- **`SessionTerminate`** frame (server → client, standardized cause codes):
  - `balance-exhausted` | `max-duration-reached` | `server-shutdown` | `client-disconnect` | `policy-violation`

### Lifecycle

1. Client requests resource → server returns `SessionRequirementsResponse`.
[...]

Issue #2126, opened 2026-04-24
[Proposal]: credit-bundle — a fiat-payment fallback scheme for x402


orgtom78:
# Proposal: `credit-bundle` — a fiat-payment fallback scheme for x402

**Target:** `x402-foundation/x402`
**Type:** Feature discussion / scheme proposal
**Author:** Tobias Pfuetze, *[delegare.dev](https://delegare.dev/)*

---

## Problem

The x402 spec currently defines a single payment rail: on-chain USDC via EVM-compatible wallets.
This creates a hard blocker for the largest segment of potential enterprise API consumers:
**corporate buyers who can only pay by credit card or invoice and have no crypto infrastructure**.

In practice, a client hitting a 402 without a wallet has two choices today:
1. Get an unhelpful error with no recovery path.
2. Skip the gated resource entirely.

Neither is acceptable for production B2B deployments. The spec already states it aims to be
"network agnostic" — this proposal extends that to include **fiat payment rails**.

---

### Core idea

Add a `credit-bundle` entry to the `paymentMethods` array in the 402 response body, alongside
the existing USDC `accepts` array. Clients that cannot pay on-chain can purchase a pre-paid
credit bundle via Stripe/Paddle/any fiat processor, then authenticate subsequent requests with
a short-lived `X-Bundle-Token` JWT.

### Why pre-paid bundles and not per-request card charges?

- Stripe's minimum fee is ~$0.30/transaction. Per-request card charging is economically
  unviable below ~$5 per call.
- Corporate AP teams want a single line-item invoice, not thousands of micro-charges.
- Pre-paid bundles give merchants predictable cash flow and clients predictable cost.
- A `$10 minimum = 500 requests` model lands at $0.02/request — commercially sensible.

---

### 402 response body (backward compatible)

```json
{
  "x402Version": 1,
  "accepts": [
    {
      "scheme": "exact",
      "network": "base",
      "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
      "maxAmountRequired": "0.05",
      "payTo": "0xMerchantWallet",
      "resource": "/api/v1/classify",
      "description": "USDC payment for /api/v1/classify",
      "mimeType": "application/json",
      "maxTimeoutSeconds": 3600
    }
  ],
  "paymentMethods": [
    {
      "type": "x402-usdc",
      "details": { "...": "same as accepts[0]" }
    },
    {
      "type": "credit-bundle",
[...]

Issue #452, opened 2025-10-11
Define idempotent retry semantics for `/verify` and `/settle`


winor30:
## Summary
- `/verify` and `/settle` are POST endpoints but the spec never states how facilitators must deal with duplicate requests, so recoverable network errors can produce inconsistent retries.
- Each `PaymentPayload` already carries a unique EIP-3009 nonce; using the canonical payload as the idempotency key would let facilitators return the original response instead of surfacing contract-level nonce errors.
- Clarifying the requirement in [specs/x402-specification.md](https://github.com/coinbase/x402/blob/main/specs/x402-specification.md) and aligning the reference tooling gives integrators a predictable retry contract without inventing new protocol surface.

## Background
- [specs/x402-specification.md](https://github.com/coinbase/x402/blob/main/specs/x402-specification.md) `POST /verify` and `POST /settle` but omits expected behaviour for repeated submissions of the same payload.
- Once settlement succeeds, the authorization nonce is consumed on-chain. Retrying the same payload after a transient failure eventually triggers an `AuthorizationUsed` revert in the ERC-3009 contract, which current facilitators surface as `success: false`.
- Reference implementations (e.g. typescript/packages/x402/src/schemes/exact/evm/facilitator.ts) do not persist prior results, so applications cannot distinguish “already settled” from genuine failures.

## Proposal
1. Spec change
   - Update `POST /verify` and `POST /settle` to state that facilitators MUST treat the canonical `paymentPayload` as the idempotency key. The first submission performs validation/settlement; subsequent identical submissions return the original `isValid` / `success` payload (including `transaction` when available).
   - Add a normative error code (e.g. `already_settled`) that facilitators return when they detect the payload has already succeeded, so clients can stop retrying without triggering compensation flows.

## Notes
- Treating the signed payload as the idempotency key keeps the protocol self-contained across HTTP, MCP, and any other transport that forwards the same payload.

shafu0x:
+1

junbeomlee:
+1

fabrice-cheng:
Thank you all for your input. My two cents:
Idempotency for duplicate payments tied to the same request should be handled at the application level, which is responsible for bypassing /verify and /settle altogether.

Only the application has enough context and business logic to determine what constitutes a new payable request versus one that should reuse a previously executed payment.

The aim is for facilitators to remain stateless.

Let me know if you disagree

winor30:
@fabrice-cheng thank you for your comment.
[...]

PR #2612, opened 2026-06-12
specs: add credit network binding for batch-settlement (general API monetisation, fiat-native)

orgtom78:
> **Status: proposal / discussion draft — not yet implemented as written.** This specifies a *new* binding. The shipped [`@delegare/x402`](https://www.npmjs.com/package/@delegare/x402) implements the on-chain `exact` scheme (EIP-3009 USDC on Base) plus a mandate/session-key authorization layer — **not** this binding's attestation/`/authorize` flow. Wire format will be aligned to this spec as it stabilizes. Please review it as a design proposal, not a description of shipped code.

Adds a credit-backed network binding for `batch-settlement` targeting general REST API monetisation with no onchain capital requirement, per the direction in #2126 (`accepts`-native, batch-settlement as the structural home — thanks @phdargen).

Fills the gap between the two existing bindings: `cloudflare:402` is credit-backed but specific to crawler/content monetisation; EVM covers general APIs but is capital-backed. This binding: pre-purchased fiat credit bundles, short-TTL JWS attestations, atomic decrement-on-authorize, PSP-agnostic settlement (Stripe, Adyen, …). **The bundle authority is expressed as the standard x402 facilitator** (`/verify`, `/settle` with binding semantics); the only new surface is a client-facing `/authorize` endpoint. AP2 mandate verification is included as an informative (non-normative) authorization profile, keeping the binding rail-focused per the discussion in #2452.

### Design notes for reviewers

- The binding trades one client→authority round trip per request for pre-spend atomic budget enforcement — the opposite trade from the EVM binding's locally-signed vouchers. An explicit design-choice section covers when each fits.
- Payment guarantee is explicit: synchronous settle (standard middleware flow, zero risk) or deferred settle bounded by a `holdExp` settlement window (RECOMMENDED ≥24h).
- v1 scope is consume-only: bundle purchase, top-up, and refund are client↔authority operations outside the x402 exchange.
- Relation to #2273 (metered-session): that proposal is an extension for prepaid metered sessions on streaming transports; this is a network binding of an existing scheme for request/response APIs. Complementary, not overlapping.

### Open question for the foundation

Every existing binding is concrete (`cloudflare:402` is one named network; EVM has deployed contracts), while this one is parameterized (`<authority>:credit`). If the foundation prefers concrete bindings, this can be reshaped as a single named network with the generic parts kept as guidance — happy to go either way. Since x402 v2 uses CAIP-2, there's also a question of ordering (`credit:<authority>` vs `<authority>:credit`) and whether strict CAIP-2 applies to off-chain bindings at all (`cloudflare:402` is also non-CAIP-2) — expanded in the thread below.

### Reference implementation

A production system implementing this binding's *concept* — Stripe-funded credit bundles and on-chain USDC accepted at the same endpoints — runs behind [`@delegare/x402`](https://www.npmjs.com/package/@delegare/x402) (npm). To be precise about scope: what ships today is the on-chain `exact` scheme (EIP-3009 USDC on Base) plus a mandate/session-key authorization layer; the credit-attestation flow specified here (`/authorize`, JWS attestations, `holdExp`, atomic decrement) is proposed, not yet implemented. Its wire format will be aligned to this spec as it stabilizes. Sandbox for testing: https://sandbox.delegare.dev

### Update — 2026-07 revision (`d81c85e`)

Pushed a revision addressing review feedback:

- **Idempotency is now normative** on `/authorize` (coalesce concurrent redemptions; retain the key binding through `holdExp`) and `/settle` (`jti`, concurrency-safe).
- **`/settle` uses the standard `SettleRequest`** — the actual charge is carried in the settlement-time `paymentRequirements.amount` (following the `upto` precedent), dropping the non-standard top-level `chargedAmount`/`status`.
- **`aud` and `resource` attestation claims are now REQUIRED**, closing cross-audience and cross-resource replay of a bearer attestation.
- Clarified that verification prevents double-*charge* but not concurrent double-*execution* — servers MUST coalesce fulfillment by `jti`.

Remaining open decisions (network-identifier/CAIP-2, whether a zero-cost *fulfilled* charge must be distinguishable from *failed*, and policy-versioning) are in the thread below.

---

Disclosure per CONTRIBUTING: drafted with AI assistance, reviewed and verified by the author.

Marked as draft — discussion surface, not a merge request.

### DIFF
--- a/specs/schemes/batch-settlement/scheme_batch_settlement_credit.md
+++ b/specs/schemes/batch-settlement/scheme_batch_settlement_credit.md
@@ -1,0 +1,335 @@
+# Scheme: `batch-settlement` `credit`
+
+## Summary
+
+The `batch-settlement` scheme on a `credit` network binding enables access to general API resources through pre-purchased fiat credit bundles. The client purchases credits from a **bundle authority** through a PSP (Stripe, Adyen, or any fiat rail), then presents a short-lived signed **attestation** per request. The authority atomically decrements the bundle balance at authorization time; settlement to the resource server happens later through off-chain financial rails on a defined schedule.
+
+**The bundle authority is the facilitator for this binding.** It exposes the standard x402 facilitator endpoints (`/verify`, `/settle`) toward resource servers, plus one binding-specific client-facing endpoint (`/authorize`) that issues attestations. No parallel server↔authority interface is introduced; existing x402 server middleware that knows how to call a facilitator works unchanged.
+
+**Network Identifier**: `<authority>:credit` (e.g. `acme:credit`)
+**Trust Model**: Credit-backed (see [scheme_batch_settlement.md](./scheme_batch_settlement.md#credit-backed))
+
+This binding fills the gap between the two existing `batch-settlement` bindings:
+
+- [`cloudflare:402`](./scheme_batch_settlement_cloudflare.md) is credit-backed but purpose-built for content monetisation via crawler-identity infrastructure (HTTP Message Signatures, `.well-known` key directories). Its registration and auth flow do not translate to general REST API routes.
+- [`EVM`](./scheme_batch_settlement_evm.md) covers general API monetisation but is capital-backed: it requires onchain deposits and wallet infrastructure from the client.
+
+The `credit` binding targets the remaining cell: **general API monetisation with no onchain capital requirement** — AI agents and corporate clients paying for API access through standard fiat billing.
+
+The scheme supports **dynamic pricing**: the client authorizes a maximum per request (`PaymentRequirements.amount`), and the server charges the actual cost within that ceiling, communicated via `PAYMENT-RESPONSE`.
+
+## Design choice: central authorization vs local signing
+
+This binding deliberately diverges from the EVM binding's locally-signed cumulative vouchers. There, the client signs vouchers offline with zero extra round trips, and the onchain channel guarantees the server can claim what was signed. That model is optimal for throughput but provides no hard cross-merchant budget cap: a client signing locally can oversign across many merchants in parallel, and no single party sees the cumulative total until settlement.
+
+The `credit` binding makes the opposite trade. Every authorization is an atomic decrement against a single authority, which means:
+
+- A budget cap (e.g. an agent's spending mandate) is enforced **before** spend, across all merchants drawing on the same bundle, with no read-modify-write race.
+- The cost is one client→authority round trip per request. (The server→facilitator `/verify` and `/settle` calls are the standard x402 server flow, not an additional burden of this binding, and `/settle` MAY be deferred; see [Settlement timing](#settlement-timing).)
+
+For high-frequency sub-cent calls where throughput dominates and the client is trusted with its own budget, the EVM binding is the better fit. For agentic clients operating under a principal's spending cap — where overspend is a liability question, not a performance question — pre-spend atomic enforcement is the point of the binding. Implementations MAY amortize the client round trip by authorizing a batch ceiling in one call where the use case tolerates it.
+
+## Roles
+
+- **Client** — an agent or application consuming paid API resources.
+- **Resource Server** — the API being monetised.
+- **Bundle Authority** — the facilitator for this binding. Sells credit bundles, holds balances, issues attestations via `/authorize`, verifies payments via `/verify`, accepts charge settlement via `/settle`, and settles with the resource server through the PSP rail. The authority may be operated by the resource server itself or by a third party.
+- **PSP** — the payment service provider (Stripe, Adyen, etc.) through which bundles are purchased. The PSP is invisible to the x402 protocol; only the authority interacts with it.
+
+## Scope
+
+The x402-visible surface of this binding is a single payload type: `consume`. Bundle lifecycle operations — purchase, top-up, and refund of unused credits — are client↔authority operations outside the x402 exchange, reachable from the `extra.purchaseUrl` in the `402` response. The server↔authority surface is the standard facilitator interface with binding-specific semantics specified below; the only novel endpoint is the client-facing `/authorize`.
+
+## Authority identity and trust binding
+
+Two identifiers refer to the authority; the spec binds them as follows:
+
+- **`extra.authority`** (a URL origin, e.g. `https://credits.acme.com`) is the trust anchor. It is the facilitator base URL, the key-discovery origin, and the attestation `iss` claim MUST equal it exactly.
+- **The `network` namespace** (e.g. `acme` in `acme:credit`) is a display and routing label. To avoid collisions, the namespace SHOULD be derived from the authority's registrable domain (e.g. `credits.acme.com` → `acme`). The resource server's declaration of `extra.authority` in the `accepts` entry is what binds the label to an origin for that resource.
+
+Clients and servers MUST anchor trust decisions to the `extra.authority` origin, never to the namespace label. A server lists an `accepts` entry for an authority precisely when it has a settlement relationship with that origin; a client buys credits from that origin precisely when it is willing to fund requests against servers that accept it.
+
+## Credits denomination
+
+Credits are denominated in the smallest unit of the bundle's `asset` (e.g. cents for USD). A bundle balance of `8420` with `asset: "USD"` is $84.20 of spending power. Purchase tiers MAY grant more credits than the purchase price (a volume bonus): `{ "credits": "10000", "price": "9000" }` means pay $90.00, receive $100.00 of credit. Attestation `amount`, `chargedAmount`, and `bundleState.balance` are all in the same unit; there is no separate abstract credit unit.
+
+## Protocol Flow
+
+### First-time setup (Client)
+
+1. Client receives a `402` with the `credit` binding in `accepts` and `extra.purchaseUrl`.
+2. Client (or the human principal behind it) completes a PSP-hosted checkout at `purchaseUrl`, purchasing a credit bundle tier.
+3. The authority issues the client a long-lived **bundle credential** (out of band — API key, OAuth client, or signed credential per the authority's account model).
+4. For autonomous top-ups, the principal may additionally establish a PSP off-session mandate during checkout, letting the agent top up the bundle through the authority's API without further human interaction.
+
+### Payment flow (per request)
+
+1. Client calls the authority's [`/authorize`](#post-authorize-client-facing) endpoint. The authority **atomically decrements** the bundle balance and returns a signed attestation in the same call, or rejects with `insufficient_credits`. There is no read-only balance query that is safe to act on; the idempotency key makes retries of the same logical request safe against double-decrement.
+2. Client sends the HTTP request with the payment payload (type `consume`) carrying the attestation.
+3. Resource server verifies the payment — via facilitator [`/verify`](#post-verify-standard-facilitator-endpoint), or locally by signature check (see [Verification](#verification)).
+4. On success, the server serves the resource and returns `PAYMENT-RESPONSE` with the actual charged amount and a bundle state snapshot.
+5. The server settles the actual charge via facilitator [`/settle`](#post-settle-standard-facilitator-endpoint) — synchronously before responding, or deferred within the hold window. The difference between authorized and charged amounts is re-credited to the bundle. If fulfillment fails, the server settles with `status: "failed"` (or lets the hold lapse at `holdExp`).
+6. The authority settles accumulated charges with the resource server on a billing cycle through the PSP rail.
+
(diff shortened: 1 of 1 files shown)


egoriklok:
Free public readiness snapshot for specs: add credit network binding for batch-settlement (general API monetisation, fiat-native):

- Public signal: Public GitHub candidate x402-foundation/x402#2612: specs: add credit network binding for batch-settlement (general API monetisation, fiat-native). Matched R1 terms: api, auth, crawl, payment, session, sessions, stripe, verified.
- R1 fit: api, auth, crawl, payment, session, sessions, stripe, verified, x402.
- Buyer-agent check I would add before autonomous spend: explicit price/cap unit, policy/version hash, receipt or charge-evidence reference, revocation/dispute path, and allow/caution/block preflight result.
- This separates "interesting paid surface" from "safe for an agent wallet to use".

No invoice, no payment link, no delivery link, no private endpoint; this is only a free public snapshot.

javimosch:
Following #2126 and this PR with a lot of interest — I run a small production rail that
is, structurally, a second implementation of exactly this shape, so sharing data points
in case they're useful while the binding is still being drafted.

**peage** (https://peage.intrane.fr) is a fiat pay-per-call rail for agents: prepaid
wallets funded by card via Stripe Checkout (EUR, cents as the atomic unit), merchants
settle with one HTTP call, every settlement returns an HMAC-signed receipt that anyone
can verify without auth. Live since July 2026 with three production merchants
(lead-gen API, artifact host, an OSS SerpApi alternative that embeds the merchant hook
behind a build flag).

Points of contact with this PR's design that might inform the binding:

- **Credit-backed commitment model works in practice.** Prepaid balance, atomic
  decrement on settle, no chain involved. Our receipt id is what we'd put in
  `transaction`, matching the spec's "commitment/receipt identifier for fiat bindings".
- **Both fixed and variable pricing map cleanly**: we ship synchronous exact-amount
  settlement *and* a hold→capture escrow (reserve max, settle actual, remainder
  auto-refunds on TTL) — i.e. `exact` and `upto` semantics on a fiat credit rail.
- **Idempotency needs to be in the binding.** Upstream /verify + /settle retry semantics
  are unspecified (#452); we require merchant-supplied idempotency keys so a retried
  settle can never double-bill. I'd argue the fiat-credit binding should make that
  normative rather than optional.
- **Caller-side spend caps** (per-charge + per-merchant daily, enforced by the wallet,
  not the merchant) turned out to be the thing that makes agents/operators comfortable
  handing a token to unknown merchants. Worth considering as a binding-level concept.
- **Auditability**: we expose a public solvency endpoint recomputing the ledger
  invariant per request (https://peage.intrane.fr/v1/solvency) — funded == float +
  escrow + merchant balances + fees. For credit-backed bindings where a trusted
  intermediary holds balances, something like this seems close to table stakes.

We've mapped our rail onto the v2 wire format (facilitator-shaped: /verify, /settle,
/supported; `network: peage:eur`, `asset: "EUR"`, `payTo: "merchant"`) and plan to
implement wire-compat once this binding stabilizes — we'd rather align with the
canonical naming from this PR than invent a parallel one. Happy to be an interop
test target, and to share ledger-design details if useful. Docs (written for agents):
https://peage.intrane.fr/llms.txt · repo: https://github.com/javimosch/peage

sunsetpatrol:
+1 to @javimosch's point that idempotency should be normative rather than optional for credit-backed bindings — with retry semantics unspecified upstream (#452), a duplicated redeem on a fiat rail is a real double charge, not a reverted transaction. To make the requirement concrete, here is the layering that survived testing in the Stipendex project (a shipped paid-tools rail with the same accumulate-then-redeem shape); all three layers turned out necessary, and each catches a failure mode the others miss:

1. **In-flight coalescing, keyed by the payment/commitment id.** Concurrent duplicate redeems of the same commitment collapse into a single execution — one verifier call, one charge, all callers get the same response. Without this, two racing retries both pass the "not yet redeemed" check.
2. **Completed-redemption cache.** A replay arriving *after* completion returns the cached result with the same receipt/settlement identifier, and the charge happens exactly once. This is the layer a client retry after a network timeout actually hits.
3. **First-claimer-wins binding.** The commitment id is claimed by the first concrete payment id that redeems it; a *different* payment id presenting the same commitment is rejected, and rollback of a failed redeem is only allowed to the holder of the claim. This blocks the cross-payment reuse case that layers 1–2 don't see.

Suggested normative shape for the binding: "A network binding MUST define an idempotency key for redemption; facilitators MUST coalesce concurrent redemptions on that key, MUST replay the original success response for repeated redemptions, and MUST NOT charge more than once per key."

Test vectors for the conformance set: (a) concurrent duplicate redeem → single charge, identical settlement identifiers in both responses; (b) late replay → same identifiers, balance unchanged. Happy to write these out fully if the direction lands.

orgtom78:
Pushed a revision (`d81c85e`) folding in the settled feedback — full details in the updated PR description. Short version: idempotency is now **normative** on both `/authorize` (coalesce concurrent redemptions; retain the key binding through `holdExp`) and `/settle` (`jti`, concurrency-safe); `/settle` now uses the **standard `SettleRequest`** with the actual charge carried in settlement-time `paymentRequirements.amount` (following `upto`), dropping the non-standard top-level `chargedAmount`/`status`; and `aud` + `resource` attestation claims are now **REQUIRED** to close cross-audience/cross-resource replay. I also added a note that verification prevents double-*charge* but not concurrent double-*execution* (servers MUST coalesce fulfillment by `jti`).

Three decisions I'd like input on before going further:

**1. Network identifier / CAIP-2.** The draft uses `<authority>:credit`. Since x402 v2 uses CAIP-2, that's arguably backwards — it puts each authority in the namespace (class) slot, whereas `credit:<authority-reference>` matches CAIP-2's namespace = class / reference = instance shape. But `cloudflare:402` is also non-CAIP-2, so the real question is whether strict CAIP-2 applies to off-chain bindings at all, or just colon-delimited x402 identifiers. If strict, both this binding and `cloudflare:402` need reconciling. Which way does the foundation want to go?

**2. Zero-cost fulfilled vs. failed at `/settle`.** Following `upto`, `amount: "0"` releases the hold — which collapses "fulfilled at zero cost" and "fulfillment failed" into one ledger operation. If they need to be distinguishable for audit, the status can't ride on `amount` alone; it would go in `paymentRequirements.extra`, which means extending facilitator settlement-override support. Worth it, or is a ledger-level release sufficient and the fulfilled/failed distinction belongs in server-side audit records?

**3. Policy/version signal (for @egoriklok).** To be precise about the earlier exchange: `amount`/`asset` are available pre-spend and `commitmentId` (the `jti`) is a post-settlement audit handle, but this binding has **no explicit policy/version identifier** today — and the informative AP2 `evidence` claim (a mandate-body digest) isn't equivalent to one. Should policy versioning live in this binding, or in a separate authorization extension?

_(Status note, also added to the description: this is a design proposal — the shipped `@delegare/x402` implements the on-chain `exact` scheme plus a mandate/session-key layer, not this binding's attestation flow yet.)_

javimosch:
Thanks @orgtom78 for folding the idempotency feedback into the revision — the normative `/authorize` coalescing + `/settle` `jti` is exactly the shape we needed. Aligning peage against the July revision:

**What already maps cleanly (no changes needed on our side):**

- **Standard `SettleRequest` with settlement-time `amount`** — peage's `/x402/settle` already carries the actual charge in `paymentRequirements.amount` and returns it in the `SettleResponse.amount` field. We never had a non-standard `chargedAmount`/`status`, so the revision is a no-op for us. Good.
- **Idempotency on `/settle` via `jti`** — we've had this since launch: `idempotency_key` in the scheme payload, `INSERT OR IGNORE` on `(merchant_id, idem)` in the charges table. A retried settle returns the same `transaction` (our signed receipt) with no double debit. The `jti` naming is new but the semantics are identical — we'll alias.
- **`exact` scheme on a named network** — peage speaks `scheme: "exact"`, `network: "peage:eur"`, `asset: "EUR"` (ISO 4217, cents), `payTo: "merchant"`. The `/x402/supported` endpoint advertises both v1 and v2. This is already live and tested (the test suite exercises the full v2 + v1-tolerant flow including the `PAYMENT-REQUIRED`/`PAYMENT-RESPONSE` header dance on a demo resource).

**Where the revised spec introduces concepts we don't have yet (and whether they matter for us):**

- **`/authorize` endpoint with JWS attestations + `holdExp`** — peage doesn't have an `/authorize` step. Our flow is direct: the merchant calls `/x402/settle` with the wallet token in the scheme payload, and the charge happens synchronously. The attestation model (short-TTL JWS, `aud` + `resource` claims) is designed for a pre-spend authorization round trip that decouples "can this agent pay?" from "charge it now." For a fiat credit rail where the wallet balance is checked synchronously at settle time and the whole thing is one HTTP call, the extra round trip doesn't buy much — but I can see it mattering for high-throughput merchants who want to reject unfunded callers before doing work. We'd implement it as a thin wrapper over our existing `/x402/verify` (which already checks balance + caps without moving money) plus a TTL'd attestation. Not blocking for us, but we could add it as an interop test target if the binding stabilizes.
- **`aud` + `resource` attestation claims (REQUIRED in the revision)** — these prevent cross-audience/cross-resource replay of a bearer attestation. peage's equivalent is the per-merchant `Bearer` key (the merchant identity is resolved from the key, not from the payload) plus the `idempotency_key` scoping. Different mechanism, same goal. If we add `/authorize`, we'd carry these claims.
- **Concurrent double-execution coalescing by `jti`** — the revision correctly notes that verification prevents double-*charge* but not concurrent double-*execution*, and servers MUST coalesce fulfillment. peage is a payment rail, not a fulfillment layer, so this is the merchant's responsibility — but our `idempotency_key` gives them the handle to do it.

**One concrete data point on the `holdExp` settlement window:**

We ship a hold→capture escrow (`POST /v1/holds` → `POST /v1/holds/capture`) that is structurally the same as the revision's `holdExp` concept: reserve the max, settle the actual, remainder auto-refunds on TTL expiry. Our default TTL is 900s (15 min), not 24h — the revision's RECOMMENDED `≥24h` makes sense for deferred-settle flows where the merchant batches at end of day, but for synchronous per-call settlement (our default), shorter is better because it bounds the time funds are locked. Worth noting in the spec that the window should be scheme-specific, not a flat minimum.

**Network identifier question:**

The PR asks whether to use `<authority>:credit` or `credit:<authority>`. peage uses `peage:eur` — which is `<authority>:<asset>`, not `<authority>:credit`. This is because peage is a concrete binding (one named rail, one asset), not a parameterized credit-bundle authority. If the foundation prefers concrete bindings (the PR's first option), `peage:eur` already fits. If the parameterized form wins, we'd register as a concrete instance under it. Either way, no change to our wire format.

Happy to run the revised spec's test vectors against our `/x402/verify` + `/x402/settle` once they're published. Our test suite already covers the v2 header flow end-to-end (including `unsupported_scheme` and `invalid_network` rejection paths), so adding conformance assertions for the attestation claims would be straightforward.

orgtom78:
Thanks @javimosch — a second production rail confirming the standard-`SettleRequest` + `jti`-idempotency shape is exactly the signal this needed, and the escrow/TTL data is genuinely useful.

**On `holdExp` — you've convinced me, and I've revised it (`c24f121`).** You're right that a flat RECOMMENDED ≥24h is wrong: `holdExp` is a *ceiling*, not a fixed lock, so the only place a long window bites is failure-reclaim (authorize → crash-before-settle → funds held until `holdExp`). The spec now makes the window **mode-specific**: synchronous settle SHOULD use a short window (minutes) to bound that reclaim case — your 900s is cited as the reference — while the ≥24h RECOMMENDED is scoped to deferred settle where it covers the background-retry horizon.

**On `/authorize`:** agreed it feels optional for a synchronous rail. Its point isn't rejecting unfunded callers (that's a side benefit) — it's atomic pre-*work* reservation and cross-merchant budget enforcement under a principal's cap, which a check-at-settle model can't do race-free once an agent fans out across merchants in parallel. For a purely synchronous rail like peage, collapsing authorize+settle into one call is reasonable, and your `/x402/verify` (balance + caps without moving money) is already most of the way there. I'll note a "synchronous profile" that allows this explicitly rather than requiring the extra round trip.

**On naming:** `peage:eur` is a useful third data point — it, `cloudflare:402`, and this draft are all **authority-first**, which is real ecosystem-convention evidence to weigh against the CAIP-2 "namespace = class" argument in open question #1 above. The concrete-binding path fits `peage:eur` as-is; the parameterized path would register it as a concrete instance. Either way no wire change for you.

Would love to have peage as an interop test target alongside Stipendex once the conformance vectors are written — and yes, please, the fully-written vectors (concurrent duplicate redeem → single charge; late replay → same identifiers, balance unchanged) would be a great contribution to the conformance set.

PR #2614, opened 2026-06-12
Validate Retry After delta seconds

samsamtrum:
The Retry After header defines delta seconds as nonnegative decimal integers. The current parser accepts fractional values like `1.5` via `Number()`, which can produce nonstandard retry delays instead of falling back to the existing backoff path.

This tightens delta seconds parsing to integer digits only and keeps HTTP date parsing unchanged.


phdargen (approved):
Thanks for the contribution @samsamtrum

PR #340, opened 2025-08-14
Add sui support

hayes-mysten:
## Description

This is still very much a work in progress, and is based off of #283 (and therefor the diff here is not very useful).

This PR is not really reviewable as is, because it contains all the changes from #283.  You can see all the sui specific changes in one commit [here](https://github.com/coinbase/x402/pull/340/commits/898b7524b721be4ff0e8983ae0f62e999d463151)

Some quick demos of the browser wallet and fetch examples working with sui:

https://github.com/user-attachments/assets/f10b4daa-e7ae-4f82-b8dd-51639e88d842

Screenshot 2025-08-14 at 3 28 22 PM

## Checklist

- [ ] I have formatted and linted my code
- [ ] All new and existing tests pass
- [ ] My commits are signed (required for merge)

PR #2615, opened 2026-06-12
spec: gasless Address Balances for exact-on-Sui

Sceat:
This amends [`scheme_exact_sui.md`](specs/schemes/exact/scheme_exact_sui.md) to complete the spec's **own stated follow-up**. The merged spec's Recommendation reads, verbatim:

> "In a follow up, once the 'Address Balances' feature has been implemented and rolled out, leverage it to be able to reduce the required gas fees, simplify the sponsorship flow, as well as investigate how `EIP-3009` and/or `EIP-2612` style authorizations could be implemented on top of 'Address Balances'."

Address Balances is now rolled out — Sui protocol v125, [Gasless Stablecoin Transfers](https://docs.sui.io/develop/transaction-payment/gasless-stablecoin-transfers). This PR adds the gasless construction the spec asked for, keeps the classic `Coin<T>` path as the `coin` method for back-compat and sub-minimum amounts, and leaves the `EIP-3009`/`EIP-2612` authorizations as the remaining Future Work item.

Follows up on **#340** — @phdargen's 2026-03-10 invitation to land a v2-native Sui mechanism; this is the spec half.

cc @bmwill — as the spec's author and the person who wrote the follow-up this implements.

---

## What changes

- **Two `extra.assetTransferMethod` values (added in review round 1)** — `address-balance` (gasless) and `coin` (classic), mirroring the EVM scheme layout, each with its own construction and verification section. The facilitator derives the effective method from the decoded bytes (`gasPayment == []` is `address-balance`) and rejects contradictions with the declared or echoed method. Sponsored transactions are the `coin` method's standard sponsorship flow: capability announced by the facilitator in `supported/`, gas station URL relayed by the resource server via `extra.gasStation`.
- **Gasless Construction (Address Balances)** — a new section: `gasPayment = []` ∧ `gasPrice = 0`, allowlisted ops only (the four functions that exist on-chain — `0x2::balance::{send_funds,redeem_funds}`, `0x2::coin::{send_funds,into_balance}` — plus the native `SplitCoins` / `MergeCoins` a `Coin<T>`-object source emits), no object writes, allowlisted stablecoins. The `0.01` minimum is noted as a **protocol parameter, not a security boundary** (the chain did not reject a sub-minimum transfer on testnet). The ~2-epoch `ValidDuring` window is the natural expiry.
- **Declared Outputs (`extra.outputs`, optional)** — requirements MAY declare `outputs[]` with `sum(outputs) == amount`; verification then anchors on the exact payer debit + exact per-output credits + no undeclared recipient. Absent → the single-`payTo` rule is unchanged. This is chain-native atomic routing (a PTB already expresses N transfers in one transaction): marketplace routing, affiliate / revenue splits, tax withholding.
- **Verification + Settlement** — extends the checklist (`x402Version` / scheme / network / payload-shape, signer-equals-sender binding, the gasless-shape assertions, outputs matching, replay / expiry) and notes keyless third-party broadcast (the facilitator holds no payer key) and idempotent re-broadcast (same digest, no double charge).
- **Future Work + Security Considerations** — removes the now-completed Address-Balances item (keeps `EIP-3009`/`EIP-2612`), and adds a Security Considerations section per `CONTRIBUTING.md`: replay prevention, authorization scope, settlement atomicity, and the no-object-writes bound.

RFC-2119 keywords throughout; the JSON examples use the exact v2 field names.

---

## Empirical evidence (real Sui testnet, re-verified on-chain)

Every transaction below settled gaslessly — `gasData.price = 0`, `gasPayment = []` — from a local CLI active address; recipients are ephemeral receive-only addresses. Inspect any digest at `https://suiscan.xyz/testnet/tx/<digest>`.

| The amendment claims… | Testnet digest | …and the chain shows |
| --- | --- | --- |
| single-output gasless `send_funds` settles | `B8XnHLauQt2V2ehe4pWGTk6RXc4vKA8CLE9FNoADW3NJ` | `gasPrice = 0`, `gasPayment = []`, budget 0 |
| a two-output split in one gasless PTB | `7JR6LzfYqupyX57kuW1eovGg8Pro6jNBGvSmDSYP2hAB` | payer −total, two exact credits |
| sub-`0.01` is NOT rejected (it's a protocol param, not a security boundary) | `G94L1nqGrzuMG3CQvsRKX3eosBuCFxSvzq7srwuLcVr7` | a below-min output committed, `gasPrice = 0` |
| keyless third-party broadcast (the facilitator never holds the payer key) | `ABhoH9CG4EbWQjfvQ8GoCciweNLLcfcpvgCDaXnc9P3A` | signed bytes broadcast from a separate keyless process |
| a stale signed tx expires (~2-epoch `ValidDuring` TTL) | `Cgv8wqkDzWru4mZEddWDNhi8174LTELHXACiYSMCr9SK` | replay after 84 s fails; TTL captured in the bytes |
| full E2E: 402 → gasless 2-output pay → verify(outputs exact) → settle → served | `5qDTpC88uAdMfetEUs1DEaV6973qeC8FPbSHJt2AqSHi` (self-built) · `4uPkvJjxfFqcDZzX6pFCebhBRPMHMFLkVsi9QA88gRey` (prebuilt bytes; feature removed from the spec in review round 1) | cheat rejected (`invalid_exact_sui_payload_outputs_mismatch`); replay idempotent (same digest); wire = v2 spec |

A live conformance facilitator (testnet) and a one-command repro accompany the reference-implementation PR; this spec PR is intentionally independent of it.

---

## AI assistance disclosure (per `specs/CONTRIBUTING.md`)

Drafting was AI-assisted (Claude Fable 5 + Opus 4.8) and then reviewed through multiple consensus rounds. Every protocol constraint cited — the allowlisted ops, `gasPrice = 0` / `gasPayment = []`, the `0.01` minimum, the `ValidDuring` TTL — is sourced from the Sui gasless docs and independently confirmed on live testnet; the digests above are real transactions with real USDC, not generated examples. Hallucinated constraints do not settle on-chain.


DrVelvetFog:
Strong call landing this as the spec's own stated follow-up rather than a side path. One thing for the Security Considerations section: the verification checklist here yields a settlement that's fully recomputable from chain (digest + per-output `balanceChanges` vs declared `extra.outputs`), and the gasless path has no gas-payment side effects to net out, so an independent verifier reads exactly the declared `amount`. That makes it a clean anchor for settlement-receipt binding — worth a forward-reference to #2666 so the two compose explicitly. (More detail plus a conformance-vector offer on the impl PR.)

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+    "transaction": "AAAIAQDi1HwjSnS6M..."
+  },
+  "extensions": {}
+}
+```
+
+## Gasless Construction (Address Balances)
+
+When the asset is an allowlisted stablecoin and the per-transfer amount meets the protocol
+minimum, the payer SHOULD construct a **gasless** transaction against its Address Balance.
```
phdargen:
```suggestion
minimum of 0.01, the payer SHOULD construct a **gasless** transaction against its Address Balance.
```
according to https://docs.sui.io/develop/transaction-payment/gasless-stablecoin-transfers

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+   **protocol parameter, NOT a security boundary**, and verification MUST NOT rely on it: an
+   empirical testnet probe settled a sub-`0.01` and a `1`-base-unit gasless transfer that the
+   chain did NOT reject (testnet digest `G94L1nqGrzuMG3CQvsRKX3eosBuCFxSvzq7srwuLcVr7`).
+   Verification anchors only on the exact `amount`/`outputs` (see Verification). Below the
+   minimum on an enforcing network, the payer MUST fall back to the classic `Coin<T>` path.
+6. **Natural expiry (TTL).** When `gasPayment` is empty the protocol stamps the transaction
+   with a `ValidDuring { minEpoch, maxEpoch }` window of approximately two epochs. This is the
+   transaction's natural expiry: a stale signed transaction ceases to be executable after the
+   window, and verification's simulation step fails for an expired transaction (testnet
+   staleness probe digest `Cgv8wqkDzWru4mZEddWDNhi8174LTELHXACiYSMCr9SK`).
```
phdargen: Empirical tests shouldn't be placed in a normative spec, please remove. You can add this info to the PR description instead

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+      { "to": "0x6fa6e5271e1d5eb02721930843a275aa1f9f4d6ba0dff88c5f8dcd367d9edceb", "amount": "49000" },
+      { "to": "0x369cf71b2f13b63c165c207bc3f263b1df9da754ca6a2d09300d9b4bf45f62f0", "amount": "1000" }
+    ]
   }
 }
 ```
 
+The matching gasless transaction is a single PTB with one `0x2::balance::send_funds<USDC>`
+command per declared output. A two-output gasless split settled on testnet with `gasPrice = 0`
+and `gasPayment = []` (digest `7JR6LzfYqupyX57kuW1eovGg8Pro6jNBGvSmDSYP2hAB`).
```
phdargen: please remove
```suggestion
The matching gasless transaction is a single PTB with one `0x2::balance::send_funds<USDC>`
command per declared output.
```

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+   sign bytes whose decoded outputs, sender, gas fields, or simulated balance changes do not
+   match the agreed `PaymentRequirements`.
+3. The client signs the verified bytes and submits the resulting `PaymentPayload` exactly as
+   on the self-built path.
+
+`extra.buildUrl` is purely a convenience: a client that ignores it and self-builds produces an
+identical `PaymentPayload`. The facilitator never holds the payer key, and the client's own
+pre-sign verification — not trust in the facilitator — is what binds the signature to the
+agreed terms. A prebuilt-path payment settled on testnet (digest
+`4uPkvJjxfFqcDZzX6pFCebhBRPMHMFLkVsi9QA88gRey`).
```
phdargen:
```suggestion
agreed terms. 
```

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+[VerifyResponse](../../x402-specification-v2.md#54-verifyresponse-schema).
 
 ## Settlement
 
-Settlement is performed via the facilitator broadcasting the transaction, along with the client's signature, to the network for execution.
+Settlement is performed via the facilitator broadcasting the signed transaction to the network
+for execution. On the gasless path the broadcast is **keyless**: the facilitator holds no payer
+key and only relays the already-signed bytes — a third party can broadcast the payer's signed
+transaction without any additional signature (testnet keyless-broadcast digest
+`ABhoH9CG4EbWQjfvQ8GoCciweNLLcfcpvgCDaXnc9P3A`).
```
phdargen:
```suggestion
transaction without any additional signature.
```

specs/schemes/exact/scheme_exact_sui.md
phdargen: sequencing diagram is outdated and should be updated or removed

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -2,7 +2,26 @@
...
+
+- **Classic `Coin<T>`** — the payer forms a complete signed transaction moving a
+  `0x2::coin::Coin<T>` of the exact amount. RETAINED unchanged for back-compat and for amounts
+  below the gasless minimum.
+- **Gasless (Address Balances)** — the payer forms a transaction composed exclusively of
+  protocol-allowlisted stablecoin operations against its Address Balance, with
+  `gasPayment = []` and `gasPrice = 0`. This eliminates the coin-object storage cost and the
+  interactive gas-station round trip, riding Sui's
+  [gasless stablecoin transfers](https://docs.sui.io/develop/transaction-payment/gasless-stablecoin-transfers)
+  (protocol v125).
```
phdargen: I think these should be treated as 2 distinct assetTransferMethods similar to EVM: https://github.com/Sceat/x402/blob/a4ea5a3ff1fc68c8069689e3fb0dad855d1e166c/specs/schemes/exact/scheme_exact_evm.md

Please add `extra.assetTransferMethod` with two values and clearly separate the facilitator verify rules for each

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
 ## Appendix
 
-### Sponsored Transactions
+### Sponsored Transactions (classic `Coin<T>` path)
 
-Sui supports sponsored or gas-less transaction via an interactive transaction construction protocol with a gas station. If a facilitator supports sponsoring of transactions then it should communicate this to the client by providing a URL via the `PaymentRequirements.extra.gasStation` field. If a client wants to make use of transaction sponsorship, then flow will be as follows:
+Sui supports sponsored transactions via an interactive transaction-construction protocol with a
+gas station. This appendix applies to the **classic `Coin<T>` path only**; the gasless path
+needs no sponsor. If a facilitator supports sponsoring of `Coin<T>` transactions it SHOULD
+communicate this via the `PaymentRequirements.extra.gasStation` field. If a client wants
```
phdargen: `PaymentRequirements` is a server-defined, not facilitator defined. The facilitator can announce this in supported/

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+   signature over the transaction before broadcasting.
 
 ### Future Work
 
-One inefficiency in the above described spec is that the gas cost for such a payment is slightly elevated due to the need to pay for the storage cost of the newly created coin object that is sent to the resource server. The resource server will be able to get a bit more of a payment (in the case where the client pays gas) by smashing the received coin into an already existing coin it may have, recouping a majority of the storage fee.
-
-The in-development feature "Address Balances" will reduce the overall gas cost for micro-payment transaction by eliminating the need to create a coin object to send, and thus eliminate the extra storage gas cost needed. This feature will also enable the sponsorship protocol to no longer need to be interactive due to no longer needing to perform coin selection.
+The gasless Address Balances construction (above) supersedes the previously-noted coin-object
+storage-cost inefficiency and the interactive sponsorship round trip for allowlisted
+stablecoins; both are resolved for that path.
```
phdargen: can be removed

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +78,305 @@ Full `PaymentPayload` object:
...
+  authorization act, so a malicious or buggy build endpoint cannot move funds the client did
+  not approve.
+- **Minimum-amount parameter.** The `0.01` gasless minimum is a protocol parameter and MUST NOT
+  be treated as a verification invariant; verification anchors on exact amounts regardless of
+  whether a given network enforces the minimum.
 
 ## Appendix
 
-### Sponsored Transactions
+### Sponsored Transactions (classic `Coin<T>` path)
```
phdargen: Sponsored Transactions shouldnt be put in appendix, but be documented as standard flow for the respective `assetTransferMethod`

specs/schemes/exact/scheme_exact_sui.md
phdargen: `extra.buildUrl` should be removed. Clients shouldn't communicate with the facilitator directly, nor do we want to add additional facilitator endpoints besides: verify/ settle/ and supported/

DrVelvetFog:
Following @phdargen's review — the `extra.assetTransferMethod` split is the right structure, and it matches how the two flows behave in practice from the facilitator side: gasless Address-Balance and sponsored are genuinely different verify surfaces (different gas-payment shape, different signature envelope over the transaction, different failure modes), even though they settle to the same recomputable result.

Since the sponsored flow is being promoted from appendix to a first-class method: I operate the live Sui facilitator from #2619, which runs the sponsored path (Enoki gas station) in production and settled the gasless cross-check earlier in this thread unmodified (`FVejSg9ddPYXwWtxjk58TjkZg6aawJFyynZEsTmsxRQ6`). Happy to draft or co-draft the sponsored `assetTransferMethod` section and its facilitator verify rules with @Sceat — including the per-method verify separation requested — written against what a live implementation actually enforces rather than from first principles. Say the word and I'll put text together against your branch.

Sceat:
@phdargen thanks for the thorough pass. All 11 points addressed in 29f97a98:

- Restructured around `extra.assetTransferMethod` with two values, `address-balance` (gasless) and `coin` (classic), mirroring the EVM scheme layout. Each method has its own construction and verification section, and the facilitator derives the effective method from the decoded bytes (empty gas payment means `address-balance`), rejecting any contradiction with the declared or echoed method.
- Sponsored transactions moved out of the appendix and documented as the standard sponsorship flow of the `coin` method. Capability is announced by the facilitator in `supported/`; the resource server relays the gas station URL via `extra.gasStation`, so requirements stay server-defined.
- `extra.buildUrl` removed entirely, along with its security considerations bullet.
- All empirical digests removed from the normative text; they live in the PR description, which stays the home for that evidence.
- Outdated sequencing diagram removed (the PNG was orphaned, deleted in the same commit); the numbered flow is rewritten per method.
- Wording suggestions applied as given (0.01 minimum, trimmed sentences, stale future work paragraph dropped).

Method values are `address-balance` and `coin`; happy to rename if you prefer a different casing convention.

---
@DrVelvetFog appreciated, and yes. The branch now has the split: `address-balance` and `coin` are separate methods with their own verification rules, and the sponsored flow is documented as the standard flow of the `coin` method (announced via `supported/`, relayed by the server through `extra.gasStation`). That section is currently written from the merged spec plus first principles, which is exactly the part that deserves production grounding. A PR against `Sceat/x402:spec/sui-address-balances` or suggestion blocks here both work, whichever is less friction. Credit stays in the commit trail.

DrVelvetFog:
Took the pen on the sponsored `coin` section — grounded against the live facilitator from #2619, whose gas station has been running the Enoki-sponsored path in production. Two refinements from what the implementation actually enforces; both loosen the spec rather than tighten it.

**1. The sponsor signature needn't be the facilitator's own.** The text has the facilitator "provide its own signature over the transaction before broadcasting." In practice the sponsor key can live in the gas station the facilitator delegates to — in my case the facilitator holds *no* key: it relays the client-signed bytes to the station (`{digest, signature}`), which adds the sponsor signature over the same bytes and broadcasts. That preserves the `address-balance` method's non-custodial property across `coin`. Suggest: *the sponsor signature MUST come from the owner of the gas objects referenced in the gas payment — whether that's the facilitator or a gas station it delegates to; the facilitator MUST only co-sign, or relay for co-signing, gas payments under its control.*

**2. Gas-station recipient allow-listing (interop gotcha).** A gas station may guard which addresses a sponsored tx can touch (Enoki's `allowedAddresses`) and reject the sponsorship if a payee isn't listed. So the sponsored `coin` construction MUST include every credited recipient — `payTo` and any `extra.outputs[].to` — in the sponsorship request, not just the payer.

The shared verify surface — signer-equals-sender, gas-shape derived from the decoded bytes, effects-based balance-change assertion — matches what I enforce as written; the one structural difference is #1 (sponsor signature + broadcast delegated to the gas station). Happy to drop both in as inline suggestion blocks against the current lines if the wording lands.

lucky-ivanius:
**Proposal: use Address Balance *gas* sponsorship for the sponsored flow — no gas coin objects, no Enoki or gas station**

First, to disambiguate two uses of "Address Balance" in this thread: there's the `address-balance` *transfer method* this PR adds (truly gasless, `gasPrice = 0`), and there's **Address Balance *gas* sponsorship** — a normal transaction (`gasPrice > 0`) that draws its gas from the *sponsor's* SUI address balance instead of from gas coin objects ([Mysten SDK docs → Address balance sponsorship](https://sdk.mystenlabs.com/sui/transactions/signing-and-execution#address-balance-sponsorship)). This comment proposes the second one as the sponsored flow method, replacing the interactive gas-station protocol as the default.

We run a Sui facilitator in production that sponsors gas exactly this way (links below), and the gas-station flow has structural problems from the facilitator side:

1. **It's interactive by construction.** Gas object refs are part of the signed bytes, so the sponsor must reserve and communicate a gas object *before* the client can sign — the SDK docs call this out verbatim: *"The user must wait for the sponsor to set gas coins before signing, because gas coins are part of the signed transaction data."* That forced round trip is the only reason `extra.gasStation` exists.
2. **It doesn't scale without coin-pool management.** Each in-flight sponsorship pins one owned gas object until the transaction executes or expires, and double-assigning a coin risks equivocation. So the station must maintain a pre-split pool sized to peak concurrency and continuously split/merge/rotate coins. Note the `coin` method is exactly the *micropayment* path (sub-`0.01` amounts, non-allowlisted assets) — the highest-throughput case is the one being made interactive.
3. **It adds a client-facing endpoint outside `verify`/`settle`/`supported`.** The same principle that removed `extra.buildUrl` in round 1 applies to `extra.gasStation`.

**Address Balance gas sponsorship removes all three.** The client builds the *complete* transaction itself — `setGasOwner(<facilitator sponsor address>)`, `setGasPayment([])` (empty = gas drawn from the gas owner's address balance), budget within an advertised cap — signs it, and sends the standard `{ signature, transaction }` payload. The facilitator verifies the decoded bytes and adds its sponsor signature **asynchronously at settle**. No gas objects anywhere, no pool, no station, no new endpoint, and the facilitator never touches a payer key.

The one trade-off, stated plainly: the facilitator must hold its sponsor SUI as an **Address Balance** instead of as coin objects — a one-time conversion of funds it already self-custodies (e.g. `0x2::coin::send_funds` to itself). In exchange, that is the *entire* operational footprint: **no Enoki or any other gas-station service is required to run a sponsoring facilitator** — a funded Sui address is enough.

This is exactly the `exact`-on-SVM model: facilitator announces `extra.feePayer`, client sets it in the tx and signs, sponsor signs last at settle. The Sui equivalent: facilitator announces `gasOwner` (+ `gasBudget`) in `supported/`, and the resource server relays them via `PaymentRequirements.extra` — the same relay mechanism this branch already defines for `gasStation`.

Client construction:

```ts
const tx = new Transaction();
tx.setSender(payerAddress);
tx.setGasOwner(requirements.extra.gasOwner); // facilitator's sponsor address
tx.setGasPayment([]);                        // empty = gas from the gas owner's address balance
tx.setGasBudget(requirements.extra.gasBudget);
// ...exact-amount transfer commands (with useGasCoin: false, since tx.gas belongs to the sponsor)...
const bytes = await tx.build({ client });
const { signature } = await payerKeypair.signTransaction(bytes);
// -> PaymentPayload.payload = { signature, transaction: toBase64(bytes) }
```

Facilitator at settle, after re-running verification:

```ts
const sponsorSignature = await sponsorKeypair.signTransaction(txBytes);
await client.executeTransaction({ transaction: txBytes, signatures: [clientSignature, sponsorSignature] });
```

**Spec impact — and a derivation bug this exposes.** The branch currently derives the method as "`gasPayment == []` is `address-balance`, anything else is `coin`", and `coin` verification asserts `gasPayment` is non-empty. An Address-Balance-*sponsored* transaction has `gasPayment == []` **and** `gasPrice > 0`, so as written it is routed to `address-balance` verification and rejected at the `gasPrice == 0` assertion — the flow is unrepresentable in the current text even though it's a first-class protocol feature. The derivation needs to be gasPrice-aware:

- `gasPayment == [] ∧ gasPrice == 0` → `address-balance` (gasless transfer)
- `gasPayment == [] ∧ gasPrice > 0 ∧ gasOwner ≠ sender` → `coin`, sponsored (gas from the sponsor's address balance)
- `gasPayment != []` → `coin`, self-paid gas (or coin-object sponsorship, if retained)

Sponsored verification then mirrors the SVM sponsor checklist: decoded `gasOwner` MUST be one of the facilitator's announced sponsor addresses and MUST NOT equal `sender`; decoded `gasBudget` MUST NOT exceed the announced cap (the facilitator's per-transaction exposure is bounded by `gasBudget`); everything else — signer-equals-sender, digest replay guard, simulation, effects-based output verification — is unchanged. And since `gasPayment` is empty, the sponsored path inherits the same ~2-epoch `ValidDuring` natural expiry the branch already documents.

This also subsumes both refinements from the gas-station discussion above: the facilitator co-signs only over gas it controls (no station key delegation to specify), and recipient allow-listing becomes the facilitator's own verify-time policy rather than a station-config interop concern.

The implementation predates the `assetTransferMethod` split, which I agree is the right structure — this proposal slots in purely as gas sponsored flow.

Concretely, I'd suggest: make Address Balance gas sponsorship the RECOMMENDED sponsored flow (non-interactive, consistent with the SVM scheme), replace `extra.gasStation` with `extra.gasOwner`+`extra.gasBudget`, and keep the interactive gas-station flow as an OPTIONAL alternative for facilitators already operating one. Happy to draft this as suggestion blocks against `spec/sui-address-balances`.

---

**Production evidence.** This flow is running in production today. Our facilitator at **https://facilitator.tentaclepay.com/** sponsors gas from an Address Balance on `sui:mainnet` and `sui:testnet`. It settles **any Sui `Coin<T>`**, not only the protocol-allowlisted gasless stablecoins, and amounts below the `0.01` gasless minimum are simply sponsored — exactly the cases the `extra.assetTransferMethod = coin` exists for. Its `/supported` response already announces the proposed fields — `curl https://facilitator.tentaclepay.com/supported` returns `extra: { gasOwner, gasBudget }` on each network kind. All of it is open source:

- **Scheme implementation**: [tentaclepay/sui-x402](https://github.com/tentaclepay/sui-x402) — `getExtra()` exposes `gasOwner`/`gasBudget`, `settle()` co-signs asynchronously ([scheme.ts](https://github.com/tentaclepay/sui-x402/blob/main/typescript/src/exact/facilitator/scheme.ts))
- **Facilitator service**: [tentaclepay/apps → packages/facilitator](https://github.com/tentaclepay/apps/tree/main/packages/facilitator)
- **Full client + resource-server example** against the live facilitator: [tentaclepay/apps → examples/x402/exact](https://github.com/tentaclepay/apps/tree/main/examples/x402/exact)

Settled `sui:mainnet` examples, one per case:

| Case | Mainnet digest | Gas shape on chain |
| --- | --- | --- |
| `address-balance` (gasless) | [`3jaDVwegj5t85VKAyT1w9fkmBdrq98LbadyuDhgSEVkJ`](https://suivision.xyz/txblock/3jaDVwegj5t85VKAyT1w9fkmBdrq98LbadyuDhgSEVkJ) | `gasPayment = []`, `gasPrice = 0`, budget `0` |
| Sponsored — allowlisted asset **below the `0.01` minimum** (0.005 USDC) | [`65DptF4YMETYsnTzHZHMvhwg5BUKwfY1MU4e7pdBnTpe`](https://suivision.xyz/txblock/65DptF4YMETYsnTzHZHMvhwg5BUKwfY1MU4e7pdBnTpe) | `gasPayment = []`, `gasPrice = 100`, gas debited from the sponsor's SUI address balance |
| Sponsored — **native SUI** payment | [`G8ifkvbsNvEffLBgB59ZRbpc5f2Bv9XwT48BAhzbr8KY`](https://suivision.xyz/txblock/G8ifkvbsNvEffLBgB59ZRbpc5f2Bv9XwT48BAhzbr8KY) | `gasPayment = []`, `gasPrice = 100`, non-stablecoin `Coin<T>` |

The two sponsored digests are exactly the shape the current derivation rule cannot represent — `gasPayment == []` with `gasPrice > 0` — settled on mainnet, with gas drawn from the Address Balance of the same `gasOwner` the live `/supported` announces.

lucky-ivanius:
One thing worth putting on the table first: **neither method constrains where the payer's assets live.** A payer holding only `Coin<T>` objects can pay via `0x2::balance::send_funds` — the SDK can resolves the source automatically. The real discriminator is the transaction's **gas shape**: `gasPrice = 0` ∧ `gasPayment = []` versus gas-carrying.

So `extra.assetTransferMethod = address-balance` reads as an asset-location requirement that doesn't exist, and overloads a feature name both methods touch. `gasless` names the actual verifiable property (exactly what derivation decodes), matches the feature's own title.

Raising it now while the values are new in this PR, a rename is free.

DrVelvetFog:
Strong proposal, and I think it's the right default. Address Balance gas sponsorship is non-interactive and mirrors the SVM `feePayer` model — a cleaner fit than the interactive station. And the derivation-bug catch is real: on the branch, `gasPayment == []` routes to `address-balance`, whose gas-shape check asserts `gasPrice == 0`, so an Address-Balance-*sponsored* tx (`gasPayment == []` ∧ `gasPrice > 0`) is rejected before it can reach `coin`. The gasPrice-aware derivation you laid out fixes it cleanly, and your live `/supported` already announces `gasOwner`/`gasBudget` on both networks — so it's proven, not just proposed.

One operator's-eye trade-off worth keeping on the record, from running the Enoki-sponsored path in production: what a managed gas station actually buys you isn't the gas mechanism — it's the guardrails around it (per-address/day sponsorship caps, allowed move-call targets, allowed recipients). With raw Address Balance sponsorship those don't disappear, they move into the facilitator's own verify-time policy over the decoded bytes. That's a fine trade — arguably better, since it's explicit in the verification rules instead of hidden in a station's config — but it's real work a sponsoring facilitator takes on. Which is exactly why your "recommended default + keep the interactive station as an optional alternative" split is the right call: facilitators already operating a managed station get the guardrails for free, everyone else gets the simpler non-interactive path.

+1 on drafting it as suggestion blocks against the branch.

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +90,322 @@ Full `PaymentPayload` object:
...
+`PaymentRequirements` is defined by the resource server — not the facilitator — the resource
+server relays the gas-station URL to clients via `PaymentRequirements.extra.gasStation`. If a
+client wants sponsorship, the flow is:
+
+1. Client makes a request and gets a payment-required response from the service.
+2. Client constructs a partial transaction (without gas payment) to pay for the request based
+   on the provided `PaymentRequirements`.
+3. Client sends the partial transaction to the gas station at
+   `PaymentRequirements.extra.gasStation`. The gas station fills in the necessary gas
+   information (gas objects, budget, etc.) and sends back a fully formed transaction.
```
phdargen: Similar to my` extra.buildUrl` comment earlier, clients shouldnt interact with the facilitator (and more fundamentally a server provided untrusted url) directly but build the tx themselves

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +90,322 @@ Full `PaymentPayload` object:
...
+sponsorship flow below.
+
+### Sponsored transactions
+
+Sui supports sponsored transactions via an interactive transaction-construction protocol with
+a gas station. A facilitator that sponsors `coin` payments announces this capability in its
+`/supported` response (in the `extra` of the corresponding scheme/network kind). Because
+`PaymentRequirements` is defined by the resource server — not the facilitator — the resource
+server relays the gas-station URL to clients via `PaymentRequirements.extra.gasStation`. If a
+client wants sponsorship, the flow is:
```
phdargen: Its save to assume a client does want sponsorship. Thus this should not be a client choice, either the facilitator offers sponsorship, then its sponsored or it doesn't

phdargen:
Thanks a lot for the input @lucky-ivanius, I'm aligned with the direction: address balance gas sponsorship as the sponsored flow, mirroring the SVM `feePayer` model.  A few concrete refinements:

1. Align notation with SVM and others: `gasOwner` → `feePayer`. Same role as SVM's `extra.feePayer `(the account that pays fees and provides the final signature), so let's reuse that name for cross-scheme consistency

2. gasBudget is client-set within an announced cap, not facilitator-mutable. On Sui the budget is part of the signed BCS bytes, so the facilitator can't set/raise it after the client signs (it co-signs identical bytes). So: facilitator announces a ceiling in /supported (suggest maxGasBudget for explicit semantics) server mirrors it in 402, client sets a budget ≤ cap, facilitator verifies gasBudget ≤ maxGasBudget at settle (which also bounds its exposure)

3. +1 on address-balance → gasless rename

4. `feePayer` presence is the sponsorship discriminator. Announcing `feePayer` signals sponsorship is offered (SVM-style), and it also names the gas owner the client must set. The three cases then derive cleanly from requirements + decoded bytes:
- gasless → no feePayer; gasPayment == [] ∧ gasPrice == 0
- coin, sponsored → feePayer present; gasPayment == [] ∧ gasPrice > 0 ∧ gasOwner == feePayer ≠ sender
- coin, client-paid → no feePayer; gasPayment != []
Sponsored verify then mirrors the SVM sponsor checklist: decoded gasOwner MUST equal the announced feePayer and MUST NOT equal sender; gasBudget ≤ maxGasBudget

5. I'd drop the interactive gas-station flow entirely rather than keep it as an optional alternative. It reintroduces exactly the client↔untrusted-URL round trip we removed with extra.buildUrl, and maintaining two sponsorship mechanisms in one scheme isn't worth it

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +90,322 @@ Full `PaymentPayload` object:
...
+  "scheme": "exact",
+  "network": "sui:mainnet",
+  "amount": "50000",
+  "asset": "0xdba34672e30cb065b1f93e3ab55318768fd6fef66c15942c9f7cb846e2f900e7::usdc::USDC",
+  "payTo": "0x6fa6e5271e1d5eb02721930843a275aa1f9f4d6ba0dff88c5f8dcd367d9edceb",
+  "maxTimeoutSeconds": 60,
+  "extra": {
+    "outputs": [
+      { "to": "0x6fa6e5271e1d5eb02721930843a275aa1f9f4d6ba0dff88c5f8dcd367d9edceb", "amount": "49000" },
+      { "to": "0x369cf71b2f13b63c165c207bc3f263b1df9da754ca6a2d09300d9b4bf45f62f0", "amount": "1000" }
```
phdargen: Suggest to use a more network agnostic notation: `extra.splits` witth entries `{ payTo, amount } `, reusing the top-level payTo/amount field names so a split has the exact same shape as the core payment. 

The top-level `payTo` should remain the primary receiver, no duplication in `extra`. `Splits` lists optional additional receivers, the primary `payTo` implicitly receives the remainder: `amount − Σsplits.amount`

```suggestion
  "payTo": "0x6fa6e5271e1d5eb02721930843a275aa1f9f4d6ba0dff88c5f8dcd367d9edceb",
  "maxTimeoutSeconds": 60,
  "extra": {
    "splits": [
      { "payTo": "0x369cf71b2f13b63c165c207bc3f263b1df9da754ca6a2d09300d9b4bf45f62f0", "amount": "1000" }
```

Verification rules: 
- Σsplits.amount < amount (strict, so payTo always keeps a non-zero primary share); reject otherwise 
- each splits[].payTo distinct, ≠ primary payTo, amount > 0.

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +90,322 @@ Full `PaymentPayload` object:
...
+    ]
+  }
+}
+```
+
+The matching gasless transaction is a single PTB with one `0x2::balance::send_funds<USDC>`
```
phdargen, line 388: Is there any limit on the amount of splits the gasless stablecoin transfer accepts in a single tx? 
I suppose at least each individual transfer has to be above the protocol minimum of 0.01?

specs/schemes/exact/scheme_exact_sui.md
```diff
@@ -51,53 +90,322 @@ Full `PaymentPayload` object:
...
+
+Verify the outputs of the simulation against the agreed recipients, reading the asset's
+`balanceChanges`:
+
+- If `extra.outputs` is present: each declared recipient MUST be credited EXACTLY its
+  declared amount; the payer MUST be debited EXACTLY `amount` (= the sum of outputs); and
+  there MUST be no undeclared recipient of the asset. Only AddressOwner credits count;
+  ObjectOwner/Shared credits MUST be rejected. A mismatch MUST be rejected
+  (`invalid_exact_sui_payload_outputs_mismatch`).
+- If `extra.outputs` is absent: the resource server's address (`payTo`) MUST see a balance
```
phdargen: For the sponsored flow we should also restrict the number of splits the facilitator accepts

Sceat:
@phdargen all five landed in 024a532b:

1. **`feePayer` it is.** `gasOwner`/`gasBudget` as announcement fields are gone; the facilitator announces `extra.feePayer` + `extra.maxGasBudget` in `/supported`, the resource server relays them verbatim, and `feePayer` presence is the sponsorship signal, mirroring `exact` on SVM.
2. **Budget semantics as you stated.** The client sets `gasBudget` within the announced `maxGasBudget`; verification asserts `gasBudget <= maxGasBudget` and settle re-asserts it. The construction section spells out why the ceiling is the exposure bound: the budget is inside the signed BCS bytes, so the facilitator co-signs identical bytes and cannot set or raise it.
3. **`address-balance` -> `gasless` everywhere**, with @lucky-ivanius's rationale folded into the Summary: neither method constrains where the payer's assets live, the gas shape is the discriminator, so the method name names the gas property.
4. **Three-case derivation is normative** in the Summary and the `coin` gas-shape check: `gasless` (`[] ∧ price == 0`), sponsored (`[] ∧ price > 0`, valid only under an announced `feePayer`, decoded `gasOwner == feePayer != sender`, `gasBudget <= maxGasBudget`), client-paid (non-empty). The sponsored shape no longer falls through to the `gasless` assertions, closing the derivation bug.
5. **Interactive gas-station flow dropped entirely**; `extra.gasStation` is gone. @DrVelvetFog's two station refinements survive in post-station form: "only co-sign gas you control" became "decoded gas owner MUST equal the announced `feePayer`", and the guardrails a managed station provided as config are documented as the facilitator's own verify-time policy over the decoded bytes (Security Considerations), per his operator note.

Two consequences the rework surfaced, both now specified: a sponsored transaction whose asset also comes from an Address Balance has **no object inputs**, so re-simulation is not a replay guard there either; the digest lookup is documented as the uniform guard across all shapes. And since sponsored bytes carry an empty `gasPayment`, they inherit the same ~2-epoch `ValidDuring` natural expiry as `gasless`.

@lucky-ivanius the tentaclepay digests are exactly the fixture set for the sponsored path, thank you for grounding this with mainnet evidence. #2616 picks up the rename and the sponsored flow next.

phdargen:
Thanks @Sceat, the address balance sponsorship/gas-station rework looks good!

Leaves the following open from my last review (see inline comments above):
- Rename extra.outputs → extra.splits with { payTo, amount }, primary payTo keeping the remainder (amount − Σ splits) rather than duplicating it in the array
- Clarify gasless multi-recipient bounds: does each leg need ≥ 0.01, and is there a max number of legs per tx?
- Cap how many splits a sponsoring facilitator will accept

Sceat:
@phdargen all three landed in e4c91068:

1. **`extra.outputs` → `extra.splits`**, your semantics exactly: entries reuse the top-level `payTo`/`amount` names so a split has the same shape as the core payment; the primary `payTo` is never duplicated in `extra` and implicitly receives the remainder `amount − Σ splits[].amount`; `Σ splits[].amount < amount` strict, split `payTo`s distinct and ≠ primary, amounts > 0. A side effect worth noting: the old "outputs must sum to `amount`" failure mode disappears structurally, since the remainder is derived rather than declared.
2. **Gasless bounds:** the `0.01` minimum applies **per transfer**, so with splits every credited leg, the primary remainder included, must individually meet it; below that the payment falls to `coin`. Method-selection language throughout now carries the per-leg qualifier. On a count limit: there is no dedicated split-count parameter; the effective ceilings on a gasless transaction are its serialized-size cap (`get_gasless_max_tx_size_bytes`), its computation cap (`gasless_max_computation_units` × reference gas price), and the generic PTB command cap.
3. **`extra.maxSplits`**: optional announced cap (non-negative decimal integer string; `"0"` = no splits), relayed like the other fields and enforced when announced. It caps routing fan-out and per-payment verification work; the hard monetary exposure bound for a sponsor remains `maxGasBudget`.

The same pass tightened five things that review of the Sui sources surfaced, flagging them here rather than leaving them to be discovered:

- **The gasless MoveCall allowlist had grown past the text.** It is nine functions in current node releases: `0x2::balance::{send_funds, redeem_funds, split, zero}`, `0x2::funds_accumulator::withdrawal_split`, `0x2::coin::{send_funds, redeem_funds, into_balance, put}` (sui-types `GASLESS_FUNCTIONS`; allowed commands remain those MoveCalls plus `SplitCoins`/`MergeCoins`). The spec now describes the set as release-versioned and lists the current members.
- **Gasless shape requires `gasBudget == 0`**, enforced at signing ("gas_budget must be 0 for gasless transactions"); the derivation, construction, and verification now assert it alongside `gasPayment == []` and `gasPrice == 0`.
- **Non-sponsored shapes must have gas owner == `sender`**: a distinct gas owner requires a sponsor co-signature, which only the sponsored shape carries; the derivation and verification now say so.
- **Simulation does not enforce `TransactionExpiration`** — the node's input-check path is not epoch-aware; enforcement lives in the signing path only. The earlier text claimed a TTL-expired transaction fails simulation, which is wrong: it simulates cleanly and dies at broadcast, so a serve-before-settle facilitator could verify a payment that can no longer settle. Both methods now carry an explicit expiry step mirroring the signing path's full validity (epoch window, chain identifier, rejection of the unsupported timestamp fields). Credit: @Danny-Devs flagged the simulation gap on #2616; it checks out against sui-types/sui-transaction-checks.
- **Executed-first settlement must bind to the current requirements**: the idempotent success path now requires validating the committed transaction's effects (signature, payer, Recipient Verification) against the requirements being settled, since committed bytes and signatures are public and could otherwise be replayed as payment for unrelated requirements.

Also corrected in passing: the splits example now uses legs that meet the per-transfer minimum (`10000` units = 0.01 USDC), and the security-considerations language now states the exact-recipients guarantee precisely (it holds when `extra.splits` is declared; the single-recipient default asserts the `payTo` credit, with the stricter payer-debit/no-undeclared checks as a SHOULD for compatibility with the merged spec).

phdargen:
> * **Executed-first settlement must bind to the current requirements**: the idempotent success path now requires validating the committed transaction's effects (signature, payer, Recipient Verification) against the requirements being settled, since committed bytes and signatures are public and could otherwise be replayed as payment for unrelated requirements.

Settle’s executed-first “idempotent success” conflicts with core x402: success: true means a new valid payment and the server releases the resource. Otherwise a client could replay an already-committed payment (matching payTo/amount) to get another settle success and regain access without paying again. Idempotent retries belong in the [payment-identifier extension](https://github.com/x402-foundation/x402/blob/main/specs/extensions/payment_identifier.md), not at scheme level. Settle should reject instead, same as verify

phdargen:
Please also check Sui’s RPC behavior for concurrent same-digest calls to see if we need a SVM-style duplicate settlement mitigation (duplicate_settlement cache)

https://github.com/x402-foundation/x402/blob/67b1ba0a7abbd7907a28fa624670872532e0eae9/specs/schemes/exact/scheme_exact_svm.md?plain=1#L347-L362

Sceat:
@phdargen Both right, and the first is a clear correction to the round-3 text.

**Executed-first settle should reject, not return idempotent success.** You are right that at the scheme level this conflicts with what settle success means. If success releases the resource, a client can re-settle an already-committed payment whose payTo and amount still match the requirements and get a second release for free, and the recipient check does not stop that because it is the same payment. Settle will reject an already-committed digest, exactly as verify does. The mid-flight case the executed-first path was trying to serve, a settle that landed on chain but whose response was lost, belongs in the `payment-identifier` extension: the `id` gives the resource server and the facilitator an idempotency key to return a cached response rather than a second settle. I will pull the "return the original digest as success" language and replace it with a reject, and point idempotency at the extension.

**Concurrent duplicate settlement: yes, Sui needs the same mitigation.** Sui executes a transaction at most once, consensus dedups by digest and a validator returns `TransactionAlreadyExecuted` for a finalized transaction rather than re-running it, so the payment can never double-spend. But that is precisely the SVM situation you linked: the on-chain effect happens once while concurrent `/settle` calls for the same signed bytes, fired before the first commits, can each observe success, so one payment buys several resources. The `duplicate_settlement` in-memory cache closes the single-process case the same way it does on SVM, and I will add a Duplicate Settlement section mirroring the SVM one.

One thing worth making explicit in that section rather than leaving implicit: a facilitator running more than one replica cannot close the race with a per-process cache alone, since two concurrent calls may land on different replicas. For that deployment the `payment-identifier` extension is the complete answer, because it deduplicates at the resource server where the release decision is actually made. The in-memory cache stays the RECOMMENDED single-process mitigation; the extension is what a horizontally scaled facilitator leans on.

I will fold both into the next spec revision, together with the version numbers (gasless is protocol v119 on testnet and devnet, v124 on mainnet) and the "enforced in dry-run since v1.74.0 (#26794), MUST NOT rely on the node version" reword from the thread above.

phdargen:
> One thing worth making explicit in that section rather than leaving implicit: a facilitator running more than one replica cannot close the race with a per-process cache alone, since two concurrent calls may land on different replicas. For that deployment the `payment-identifier` extension is the complete answer, because it deduplicates at the resource server where the release decision is actually made. The in-memory cache stays the RECOMMENDED single-process mitigation; the extension is what a horizontally scaled facilitator leans on.

Please use similar wording as XRP here:
https://github.com/x402-foundation/x402/blob/67b1ba0a7abbd7907a28fa624670872532e0eae9/specs/schemes/exact/scheme_exact_xrpl.md?plain=1#L473

How horizontally scaled facilitators implement the cache is an implementation detail, should not point here to the `payment-identifier` extension

Sceat:
Correcting my previous comment: several of its Sui-internals claims asserted more than I had verified. The narrower, accurate position:

**What the source I cited (`authority.rs`) actually establishes.** `handle_vote_transaction` accepts a current-epoch finalized transaction (returns `Ok`) and returns `TransactionAlreadyExecuted` only for a last-epoch execution; and execution serializes same-digest transactions with a per-transaction guard (`acquire_tx_guard`) and reuses cached effects when the digest already executed. That is the guard-and-cache behavior at that path, nothing more.

**What I over-asserted and should not have.** "Consensus dedups by digest" (I did not verify the consensus path; the guard/cache is what I can cite). "A validator returns `TransactionAlreadyExecuted` for a finalized transaction" (that holds only last-epoch). And "concurrent `/settle` calls can each observe success" (the source shows the internal guard and cached-effects reuse, not what the fullnode execute-RPC returns to each concurrent caller). Whether the concurrent-duplicate race exists therefore depends on that RPC behavior, which I have not verified on Sui, so the `duplicate_settlement` concern and its mitigation are conditional on it. I will verify it before the spec revision lands.

**On the multi-replica point.** `payment-identifier` deduplicates same-`id` retries, so a fresh `id` on the same signed transaction bypasses it and does not close a digest-level race. Fully closing that race across replicas would require shared atomic digest-level coordination; I am not asserting a specific design is sufficient here.

The spec direction (settle rejects committed digests; idempotency via `payment-identifier` for same-`id` retries) is unchanged. The dry-run-since-v1.74 point rests on the introducing change (#26794), cited from that PR rather than re-derived from the snapshot.

DrVelvetFog:
@Sceat @phdargen — you flagged that whether the concurrent-duplicate race exists depends on what the fullnode execute-RPC returns to each caller, and that it was unverified. I have a testnet facilitator and a harness handy, so I ran it — offering the data toward your verification, not getting ahead of it.

One transaction, signed once. Eight identical `ExecuteTransaction` gRPC calls fired concurrently (each on its own client so nothing dedups locally), then three more re-submitted after finality. Node `sui-node/1.76.0`, testnet.

Every one of the eleven returned `status: success` with the same digest — no throw, no `TransactionAlreadyExecuted`, no duplicate error. Ground truth: the payer balance dropped by one transaction's gas (~1.02M MIST), not eleven times, so all eleven "success" replies map to a single on-chain execution.

So at the RPC-observation layer the race is real: the fullnode hands back cached success effects to every concurrent and post-finality caller of the same signed bytes. A facilitator that treated "execute didn't error" as "this is the settlement that releases the resource" would release N times under concurrency. That lines up with the direction here — settle must reject an already-committed digest rather than return idempotent success, and dedup has to live at the release point (the `payment-identifier` path), not in execute's return value.

Scope, kept honest: this is one fullnode's RPC behavior on 1.76.0 testnet, which is the layer that was open — I'm not asserting anything about the consensus path. Harness is a ~120-line script against the standard SDK gRPC client; happy to share it or run variants (higher concurrency, USDC/gasless-sponsored shape) if useful.

Sceat:
@DrVelvetFog that's exactly the layer that was open, thank you for running it. Eleven success replies on one signed tx with a single gas debit is the clean result: in your run the fullnode returned identical success effects to every concurrent and post-finality caller, so a facilitator that reads "execute didn't error" as "this is the settlement that releases the resource" could release N times. That's the case for settle rejecting an already-committed digest rather than returning idempotent success. The one place I'd draw the line differently: the concurrent window (callers that arrive before the first commit lands) needs an atomic digest-level check-and-record, and `payment-identifier` only covers same-id retries, so it isn't sufficient for that race on its own across replicas.

One scope line before we call the RPC layer closed: your run used the standard gas-paying shape, which doesn't reach the gasless Address-Balance path this mechanism settles on (empty gas payment, zero gas price, a different validity path in the node). So it doesn't establish the gasless shape's RPC behavior, and since this is an RPC-response-layer fact (the chain shows one execution either way, observable only by being the concurrent caller) we'd rather verify it than carry the gas-paying result across. We're running our own concurrent harness on the gasless Address-Balance shape to read the N-success-versus-1-execution count directly. If you're up for pointing yours at that same shape, a second independent run would be welcome corroboration.

DrVelvetFog:
@Sceat ran it on the gasless shape — same result as the gas-paying one, so the race isn't specific to the gas-paying path.

Built the payment the way your client does: one `0x2::balance::send_funds` drawn via `tx.balance()`, `setGasBudget(0n)` to elect gaslessness (no manual gasPrice/gasPayment). Confirmed gasless before firing and again on-chain — `gasData.price = 0`, `payment = []`, payer SUI delta 0 (gas fully rebated). USDC on testnet, node 1.76.0.

One signed tx, 8 identical `ExecuteTransaction` calls concurrently (each on its own gRPC client), then 3 more after finality:

- all 11 returned `status: success`, same digest, no throw, no `TransactionAlreadyExecuted`
- ground truth from balances: payer −10000, seller +10000 USDC, exactly once — so 11 "success" replies map to one on-chain execution
- reproduced twice (digest `6N9DhoZt8LCpdrQzWB1Rfk4SdpmaX3QRSHhKCBqiBPdi`)

So the gasless Address-Balance path behaves the same as the gas-paying one at the RPC-observation layer: the fullnode hands cached success effects to every concurrent and post-finality caller of the same signed bytes. Same conclusion — a facilitator that reads "execute didn't error" as "this settlement releases the resource" releases N times, so settle has to reject an already-committed digest and dedup at the release point rather than trust execute's return value. The `duplicate_settlement` section applies to Sui's gasless path too.

Scope kept honest: one fullnode, 1.76.0 testnet, the USDC gasless Address-Balance shape — the RPC-response layer, which is the part that was open; not asserting anything about the consensus path. Same ~150-line probe pointed at the gasless build; happy to share it or run variants (higher concurrency, the SUI-asset shape).

hayes-mysten:
Hey, sorry for coming into this PR late. I left a series of comments on https://github.com/x402-foundation/x402/pull/2616 before realizing this PR is probably the more appropriate place for some of that discussion. I recently came across these PRs and initially planned to just review implementation details to make sure things aligned with our plans for the Sui TS SDK, but that led me to some broader questions that are better raised here.

Before getting into the weeds on the specifics of the implementation or the spec, I think it would be good to align on some top-level concepts:

## Address balances vs gasless transactions

Currently we're splitting into two assetTransferMethods. That may be okay, but I would strongly recommend not tying gasless and address balances together. Gasless transactions are only available for specific stablecoins and come with a lot of restrictions that hurt composability and flexibility.

My understanding of the intent behind assetTransferMethod is that it's about how the server receives its funds, since the server is the one declaring it in their requirements. Expecting address-balance payments makes sense; requiring that they be sent in a way that doesn't require gas seems overly specific.

## Transaction validation

Both the implementation and the spec seem very concerned with validating the exact shape of the transaction, which I think is fundamentally the wrong approach on Sui.

For validation, I would strongly recommend focusing on transaction effects (primarily balance changes) as the only thing validated for payments. This leaves a lot more flexibility for the client. Transactions on Sui are intended to be composable and multi-functional, and it would be much better if payments could be sourced arbitrarily (on-demand swaps, withdrawals from contracts with balance limits, etc.) rather than requiring a very specific gasless-eligible payment pattern.

This is another reason I think making "gasless" an assetTransferMethod is the wrong approach. If a client builds the transactions, many payments will naturally be gasless, but there's no reason to require this from a validation perspective. It could be a client setting (only sign transactions that don't require gas), but I wouldn't enforce it during validation.

## Coins vs balances

Sui is moving toward address balances. I can see an argument for a coin vs address-balance assetTransferMethod, but I'm not convinced the distinction is worthwhile. You can validate the effects and accept payment in either form.  SDKs, wallets, explorers, etc. all merge these together, and using them interchangeably is trivial. Unless there's a very good reason, it may be easier to keep things together and simplify validation down to "did I receive the funds I requested," without getting too specific about how the funds arrived.


I don't want to derail something that clearly has already had a lot of time and effort put into it, but I also want to make sure we are going down the right path here before this is too broadly adopted and it becomes harder to pivot.

I'd love to get some thoughts on these topics, and I don't want to be over prescriptive.  Happy to also help out with the current direction if these changes seem too far out of line with your vision for x402

DrVelvetFog:
@hayes-mysten on the replay-prevention piece — some measured behavior from earlier in this thread that might save you time ([gas-paying run](https://github.com/x402-foundation/x402/pull/2615#issuecomment-5033678419), [gasless run](https://github.com/x402-foundation/x402/pull/2615#issuecomment-5045035737)):

One signed tx, 8 concurrent ExecuteTransaction calls plus 3 re-submits after finality: all 11 returned success with the same digest, zero throws, on both the gas-paying and gasless shapes. Balance changes confirm exactly one execution. So "execute didn't error" carries no settlement information on Sui — a settle has to dedup on the digest and bind the *committed* effects to the requirements it's settling, not read execution status as an outcome.

One new wrinkle since those runs: the public testnet fullnode has already pruned the gasless-run digests — getTransaction returns NOT_FOUND today for `6N9DhoZt8LCpdrQzWB1Rfk4SdpmaX3QRSHhKCBqiBPdi`, the digest cited above. So a digest-lookup replay guard is only load-bearing inside the node's retention window. That works out fine as long as the transaction's own replay protection (expiration for gasless, owned-input versions otherwise) expires it well inside retention, but it's an assumption worth stating in the spec rather than inheriting silently.

Both points line up with validating effects of the executed transaction. FWIW that's how the facilitator I run does it — net balance change to payTo, then the same check re-run against committed effects at settle — and I'm happy to share the duplicate-submission harness if it's useful for the redesign.

_Per the contributing guide's AI section: drafted with AI assistance; the duplicate-submission results and today's pruning check were run against live testnet and the outputs above are observed, not projected._

hayes-mysten:
@DrVelvetFog I probably should have been more precise about the replay prevention piece.  Sui will prevent replaying transactions, and there is no concern that the same signed transaction can be replayed to charge the sender multiple times.

The concern is that there is no mechanism to prevent parallel use of the same transaction in multiple transactions.  Sui's idempotent transaction execution model does mean that it is more challenging to assert that the same payment has not already been received without some state in the server.

It looks like the implementation for some chains allows signing an intent on the client with the server creating a separate transaction on each request, with the intent holding a non reusable nonce to prevent replays.  Sui does not have that type of mechanism.  

Checking to see if the tx has already been executed is reasonable, but the spec should definitely include a way to have something along the lines of an invoice id or other identifier the server can pass in extra that must be embeded in the ValidDuring nonce or an unused pure input and can be verified by the validator.  This can probably be optional, but I think its critical to have something like that to prevent using the same payment in multiple parallel requests

DrVelvetFog:
Right, thanks for the clarification — that matches what I see operating my facilitator. Digest-level dedup solves re-settlement, but it makes settle *idempotent*: every presenter of the same bytes gets the original success back, so one payment presented against N parallel requests reads as N successful settles unless the resource server itself tracks which digest paid for which request. Moving the binding into the signed bytes is where it belongs.

Two mechanics notes from checking the types:

- `ValidDuring.nonce` is a `u32` (per the SDK's BCS structs, which track sui-types `transaction.rs`). That's enough to disambiguate one merchant's outstanding requests inside the validity window — the payment already binds payTo/asset/amount, so the nonce only has to separate parallel requests to the same merchant — but it won't carry a content-derived id like a hash of an offer. If the spec wants that, the unused-pure-input route is the one that fits arbitrary bytes. Either way it's in the signed bytes and recoverable from the payload, so verify can check it against `extra` statelessly.
- If the id rides in `ValidDuring`, the spec should require a `ValidDuring` expiration whenever the server requests an id, independent of gas shape — a coin-sourced payment can legally carry `Expiration::None` (replay-protected by its owned inputs), and the binding shouldn't silently vanish with the expiration.

Optional for the field seems right, but I'd make the verify rule MUST-when-requested: if `extra` carries an id and the transaction doesn't embed it, reject before settle.

The verify side is a decode-and-compare; happy to validate it on my live facilitator if observed behavior would help the spec text.

_Per the contributing guide's AI section: drafted with AI assistance; the type check was verified against the SDK source before posting._

hayes-mysten:
I've opened https://github.com/x402-foundation/x402/pull/3081 as an alternative (and an implementation in https://github.com/x402-foundation/x402/pull/3082).   I know there has already been a lot of iteration and discussion here, so feel free to discard my version if these changes do not make sense.  I am hoping this is both simpler and more flexible and can address some of the remaining open issues in the current solution.  Happy to talk more about it if there are any concerns or questions I can help with

Sceat:
Most of the discussion here was probably claude talking to claude and we'll need a deeper look by seniors on the matter for final reviews (like you did), I'm currently out of bandwidth so can't look into it atm, but hopefully there is some interesting key parts

PR #2616, opened 2026-06-12
@x402/sui — exact scheme mechanism for Sui (gasless transfers)

Sceat:
A `@x402/sui` mechanism package implementing the `exact` scheme for Sui, mirroring `@x402/aptos` in shape.

The headline: on Sui this payment is **gasless**. It rides Sui's **Address Balances** (protocol v125), so the transaction carries `gasPayment: []` and `gasPrice: 0` and is built entirely from protocol-allowlisted stablecoin operations.

No gas token, no sponsor, no gas-station round trip, no coin-object storage cost — the spec's entire Sponsored Transactions appendix is unnecessary on this path. The classic `Coin<T>` construction stays supported as the fallback for amounts below the protocol's `0.01` minimum.

This implements [`scheme_exact_sui.md`](specs/schemes/exact/scheme_exact_sui.md) and follows up on **#340** — @phdargen's invitation to land a v2-native Sui mechanism. The spec amendment ships as its **own PR (#2615)**; this PR carries only the implementation — the two review together.

cc @bmwill for spec-conformance.

---

## Structure

Mirrors `@x402/aptos` one-to-one — same `SchemeNetworkClient` / `SchemeNetworkServer` / `SchemeNetworkFacilitator` split, package layout, configs, and exports map. No changes to core. The payload is spec-exact: `{ signature, transaction }` (base64 signature, base64-BCS transaction).

```
typescript/packages/mechanisms/sui/
  src/{constants,types,signer,utils,index}.ts
  src/exact/{client,server,facilitator}/{scheme,index}.ts
  test/unit/{constants,types,signer,index,facilitator-verify}.test.ts
  test/integrations/exact-sui.test.ts      (env-gated, like aptos)
  test/fixtures/{gen,fixtures}.ts          (real testnet-signed fixtures)
```

---

## How the facilitator verifies a payment

Seven checks, each with a typed `invalidReason` and a dedicated unit test:

1. **Protocol envelope** — x402 version 2, scheme `exact`, matching network, and a well-formed `{ signature, transaction }` payload.
2. **Signature binds the payer** — the signature verifies over the transaction *and* the recovered address is the transaction's `sender`, so the debit can only come from the signer.
3. **Gasless shape** — the decoded transaction must carry `gasPrice: 0` and `gasPayment: []`, and every command must be an allowlisted stablecoin call (`0x2::balance::{send_funds,redeem_funds}`, `0x2::coin::{send_funds,into_balance}`) or native `SplitCoins` / `MergeCoins` change-making. `TransferObjects` and everything else is rejected. With zero gas and no object writes, there's no facilitator gas to drain and nothing to grief.
4. **Replay guard** — the digest is recomputed from the signed bytes and looked up on-chain; an already-committed transaction is refused. *(Simulation can't be the guard here: a gasless tx has no object inputs, so replaying executed bytes still simulates clean.)*
5. **Expiry** — the dry-run must succeed; a TTL-expired transaction fails at this step.
6. **Exact balance match** — every declared output (`extra.outputs`, defaulting to a single `payTo`) is credited to the cent, the payer is debited exactly `amount`, and no undeclared recipient of the asset is allowed.

Settlement is **executed-first and idempotent**: an already-committed tx returns its original digest without re-broadcasting; otherwise it re-verifies (defence in depth), broadcasts keylessly, and waits for finality. There is no `getExtra()` step — gasless is sponsor-free, which is simpler than the Aptos/SVM mechanisms and a feature in its own right.

---

## Tests

**67 unit tests, fully network-free** (`pnpm --filter @x402/sui test`) — every rejection branch (bad version / scheme / network, malformed payload, bad signature, sender ≠ signer, the gasless-shape guard, the `assetTransferMethod` gate, the coin-object plumbing accept and the `TransferObjects` reject, the already-executed replay, simulation failure, recipient mismatch, outputs mismatch, undeclared-recipient skim, payer-debit mismatch, asset mismatch), plus single- and two-output happy paths and the idempotent settle. Fixtures are **real testnet-signed** transactions captured once by `test/fixtures/gen.ts`.

**7 integration tests on live testnet** (`test:integration`, gated on `SUI_CLIENT_PRIVATE_KEY`) — full client → server → facilitator round-trips for single-output, two-output split, idempotent re-settle, a cheat rejection, a coin-object payer settling end-to-end, the replay rejection, and the `assetTransferMethod` gate refusing a declared `coin` method.

**Live conformance evidence** (Sui testnet, settled for this submission):

| Scenario | Digest | On-chain effect |
| --- | --- | --- |
| Gasless single-output | `AUEMarkDbo1L4xW4e9K5pRMorXtaQwWXgK6XviBQmcVo` | `gasPrice=0`, `gasPayment=[]`; USDC −10000 / +10000 |
| Gasless two-output split | `61HtkapYg1DaFnuZhDZCsrWLxcAyaYDc4ZaF7TzcqnaC` | `gasPrice=0`, `gasPayment=[]`; USDC −10000 / +200 / +9800 |

Both are verifiable on any public testnet fullnode via `sui_getTransactionBlock`: zero gas price, empty gas payment, and balance changes matching the declared outputs exactly.

---

## Credits & provenance

Builds on @Danny-Devs's [`x402-sui`](https://github.com/Danny-Devs/x402-sui) (Apache-2.0) — the signer, utils, constants, and server scheme derive from his work, adapted to `@mysten/sui` 2.17 (gRPC for the gasless build, JSON-RPC for simulate / execute) and the `@x402/core` interfaces. Derived files carry an Apache-2.0 §4 header and the commits are co-authored to him.

| Area | Source |
| --- | --- |
| signer / utils / constants / server scheme / test harness | @Danny-Devs `x402-sui` (Apache-2.0), adapted |
| gasless multi-output builder, exact-fee matcher, `assetTransferMethod` gate, cheat / replay cases | @Sceat (contributed here) |
| net-new | version guard, BCS command-shape allowlist, `getAssetDecimals`, the 7-step verify, idempotent keyless settle |
| spec + Address-Balances follow-up | @bmwill (Mysten) |
| the Transaction + Address-Balance APIs this rides on | `@mysten/sui` — thanks @hayes-mysten |

---

## AI assistance disclosure (per CONTRIBUTING)

Most of this implementation was AI-generated (Claude Fable 5 + Opus 4.8) and then reviewed through multiple consensus rounds. Every payment, signature, and settlement path was checked against the spec and exercised on live testnet with real USDC before submission — the unit-test assertions derive from **on-chain effects** (committed real dry-run `balanceChanges` and signed transactions), not from generated expectations. Hallucinated code doesn't settle USDC on-chain; the digests above are the proof.

---

## Local gauntlet

```
pnpm --filter @x402/core --filter @x402/sui build
pnpm --filter @x402/sui test                                      # 64 passed
pnpm --filter @x402/sui lint:check
pnpm --filter @x402/sui format:check
SUI_CLIENT_PRIVATE_KEY=… pnpm --filter @x402/sui test:integration  # 7 passed (testnet)
```

## Checklist

- [x] Formatted and linted (`lint:check` + `format:check` green)
- [x] All new and existing tests pass (64 unit + 7 testnet integration)
- [x] Changeset added (`typescript/.changeset/sui-exact-mechanism.md`)

cc @phdargen @bmwill @CarsonRoscoe @hayes-mysten @Danny-Devs

### DIFF
--- a/typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
+++ b/typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
@@ -1,0 +1,276 @@
+import { SuiGrpcClient } from "@mysten/sui/grpc";
+import { Transaction } from "@mysten/sui/transactions";
+import { fromBase64 } from "@mysten/sui/utils";
+import type {
+  PaymentRequirements,
+  PaymentPayloadResult,
+  SchemeNetworkClient,
+} from "@x402/core/types";
+import type { ClientSuiSigner } from "../../signer";
+import type { ExactSuiPayload, SuiOutput } from "../../types";
+import {
+  GASLESS_ALLOWED_TARGETS,
+  GASLESS_ALLOWED_NON_MOVECALL,
+  TESTNET_RPC_URL,
+  MAINNET_RPC_URL,
+  DEVNET_RPC_URL,
+  SUI_MAINNET_CAIP2,
+  SUI_TESTNET_CAIP2,
+  SUI_DEVNET_CAIP2,
+  normalizeMoveTarget,
+} from "../../constants";
+import { createSuiClient, matchBalanceChanges, outputsOf } from "../../utils";
+
+/**
+ * Sui client implementation for the Exact payment scheme.
+ *
+ * Builds a GASLESS Address-Balance payment PTB — one `0x2::balance::send_funds`
+ * per declared output, drawn from the payer's Address Balance (the SDK's
+ * `tx.balance()` input resolves the coin source), built over gRPC so the gasless
+ * params (`gasPrice = 0`, `gasPayment = []`) resolve, with `setGasBudget(0n)` to
+ * force the resolver's gasless election deterministically. Signs but does NOT
+ * execute — the facilitator broadcasts during settlement.
+ *
+ * When the requirements advertise `extra.buildUrl`, the client MAY instead fetch
+ * the facilitator's prebuilt unsigned bytes — but it independently verifies them
+ * (sender, gasless gas fields, allowlisted commands) BEFORE signing.
+ */
+export class ExactSuiScheme implements SchemeNetworkClient {
+  readonly scheme = "exact";
+
+  /**
+   * Creates a new ExactSuiScheme client instance.
+   *
+   * @param signer - The client signer for signing transactions
+   * @param config - Optional config object
+   * @param config.rpcUrl - Optional custom gRPC RPC URL
+   */
+  constructor(
+    private readonly signer: ClientSuiSigner,
+    private readonly config?: { rpcUrl?: string },
+  ) {}
+
+  /**
+   * Creates a payment payload by building (or fetching) and signing a gasless PTB.
+   *
+   * @param x402Version - The x402 protocol version
+   * @param paymentRequirements - The payment requirements (amount, asset, payTo, network, extra)
+   * @returns Promise resolving to a signed payment payload
+   */
+  async createPaymentPayload(
+    x402Version: number,
+    paymentRequirements: PaymentRequirements,
+  ): Promise<PaymentPayloadResult> {
+    const { asset, network } = paymentRequirements;
+    if (!asset) {
+      throw new Error("Asset is required");
+    }
+
+    const outputs = outputsOf(paymentRequirements);
+    this.assertOutputsSumToAmount(outputs, paymentRequirements.amount);
+
+    const client = this.grpcClient(network);
+
+    // PATH B: a prebuilt-transaction URL was advertised — fetch, VERIFY, then sign.
+    const buildUrl = paymentRequirements.extra?.buildUrl;
+    const bytes =
+      typeof buildUrl === "string"
+        ? await this.fetchAndVerifyBuildUrl(buildUrl, paymentRequirements, outputs)
+        : await this.buildGaslessOutputs(client, asset, outputs);
+
+    const tx = Transaction.from(fromBase64(bytes));
+    const { signature, bytes: signedBytes } = await this.signer.signTransaction(tx);
+
+    const payload: ExactSuiPayload = {
+      signature,
+      transaction: signedBytes,
+    };
+
+    return {
+      x402Version,
+      payload,
+    };
+  }
+
+  /**
+   * Build the gasless Address-Balance payment PTB: one `0x2::balance::send_funds`
+   * per declared output, `setGasBudget(0n)` to force the gasless election.
+   *
+   * @param client - A gRPC client (gasless eligibility resolves over gRPC)
+   * @param asset - The coin type to transfer
+   * @param outputs - The declared `{ to, amount }` outputs
+   * @returns Base64-encoded unsigned transaction bytes
+   */
+  private async buildGaslessOutputs(
+    client: SuiGrpcClient,
+    asset: string,
+    outputs: SuiOutput[],
+  ): Promise<string> {
+    const tx = new Transaction();
+    tx.setSender(this.signer.address);
+    for (const o of outputs) {
+      tx.moveCall({
+        target: "0x2::balance::send_funds",
+        typeArguments: [asset],
+        arguments: [tx.balance({ type: asset, balance: BigInt(o.amount) }), tx.pure.address(o.to)],
+      });
+    }
+    // Force the gasless election (the resolver's `budget === 0n` branch) so
+    // gaslessness does not depend on node-rebate behavior. Do NOT also set
+    // gasPrice/gasPayment manually — the protocol rejects those.
+    tx.setGasBudget(0n);
+    const built = await tx.build({ client });
+    return Buffer.from(built).toString("base64");
+  }
+
+  /**
+   * Fetch unsigned bytes from a facilitator `buildUrl` and INDEPENDENTLY verify
+   * them before signing: decode, assert the sender is this client, assert the
+   * gasless gas fields, assert every command is allowlisted, and assert the
+   * declared recipients/amounts match. The client MUST NOT sign bytes that fail.
+   *
+   * @param buildUrl - The facilitator's prebuilt-transaction endpoint
+   * @param requirements - The agreed payment requirements
+   * @param outputs - The declared `{ to, amount }` outputs
+   * @returns Base64-encoded, verified unsigned transaction bytes
+   */
+  private async fetchAndVerifyBuildUrl(
+    buildUrl: string,
+    requirements: PaymentRequirements,
+    outputs: SuiOutput[],
+  ): Promise<string> {
+    if (!/^https:\/\//.test(buildUrl)) {
+      throw new Error(`buildUrl must be an absolute https URL: ${buildUrl}`);
+    }
+
+    const res = await fetch(buildUrl, {
+      method: "POST",
+      headers: { "content-type": "application/json" },
+      body: JSON.stringify({ sender: this.signer.address, requirements }),
+    });
+    if (!res.ok) {
+      throw new Error(`buildUrl returned ${res.status}`);
+    }
+    const body = (await res.json()) as { transaction?: string; bytes?: string };
+    const bytes = body.transaction ?? body.bytes;
+    if (typeof bytes !== "string") {
+      throw new Error("buildUrl response missing transaction bytes");
+    }
+
+    await this.assertBytesSafe(bytes, requirements.network, requirements.asset, outputs);
+    return bytes;
+  }
+
+  /**
+   * Decode prebuilt (facilitator-built) bytes and assert they are safe to sign — the
+   * ENTIRE threat model of `buildUrl` is a malicious or buggy facilitator, so this is
+   * the one guard that matters. It asserts, in order: (1) the sender is this client,
+   * (2) the gasless gas fields (`gasPrice == 0` ∧ `gasPayment == []`), (3) every
+   * command is an allowlisted gasless op or tolerated coin plumbing, and (4) — the
+   * load-bearing check the old code SKIPPED — a DRY-RUN proves the exact-fee split:
+   * each declared recipient is credited exactly, this client is debited exactly the
+   * total, and NO undeclared address receives the asset. Without (4) a facilitator
+   * could slip a hidden recipient into bytes the client then blindly signs.
+   *
+   * Uses the same exact-fee balance-change matcher as the facilitator's verify,
+   * with `expectedPayer = sender`.
+   *
+   * @param bytes - Base64-encoded unsigned transaction bytes
+   * @param network - CAIP-2 network identifier (for the dry-run client)
+   * @param asset - The required coin type
+   * @param outputs - The declared `{ to, amount }` outputs
+   */
+  private async assertBytesSafe(
+    bytes: string,
+    network: string,
+    asset: string,
+    outputs: SuiOutput[],
+  ): Promise<void> {
+    const data = Transaction.from(fromBase64(bytes)).getData();
+
+    const sender = (data.sender ?? "").toLowerCase();
+    if (sender !== this.signer.address.toLowerCase()) {
+      throw new Error(`buildUrl sender mismatch: ${data.sender} ≠ ${this.signer.address}`);
+    }
+
+    const price = data.gasData?.price;
+    const payment = data.gasData?.payment;
+    const gasless =
+      (price === "0" || price === null || price === undefined) &&
+      (payment === null ||
+        payment === undefined ||
+        (Array.isArray(payment) && payment.length === 0));
+    if (!gasless) {
+      throw new Error(`buildUrl bytes are not gasless: ${JSON.stringify(data.gasData ?? {})}`);
+    }
+
+    for (const cmd of data.commands) {
+      if (cmd.$kind === "MoveCall" && cmd.MoveCall) {
+        const target = normalizeMoveTarget(
+          cmd.MoveCall.package,
+          cmd.MoveCall.module,
+          cmd.MoveCall.function,
+        );
+        if (!GASLESS_ALLOWED_TARGETS.has(target)) {
+          throw new Error(`buildUrl bytes contain a disallowed target: ${target}`);
+        }
+      } else if (!GASLESS_ALLOWED_NON_MOVECALL.has(cmd.$kind)) {
+        throw new Error(`buildUrl bytes contain a disallowed command: ${cmd.$kind}`);
+      }
+    }
+
+    // (4) The exact-fee match — dry-run the UNSIGNED bytes and verify the recipients
+    // and amounts BEFORE signing. `expectedPayer` is the sender we just matched.
+    const sim = await createSuiClient(network as never).dryRunTransactionBlock({
+      transactionBlock: bytes,
+    });
+    if (sim.effects?.status?.status !== "success") {
+      throw new Error(`buildUrl bytes fail dry-run: ${sim.effects?.status?.error ?? "unknown"}`);
+    }
+    const problems = matchBalanceChanges(
+      sim.balanceChanges ?? [],
+      asset,
+      outputs,
+      this.signer.address,
+    );
+    if (problems.length > 0) {
+      throw new Error(`buildUrl bytes do not pay the declared split: ${problems.join("; ")}`);
+    }
+  }
+
+  /**
+   * Assert that declared outputs sum to the total `amount`.
+   *
+   * @param outputs - The declared `{ to, amount }` outputs
+   * @param amount - The total atomic-unit amount
+   */
+  private assertOutputsSumToAmount(outputs: SuiOutput[], amount: string): void {
+    const total = outputs.reduce((s, o) => s + BigInt(o.amount), 0n);
+    if (total !== BigInt(amount)) {
+      throw new Error(`declared outputs sum ${total} ≠ amount ${amount}`);
+    }
+  }
+
+  /**
+   * Create a gRPC client for the given network (the transport gasless build needs).
+   *
+   * @param network - CAIP-2 network identifier
+   * @returns A SuiGrpcClient for the network
+   */
+  private grpcClient(network: string): SuiGrpcClient {
+    if (this.config?.rpcUrl) {
+      const ref = network.split(":")[1] as "testnet" | "mainnet" | "devnet";
+      return new SuiGrpcClient({ network: ref, baseUrl: this.config.rpcUrl });
+    }
+    switch (network) {
+      case SUI_MAINNET_CAIP2:
+        return new SuiGrpcClient({ network: "mainnet", baseUrl: MAINNET_RPC_URL });
+      case SUI_TESTNET_CAIP2:
+        return new SuiGrpcClient({ network: "testnet", baseUrl: TESTNET_RPC_URL });
+      case SUI_DEVNET_CAIP2:
+        return new SuiGrpcClient({ network: "devnet", baseUrl: DEVNET_RPC_URL });
+      default:
+        throw new Error(`Unsupported Sui network: ${network}`);
+    }
+  }
+}
--- a/typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
+++ b/typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
@@ -9,17 +9,14 @@ import type {
 import type { ClientSuiSigner } from "../../signer";
 import type { ExactSuiPayload, SuiOutput } from "../../types";
 import {
-  GASLESS_ALLOWED_TARGETS,
-  GASLESS_ALLOWED_NON_MOVECALL,
   TESTNET_RPC_URL,
   MAINNET_RPC_URL,
   DEVNET_RPC_URL,
   SUI_MAINNET_CAIP2,
   SUI_TESTNET_CAIP2,
   SUI_DEVNET_CAIP2,
-  normalizeMoveTarget,
 } from "../../constants";
-import { createSuiClient, matchBalanceChanges, outputsOf } from "../../utils";
+import { outputsOf } from "../../utils";
 
 /**
  * Sui client implementation for the Exact payment scheme.
@@ -31,9 +28,10 @@ import { createSuiClient, matchBalanceChanges, outputsOf } from "../../utils";
  * force the resolver's gasless election deterministically. Signs but does NOT
  * execute — the facilitator broadcasts during settlement.
  *
- * When the requirements advertise `extra.buildUrl`, the client MAY instead fetch
- * the facilitator's prebuilt unsigned bytes — but it independently verifies them
- * (sender, gasless gas fields, allowlisted commands) BEFORE signing.
+ * Implements the `address-balance` asset transfer method. When the requirements
+ * declare `extra.assetTransferMethod`, the payment MUST use that method — a
+ * declared `coin` method is rejected here (this client does not build the
+ * classic gas-paying path).
  */
 export class ExactSuiScheme implements SchemeNetworkClient {
   readonly scheme = "exact";
@@ -66,17 +64,20 @@ export class ExactSuiScheme implements SchemeNetworkClient {
       throw new Error("Asset is required");
     }
 
+    // A declared method is binding (spec "Method selection rules"); this client
+    // implements the gasless `address-balance` method only.
+    const method = paymentRequirements.extra?.assetTransferMethod;
+    if (method !== undefined && method !== "address-balance") {
+      throw new Error(
+        `unsupported assetTransferMethod: ${String(method)} (this client implements address-balance)`,
+      );
+    }
+
     const outputs = outputsOf(paymentRequirements);
     this.assertOutputsSumToAmount(outputs, paymentRequirements.amount);
 
     const client = this.grpcClient(network);
-
-    // PATH B: a prebuilt-transaction URL was advertised — fetch, VERIFY, then sign.
-    const buildUrl = paymentRequirements.extra?.buildUrl;
-    const bytes =
-      typeof buildUrl === "string"
-        ? await this.fetchAndVerifyBuildUrl(buildUrl, paymentRequirements, outputs)
-        : await this.buildGaslessOutputs(client, asset, outputs);
+    const bytes = await this.buildGaslessOutputs(client, asset, outputs);
 
     const tx = Transaction.from(fromBase64(bytes));
     const { signature, bytes: signedBytes } = await this.signer.signTransaction(tx);
@@ -121,121 +122,6 @@ export class ExactSuiScheme implements SchemeNetworkClient {
     tx.setGasBudget(0n);
     const built = await tx.build({ client });
     return Buffer.from(built).toString("base64");
-  }
-
-  /**
-   * Fetch unsigned bytes from a facilitator `buildUrl` and INDEPENDENTLY verify
-   * them before signing: decode, assert the sender is this client, assert the
-   * gasless gas fields, assert every command is allowlisted, and assert the
-   * declared recipients/amounts match. The client MUST NOT sign bytes that fail.
-   *
-   * @param buildUrl - The facilitator's prebuilt-transaction endpoint
-   * @param requirements - The agreed payment requirements
-   * @param outputs - The declared `{ to, amount }` outputs
-   * @returns Base64-encoded, verified unsigned transaction bytes
-   */
-  private async fetchAndVerifyBuildUrl(
-    buildUrl: string,
-    requirements: PaymentRequirements,
-    outputs: SuiOutput[],
-  ): Promise<string> {
-    if (!/^https:\/\//.test(buildUrl)) {
-      throw new Error(`buildUrl must be an absolute https URL: ${buildUrl}`);
-    }
-
-    const res = await fetch(buildUrl, {
-      method: "POST",
-      headers: { "content-type": "application/json" },
-      body: JSON.stringify({ sender: this.signer.address, requirements }),
-    });
-    if (!res.ok) {
-      throw new Error(`buildUrl returned ${res.status}`);
-    }
-    const body = (await res.json()) as { transaction?: string; bytes?: string };
-    const bytes = body.transaction ?? body.bytes;
-    if (typeof bytes !== "string") {
-      throw new Error("buildUrl response missing transaction bytes");
-    }
-
-    await this.assertBytesSafe(bytes, requirements.network, requirements.asset, outputs);
-    return bytes;
-  }
-
-  /**
-   * Decode prebuilt (facilitator-built) bytes and assert they are safe to sign — the
-   * ENTIRE threat model of `buildUrl` is a malicious or buggy facilitator, so this is
-   * the one guard that matters. It asserts, in order: (1) the sender is this client,
-   * (2) the gasless gas fields (`gasPrice == 0` ∧ `gasPayment == []`), (3) every
-   * command is an allowlisted gasless op or tolerated coin plumbing, and (4) — the
-   * load-bearing check the old code SKIPPED — a DRY-RUN proves the exact-fee split:
-   * each declared recipient is credited exactly, this client is debited exactly the
-   * total, and NO undeclared address receives the asset. Without (4) a facilitator
-   * could slip a hidden recipient into bytes the client then blindly signs.
-   *
-   * Uses the same exact-fee balance-change matcher as the facilitator's verify,
-   * with `expectedPayer = sender`.
-   *
-   * @param bytes - Base64-encoded unsigned transaction bytes
-   * @param network - CAIP-2 network identifier (for the dry-run client)
-   * @param asset - The required coin type
-   * @param outputs - The declared `{ to, amount }` outputs
-   */
-  private async assertBytesSafe(
-    bytes: string,
-    network: string,
-    asset: string,
-    outputs: SuiOutput[],
-  ): Promise<void> {
-    const data = Transaction.from(fromBase64(bytes)).getData();
-
-    const sender = (data.sender ?? "").toLowerCase();
-    if (sender !== this.signer.address.toLowerCase()) {
-      throw new Error(`buildUrl sender mismatch: ${data.sender} ≠ ${this.signer.address}`);
-    }
-
-    const price = data.gasData?.price;
-    const payment = data.gasData?.payment;
-    const gasless =
-      (price === "0" || price === null || price === undefined) &&
-      (payment === null ||
-        payment === undefined ||
-        (Array.isArray(payment) && payment.length === 0));
-    if (!gasless) {
-      throw new Error(`buildUrl bytes are not gasless: ${JSON.stringify(data.gasData ?? {})}`);
-    }
-
-    for (const cmd of data.commands) {
-      if (cmd.$kind === "MoveCall" && cmd.MoveCall) {
-        const target = normalizeMoveTarget(
-          cmd.MoveCall.package,
-          cmd.MoveCall.module,
-          cmd.MoveCall.function,
-        );
-        if (!GASLESS_ALLOWED_TARGETS.has(target)) {
-          throw new Error(`buildUrl bytes contain a disallowed target: ${target}`);
-        }
-      } else if (!GASLESS_ALLOWED_NON_MOVECALL.has(cmd.$kind)) {
-        throw new Error(`buildUrl bytes contain a disallowed command: ${cmd.$kind}`);
-      }
-    }
-
-    // (4) The exact-fee match — dry-run the UNSIGNED bytes and verify the recipients
-    // and amounts BEFORE signing. `expectedPayer` is the sender we just matched.
-    const sim = await createSuiClient(network as never).dryRunTransactionBlock({
-      transactionBlock: bytes,
-    });
-    if (sim.effects?.status?.status !== "success") {
-      throw new Error(`buildUrl bytes fail dry-run: ${sim.effects?.status?.error ?? "unknown"}`);
-    }
-    const problems = matchBalanceChanges(
-      sim.balanceChanges ?? [],
-      asset,
-      outputs,
-      this.signer.address,
-    );
-    if (problems.length > 0) {
-      throw new Error(`buildUrl bytes do not pay the declared split: ${problems.join("; ")}`);
-    }
   }
 
   /**
--- a/typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts
@@ -1,0 +1,335 @@
+import { Transaction } from "@mysten/sui/transactions";
+import { fromBase64 } from "@mysten/sui/utils";
+import type {
+  Network,
+  PaymentPayload,
+  PaymentRequirements,
+  SchemeNetworkFacilitator,
+  SettleResponse,
+  VerifyResponse,
+} from "@x402/core/types";
+import type { FacilitatorSuiSigner } from "../../signer";
+import type { ExactSuiPayload } from "../../types";
+import {
+  GASLESS_ALLOWED_TARGETS,
+  GASLESS_ALLOWED_NON_MOVECALL,
+  normalizeMoveTarget,
+} from "../../constants";
+import { matchBalanceChanges, outputsOf } from "../../utils";
+
+/**
+ * Sui facilitator implementation for the Exact payment scheme.
+ *
+ * Verifies a signed-but-not-executed gasless transaction by binding the
+ * signature to the sender, asserting the gasless BCS shape (zero gas + an
+ * allowlisted-command-only PTB), rejecting an already-executed transaction (the
+ * stateless replay guard — simulation alone is NOT one for Address Balances),
+ * simulating it, and matching the asset balance changes against the declared
+ * outputs EXACTLY. Settles by keyless broadcast (idempotent, executed-first).
+ */
+export class ExactSuiScheme implements SchemeNetworkFacilitator {
+  readonly scheme = "exact";
+  readonly caipFamily = "sui:*";
+
+  /**
+   * Creates a new ExactSuiScheme facilitator instance.
+   *
+   * @param signer - The facilitator signer for verification, simulation, and broadcast
+   */
+  constructor(private readonly signer: FacilitatorSuiSigner) {}
+
+  /**
+   * Get mechanism-specific extra data for the supported kinds endpoint. The
+   * gasless path is sponsor-free, so there is no feePayer to advertise.
+   *
+   * @param _ - The network identifier (unused)
+   * @returns Always undefined — gasless needs no sponsor metadata
+   */
+  getExtra(_: Network): Record<string, unknown> | undefined {
+    void _;
+    return undefined;
+  }
+
+  /**
+   * Get the facilitator's broadcast identities for a network. Empty on the
+   * pure gasless path (keyless broadcast — there is no sponsor key).
+   *
+   * @param _ - The network identifier (unused)
+   * @returns Array of signer addresses
+   */
+  getSigners(_: string): string[] {
+    void _;
+    return [...this.signer.getAddresses()];
+  }
+
+  /**
+   * Verify a payment payload (the spec's `exact`-Sui verification):
+   * version/scheme/network/shape → signature binds the sender → gasless command
+   * shape → not-already-executed (the stateless replay guard) → simulate succeeds →
+   * exact balance-change match against the declared outputs.
+   *
+   * @param payload - The payment payload to verify
+   * @param requirements - The payment requirements
+   * @returns Promise resolving to verification response
+   */
+  async verify(
+    payload: PaymentPayload,
+    requirements: PaymentRequirements,
+  ): Promise<VerifyResponse> {
+    try {
+      // Step 1: version + scheme + network + payload shape.
+      if (payload.x402Version !== 2) {
+        return { isValid: false, invalidReason: "invalid_x402_version", payer: "" };
+      }
+      if (payload.accepted.scheme !== "exact" || requirements.scheme !== "exact") {
+        return { isValid: false, invalidReason: "invalid_scheme", payer: "" };
+      }
+      if (payload.accepted.network !== requirements.network) {
+        return { isValid: false, invalidReason: "invalid_network", payer: "" };
+      }
+      const suiPayload = payload.payload as ExactSuiPayload;
+      if (!suiPayload?.transaction || !suiPayload?.signature) {
+        return { isValid: false, invalidReason: "invalid_payload", payer: "" };
+      }
+
+      // Step 2: signature is valid AND the recovered address is the tx sender.
+      let payer: string;
+      try {
+        payer = await this.signer.verifySignature(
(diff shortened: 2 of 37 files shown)


DrVelvetFog:
I run a non-custodial x402 facilitator on Sui (the one in #2619), on the Enoki-sponsored path — so this is the construction I didn't take, and it's clean. Riding Address Balances to drop `gasPayment` and `gasPrice` to zero makes the whole sponsorship round-trip unnecessary for stablecoins at or above the protocol minimum, which is most real payment traffic. Worth saying out loud for readers: sponsorship doesn't disappear, it becomes the fallback rail for the `Coin<T>` path (sub-minimum amounts, non-allowlisted assets), and a production facilitator wants both. Your PR keeps the classic path for exactly that, which is the right call.

The part I want to flag from the settlement side: your verify recomputes the digest from the signed bytes, and your settle is executed-first and idempotent, returning the original digest rather than re-broadcasting. That plus the exact per-output `balanceChanges` against declared `extra.outputs` is precisely the substrate for binding a signed execution receipt to the settlement — the settled result is recomputable from chain alone, no trust in the facilitator, which is the property the settlement-receipt binding extension (#2666) is built on. The gasless shape makes it cleaner than the sponsored path: no gas-payment noise in the balance delta, so the net to `payTo` is exactly the declared amount.

Two things I can contribute if useful: (1) cross-check the round-trip against my live testnet facilitator and report digests back here, and (2) add an exact-on-Sui settlement-receipt-binding conformance vector to #2666's set, so the Sui exact scheme ships with independent recompute coverage the same way the generic and SVM vectors do. Either way, good to see a v2-native gasless Sui mechanism land — this is the right shape.

0xkurious:
Nice work on the gasless Address Balance mechanism — ran it against my own AWS KMS setup to check something adjacent to what @DrVelvetFog raised about settlement-receipt binding: can `ExactSuiScheme` (client) be driven by a signer whose private key never touches process memory?

`toClientSuiSigner()` types its `keypair` param as the abstract `Signer` class from `@mysten/sui/cryptography`, not the concrete `Ed25519Keypair` shown in the README example. Turns out `@mysten/aws-kms-signer@0.3.0` (Ed25519 support just landed upstream in [ts-sdks#1118](https://github.com/MystenLabs/ts-sdks/pull/1118)) `extends Signer`, so it's a drop-in — zero adapter code needed:

```ts
const kmsSigner = await AwsKmsSigner.fromKeyId(KMS_KEY_ID, {
  region: "us-east-1",
  credentials: fromNodeProviderChain(),
});
const clientSigner = toClientSuiSigner(kmsSigner, suiClient);
x402Client.register("sui:testnet", new ExactSuiScheme(clientSigner));
```
Ran two real gasless payments end-to-end on testnet (self-facilitation server, this PR's branch built locally) — private key never left KMS in either:

- CwwwoqB4sP53vhQFtKuL8QnHwqA4rYcZQYY6tKs9LP8A (https://suiscan.xyz/testnet/tx/CwwwoqB4sP53vhQFtKuL8QnHwqA4rYcZQYY6tKs9LP8A) — USDC -10000/+10000
- 2cFM63YZAMrsmaJYcB2tUe6L15YZkFSiqnB7E6sZLzMz (https://suiscan.xyz/testnet/tx/2cFM63YZAMrsmaJYcB2tUe6L15YZkFSiqnB7E6sZLzMz) — USDC -10000/+10000, gasUsed at zero across the board

Repro + full writeup: https://github.com/0xkurious/x402-sui-kms-signer

Worth calling out: nobody coordinated this compatibility. @mysten/sui's Signer was already an abstract class before this PR existed, and toClientSuiSigner happened to type against it instead of Ed25519Keypair. That's what makes AWS KMS custody (or a future TEE/HSM/passkey signer) a drop-in here with zero glue code — the composability came for free from good abstraction boundaries on both sides, not from any explicit design-for-HSM effort. 

Might be worth a line in the README calling out Signer as the actual contract, so this is discoverable without reading signer.ts.

Sceat:
@DrVelvetFog yes to both, go ahead. Cross checked digests from your facilitator posted here would be the best third party coverage this PR can get. Same for the sui vector in #2666: verify recomputes the digest from the signed bytes and settle returns the executed digest with per output `balanceChanges` asserted against the declared `extra.outputs`, so the recompute should work from chain data alone. If any wire shape turns out ambiguous while you build the vector, flag it here and I'll tighten it on my side. You read the split exactly as intended: gasless is the default rail for allowlisted stables at or above the protocol minimum, the classic `Coin<T>` path stays in the spec as the fallback for everything else, which is the half your facilitator already covers. Following your comment on #2615 I added an "independently recomputable settlement" note to the spec's security considerations. I kept the #2666 reference out of the spec text for now since there is no merged file to point at yet, the property stands on its own either way.

@0xkurious verified both digests on testnet, clean gasless shape, no gas noise in the balance changes. Typing `toClientSuiSigner` against the abstract `Signer` was intentional (signer.ts says any sui signer works) but KMS custody is a path I never tested myself, so this is a genuinely useful datapoint. Added your suggested line to the README calling out `Signer` as the actual contract, with remote custody signers like `@mysten/aws-kms-signer` as the example.

Also rebased on main: concordium and keeta landed in the meantime and conflicted with the e2e network registry, the workspace manifest and the lockfile.

DrVelvetFog:
Ran the round-trip through my independent Sui facilitator (the non-custodial one from #2619) and it settled your gasless shape cleanly, no facilitator changes needed. Built the payment the way your client does — one `0x2::balance::send_funds` per output via `tx.balance()`, over gRPC, `setGasBudget(0n)` to force the gasless election — signed it, and pushed it through `/verify` → `/settle`.

Single-output gasless USDC on testnet:

- digest: [`FVejSg9ddPYXwWtxjk58TjkZg6aawJFyynZEsTmsxRQ6`](https://testnet.suivision.xyz/txblock/FVejSg9ddPYXwWtxjk58TjkZg6aawJFyynZEsTmsxRQ6)

Then I recomputed the settlement from chain alone (`getTransactionBlock`, `showBalanceChanges`/`showEffects`/`showInput`), no trust in the facilitator's response:

- payTo credited **exactly** +10000, payer debited **exactly** −10000 (USDC), no undeclared recipient — the exact-fee, no-skim balance-change match your `matchBalanceChanges` enforces, confirmed against the executed effects.
- gasless on chain: `gasData.price = 0`, `gasData.payment = []`, `gasUsed` fully rebated. Payer SUI delta is **0** — no gas noise in the asset delta, so the net to `payTo` is precisely the declared amount, exactly as you called out.

That's the property #2666 binds to: the settled result is recomputable from committed chain data, not from anything the facilitator asserts. Two independent implementations (yours + mine) now produce the same recomputable result for the gasless exact-Sui mechanism.

Next I'll add an exact-on-Sui settlement-receipt-binding conformance vector to #2666's set anchored to this digest, so the Sui exact scheme ships with the same independent-recompute coverage as the generic and SVM vectors. This run is the single-output path; happy to extend the same cross-check to a multi-recipient `extra.outputs` fee-split if you want that leg covered too — flag any wire-shape detail you'd want pinned and I'll match it.

DrVelvetFog:
Landed the exact-on-Sui conformance vector I mentioned, anchored to this cross-check's digest ([`FVejSg9…`](https://testnet.suivision.xyz/txblock/FVejSg9ddPYXwWtxjk58TjkZg6aawJFyynZEsTmsxRQ6)) — two rails, byte-identical checker, `python3 _check_independent.py` exit 0, and cited in #2666 §5 as a live-settlement reproduction:
https://github.com/DrVelvetFog/sui-x402-facilitator/tree/087df5232206ea8c5105ce58f6f42063315ca744/conformance/x402_settlement_v0_sui_gasless

So the gasless mechanism here now has an independent-recompute vector to go with the digest cross-check. Offer still stands to add the multi-recipient `extra.outputs` fee-split leg if you want that path covered too.

Danny-Devs:
Thanks for carrying the Sui exact work into the monorepo — good to see it land. The replay-guard reasoning is the part I like most: gasless txs have no object inputs, so executed bytes still re-simulate clean, which means digest-recompute really is the only stateless guard. That's a genuine sharpening of the merged spec's "simulate to ensure not already executed" wording. Pinning the allowlist to the four functions that exist on-chain is the right call too.

Four things I hit going through the facilitator, all fixable:

1. **Method name mismatch (interop break).** #2615 mandates `assetTransferMethod` be `"gasless"` or `"coin"`, but this package types, announces, and gates on `"address-balance"` — and rejects `"gasless"`. A spec-conformant counterparty gets bounced. One side needs a rename; not sure which you want canonical.

2. **Balance matcher misses non-canonical addresses.** `matchBalanceChanges` compares with `toLowerCase()` but not `normalizeSuiAddress`, so a `payTo` / `outputs[].to` with stripped leading zeros won't match the padded `AddressOwner` the chain returns, and a valid payment gets rejected. Normalizing both sides fixes it.

3. **Verify trusts simulation to catch expiry, but Sui simulation doesn't check it.** The step-5 comment says simulate "proves it is not expired (TTL)," but `dryRunTransactionBlock` / `SimulateTransaction` skip `TransactionExpiration` — it's enforced only in the epoch-aware `validity_check` (`sui-types/src/transaction.rs:3135`), not the `validity_check_no_gas_check` simulation calls (`:3449`). The ~2-epoch `ValidDuring` bounds it, so it isn't unbounded, but a stale payload still passes verify and fails only at settle — so a serve-first flow can deliver against a payment that can't settle. Decoding the expiration and rejecting it directly closes it.

4. **`settle()`'s idempotency shortcut returns success without binding to these requirements.** The executed-first branch returns `success` + digest on `isTransactionExecuted(digest)` before verify runs — no signature check, no `matchBalanceChanges`, no check that the executed tx paid this `payTo` / `asset` / `amount`. Raw bytes + signatures of any committed tx are public via `getTransactionBlock`, so a historical gasless transfer can be wrapped as a fresh payload. Verify-first flows catch it, but a facilitator can't assume that ordering. Running `matchBalanceChanges` against the executed effects in that branch keeps re-settle idempotent and closes the replay.

I build s402 (Sui-native, x402-compatible), so seeing Sui land upstream in this shape is genuinely good news — I'd rather help these get fixed here than fork the effort, and I'm glad to send suggestions for any of them.

_Per the contributing guide's AI section: I drafted this with AI assistance and checked each finding against the PR source and the Sui codebase myself before posting._

Sceat:
@Danny-Devs thanks for the careful pass — all four findings confirmed. State of each; none of the code fixes are committed yet, they land as one alignment series against the round-3 spec:

1. **Method name:** #2615 is canonical — the value was renamed to `gasless` in the round-2 review (the gas shape, not the asset location, is what the method names). This package will be renamed to match in the alignment series, together with the new `feePayer` sponsored flow and `extra.splits` from the same review rounds.
2. **Confirmed.** `matchBalanceChanges` needs `normalizeSuiAddress` on both sides; lowercasing alone misses stripped leading zeros against the chain's padded `AddressOwner`. Will fix with a stripped-zero address test vector.
3. **Confirmed against current main.** `check_transaction_input_inner` (sui-transaction-checks) never receives an epoch and never runs the epoch-aware `validity_check`; `TransactionExpired` is raised only in the authority signing path, so dry-run passes expired bytes and the failure surfaces at broadcast. Verify will decode the expiration and mirror the signing path's validity in full (epoch window, chain identifier, rejection of the unsupported timestamp fields). The #2615 text that claimed simulation catches TTL expiry is corrected in the same round, with credit to you.
4. **Confirmed.** The executed-first idempotency branch must bind the committed transaction to the requirements being settled before returning success; the fix re-runs payer recovery and recipient matching against the executed transaction's effects, which keeps re-settle idempotent while closing the historical-transaction replay. The round-3 spec now states this requirement explicitly so it binds every implementation, not just this one.

One more item in the same series, found while re-checking the protocol sources for your #3: the package's gasless function allowlist (four functions in `constants.ts`) is stale against current node releases, which allow nine (`balance::{send_funds, redeem_funds, split, zero}`, `funds_accumulator::withdrawal_split`, `coin::{send_funds, redeem_funds, into_balance, put}`); the constant and its tests will be refreshed with the rest.

Suggestion blocks are welcome on any of these, test vectors from your s402 work especially — the findings will be credited in the commits.

Danny-Devs:
Following up with a correction to one of my own findings, plus two new items from a deeper verification pass. Apologies in advance for the churn on finding 3 — details below, sources pinned.

**Correction to my finding 3 — Sui moved, and my claim is now stale.** "Simulation doesn't check `TransactionExpiration`" was true when I verified it, but Sui #26794 (`44cbd478cffe`, merged 2026-06-09, "Unify validity_check and validity_check_no_gas_check for dry_run") changed it: `AuthorityState::simulate_transaction` now calls the epoch-aware `validity_check` unconditionally (`crates/sui-core/src/authority.rs:2593` at `b4ea45ea`), enforcing the epoch window, chain-identifier match, and timestamp-field rejection in dry-run. I confirmed containment in testnet-v1.74.0, mainnet-v1.74.1, v1.75.2, and devnet-v1.76.0 (testnet-v1.73.0 predates it). I had verified against `sui-transaction-checks` — which indeed still has no expiry check — and missed the newer unified path in sui-core. My mistake, apologies.

Two knock-ons:

1. The round-3 spec states the stale version as fact ("Simulation enforces none of this…" in gasless step 5, and the coin/sponsored step-5 equivalent). Suggest rewording to "enforced in dry-run since v1.74.0 (#26794); implementations MUST NOT rely on the node version."
2. The explicit verify-side expiry check is still the right call, for better reasons than I originally gave: (a) pre-1.74 nodes get no check from simulation; (b) nothing anywhere enforces `maxTimeoutSeconds` — simulation only checks the window against the *current* epoch, so a far-future expiration passes; (c) a coin-object-sourced gasless tx — which the facilitator accepts (SplitCoins/MergeCoins tolerated) even though this package's client never builds it — satisfies replay protection through owned inputs (`check_replay_protection`, `crates/sui-transaction-checks/src/lib.rs:255`) and may legally carry `Expiration::None`: no TTL at all.

**New finding A — `isTransactionExecuted` is existence-only, so a committed-but-failed transaction settles as success.** `src/signer.ts:202-211` returns true on `getTransaction` presence with no effects-status check, and both the settle idempotency branch (`src/exact/facilitator/scheme.ts:246-256`) and the race-catch (`:288-299`) trust it. But a failed address-balance withdrawal still *commits*, with effects — Sui's own e2e tests prove it (`crates/sui-e2e-tests/tests/address_balance_tests.rs`: `test_withdraw_insufficient_balance` at `:457`; `InsufficientFundsForWithdraw` asserted against committed effects at `:1804`, `:1911`). So a settle whose broadcast lands as a failed commit — via the window between settle's own re-verify and broadcast, or a re-settle of an already-committed-failed digest — returns `success: true` while no funds moved. Fix: gate both branches on effects status == success, plus the committed-effects recipient check the round-3 spec already prescribes for the executed-first path (which would close my finding 4 in the same stroke).

**New finding B — the not-found matcher over-matches.** `isNotFound` (`src/signer.ts:150-155`) tests `/not found/i` against the full message, so a JSON-RPC "Method not found" or an HTML 404 body classifies as "transaction never executed" — and the replay guard then passes an already-settled payment, the opposite of the propagate-errors intent stated in the adjacent comment. Whether it's reachable depends on the deployment surfacing such an error shape on that path, but it's a correctness bug either way. Matching only the genuine miss shape ("Transaction <digest> not found" / the NOT_FOUND code you already handle separately) closes it.

Minor, defense-in-depth: `assertGaslessShape` (`:332-358`) omits `gasBudget == 0` and gas-owner == sender. Not exploitable — the node enforces both at simulation ("gas_budget must be 0 for gasless transactions") — but worth asserting locally so shape validation doesn't depend on the transport. Two spec nits from the same pass: gasless is protocol v119 on testnet/devnet and v124 on mainnet, not v125; and the ~2-epoch `ValidDuring` is set by the SDK, not "stamped by the protocol" — the protocol *requires* replay protection (protected expiration OR gas payment OR owned inputs) and validates the window, but never stamps one.

On the vectors: happy to send the expiration suite (fail-closed on missing expiration + a `maxTimeoutSeconds` bound at epoch granularity, with an offline BCS fixture builder — no funded key or network needed) and a short-form-address vector for the `normalizeSuiAddress` fix — as suggestion blocks or a small patch series against the alignment branch once it's up, whichever is easier to pull.

_Per the contributing guide's AI section: drafted with AI assistance; I checked each claim against the cited sources (SHAs above) before posting._

Danny-Devs:
As promised — the test vectors, delivered now so they can lead the alignment series rather than trail it. Everything below was executed against this PR's head (`75408fa`), not just written: the new file runs **6 failing / 1 passing on exactly the intended assertions** (no import or type errors), the existing 67-test suite stays green beside it, and under a local proof-of-concept fix (expiration check + `normalizeSuiAddress`) the full run goes 74/74. Adapted from my implementation of the same mechanism (Apache-2.0, attribution in the file header).

One framing correction that sharpens the expiration story, found while building the offline fixtures: your client's payloads are **not** unbounded — the SDK's default build path stamps `ValidDuring(currentEpoch, currentEpoch+1)` automatically (all four committed fixtures carry it). The actual gap is narrower and still real: `verify()` never *reads* the expiration field, and the payer controls the bytes — a hand-built payload with `Expiration::None`, a legacy `Epoch(n)`, or a far-future `ValidDuring` window all pass verify today, so `requirements.maxTimeoutSeconds` is silently unenforced for gasless payments (a pre-signed payment stays broadcastable ~indefinitely). Conformance/liveness-bound severity, not fund theft. The suite covers all three variants, plus an ACCEPT guard so the bound can't be over-tightened (your client's own `currentEpoch+1` window must always pass a 60s timeout — epoch granularity rounds up).

Notes for pulling these in:

- The `invalidReason` strings in the two REJECT assertions are **proposals** (taken from my implementation); if you prefer your `invalid_transaction_state` vocabulary, the load-bearing assertions are the `isValid` ones — rename freely.
- **Fix placement:** the expiration-required check needs to sit *after* your step-3 gasless-shape check — `BAD_GAS_PAYING_TX` and `TRANSFER_OBJECTS_TX` both carry `Expiration::None`, so an earlier check breaks two existing step-3 tests (verified both orderings).
- `ValidDuring` also carries `minTimestamp`/`maxTimestamp` — so a fix could enforce `maxTimeoutSeconds` *precisely* via `maxTimestamp` where present, with the epoch bound as the floor. The vectors only pin the epoch-bound floor, so they stay valid under either design.
- The normalization vectors use explicitly constructed short-form/padded pairs (a leading-zero-nibble address has two legitimate encodings; the chain always reports the padded one). Fail-closed false rejection of a correct payment — liveness/interop, not fund loss.
- If the fix reads the current epoch from the dry-run, your existing mock in `facilitator-verify.test.ts` needs one line inside `effects` — `executedEpoch: "1128"` (that exact value keeps all fixture happy-paths green: the committed fixtures' `maxEpoch` values are 1128/1129). The contributed file carries its own mock, so it stands alone either way.

<details>
<summary><b>test/unit/contributed-vectors.test.ts</b> (complete file — 6 red on this head, 74/74 with the fix)</summary>

```typescript
// Contributed test vectors for PR #2616 — expiration bound + address normalization.
// Both suites are fully OFFLINE (no network, no funded key): the expiration vectors
// are built with an offline BCS fixture builder (`bcs.TransactionData.serialize` of a
// gasless-shaped PTB), and the normalization vectors drive `matchBalanceChanges` /
// `verify()` with explicit short-form vs padded address pairs.
//
// Adapted from https://github.com/Danny-Devs/x402-sui (Apache-2.0), commit 7e2fbf5b.

import { describe, it, expect } from "vitest";
import { bcs } from "@mysten/sui/bcs";
import type { DryRunTransactionBlockResponse } from "@mysten/sui/jsonRpc";
import { toBase64, normalizeSuiAddress } from "@mysten/sui/utils";
import type { PaymentPayload, PaymentRequirements } from "@x402/core/types";
import { ExactSuiScheme as ExactSuiFacilitator } from "../../src/exact/facilitator/scheme";
import type { FacilitatorSuiSigner } from "../../src/signer";
import { matchBalanceChanges } from "../../src/utils";
import {
  VALID_SINGLE_BALANCE_CHANGES,
  PAYER,
  RECIPIENT_1,
  ASSET,
  NETWORK,
} from "../fixtures/fixtures";

// The testnet chain identifier baked into the committed fixtures' ValidDuring
// expirations (the SDK's default build path binds it automatically).
const TESTNET_CHAIN_ID = "69WiPg3DAQiwdxfncX6wYQ2siKwAe6L9BZthQea3JNMD";

type FixtureExpiration =
  | { None: true }
  | { Epoch: string }
  | {
      ValidDuring: {
        minEpoch: string | null;
        maxEpoch: string | null;
        minTimestamp: string | null;
        maxTimestamp: string | null;
        chain: string;
        nonce: number;
      };
    };

/**
 * Offline BCS fixture builder: a gasless-shaped (gasPrice 0, gasPayment [])
 * empty PTB from PAYER with a chosen TransactionExpiration. The PTB carries no
 * commands (the money movement is supplied by the mock's balanceChanges), so it
 * passes every existing verify() step — isolating the EXPIRATION field as the
 * only variable under test. A malicious client controls these bytes end to end,
 * so any variant it can serialize is a variant verify() must handle.
 */
function gaslessTxWithExpiration(expiration: FixtureExpiration): string {
  return toBase64(
    bcs.TransactionData.serialize({
      V1: {
        kind: { ProgrammableTransaction: { inputs: [], commands: [] } },
        sender: PAYER,
        gasData: { payment: [], owner: PAYER, price: "0", budget: "0" },
        expiration,
      },
    }).toBytes(),
  );
}

/** ValidDuring epoch window [min, max] in the shape the SDK's build path emits. */
function validDuring(minEpoch: string, maxEpoch: string): FixtureExpiration {
  return {
    ValidDuring: {
      minEpoch,
      maxEpoch,
      minTimestamp: null,
      maxTimestamp: null,
      chain: TESTNET_CHAIN_ID,
      nonce: 42,
    },
  };
}

// Same shape as the mock in facilitator-verify.test.ts, plus `executedEpoch` on
// the dry-run effects (the dry-run reports the network's current epoch there —
// it is what an epoch-granularity expiration bound must compare against).
function mockSigner(opts: {
  payer?: string;
  balanceChanges?: unknown[];
  executedEpoch?: string;
}): FacilitatorSuiSigner {
  return {
    getAddresses: () => [],
    verifySignature: async () => opts.payer ?? PAYER,
    isTransactionExecuted: async () => false,
    simulateTransaction: async () =>
      ({
        effects: {
          status: { status: "success" },
          executedEpoch: opts.executedEpoch ?? "1000",
        },
        balanceChanges: opts.balanceChanges ?? [],
      }) as unknown as DryRunTransactionBlockResponse,
    executeTransaction: async () => "0xdigest",
    waitForTransaction: async () => {},
  };
}

function buildPayload(tx: string, sig: string, network = NETWORK): PaymentPayload {
  return {
    x402Version: 2,
    accepted: { scheme: "exact", network } as PaymentRequirements,
    payload: { transaction: tx, signature: sig },
  } as PaymentPayload;
}

function requirements(over: Partial<PaymentRequirements> = {}): PaymentRequirements {
  return {
    scheme: "exact",
    network: NETWORK,
    asset: ASSET,
    amount: "10000",
    payTo: RECIPIENT_1,
    maxTimeoutSeconds: 60,
    extra: {},
    ...over,
  };
}

describe("ExactSuiFacilitator.verify() — expiration bound (maxTimeoutSeconds)", () => {
  // A gasless tx has NO object inputs, so nothing is consumed and nothing goes
  // stale: the signed bytes stay broadcastable until their TransactionExpiration.
  // The client's own build path emits ValidDuring(currentEpoch, currentEpoch+1)
  // — the committed fixtures all carry it — but verify() never READS the
  // expiration, and the payer's client controls the bytes: it can serialize
  // Expiration::None (broadcastable FOREVER), a legacy Epoch bound, or a
  // ValidDuring window ending far in the future. In all three cases
  // `requirements.maxTimeoutSeconds` is silently unenforced. Fail closed:
  // require a bounded expiration and check it against the current epoch the
  // dry-run reports (`effects.executedEpoch`), rounding the epoch bound UP so
  // the client's own currentEpoch+1 window always passes.

  it("REJECTS Expiration::None (unbounded lifetime — maxTimeoutSeconds unenforceable)", async () => {
    const f = new ExactSuiFacilitator(mockSigner({ balanceChanges: VALID_SINGLE_BALANCE_CHANGES }));
    const tx = gaslessTxWithExpiration({ None: true });
    const r = await f.verify(buildPayload(tx, "c2ln"), requirements());
    expect(
      r.isValid,
      "Expiration::None must be rejected: the signed bytes stay broadcastable indefinitely, so maxTimeoutSeconds cannot be enforced",
    ).toBe(false);
    expect(r.invalidReason).toBe("invalid_exact_sui_payload_transaction_expiration_missing");
  });

  it("REJECTS a ValidDuring window ending further ahead than maxTimeoutSeconds allows", async () => {
    // Current epoch (dry-run executedEpoch) = 1000; maxTimeoutSeconds = 60 →
    // ceil(60 / ~86400s-per-epoch) rounds up to 1 epoch ahead → allowed ≤ 1001.
    // maxEpoch 1005 keeps the payment broadcastable ~4 days past the timeout.
    const f = new ExactSuiFacilitator(
      mockSigner({ balanceChanges: VALID_SINGLE_BALANCE_CHANGES, executedEpoch: "1000" }),
    );
    const tx = gaslessTxWithExpiration(validDuring("1000", "1005"));
    const r = await f.verify(buildPayload(tx, "c2ln"), requirements({ maxTimeoutSeconds: 60 }));
    expect(
      r.isValid,
      "ValidDuring maxEpoch 1005 with current epoch 1000 and maxTimeoutSeconds=60 must be rejected (≈4 days too broadcastable)",
    ).toBe(false);
    expect(r.invalidReason).toBe("invalid_exact_sui_payload_transaction_expiration_too_far");
  });

  it("REJECTS a legacy Epoch expiration further ahead than maxTimeoutSeconds allows", async () => {
    // Same bound, oldest variant: a hand-built tx can still use Epoch(n).
    const f = new ExactSuiFacilitator(
      mockSigner({ balanceChanges: VALID_SINGLE_BALANCE_CHANGES, executedEpoch: "1000" }),
    );
    const tx = gaslessTxWithExpiration({ Epoch: "1005" });
    const r = await f.verify(buildPayload(tx, "c2ln"), requirements({ maxTimeoutSeconds: 60 }));
    expect(
      r.isValid,
      "Epoch(1005) with current epoch 1000 and maxTimeoutSeconds=60 must be rejected (≈4 days too broadcastable)",
    ).toBe(false);
    expect(r.invalidReason).toBe("invalid_exact_sui_payload_transaction_expiration_too_far");
  });

  it("ACCEPTS the client's own currentEpoch+1 window even for a small timeout (bound rounds UP)", async () => {
    // Epoch granularity is the floor: the SDK's build path emits
    // ValidDuring(current, current+1) — exactly what the committed fixtures
    // carry — and a 60s timeout cannot be expressed in whole epochs. The bound
    // must round up to at least one epoch ahead, or the facilitator would
    // reject every payment its own client builds. Guards against over-tightening.
    const f = new ExactSuiFacilitator(
      mockSigner({ balanceChanges: VALID_SINGLE_BALANCE_CHANGES, executedEpoch: "1000" }),
    );
    const tx = gaslessTxWithExpiration(validDuring("1000", "1001"));
    const r = await f.verify(buildPayload(tx, "c2ln"), requirements({ maxTimeoutSeconds: 60 }));
    expect(r.isValid, r.invalidMessage).toBe(true);
  });
});

// ─────────────────────────────────────────────────────────────────────────────
// Address normalization: a Sui address whose full form has a leading zero nibble
// has a VALID short (stripped) encoding — "0x0abc…" ≡ "0xabc…". The chain always
// reports padded 64-hex AddressOwner values, but declared outputs / payTo /
// payer come from requirements JSON, where either encoding is legitimate.
// `toLowerCase()`-only comparison treats the two encodings as different
// addresses → a fail-CLOSED false rejection of a correct payment (liveness /
// interop, not fund loss). Normalizing BOTH sides (normalizeSuiAddress) fixes it.
// ─────────────────────────────────────────────────────────────────────────────

// A recipient whose full form has a leading zero nibble, in both encodings.
const SHORT_TO = "0xc0ffee" + "0".repeat(55) + "42"; // 63 hex chars — short form
const PADDED_TO = normalizeSuiAddress(SHORT_TO); // "0x0c0ffee…42" — as the chain reports it

// PAYER's full form starts "0x087aa8…", so it too has a valid stripped encoding.
const SHORT_PAYER = "0x" + PAYER.slice(3); // strip the leading zero nibble

const PADDED_TO_BALANCE_CHANGES = [
  { coinType: ASSET, owner: { AddressOwner: PAYER }, amount: "-10000" },
  { coinType: ASSET, owner: { AddressOwner: PADDED_TO }, amount: "10000" },
];

describe("matchBalanceChanges — short-form vs padded address equivalence", () => {
  it("matches a declared output given in short form against the chain's padded AddressOwner", () => {
    const problems = matchBalanceChanges(
      PADDED_TO_BALANCE_CHANGES,
      ASSET,
      [{ to: SHORT_TO, amount: "10000" }],
      PAYER,
    );
    expect(
      problems,
      `"${SHORT_TO}" and "${PADDED_TO}" are the same Sui address — a correct payment must not be rejected`,
    ).toEqual([]);
  });

  it("matches a short-form payer against the chain's padded debit entry", () => {
    const problems = matchBalanceChanges(
      VALID_SINGLE_BALANCE_CHANGES, // real fixture: padded PAYER debited -10000
      ASSET,
      [{ to: RECIPIENT_1, amount: "10000" }],
      SHORT_PAYER,
    );
    expect(
      problems,
      `"${SHORT_PAYER}" and "${PAYER}" are the same Sui address — the payer debit must be recognized`,
    ).toEqual([]);
  });
});

describe("ExactSuiFacilitator.verify() — short-form payTo (end-to-end normalization)", () => {
  it("ACCEPTS a correct payment whose requirements.payTo uses the short-form encoding", async () => {
    const f = new ExactSuiFacilitator(mockSigner({ balanceChanges: PADDED_TO_BALANCE_CHANGES }));
    const tx = gaslessTxWithExpiration(validDuring("1000", "1001"));
    const r = await f.verify(
      buildPayload(tx, "c2ln"),
      requirements({ payTo: SHORT_TO, amount: "10000" }),
    );
    expect(r.isValid, r.invalidMessage).toBe(true);
    expect(r.payer).toBe(PAYER);
  });
});
```

</details>

<details>
<summary>Observed output on <code>75408fa</code> (trimmed)</summary>

```
❯ test/unit/contributed-vectors.test.ts (7 tests | 6 failed)
  × verify() — expiration bound > REJECTS Expiration::None (unbounded lifetime — maxTimeoutSeconds unenforceable)
    → expected true to be false
  × verify() — expiration bound > REJECTS a ValidDuring window ending further ahead than maxTimeoutSeconds allows
    → expected true to be false
  × verify() — expiration bound > REJECTS a legacy Epoch expiration further ahead than maxTimeoutSeconds allows
    → expected true to be false
  ✓ verify() — expiration bound > ACCEPTS the client's own currentEpoch+1 window even for a small timeout (bound rounds UP)
  × matchBalanceChanges > matches a declared output given in short form against the chain's padded AddressOwner
    → expected [ …(2) ] to deeply equal []
  × matchBalanceChanges > matches a short-form payer against the chain's padded debit entry
    → expected [ 'payer expected -10000 got 0' ] to deeply equal []
  × verify() — short-form payTo (end-to-end) > ACCEPTS a correct payment whose requirements.payTo uses the short-form encoding
    → expected false to be true

Test Files  1 failed | 5 passed (6)
     Tests  6 failed | 68 passed (74)
```

</details>

Happy to reshape any of it to fit the alignment series — or if it's easier, I can send this as a PR against your branch once it's up.

_Per the contributing guide's AI section: drafted with AI assistance; the suite was run against the PR head locally and the outputs above are observed, not projected._

Sceat:
@Danny-Devs This is an exceptional review, and the self-correction on finding 3 is exactly the rigor that makes the scheme better. I verified each claim against the sources independently before replying; all three confirm.

**Finding 3 correction (dry-run enforces expiration since v1.74), confirmed.** `AuthorityState::simulate_transaction` calls the epoch-aware `validity_check(&epoch_store.tx_validity_check_context())` unconditionally (sui-core `authority.rs:2592-2593`, the call at 2593 under a "Full validity check including gas budget and price" comment), so on v1.74+ the dry-run does enforce the epoch window, the chain-identifier match, and the timestamp-field rejection. The round-3 spec's "Simulation enforces none of this" is version-stale. It will reword to "enforced in dry-run since v1.74.0 (#26794); implementations MUST NOT rely on the node version," and keep the explicit verify-side check for the three reasons you give: pre-1.74 nodes get no check, nothing enforces `maxTimeoutSeconds` (the dry-run only checks the window against the current epoch, so a far-future bound passes), and a coin-object-sourced gasless transaction satisfies replay protection through owned inputs and may legally carry `Expiration::None`.

**Finding A (existence-only execution check), confirmed and real.** `isTransactionExecuted` (`signer.ts:202-211`) returns `true` on `getTransaction` presence with no effects-status check, and a failed address-balance withdrawal still commits with effects, so a committed-failed digest settles as success while no funds moved. Fix as prescribed: gate both the settle idempotency branch and the race-catch on `effects.status == "success"`, plus the committed-effects recipient check the round-3 spec already mandates for the executed-first path (which closes your earlier finding 4 in the same stroke).

**Finding B (not-found over-match), confirmed.** `isNotFound` (`signer.ts:150-155`) falls back to `/not found/i` over the message after the NOT_FOUND code check, so a JSON-RPC "Method not found" or a 404 body classifies as never-executed and the replay guard passes an already-settled payment, the opposite of the adjacent comment's stated intent. Fix: match only the genuine miss shape and lean on the NOT_FOUND code already handled separately.

The gasless-shape `gasBudget == 0` and gas-owner assertions and the "the SDK sets the ValidDuring window, the protocol requires replay protection and validates it rather than stamping one" framing are well-taken and land in the same pass. On the versions: confirmed, gasless enables at protocol v119 on testnet and devnet (`119 => { if chain != Chain::Mainnet ... enable_gasless = true }`) and v124 on mainnet, not v125; the spec will correct it.

On the vectors: yes, please. The offline BCS-fixture approach with no funded key or network, and the `maxTimeoutSeconds` ACCEPT-guard against over-tightening, are exactly right, and the fix-placement note (the expiration-required check sits after the step-3 shape check, since the `None`-carrying shape-reject fixtures precede it) is a real gotcha. I will open the alignment branch shortly and pull `contributed-vectors.test.ts` in with your Apache-2.0 header and attribution intact. A suggestion-block series or a PR against that branch both work, whichever is less friction for you.

typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,162 @@
...
+   * Build the gasless Address-Balance payment PTB: one `0x2::balance::send_funds`
+   * per declared output, `setGasBudget(0n)` to force the gasless election.
+   *
+   * @param client - A gRPC client (gasless eligibility resolves over gRPC)
+   * @param asset - The coin type to transfer
+   * @param outputs - The declared `{ to, amount }` outputs
+   * @returns Base64-encoded unsigned transaction bytes
+   */
+  private async buildGaslessOutputs(
+    client: SuiGrpcClient,
```
hayes-mysten, line 106: I would recommend `ClientWithCoreApi` as the type used across any SDK surface (this allows this to work with grpc or graphql

typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    const outputs = outputsOf(paymentRequirements);
+    this.assertOutputsSumToAmount(outputs, paymentRequirements.amount);
+
+    const client = this.grpcClient(network);
+    const bytes = await this.buildGaslessOutputs(client, asset, outputs);
+
+    const tx = Transaction.from(fromBase64(bytes));
+    const { signature, bytes: signedBytes } = await this.signer.signTransaction(tx);
```
hayes-mysten, line 83: this seems a little odd to me.  The sui Signer class takes bytes, so we get bytes above, parse the to transaction, pass them to the `ClientSuiSigner` which immediately re-serializes again before the real signer signs the bytes

typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,162 @@
...
+      tx.moveCall({
+        target: "0x2::balance::send_funds",
+        typeArguments: [asset],
+        arguments: [tx.balance({ type: asset, balance: BigInt(o.amount) }), tx.pure.address(o.to)],
+      });
+    }
+    // Force the gasless election (the resolver's `budget === 0n` branch) so
+    // gaslessness does not depend on node-rebate behavior. Do NOT also set
+    // gasPrice/gasPayment manually — the protocol rejects those.
+    tx.setGasBudget(0n);
```
hayes-mysten, line 122: this was needed on json rpc, but on grpc its no longer required.  Is only supporting gasless rather than just getting gasless transactions when eligible the preferred option here?

I think just allowing any coin type, and letting the network set the gas cost to 0 for eligible transactions is more flexible

typescript/packages/mechanisms/sui/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,162 @@
...
+    const built = await tx.build({ client });
+    return Buffer.from(built).toString("base64");
```
hayes-mysten, line 124: is Buffer the standard base64 method here?  The sui sdk exposes a toBase64 method that does not require a transform or polyfill to work in browsers

typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,359 @@
...
+      } catch (error) {
+        return {
+          isValid: false,
+          invalidReason: "invalid_exact_sui_payload_signature",
+          invalidMessage: error instanceof Error ? error.message : String(error),
+          payer: "",
+        };
+      }
+
+      const data = Transaction.from(fromBase64(suiPayload.transaction)).getData();
```
hayes-mysten, line 136: Transaction.from already accepts base64 encoded transactions directly

typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,359 @@
...
+    requirements: PaymentRequirements,
+  ): Promise<SettleResponse> {
+    const suiPayload = payload.payload as ExactSuiPayload;
+
+    // Idempotency: an already-committed transaction settles to its original digest
+    // (no re-broadcast, no double charge). The signature/payload may be malformed,
+    // so guard the decode and fall through to verify on any failure.
+    let knownDigest: string | undefined;
+    try {
+      knownDigest = await Transaction.from(fromBase64(suiPayload.transaction)).getDigest();
```
hayes-mysten, line 245: same thing here

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+      const publicKey = await verifyTransactionSignature(txBytes, signature);
+      return publicKey.toSuiAddress();
+    },
+
+    async simulateTransaction(
+      transactionBytes: string,
+      network: string,
+    ): Promise<DryRunTransactionBlockResponse> {
+      const client = getClient(network);
+      const sim = await client.simulateTransaction<{ balanceChanges: true }>({
```
hayes-mysten, line 174: explicit generic here is not needed

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+      return {
+        effects: {
+          status: { status: success ? "success" : "failure", error },
+        },
+        balanceChanges: tx.balanceChanges.map(change => ({
+          coinType: change.coinType,
+          amount: change.amount,
+          owner: { AddressOwner: change.address },
+        })),
+      } as unknown as DryRunTransactionBlockResponse;
```
hayes-mysten, line 199: we almost definitely do not want to try to map back to the deprecated json rpc formats with unsafe casts.  Probably better to standardize on the new SuiClientType shapes instead

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+    },
+
+    async isTransactionExecuted(digest: string, network: string): Promise<boolean> {
+      const client = getClient(network);
+      try {
+        await client.getTransaction({ digest });
+        return true;
```
hayes-mysten, line 206: this is used as if this implies success, aborted/failed transactions will still be fetchable via their digest

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+      // Broadcast over gRPC. Re-broadcasting an already-executed gasless tx THROWS
+      // (unlike JSON-RPC) — that throw propagates to the scheme's settle, which
+      // treats it as a race and re-checks the chain. No retry/swallow here.
+      const result = await client.executeTransaction({
+        transaction: fromBase64(transaction),
+        signatures,
+      });
+
+      const tx = result.$kind === "Transaction" ? result.Transaction : result.FailedTransaction;
+      if (tx.status.success !== true) {
```
hayes-mysten, line 230: switch on kind directly instead

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+ * Convert any Sui keypair to a ClientSuiSigner. Works with Ed25519Keypair,
+ * Secp256k1Keypair, etc. Signs but never executes — the facilitator broadcasts.
+ *
+ * @param keypair - Any Sui cryptographic keypair
+ * @param client - A Sui client (JSON-RPC or gRPC) for building transactions
+ * @returns A ClientSuiSigner instance
+ */
+export function toClientSuiSigner(
+  keypair: Signer,
+  client: SuiJsonRpcClient | SuiGrpcClient,
```
hayes-mysten, line 259: ClientWithCoreApi

typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,359 @@
...
+   * This also avoids re-verifying an already-executed payment, which the verify replay
+   * guard (step 4) now rejects — a legitimate re-settle must still succeed.
+   *
+   * For a NOT-yet-executed payment it re-verifies (defense in depth) before broadcast.
+   *
+   * @param payload - The payment payload to settle
+   * @param requirements - The payment requirements
+   * @returns Promise resolving to settlement response
+   */
+  async settle(
```
hayes-mysten, line 234: this whole verify and settle flow looks unreliable and does not provide the right guarantees (would be trivial to bypass payments).  verify only validates intent and unexecuted transactions.

I need to look more closely at this, as well as the replay prevention mechanism, but I think this needs to be carefully redesigned, and at the very least should be looking at effects of the executed transaction rather than just the simulation.

typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,359 @@
...
+   * object-leak vector — and every other command stay rejected, and (b) the exact-fee
+   * balance-change match (step 5, `matchBalanceChanges`) binds the ACTUAL money
+   * movement: every declared output is credited exactly, the payer is debited exactly,
+   * and no undeclared address receives the asset — so the plumbing commands can move
+   * nothing the exact-fee check does not already account for.
+   *
+   * @param data - The decoded transaction data (`Transaction.getData()`)
+   * @returns A problem description, or "" when the shape is gasless-safe
+   */
+  private assertGaslessShape(data: ReturnType<Transaction["getData"]>): string {
```
hayes-mysten, line 332: this implementation is very brittle and there are already some planned sdk changes that would likely break this by expanding how balances are sourced/converted for some types of transactions.

I don't think trying to constrain the transaction by what is gasless eligible makes a lot of sense here.  I personally don't think restricting to free tier at all is the right move, but if that is the goal checking just the gas price/budget is probably sufficient, and allowing the network the fail execution if its not eligible instead

typescript/packages/mechanisms/sui/test/fixtures/gen.ts
```diff
@@ -0,0 +1,118 @@
...
+        tx.balance({ type: USDC_TESTNET, balance: BigInt(o.amount) }),
+        tx.pure.address(o.to),
+      ],
+    });
+  }
+  tx.setGasBudget(0n);
+  const built = await tx.build({ client: grpc });
+  const transaction = toBase64(built);
+  const { signature } = await signer.signTransaction(built);
+  const sim = await jsonRpc.dryRunTransactionBlock({ transactionBlock: transaction });
```
hayes-mysten, line 54: there probably should not be anything using json rpc in this PR anymore

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+  return {
+    getAddresses(): readonly string[] {
+      if (!keypair) return [];
+      return [keypair.toSuiAddress()];
+    },
+
+    async verifySignature(transactionBytes: string, signature: string, _: string): Promise<string> {
+      const txBytes = fromBase64(transactionBytes);
+      const publicKey = await verifyTransactionSignature(txBytes, signature);
+      return publicKey.toSuiAddress();
```
hayes-mysten, line 166: this is not actually verifying anything other than that the signature is for the tx, but it doesn't verify it matches the sender

typescript/packages/mechanisms/sui/src/signer.ts
```diff
@@ -0,0 +1,271 @@
...
+      } as unknown as DryRunTransactionBlockResponse;
+    },
+
+    async isTransactionExecuted(digest: string, network: string): Promise<boolean> {
+      const client = getClient(network);
+      try {
+        await client.getTransaction({ digest });
+        return true;
+      } catch (error) {
+        if (isNotFound(error)) return false;
```
hayes-mysten, line 208: this is really broad and very unreliable.  I don't think isExecuted is a good way to handle replay prevention

hayes-mysten (changes requested):
This looks like a big step in the right direction, and I think there are a lot of solid pieces here, but I think there are also a few big things that I would recommend improving on before trying to get this merged:

* This should adopt a nonce pattern for replay prevention similar to the svm implementation (gasless transactions can have 1 unused pure input of up to 32 bytes)
* The facilitator's verify/settle logic needs to be rethought/rewritten to properly verify payments
* We should probably not restrict to just gasless payments, and allow the sender to pay for gas for now (all eligible transactions should still end up as gasless)
* Adding support for sponsored flows would be great (providing a sponsor address in the payment requirements, and allowing the facilitator to sign during execution. Using the [sponsor sdk](https://sdk.mystenlabs.com/sponsor) would probably work well here)

phdargen:
Thanks a lot for your input @hayes-mysten! I think some of your comments would be better suited for the spec level discussion in https://github.com/x402-foundation/x402/pull/2615. Could you please review the spec, this is what we should align on first

hayes-mysten:
sorry, somehow I completely missed that and dove directly into the implementation. I'll take a look there

Issue #2603, opened 2026-06-11
FastAPI payment middleware silently swallows unexpected exceptions as an empty-body 402 (no logging)


moped2110:
**Body:**

### Summary

The FastAPI payment middleware catches any unexpected exception on the
settlement path and returns an empty-body **HTTP 402 Payment Required** with no
logging. An internal server error (RPC failure, DB error, a bug) is therefore
indistinguishable, to both clients and operators, from a normal "payment
required" response.

### Location

`python/x402/http/middleware/fastapi.py` — the final catch-all in the
`payment-verified` branch:

```python
            except FacilitatorResponseError as error:
                return _facilitator_error_response(error)
            except Exception:
                return JSONResponse(content={}, status_code=402)
```

The module imports/uses **no logging** at all (`grep -c "logg" returns 0`).

### Why it matters

- **Misleading status.** Per the core spec (§4 Response Types), an internal
  failure should map to **Server Error → 500**, not `402 Payment Required`.
  Returning 402 tells a paying client/agent "you still need to pay" when the
  server actually crashed — an autonomous agent may retry and pay again into a
  broken endpoint.
- **Zero observability.** With no log line, an operator has no signal that an
  exception occurred; the failure is completely silent.

### Reproduction

Cause any exception in the post-verification settlement path (e.g. the
facilitator becomes unreachable mid-settle, or the resource handler raises after
verification). The client receives `402` with an empty JSON body `{}` and the
server logs nothing.

### Suggested fix

1. Log the exception (`logger.exception(...)`) so failures are diagnosable.
2. Return `500` (Server Error) for unexpected exceptions, reserving `402` for
  genuine payment-required / payment-failed cases. Optionally include a minimal
  error body distinguishable from a payment response.

### How it was found

Surfaced while building [[x402-conformance](https://github.com/moped2110/x402-conformance)](https://github.com/moped2110/x402-conformance),
a black-box conformance suite for x402 V2 endpoints.

evidai:
This is one of the more dangerous defaults in agent payments, and worth fixing past the status code. `402` is the *one* status an autonomous client is built to resolve **by paying** — so collapsing an internal 5xx into an empty 402 doesn't just mislead an operator, it tells the agent "retry and pay again" into an endpoint that actually crashed. A server bug becomes an automatic billing loop.

From running a verify/settle path in production, two things I'd add to your fix:

[...]

PR #2622, opened 2026-06-13
fix(python/fastapi): return 500 (and log) on unexpected settlement er…

moped2110:
```markdown
### Problem
The FastAPI payment middleware's final catch-all on the settlement path returns
an empty-body `402 Payment Required` for *any* unexpected exception, with no
logging:

    except Exception:

An internal failure (RPC error, DB error, a bug) is therefore indistinguishable
from a genuine "payment required" — a paying client/agent is told to pay again
into a broken endpoint, and operators get no signal at all.

### Change
- Add a module logger and `logger.exception(...)` in the catch-all so the failure
  is diagnosable.
- Return `500` for unexpected exceptions, reserving `402` for genuine
  payment-required / payment-failed cases. The response carries a minimal error
  body distinguishable from a payment response.

This only touches the **unexpected-exception** branch. The legitimate
`settle_result.success is False` path (a real 402) and the
`FacilitatorResponseError` path (a 502) are unchanged.

### Tests
- New: `test_unexpected_settlement_error_returns_500_not_silent_402` — forces a
  generic error in `process_settlement` and asserts `500`, not `402`.
- `python -m pytest x402/tests/unit/http/middleware/test_fastapi.py` → 32 passed.

### How it was found
Surfaced while building a black-box conformance suite for x402 V2 endpoints.
```

---

### DIFF
--- a/python/x402/http/middleware/fastapi.py
+++ b/python/x402/http/middleware/fastapi.py
@@ -32,8 +32,6 @@ from ..types import (
 )
 from ..x402_http_server import PaywallProvider, x402HTTPResourceServer
 
-logger = logging.getLogger(__name__)
-
 if TYPE_CHECKING:
     from ...server import x402ResourceServer
 
@@ -51,6 +49,8 @@ from ._bazaar_utils import (
 from ._bazaar_utils import (
     validate_bazaar_extensions as _validate_bazaar_extensions,
 )
+
+logger = logging.getLogger(__name__)
 
 # ============================================================================
 # FastAPI Adapter
@@ -377,16 +377,14 @@ def payment_middleware(
 
             except FacilitatorResponseError as error:
                 return _facilitator_error_response(error)
-            except Exception as error:
+            except Exception:
                 # An unexpected error here (RPC failure, bug, ...) is a
                 # server-side failure, not a payment problem. Log it so
                 # operators get a signal (the module otherwise logs nothing),
                 # and surface it as a settle failure (402 + PAYMENT-RESPONSE,
                 # success=False) - consistent with the not-settle_result.success
                 # path and distinguishable from a genuine "payment required".
-                logger.exception(
-                    "x402: unexpected error while settling a verified payment"
-                )
+                logger.exception("x402: unexpected error while settling a verified payment")
                 settle_response = SettleResponse(
                     success=False,
                     error_reason="unexpected_settle_error",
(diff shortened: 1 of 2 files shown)


python/x402/http/middleware/fastapi.py
```diff
@@ -375,7 +378,17 @@ async def middleware(
...
+                # error, not a payment problem. Returning an empty-body 402
+                # silently tells a paying client "still unpaid" and logs nothing.
+                logger.exception(
+                    "x402: unexpected error while settling a verified payment; "
+                    "returning 500 instead of a misleading 402"
+                )
+                return JSONResponse(
+                    content={"error": "internal server error during settlement"},
+                    status_code=500,
+                )
```
phdargen: Thanks for the suggestion @moped2110!

500 misaligns with the HTTP spec's settle-failure shape and with the normal !settle_result.success path.
So I think this should return 402 + PAYMENT-RESPONSE (success: false) + {}, not a middleware-owned error body
```suggestion
            except Exception as error:
                settle_response = SettleResponse(
                    success=False,
                    error_reason=str(error),
                    error_message=str(error),
                    transaction="",
                    network=result.payment_requirements.network,
                )
                settle_headers = http_server._create_settlement_headers(
                    settle_response, result.payment_requirements
                )
                return JSONResponse(
                    content={},
                    status_code=402,
                    headers={"Content-Type": "application/json", **settle_headers},
                )
```

moped2110:
Good call on aligning with the settle-failure shape — applied. The catch-all now
returns 402 + PAYMENT-RESPONSE (success=False, error_reason) via
_create_settlement_headers, consistent with the not settle_result.success
path, instead of a middleware-owned 500. This also matches the TS and Go cores
(per @kakedashi3's note on #2603, both route unexpected settle errors through the
402 settlement-failure path), so 402 is the cross-language-consistent choice.

I kept the logger.exception(...) though — the other half of the original issue
was that the failure is completely silent (the module logs nothing), so an
operator still needs a signal even if the client-facing response is a clean
settle failure. Test updated to assert both the PAYMENT-RESPONSE header and that
the exception is logged.

(Minor: error_reason=str(error) surfaces the raw exception text to the client —
happy to swap it for a generic reason and keep the detail in the log only, if you
prefer not to leak internals.)

phdargen (approved):
Thanks for the update @moped2110, LGTM

PR #2627, opened 2026-06-14
fix(python): surface FastAPI settlement errors

### LINKED ISSUE #2603: FastAPI payment middleware silently swallows unexpected exceptions as an empty-body 402 (no logging)

go165:
## Summary
- Fixes #2603
- Log unexpected FastAPI settlement exceptions with `logger.exception`
- Return HTTP 500 with a minimal error body for unexpected settlement failures instead of an empty HTTP 402
- Preserve existing 402 behavior for normal settlement failures and 502 behavior for invalid facilitator responses
- Add a regression test and changelog fragment

## Verification
- `/tmp/x402-fastapi-test-venv/bin/python -m pytest python/x402/tests/unit/http/middleware/test_fastapi.py`
- `/tmp/x402-fastapi-test-venv/bin/python -m ruff check python/x402/http/middleware/fastapi.py python/x402/tests/unit/http/middleware/test_fastapi.py`
- `/tmp/x402-fastapi-test-venv/bin/python -m compileall -q python/x402/http/middleware/fastapi.py python/x402/tests/unit/http/middleware/test_fastapi.py`
- `git diff --check`

### DIFF
--- a/python/x402/changelog.d/2603.bugfix.md
+++ b/python/x402/changelog.d/2603.bugfix.md
@@ -1,0 +1,1 @@
+Log unexpected FastAPI settlement exceptions and return HTTP 500 instead of an empty 402 response.
--- a/python/x402/http/middleware/fastapi.py
+++ b/python/x402/http/middleware/fastapi.py
@@ -6,6 +6,7 @@ Provides payment-gated route protection for FastAPI applications.
 from __future__ import annotations
 
 import asyncio
+import logging
 from collections.abc import Awaitable, Callable
 from typing import TYPE_CHECKING, Any
 
@@ -48,6 +49,8 @@ from ._bazaar_utils import (
 from ._bazaar_utils import (
     validate_bazaar_extensions as _validate_bazaar_extensions,
 )
+
+logger = logging.getLogger(__name__)
 
 # ============================================================================
 # FastAPI Adapter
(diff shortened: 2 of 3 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/2622

PR #2634, opened 2026-06-15
Specs(exact): propose Canton exact scheme for x402 (spec-only)

### LINKED ISSUE #2632: Proposal: Canton Network exact-canton scheme
## Canton Network

- Privacy-preserving institutional settlement network using the Global Synchronizer
- Daml smart contracts as settlement primitives
- Stakeholder-scoped contract visibility

## Proposed scheme: `exact-canton`

The client signs a `TransferCommand` intent. The facilitator, acting as `delegate`, exercises `TransferCommand_Send` — submitting the transaction and paying the Global Synchronizer traffic fee.

**Flow:**
1. Client reads nonce from SV Scan
2. Client exercises `ExternalPartyAmuletRules_CreateTransferCommand` via interactive submission (prepare → sign → execute)
3. Client sends `PAYMENT-SIGNATURE` with `transferCommandCid`, `payerParty`, `nonce`
4. Facilitator validates `TransferCommand` from ACS, verifies against `PaymentRequirements`
5. Facilitator exercises `TransferCommand_Send` — ledger atomically moves CC and archives the command

**Key properties:**
- Client never submits to the ledger — only signs an intent
- Facilitator pays GS traffic fee
- Native replay protection via contract archival
- Canton Coin, live on Canton MainNet

Spec PR follows.

Denend:
## Description

Adds `exact-canton` scheme spec for Canton Network under
`specs/schemes/exact/`.

The scheme uses Ed25519 external-party signing: the client creates a
`Splice.ExternalPartyAmuletRules.TransferCommand` on-ledger; the
facilitator, acting as `delegate`, exercises `TransferCommand_Send` —
submitting the transaction and sponsoring sequencer traffic fees.

Closes #2632

## Tests

Spec-only change. Reference implementation verified live on Canton MainNet.

## Checklist


Denend:
@phdargen , our team is responsible for the x402 implementation on Canton Network, and we're running a little behind schedule, so we'd really appreciate a review when someone has a chance:)

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+    `TransferCommand`.
+12. **Facilitator** returns `SettlementResponse` with the ledger `updateId`.
+13. **Resource Server** grants the **Client** access and echoes settlement
+    metadata in the `PAYMENT-RESPONSE` header.
+
+## `PaymentRequirements` for `exact-canton`
+
+```json
+{
+  "scheme": "exact-canton",
```
phdargen:
```suggestion
  "scheme": "exact",
```

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+  "extra": {
+    "transferMethod": "external-party-amulet-rules",
+    "facilitatorParty": "ftp_facilitator::1220def...",
+    "synchronizerId": "global-domain::1220xyz...",
+    "merchantContractCid": "00mc..."
+  }
+}
+```
+
+- `amount`: Daml Decimal string, 10 decimal places. `"0.1000000000"` = 0.1 CC,
```
phdargen: should use atomic units

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+```json
+{
+  "scheme": "exact-canton",
+  "network": "canton:mainnet",
+  "amount": "0.1000000000",
+  "asset": "canton-coin",
+  "payTo": "merchant_party::1220abc...",
+  "maxTimeoutSeconds": 60,
+  "extra": {
+    "transferMethod": "external-party-amulet-rules",
```
phdargen: align with notation from other networks
```suggestion
    "assetTransferMethod": "external-party-amulet-rules",
```

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+    "facilitatorParty": "ftp_facilitator::1220def...",
```
phdargen:
```suggestion
    "feePayer": "ftp_facilitator::1220def...",
```

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+
+- `amount`: Daml Decimal string, 10 decimal places. `"0.1000000000"` = 0.1 CC,
+  `"1.0000000000"` = 1 CC. Must match exactly (string equality) what the Daml
+  contract encodes.
+- `asset`: `"canton-coin"`. This transfer method settles Canton Coin only.
+- `payTo`: Receiver party id in canonical form `"<name>::<fingerprint>"`.
+- `extra.transferMethod`: MUST be `"external-party-amulet-rules"`.
+- `extra.facilitatorParty`: The facilitator's Canton party id. MUST be set as
+  `delegate` in the `TransferCommand`. Clients MUST NOT alter this value.
+- `extra.synchronizerId`: The Global Synchronizer the transfer settles on.
```
phdargen, line 117: is this needed? is this not a fixed property?

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+    metadata in the `PAYMENT-RESPONSE` header.
+
+## `PaymentRequirements` for `exact-canton`
+
+```json
+{
+  "scheme": "exact-canton",
+  "network": "canton:mainnet",
+  "amount": "0.1000000000",
+  "asset": "canton-coin",
```
phdargen: consider token symbol instead
```suggestion
  "asset": "CC",
```

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+
+```json
+{
+  "success": false,
+  "errorReason": "invalid_exact_canton_amount_mismatch",
+  "transaction": ""
+}
+```
+
+## Facilitator Verification Rules (MUST)
```
phdargen, line 231: Verification checks are so far only structural payload checks but should also check if the tx would actually succeed onchain, ideally with a full tx simulation. Or if thats not possible, at least some targeted checks like sufficient client token balance etc

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+
+This document specifies the `exact-canton` payment scheme for the x402 protocol
+on Canton Network. The client creates a
+`Splice.ExternalPartyAmuletRules.TransferCommand` via Ed25519 external-party
+signing; the facilitator, acting as `delegate`, exercises `TransferCommand_Send`
+— submitting the transaction and sponsoring sequencer traffic.
+
+## Scheme Name
+
+`exact-canton`
```
phdargen:
```suggestion
`exact`
```

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
+# Exact Payment Scheme for Canton Network (`exact-canton`)
```
phdargen:
```suggestion
# Exact Payment Scheme for Canton Network (`exact`)
```

specs/schemes/exact/scheme_exact_canton.md
phdargen: could be dropped as seems the only option

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+7. **Receiver.** `TransferCommand.receiver` MUST equal
+   `PaymentRequirements.payTo`.
+
+8. **Nonce.** `TransferCommand.nonce` MUST be ≥ `nextNonce` from the sender's
+   `TransferCommandCounter` on SV Scan. A missing counter (first payment)
+   defaults to `nextNonce = 0`. Reject with
+   `invalid_exact_canton_nonce_reuse` if stale.
+
+9. **Description.** `TransferCommand.description` MUST parse as a JSON object
+   containing a non-empty `paymentId` string and `resourceUrl` equal to
```
phdargen: `paymentId` is unspecified, where does this come from?

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+   `resource.url`. Reject with `invalid_exact_canton_resource_url_mismatch`
+   if absent, unparseable, or mismatched.
```
phdargen: how and why `resource.url` is added to the TransferCommand? 

Whats the purpose of check 9 as a whole? I don't see an obvious benefit, rather introducing some privacy concerns if this info is committed onchain

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+
+```json
+{
+  "success": false,
+  "errorReason": "invalid_exact_canton_amount_mismatch",
+  "transaction": ""
+}
+```
+
+## Facilitator Verification Rules (MUST)
```
phdargen, line 231: who sets the tx fee? client or facilitator? if the latter, there should be a fee cap

phdargen, line 231: Should add a safety check that facilitator can't send away its own funds (sender != delegate)

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,249 @@
...
+      "transferMethod": "external-party-amulet-rules",
+      "facilitatorParty": "ftp_facilitator::1220def...",
+      "synchronizerId": "global-domain::1220xyz..."
+    }
+  },
+  "payload": {
+    "transferMethod": "external-party-amulet-rules",
+    "transferCommandCid": "00abc...",
+    "payerParty": "agent_party::1220...",
+    "nonce": 42
```
phdargen: hm sequential nonces are problematic. Client account would be effectively locked until server settles (could be many mins, in principle unbounded). And if client does another tx in-between verify and settle, settle fails and the server wasted work

phdargen:
Any plans for general token support besides native?

specs/schemes/exact/scheme_exact_canton.md
Denend: All three are needed, but only two must be provided explicitly:

- transferMethod - discriminates the Canton transfer mechanism; future methods (e.g. CIP-56) would use a different value, so clients must know which path to take.
- facilitatorParty - unique per facilitator deployment (open-source, anyone can run one); client must set this exact party as delegate in the TransferCommand or the facilitator cannot see the contract.
- synchronizerId - derivable from `AmuletRules.domain_id` (one per network), so it can be sourced from `/supported` instead of being included directly

specs/schemes/exact/scheme_exact_canton.md
Denend: Full tx simulation via PrepareSubmission isn't available for the facilitator (it's an external-party signing flow, not a dry-run API).

Balance check is feasible though: agents onboard via our relay and are hosted on the facilitator's own participant, so their Amulet holdings are in local ACS. Able to add that as a targeted check at /verify

specs/schemes/exact/scheme_exact_canton.md
Denend: paymentId is a UUID generated by the client at payment time (`randomUUID()`) and embedded in `TransferCommand.description` as JSON. Will add that to the spec.

specs/schemes/exact/scheme_exact_canton.md
Denend: resourceUrl is embedded by the client in TransferCommand.description
to bind the command to a specific endpoint - prevents reusing a
TransferCommand for a different route of the same merchant at the
same price.

Privacy concern is valid: the URL is committed onchain in description,
visible to all stakeholders. Could replace with H(resourceUrl) for
the same binding guarantee without leaking the URL. Open to dropping
it entirely - receiver + amount + delegate already prevent cross-merchant
and cross-amount reuse

specs/schemes/exact/scheme_exact_canton.md
Denend: Facilitator pays the GS traffic fee. It is determined by the Canton sequencer based on transaction size - not set by either party and not configurable. No fee cap needed: the fee is constant per transaction regardless of payment amount

specs/schemes/exact/scheme_exact_canton.md
Denend: Will update amount to atomic units (integer string). 1 CC = 10^10 atomic units - fixed by Daml Decimal type across all CIP-56 tokens per the token standard

specs/schemes/exact/scheme_exact_canton.md
Denend: In practice not exploitable - TransferCommand has signatory sender, so the agent can only set itself as sender (it signs with its own key). The check is a good defensive measure nonetheless - will add it

specs/schemes/exact/scheme_exact_canton.md
Denend: confirmed - TransferCommand_Send aborts without archiving if nextNonce < nonce,
so a second tx cannot settle until the first completes. Will document as a known behavior, by design.

[1] https://github.com/hyperledger-labs/splice/blob/main/daml/splice-amulet/daml/Splice/ExternalPartyAmuletRules.daml

Denend:
@

> Any plans for general token support besides native?

Initial version will only support the native token, however we are planning to add all CIP-56 tokens in the future 

Here is the token list: https://lighthouse.cantonloop.com/tokens

Denend:
All comments answered and changes applied 🫡 @phdargen

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+- `amount`: Integer string representing atomic units (1 CC = 10^10 units).
+  `"1000000000"` = 0.1 CC, `"10000000000"` = 1 CC. Must match exactly (string
+  equality) what the Daml contract encodes.
+- `asset`: `"CC"`. This transfer method settles Canton Coin only.
+- `payTo`: Receiver party id in canonical form `"<name>::<fingerprint>"`.
+- `extra.assetTransferMethod`: MUST be `"external-party-amulet-rules"`.
+- `extra.facilitatorParty`: The facilitator's Canton party id. MUST be set as
```
phdargen: Please rename
```suggestion
- `extra.feePayer`: The facilitator's Canton party id. MUST be set as
```

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,266 @@
...
+    "extra": {
+      "assetTransferMethod": "external-party-amulet-rules",
+      "facilitatorParty": "ftp_facilitator::1220def...",
+      "synchronizerId": "global-domain::1220xyz..."
+    }
+  },
+  "payload": {
+    "transferMethod": "external-party-amulet-rules",
+    "transferCommandCid": "00abc...",
+    "payerParty": "agent_party::1220...",
```
phdargen: Please rename, aligning with SettleResponse convention
```suggestion
    "payer": "agent_party::1220...",
```

specs/schemes/exact/scheme_exact_canton.md
phdargen: I think this can be dropped entirely

Or use an optional invoiceId/nonce, similar to SVM memo: https://github.com/x402-foundation/x402/blob/17bd7e59f95b15f45f93f363bcd99c2b20f0c043/specs/schemes/exact/scheme_exact_svm.md?plain=1#L71

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,266 @@
...
+## Scheme Name
+
+`exact`
+
+## Networks
+
+| Network | Identifier |
+|---|---|
+| Canton MainNet | `canton:mainnet` |
+| Canton TestNet | `canton:testnet` |
```
phdargen, line 24: Are those the officially adopted CAIP2 identifiers? Please double check with the Canton team
https://namespaces.chainagnostic.org/

specs/schemes/exact/scheme_exact_canton.md
phdargen: This is a significant limitation for clients and increased risk for servers. Will need to discuss it with the team

specs/schemes/exact/scheme_exact_canton.md
Denend: Going with optional extra.memo. paymentId adds no real protection (no dedup store, only non-empty check) - replay protection already comes from contract archival. Dropping the mandatory resourceUrl/paymentId check, replacing with optional extra.memo (seller-defined, no facilitator-side matching), same as SVM

Denend:
As far as I understand, there are only two things left to clarify:

1. [Nonces](https://github.com/x402-foundation/x402/pull/2634#discussion_r3441641359)

 I've already reached out to the Canton team. Please ping me as well if you get any updates from your team @phdargen 

2. [CAIP-2 identifiers](https://github.com/x402-foundation/x402/pull/2634#discussion_r3441632467)

   I am waiting for a response from the Canton team

specs/schemes/exact/scheme_exact_canton.md
Denend: I dug a bit deeper and found what seems to be an alternative:

CC supports the Allocation API (CIP-56 Token Standard) as an alternative to TransferCommand.

Flow: client exercises AllocationFactory_Allocate on ExternalPartyAmuletRules (controller = sender) → creates AmuletAllocation backed by LockedAmulet, with settlement.executor set to the facilitator's party. Facilitator is observer on the contract and exercises Allocation_ExecuteTransfer (controller includes executor) to move the funds.

Refs:

- AllocationV1.daml (https://github.com/hyperledger-labs/splice/blob/main/token-standard/splice-api-token-allocation-v1/daml/Splice/Api/Token/AllocationV1.daml) - Allocation interface, SettlementInfo.executor, Allocation_ExecuteTransfer
- AllocationInstructionV1.daml (https://github.com/hyperledger-labs/splice/blob/main/token-standard/splice-api-token-allocation-instruction-v1/daml/Splice/Api/Token/AllocationInstructionV1.daml) - AllocationFactory_Allocate, controller = sender
- AmuletAllocation.daml (https://github.com/hyperledger-labs/splice/blob/main/daml/splice-amulet/daml/Splice/AmuletAllocation.daml) - CC-specific implementation, backed by LockedAmulet
- cip-0056.md (https://github.com/canton-foundation/cips/blob/main/cip-0056/cip-0056.md#L182) - "Canton Coin (CC) implements all APIs of the standard"

Same two-phase model as today (client authorizes, facilitator executes), but no nonce field anywhere in any of the three files above - concurrency is per-holding (LockedAmulet) + allocateBefore/settleBefore deadlines, not a global per-sender counter

What do you think about it?

specs/schemes/exact/scheme_exact_canton.md
Denend: Confirmed with the Canton team - canton:mainnet/canton:testnet is fine. network-id is operator-defined, no single mandated value across Canton networks - these match the names Canton Network itself uses.[^1]

[^1]: https://docs.walletconnect.network/wallet-sdk/chain-support/canton#network-/-chain-information

specs/schemes/exact/scheme_exact_canton.md
Denend: @phdargen We discussed with the Canton team - recommendation is to use Token Standard workflows instead of Amulet-only (TransferCommand). This resolves the nonce issue and also adds support for non-CC payments.

Planning to move to the Allocation API path discussed above, targeting a rollout within the next few days

Ref: CIP-0056 https://github.com/canton-foundation/cips/blob/main/cip-0056/cip-0056.md

Denend:
@phdargen , updated the spec: switched to CIP-56 Allocation API (`allocation-direct`), no sequential nonces, CAIP-2 confirmed with Canton team. All your other comments carried over

Denend:
@phdargen could you please have a look, should be even better now

phdargen:
Thanks a lot for mapping out the alternative approach @Denend, it does seem to fix the sequential nonce issue but in turn introduces a bunch of new ones:

- server needs to mint `MerchantConsent`. Only once per facilitator address but still significant onboarding friction, where servers otherwise only need an address and never worry about wallet management, signing or holding funds for gas
- client needs to lock the funds before even requesting the resource. This adds a second tx increasing per-request latency, for which clients need to pay the tx fees themselves departing from the desired gasless client experience. Only the 2nd tx releasing the escrowed funds would be facilitator sponsored
- on the first {client, server, facilitator} interaction, seems even 4 tx would be needed in total: `SenderConsent` mint + escrow lock submitted by client plus `DirectSettlementConsent` mint + `DirectSettlementConsent_Execute` submitted by facilitator

Not sure whats the better/worse approach here, seems difficult to find a good fit for this scheme given the network-level constrains

Denend:
> Thanks a lot for mapping out the alternative approach @Denend, it does seem to fix the sequential nonce issue but in turn introduces a bunch of new ones:
> 
> * server needs to mint `MerchantConsent`. Only once per facilitator address but still significant onboarding friction, where servers otherwise only need an address and never worry about wallet management, signing or holding funds for gas
> * client needs to lock the funds before even requesting the resource. This adds a second tx increasing per-request latency, for which clients need to pay the tx fees themselves departing from the desired gasless client experience. Only the 2nd tx releasing the escrowed funds would be facilitator sponsored
> * on the first {client, server, facilitator} interaction, seems even 4 tx would be needed in total: `SenderConsent` mint + escrow lock submitted by client plus `DirectSettlementConsent` mint + `DirectSettlementConsent_Execute` submitted by facilitator
> 
> Not sure whats the better/worse approach here, seems difficult to find a good fit for this scheme given the network-level constrains

Thanks for your Review:)

1. "2 tx steady state / only the settle tx is facilitator-sponsored."

Two transactions is the minimal shape on Canton. There is no off-chain spend authorization as on EVM (EIP-3009): the client's allocate is an on-ledger transaction that creates the lock, and the facilitator's execute settles it.

The gas claim is also incorrect. Agents onboard as external parties hosted on the facilitator's participant, and traffic is billed to the submitting participant. During the payment flow, the client submits SenderConsent and allocate (via the relay to our participant), while the facilitator submits the DirectSettlementConsent mint (if needed) and execute. As a result, the client is completely gasless end to end.

The only case where someone other than the facilitator pays traffic is a self-hosted merchant creating its one-time MerchantConsent during onboarding, which is outside the payment flow.

2. "Server needs to mint MerchantConsent."

MerchantConsent is a one-time contract that does not move funds and is not created per payment request.

Like agents, merchants have two onboarding options:

* External (relay-hosted): one CLI command, no node, facilitator-sponsored traffic.
* Self-hosted: if they prefer to run their own participant.

It replaces v1's TransferPreapproval, which was funded and expired, making onboarding strictly simpler rather than adding friction.

3. "First interaction seems even 4 tx: SenderConsent + lock by client, DirectSettlementConsent mint + Execute by facilitator."

Four transactions only occur in the cold-start case (new sender × new merchant):

* SenderConsent: once per sender.
* DirectSettlementConsent: once per {sender, merchant} pair.

After that, the flow amortizes to:

* 3 transactions for a new merchant.
* 2 transactions in steady state.

For relay-hosted agents, every transaction in this flow is facilitator-sponsored.

The lock also applies to a single holding rather than the entire account. The client locks one specific coin into a LockedAmulet bounded by allocateBefore and settleBefore, while the rest of the account remains spendable and the lock expires automatically. This removes v1's sequential-nonce behavior, where a single global sender nonce could block the entire account between verify and settle.

@phdargen

Denend:
@phdargen Just following up on this. Could you please have a look when you have a chance, or tag someone who can help review the PR? Thanks!

phdargen:
> Thanks for your Review:)
> 
> 1. "2 tx steady state / only the settle tx is facilitator-sponsored."
> 
> Two transactions is the minimal shape on Canton. There is no off-chain spend authorization as on EVM (EIP-3009): the client's allocate is an on-ledger transaction that creates the lock, and the facilitator's execute settles it.
> 

hm ok I missed that `CreateTransferCommand` in your v1 approach was an onchain tx, so its 2 txs in v1 vs 2-4 in v2

> The gas claim is also incorrect. Agents onboard as external parties hosted on the facilitator's participant, and traffic is billed to the submitting participant. During the payment flow, the client submits SenderConsent and allocate (via the relay to our participant), while the facilitator submits the DirectSettlementConsent mint (if needed) and execute. As a result, the client is completely gasless end to end.
> 
> The only case where someone other than the facilitator pays traffic is a self-hosted merchant creating its one-time MerchantConsent during onboarding, which is outside the payment flow.
> 

I am not familiar with this concept of a client being 'hosted on the facilitator's participant' and it isnt described anywhere in the spec. how is this hosting relationship established and by whom? Sounds like this would expand the role of the facilitator beyond payment verification/settlement

> 2. "Server needs to mint MerchantConsent."
> 
> MerchantConsent is a one-time contract that does not move funds and is not created per payment request.
> 
> Like agents, merchants have two onboarding options:
> 
> * External (relay-hosted): one CLI command, no node, facilitator-sponsored traffic.
> * Self-hosted: if they prefer to run their own participant.
> 
> It replaces v1's TransferPreapproval, which was funded and expired, making onboarding strictly simpler rather than adding friction.

The MerchantConsent/SenderConsent/DirectSettlementConsent layer appear to be custom infrastructure, unlike TransferPreapproval that is a standard building block. If these are custom delegation contracts their design/implementation would need to be speced out in more detail and would add audit surface

phdargen:
Brainstorming with Fable, it suggested to use `TransferFactory_Transfer` instead of `Allocation`:

> x402 payments are a simple one-directional transfer, not a multi-leg atomic settlement, which is what the DVP/Allocation API is for. `TransferFactory_Transfer` (`TransferInstructionV1`, already implemented for CC on `ExternalPartyAmuletRules`) gets you:
> 
> - controller is `transfer.sender` alone - client's signature is the only authorization needed, no consent contracts at all
> - completes in a single tx when receiver has `TransferPreapproval` (standard, already-tooled, not custom)
> - no escrow - `inputHoldingCids` gives replay/double-spend protection via normal contention on the referenced holdings rather than a shared counter, so no nonce/liveness issue
> - merchant onboarding reuses the existing `TransferPreapproval` flow, no bespoke consent contract
> 
> **Comparison of flows:**
> 
> V1 — TransferCommand (original PR draft, sequential nonce)
> 1. (one-time, merchant) Merchant sets up TransferPreapproval (standard).
> 2. (per payment, client, tx 1) Client exercises ExternalPartyAmuletRules_CreateTransferCommand → creates TransferCommand{sender, receiver, delegate=facilitator, amount, nonce, expiresAt}.
> 3. (per payment, facilitator, tx 2) Facilitator exercises TransferCommand_Send, which checks TransferCommandCounter.nextNonce == nonce before moving funds.
> 
> V2 — Allocation/DVP API (current PR under review)
> 1. (one-time, merchant) Merchant creates bespoke MerchantConsent{merchant, facilitator}.
> 2. (one-time, client) Client creates bespoke SenderConsent{sender, facilitator}.
> 3. (one-time per sender×merchant pair, facilitator) Facilitator mints bespoke DirectSettlementConsent{sender, merchant, facilitator}.
> 4. (per payment, client, tx 1) Client exercises AllocationFactory_Allocate → locks a coin into AmuletAllocation (LockedAmulet, allocateBefore/settleBefore).
> 5. (per payment, facilitator, tx 2) Facilitator exercises DirectSettlementConsent_Execute → transitively authorizes Allocation_ExecuteTransfer → funds move to merchant.
> 
> V3 — TransferFactory_Transfer (FOP, new proposed alternative)
> 1. (one-time, merchant) Merchant sets up TransferPreapproval (standard, same primitive as V1).
> 2. (per payment, single tx) Client prepares + signs a TransferFactory_Transfer{sender, receiver=merchant, amount, instrumentId, executeBefore, inputHoldingCids} — controller is sender alone. Facilitator relays this already-fully-authorized transaction via ExecuteSubmission. One ledger commit, TransferInstructionResult_Completed.

Does this check out? Has TransferFactory_Transfer been considered? Any downsides or tradeoffs to consider?

Denend:
@phdargen 

On the hosting point: Canton is permissioned - a party can only exist on a validator node, and agents don't run their own. So agent parties onboard as external parties on our validator; that's simply how they obtain a wallet to interact with the ledger. Keys remain client-side: the party signs its own transactions, while our participant only submits them and pays the associated traffic costs. It never holds user keys or gains the ability to spend on their behalf, so this doesn't expand the facilitator's role.


On TransferFactory_Transfer: reviewed it and agree - good call. It maps more closely to the classic 402 / EIP-3009 flow (off-ledger authorization, relayer submits and pays the fee). We've reworked the design around it: a single commit transaction, the sender as the sole controller (no custom consents), replay protection handled natively through inputHoldingCids contention (no nonce), and merchant onboarding via the standard TransferPreapproval flow. 

We've updated the PR spec.

Denend:
@phdargen 
We've added the facilitator details and logo to the PR. Could you please take a look when you have time? Thanks!

typescript/site/app/ecosystem/partners-data/ftp/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
```
phdargen: please remove metadata + logo, the ecosystem page is deprecated

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,298 @@
...
+for the Canton Coin instrument, with the merchant as receiver. The preapproval is
+what lets an incoming `TransferFactory_Transfer` resolve **direct** — accepted
+automatically and settled in a single transaction. Without it, the transfer
+resolves to a two-step `TransferInstruction` that the merchant must accept in a
+second transaction, so the facilitator cannot settle in one round-trip and MUST
+reject the payment (see Verification Rule 7).
+
+A `TransferPreapproval` is time-bounded. The merchant creates it once through its
+own wallet/validator (receiver = the merchant) and reuses it for all payers, and
+MUST renew it before expiry to keep the one-transaction path available.
```
phdargen, line 86: Expiry is inconvenient but guess no way around it. How frequent does this need to be renewed? Does the server control the expiry window?

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,298 @@
...
+   (`submissionRef`, with `preparedTxHash`). If absent, reject with
+   `invalid_exact_canton_missing_proof`.
+
+3. **Signature valid.** The referenced submission MUST be a
+   `TransferFactory_Transfer` signed by the payer, and the signature MUST verify
+   against the payer party over `preparedTxHash`. Reject with
+   `invalid_exact_canton_signature_invalid`.
+
+4. **Amount.** The transfer amount MUST equal `paymentRequirements.amount`
+   converted to on-ledger Decimal (1 CC = 1e10 atomic units). Reject with
```
phdargen, line 248: `paymentRequirements.amount` is already in atomic units, decimal conversion shouldnt be needed

Denend: we will resolve it within the next 30 minutes or less, could you please stay with us if its possible:)

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,298 @@
...
+11. **Self-payment guard.** The proven sender MUST NOT equal the facilitator /
+    `feePayer` party. Reject with `invalid_exact_canton_self_payment`.
+
+## Funds Sufficiency
+
+There is no lock or escrow step. The payer-signed transfer names the payer's
+specific input holdings; when the facilitator relays it, the ledger executes the
+transfer atomically and rejects it if those inputs do not cover the amount. The
+facilitator performs no separate off-chain balance read. Insufficient funds
+surface as a relay rejection (`invalid_exact_canton_execute_failed`).
```
phdargen: If simulation is not possible, it should at least check sufficient client balance and if insufficient throw a non-generic error: `invalid_exact_canton_insufficient_balance`

specs/schemes/exact/scheme_exact_canton.md
Denend: Yes - the merchant controls it. The receiver sets expiresAt at creation (defaults to 30 days, overridable). It's a standing per-merchant contract off the payment path, renewed before expiry with the merchant's own key; the prepaid fee scales with the window. If it lapses there's no fund risk and the payer is unaffected — the transfer just resolves to the two-step TransferInstruction, which is rejected per Rule 7, so only the 1-tx path is unavailable until renewal. Separate from the per-payment executeBefore deadline.

specs/schemes/exact/scheme_exact_canton.md
Denend: The wire amount is atomic, but the on-ledger TransferFactory_Transfer amount is a Daml Decimal (10-dp, 1 CC = 1e10), not an atomic integer — unlike exact-evm/svm where the on-chain value is the atomic integer. So Rule 4 crosses atomic<->Decimal once to compare

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,298 @@
...
+- `extra.feePayer`: The facilitator's Canton party id — the party that relays the
+  payer-signed transfer and pays its traffic fee. Clients MUST NOT alter this
+  value.
+- `extra.synchronizerId`: The Global Synchronizer the transfer settles on.
+- `extra.instrumentId`: The Canton Coin instrument identifier
+  `{ "admin": "<DSO-party>", "id": "Amulet" }`.
+- `extra.executeBeforeSeconds`: Relative deadline (seconds from request time) the
+  client uses to compute the absolute `executeBefore` timestamp in the transfer;
+  after it, the signed transfer is no longer executable.
+- `extra.memo` (optional): Seller-defined UTF-8 string, max 256 bytes. When
```
phdargen, line 123: please add facilitator verification rule to check memo when present

specs/schemes/exact/scheme_exact_canton.md
Denend: Because this is a UTXO-style flow: the payer-signed transfer names specific input holdings (inputHoldingCids) that must cover the amount, so sufficiency is already enforced - the client can't even build the transfer without enough CC, and the ledger rejects with ITR_InsufficientFunds at execution if the named inputs don't cover it. So rather than a separate off-chain balance read, we'll map that ledger rejection to a non-generic invalid_exact_canton_insufficient_balance (falling back to invalid_exact_canton_execute_failed for other relay failures). Same specific error you asked for, no extra read, and authoritative since the ledger is what enforces it.

specs/schemes/exact/scheme_exact_canton.md
Denend: Sounds good - will add: when paymentRequirements.extra.memo is set, the facilitator checks the transfer's x402.memo equals it and rejects with invalid_exact_canton_memo_mismatch

Denend:
@phdargen all the requested changed were implemented

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,298 @@
...
+    "assetTransferMethod": "transfer-factory",
+    "payer": "agent_party::1220...",
+    "submissionRef": "...",
+    "preparedTxHash": "1220..."
+  }
+}
+```
+
+- `assetTransferMethod`: `"transfer-factory"`.
+- `payer`: The client's Canton party id — the transfer sender. This is a claim;
```
phdargen: payer shouldnt be trusted and appears unused, should be removed

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,298 @@
...
+  the facilitator binds the *proven* payer from the signed submission (Rule 8).
+- `submissionRef`: Reference to the payer-signed interactive submission — the
+  prepared `TransferFactory_Transfer` together with the payer's signature — that
+  the facilitator relays. An implementation MAY instead carry the prepared
+  transaction and signature inline; the only requirement is that the facilitator
+  can obtain the payer-signed transfer in order to submit it.
```
phdargen: Too much ambiguity here that would lead to interop gaps. Please define a single canonical shape

Denend: working on that one right now

phdargen: my agent recommended: 
```
"payload": {
  "transaction": "CiVjb20uZGlnaXRhbGFzc2V0...",
  "signature": "0xa1b2c3..."
}
transaction: base64-encoded serialized PreparedTransaction, containing exactly one TransferFactory_Transfer
signature: the payer's raw signature over the hash recomputed from transaction
```

Denend: Your inline {transaction, signature} works - the facilitator derives the payer from the transaction's sender (Rule 8) and wraps the raw signature into Canton's partySignatures, so those don't need to be on the wire. One nuance: the hash is hashing-scheme-dependent, so a 2-field shape implicitly pins the scheme. That's fine today (the network is on V2), but to stay robust if a payer prepares under V1 or the network moves to V3, I'd add an optional third field, hashingSchemeVersion (default HASHING_SCHEME_VERSION_V2). Signature is ed25519, base64. So:

"payload": {
  "transaction": "<base64 PreparedTransaction, exactly one TransferFactory_Transfer>",
  "signature": "<base64 ed25519 sig over the prepared-tx hash>",
  "hashingSchemeVersion": "HASHING_SCHEME_VERSION_V2"
}

If you'd rather keep it to two fields, we can pin the scheme to V2 in the spec - either approach works.

Denend:
@phdargen , payer removed, waiting on your decision on the payload content

specs/schemes/exact/scheme_exact_canton.md
phdargen: ok sounds good, hashingSchemeVersion can be added

Denend: will be done within 40 minutes or less

phdargen:
It appears there are 2 different fees in play, and the facilitator only sponsors one of them.

`Traffic fees`: the cost of submitting the transaction to the synchronizer. This is what your spec's `feePayer` covers: the facilitator relays the transaction, so its validator burns the traffic. That part is fine 

`Amulet transfer fees`: a separate fee deducted on-ledger from the sender's input holdings as part of executing TransferFactory_Transfer. The facilitator can't sponsor these,  they're not a submission cost but part of the token's transfer logic and come out of the payer's CC.

So if the payer signs a transfer with amount = 0.1 CC, the question the spec never answers is: does the merchant receive exactly 0.1 CC with the payer's inputs also covering the Amulet fees on top, or does the merchant receive 0.1 CC minus fees?

The spec should state that `amount` denotes the amount credited to the merchant, that the payer's inputs bear all Amulet-level fees, and that verification confirms inputs cover amount + estimated fees (if possible).

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,302 @@
...
+## Funds Sufficiency
+
+There is no lock or escrow step. The payer-signed transfer names the payer's
+specific input holdings, and those holdings enforce sufficiency: the client cannot
+build the transfer without inputs that cover the amount, and when the facilitator
+relays it the ledger executes atomically and rejects it if the named inputs do not
+cover the amount. The facilitator performs no separate off-chain balance read — it
+maps the ledger's insufficient-funds rejection to
+`invalid_exact_canton_insufficient_balance`, and any other relay failure to
+`invalid_exact_canton_execute_failed`.
```
phdargen: Suggest to replace with Rule 13:

```suggestion
13. **Input sufficiency. ** The input holdings embedded in the prepared transaction MUST have distinct contract ids and sum to at least amount plus the transfer's Amulet fees. When the facilitator's participant hosts the proven sender with read access, it SHOULD additionally verify each input holding is active in its ledger view. Reject with `invalid_exact_canton_insufficient_inputs`. Absent the hosted-payer check, verification does not establish the inputs are unspent; that is enforced only at settlement. Servers should therefore be aware that a passing /verify leaves residual risk of wasted resource-handler work if settlement subsequently fails.
```

specs/schemes/exact/scheme_exact_canton.md
Denend: done

Denend:
> It appears there are 2 different fees in play, and the facilitator only sponsors one of them.
> 
> `Traffic fees`: the cost of submitting the transaction to the synchronizer. This is what your spec's `feePayer` covers: the facilitator relays the transaction, so its validator burns the traffic. That part is fine
> 
> `Amulet transfer fees`: a separate fee deducted on-ledger from the sender's input holdings as part of executing TransferFactory_Transfer. The facilitator can't sponsor these, they're not a submission cost but part of the token's transfer logic and come out of the payer's CC.
> 
> So if the payer signs a transfer with amount = 0.1 CC, the question the spec never answers is: does the merchant receive exactly 0.1 CC with the payer's inputs also covering the Amulet fees on top, or does the merchant receive 0.1 CC minus fees?
> 
> The spec should state that `amount` denotes the amount credited to the merchant, that the payer's inputs bear all Amulet-level fees, and that verification confirms inputs cover amount + estimated fees (if possible).

Good distinction, and you're right there are two fees. Traffic fee - the facilitator (feePayer) sponsors it, as you noted. Amulet transfer fees - the mechanism exists in the token (outputFees/senderChangeFee/holdingFees), but CIP-0078 reduced Canton Coin transfer/lock fees to zero, so on MainNet they're currently 0.

Semantics: amount is credited to the merchant in full, and the payer's input holdings bear any Amulet-level fees - never amount minus fees. Rule 13 already covers the verification side (inputs must sum to ≥ amount + the transfer's Amulet fees), which holds whether fees are zero (today) or non-zero.

CIP-0078: https://github.com/canton-foundation/cips/blob/main/cip-0078/cip-0078.md

Denend:
@phdargen , hashingSchemeVersion was also added as requested and Rule 13 changes applied

specs/schemes/exact/scheme_exact_canton.md
```diff
@@ -0,0 +1,315 @@
...
```
phdargen: this is now covered by and partially contradicts Rule 13, should be removed

phdargen:
OK good to know the amulet fees are currently 0, thats great

specs/schemes/exact/scheme_exact_canton.md
Denend: Before we lock in the inline shape, I wanted to share what a measurement turned up. Across 7.4k real settlements, a prepared TransferFactory_Transfer averages ~29.5 KB; once base64-encoded into the X-PAYMENT header, the payload grows to ~40 KB, exceeding the common 8-16 KB header limits on nginx, Node, and Cloudflare. So a canonical inline payload wouldn't make it through a typical merchant stack.

On Canton, the interoperability benefit of an inline payload is also limited. The payer is an external party, and interactive-submission execute is bound to the participant hosting that party, so only that facilitator can submit the signed transaction, regardless of whether the payload is inline or reference-based.

Your point about eliminating the "ref OR inline" ambiguity is still valid. One option that fits comfortably within header limits (~0.8 KB) is to make the reference form canonical and carry the signature in the payload, which also removes the extra commit round trip:

"payload": {
  "assetTransferMethod": "transfer-factory",
  "submissionRef": "<facilitator-issued ref to the prepared TransferFactory_Transfer>",
  "preparedTxHash": "<hex hash of the prepared tx the payer signed>",
  "signature": "<base64 ed25519 sig over preparedTxHash>"
}

The facilitator resolves submissionRef to the prepared transaction it holds, verifies the recomputed hash matches preparedTxHash, checks the signature against the payer, and then relays the transaction. Happy to go whichever way you prefer - I just wanted to put the measurements on the table before we lock the format in

specs/schemes/exact/scheme_exact_canton.md
Denend: will do it within 30 minutes or less

specs/schemes/exact/scheme_exact_canton.md
phdargen: good callout, yeah that header size would probably be too large for most deployments. Any insights where exactly the large payload size comes from?

Your alternative proposal would require that the facilitator hosts the client and that the facilitator stores tx it prepared, both of which should be avoided

Denend: dug more into it. of a real prepared tx (22 KB raw, consistent across 7.4k mainnet payments) ~15 KB is the daml transaction itself, 10 nodes for one transfer (splice amulet expansion), and ~7 KB is metadata.input_contracts, four embedded contract payloads the V2 hashing scheme needs so the signature covers what the tx consumes.

one update from measuring: brotli compresses it to ~3.9 KB, so compressed inline is ~8-9 KB of header. better than raw 40 KB but still above default nginx per-header limits, so agreed, impractical as canonical.

on your concerns: hosting is orthogonal to payload shape. an interactive submission can only be executed by a participant hosting the payer party, so even an inline payload comes back to the facilitator that prepared it. inline removes the stash, not the hosting.

statefulness, agreed. proposal: canonical payload = { assetTransferMethod, submissionRef, preparedTxHash, signature, hashingSchemeVersion } (~0.8 KB). the signature moves into the payload so authorization is self-contained and the commit step disappears; the prepared tx becomes a short-lived prepare cache, unknown or expired ref -> submission_not_found -> client re-prepares

specs/schemes/exact/scheme_exact_canton.md
Denend: done

Denend:
@phdargen removed the Rule 13, waiting for your reply on this one https://github.com/x402-foundation/x402/pull/2634#discussion_r3552608384

specs/schemes/exact/scheme_exact_canton.md
phdargen: The hosting discussion is confusing and conflicts with the earlier balance discussion, where you said a balance check is only possible if the facilitator hosts the client implying this is not always the case. Could you clarify whether execution requires the payer party to be hosted on the executing facilitator’s participant?

More fundamentally, x402 is an open protocol: servers may use any facilitator or self-facilitate, and that relationship is opaque to the client. The client cannot be assumed to be hosted by or hold state with the server’s facilitator. If the payment flow truly requires that coupling, we may need to revisit the design again

Denend: Yes - execution is bound to the participant hosting the payer. `prepare` builds the transaction in the payer's view, and `execute` submits it through that participant. The facilitator is not that host - it's a thin proxy in front of a Canton participant that simply calls the participant's interactive-submission API. In our deployment, the facilitator and the validator/participant even run on separate servers.

That coupling isn't imposed by x402; it's a property of Canton being permissioned. Agents don't run their own nodes, so they must onboard as external parties on some validator's participant. Which participant they use is chosen by whoever operates the agent (the merchant, a facilitator provider, or another third party) and is opaque to the server — it is not tied to the server's facilitator. Our facilitator is open source (Apache-2.0), so anyone can deploy their own participant + facilitator and onboard agents there; nothing requires using ours.

The protocol remains open where it matters: `verify` is just a read of the on-ledger transaction via any public SV Scan, so it is completely facilitator-agnostic. Only the gasless relayed settlement assumes the payer is hosted on the participant the relaying facilitator talks to, which for agents is simply the participant they chose when onboarding

Denend:
@phdargen hi, the last comment is answered, could you have a look when you have time

specs/schemes/exact/scheme_exact_canton.md
phdargen: Servers choose a facilitator, this choice is opaque to the client. There might be many facilitators. 
Clients cant and shouldnt know to which participant the relaying facilitator talks to. Servers cant and shouldnt know where clients are hosted. If a given client can only make payments to servers that use a facilitator which happens to be compatible, this would be a serious interop problem. Open-source or not.

Checked the Canton docs on this: only prepare needs a participant hosting the party, submission can happen through any node ("Submission can happen through any node, not just one that hosts the party." , https://docs.canton.network/appdev/deep-dives/external-signing). So the coupling isn't inherent, it comes from the facilitator doing the prepare, which is only necessary for the submissionRef workaround for the oversized header problem.

Client should prepare + sign via its own validator and send the complete signed prepared tx inline; then any facilitator can relay it and neither side needs to know which validator the other uses. That means dropping submissionRef entirely, as this only works when client and facilitator happen to share a participant. The facilitator should remain stateless and there should be no direct client↔facilitator interaction.

The remaining problem is to figure out how to fit the client payload in the header

Denend: We dug into the inline direction. A few findings before deciding.

Size, measured on 3,000+ real mainnet prepared txs: ~22 KB median, 27 KB max.
As base64 in `PAYMENT-SIGNATURE` that's ~29.5 KB, over any header cap, so
uncompressed inline doesn't fit HTTP. Trimming disclosed contracts doesn't help
(only ~6.8 KB of it, and dropping them breaks execute). gzip does: prepared txs
are ~5x compressible, so gzip+base64 lands at ~5.9 KB across the whole
distribution, under the 8 KB default, though some proxies cap headers lower.

On "any facilitator can relay it": looks possible. Canton has a
`CanExecuteAsAnyParty` user right, and the execute authorization path checks
`executeAs == submitterInfo.actAs` with no hosting check, so a facilitator can
execute a payer's tx given that grant on its participant, without hosting the
payer. Source:
https://github.com/digital-asset/canton/blob/main/community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/ledger/api/auth/services/InteractiveSubmissionServiceAuthorization.scala

Given that inline is ~5.9 KB compressed but plain HTTP carries the payload in a
header, how do you see this fitting the spec? Curious what you think the right
shape is: compressed payload, some reference, or another approach

Denend:
@phdargen please take a look at our last comment https://github.com/x402-foundation/x402/pull/2634#discussion_r3604831047

phdargen:
> On "any facilitator can relay it": looks possible. Canton has a `CanExecuteAsAnyParty` user right, and the execute authorization path checks `executeAs == submitterInfo.actAs` with no hosting check, so a facilitator can execute a payer's tx given that grant on its participant, without hosting the payer. Source: https://github.com/digital-asset/canton/blob/main/community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/ledger/api/auth/services/InteractiveSubmissionServiceAuthorization.scala
> 

OK, please confirm with the Canton team that a client hosted on participant A can submit a transaction via a facilitator operating through participant B, where no relationship between the client and Participant B is required.

My current understanding of the flow is:

- Participant A prepares the transaction using the client’s contract state
- The client signs the complete prepared transaction
- The facilitator sends the prepared transaction and signature to Participant B’s ExecuteSubmission
- Participant B must have canExecuteAsAnyParty API rights
- Participant B forwards the transaction without hosting or signing on behalf of the client
- Participant A, along with any other required confirming participants, validates and confirms the transaction
- Participant B receives the final completion, allowing the facilitator to return settlement success
- Participant B pays the traffic cost of submitting the transaction

phdargen:
> We dug into the inline direction. A few findings before deciding.
> 
> Size, measured on 3,000+ real mainnet prepared txs: ~22 KB median, 27 KB max. As base64 in `PAYMENT-SIGNATURE` that's ~29.5 KB, over any header cap, so uncompressed inline doesn't fit HTTP. Trimming disclosed contracts doesn't help (only ~6.8 KB of it, and dropping them breaks execute). gzip does: prepared txs are ~5x compressible, so gzip+base64 lands at ~5.9 KB across the whole distribution, under the 8 KB default, though some proxies cap headers lower.
> 
> Given that inline is ~5.9 KB compressed but plain HTTP carries the payload in a header, how do you see this fitting the spec? Curious what you think the right shape is: compressed payload, some reference, or another approach

Will discuss with the team if the gzip approach is acceptable, at a minimum this would require the following security checks: 

- Decompression bombs: attacker sends a tiny gzip value expanding enormously. Set both compressed and decompressed limits and decompress incrementally. For example, limits around 8 KiB compressed and 64 KiB decompressed
- CPU/memory exhaustion: stop immediately when the output limit is reached; also bound protobuf nesting/node count and parsing time
- Ambiguous gzip input: require exactly one gzip member, reject trailing bytes and concatenated members, and prohibit optional filename/comment metadata

Denend:
> Will discuss with the team if the gzip approach is acceptable, at a minimum this would require the following security checks:
> 
> * Decompression bombs: attacker sends a tiny gzip value expanding enormously. Set both compressed and decompressed limits and decompress incrementally. For example, limits around 8 KiB compressed and 64 KiB decompressed
> * CPU/memory exhaustion: stop immediately when the output limit is reached; also bound protobuf nesting/node count and parsing time
> * Ambiguous gzip input: require exactly one gzip member, reject trailing bytes and concatenated members, and prohibit optional filename/comment metadata

Sounds good @phdargen - solid points. We'll add those 3 as MUST checks as soon as your team confirms the gzip path is OK

phdargen:
Replaying @CarsonRoscoe feedback here:

> Confirmed no authorization flow, and fully agree on dropping submissionRef. This does put the ball in the clients court to provide the execution bundle, that's clear to me now.
> 
> Few thoughts though:
> 
> - I believe we can trim the four complete metadata.input_contracts from the payload, and have the facilitator reconstitute this state. From this payload, only the private amulet holdings genuinely have to travel inline, the AmuletRules and the mining rounds are public DSO data any facilitator can pull from an SV scan. If the executing participant can complete the disclousre set on its own/Scan view and only the private holdings ride in the header, the payload could drop 6-7kb
> - Inline zip, provided we can add reasonable protections against zip bomb
> 
> These two together should land the final payload closer to the 4kb range, which gives sufficient headroom in our 8kb header limit we strive for

Denend:
@phdargen , updated the spec to inline: the signed prepared tx now travels gzip-compressed inline in the payload (no facilitator-held ref), so any facilitator can relay it. Added a Payload Encoding & Limits section with the three checks. Lmk if the shape looks right

Denend:
> Will discuss with the team if the gzip approach is acceptable, at a minimum this would require the following security checks:
> 
> * Decompression bombs: attacker sends a tiny gzip value expanding enormously. Set both compressed and decompressed limits and decompress incrementally. For example, limits around 8 KiB compressed and 64 KiB decompressed
> * CPU/memory exhaustion: stop immediately when the output limit is reached; also bound protobuf nesting/node count and parsing time
> * Ambiguous gzip input: require exactly one gzip member, reject trailing bytes and concatenated members, and prohibit optional filename/comment metadata


Agreed on gzip; on the trim, two caveats

The trim works cryptographically, which was my worry. The V2 metadata hash covers input_contracts but hashes the decoded create node (contractId, packageName, templateId, argument, signatories, stakeholders, createdAt), not the blob bytes. A facilitator rebuilding AmuletRules and the round from Scan by cid lands on the same hash, so the payer signature still verifies

First caveat - 6-7kb is uncompressed. Measured on real mainnet txs, 22kb raw goes to 4.4kb gzipped and 5.9kb as base64, and the AmuletRules argument is the most compressible part of the payload. After gzip the trim buys roughly a kilobyte, not six. gzip alone already fits the 8kb budget

Second - the client signs one specific OpenMiningRound cid and the facilitator has to rebuild that exact one or the hash diverges. Scan serves current rounds, so a round rolling between prepare and settle leaves nothing to rebuild and the payment fails where it works today. Narrow window, not zero, wants a documented fallback. Reconstitution also turns "any facilitator can relay this" into "any facilitator with a Scan view", worth stating in the spec rather than leaving implicit

We could land gzip now with the limits already written in, and I'll bring a measured gzip-with vs gzip-without delta so the trim gets decided on compressed bytes. Open either way - whether the executing participant needs anything outside the hashed set (driver metadata on the disclosed contracts) to complete the disclosure. a live test answers that

Denend:
@phdargen could you please take a look at the last changes when you have time

Denend:
@CarsonRoscoe , could we please help us to close the remaining design decision on this PR?

Denend:
To sum up - I think the spec is ready. The current shape is gzip-inline + the three MUST security checks, self-contained (no trim) so any facilitator can relay without a Scan view.

On the trim: we left it out on purpose (after gzip it saves ~1KB not 6, and it adds a round-roll failure mode + turns "any facilitator" into "any facilitator with a Scan view"). Happy to revisit it later as an optimization, but it doesn't block merge or implementation. JFYI.

 If you disagree with that take, let me know, otherwise I think this is good to merge.

phdargen:
> To sum up - I think the spec is ready. The current shape is gzip-inline + the three MUST security checks, self-contained (no trim) so any facilitator can relay without a Scan view.
> 
> On the trim: we left it out on purpose (after gzip it saves ~1KB not 6, and it adds a round-roll failure mode + turns "any facilitator" into "any facilitator with a Scan view"). Happy to revisit it later as an optimization, but it doesn't block merge or implementation. JFYI.
> 
> If you disagree with that take, let me know, otherwise I think this is good to merge.

OK agree its probably not worth the added complexity to trim, dont think we could reliably land under the 8kb budget either way. 

- Prepared transaction: ~22 KB - 27 KB
- gzip(Prepared transaction): 4.4 - 5.4 KB 
- base64(gzip(Prepared transaction)): ~5.9 KB - 7.2 KB 
- PaymentPayload JSON: ~6.6–7.9 KB (assuming minimal resource + accepted metadata, no extensions)
- Final base64-encoded header: ~8.8–10.5 KB

Even if we trim, it would typically exceed and leave little to no headroom for additional extension metadata. 

Don't really see a way around it, lets at least add sth like this to the spec:
`HTTP servers and intermediaries supporting this scheme SHOULD accept a PAYMENT-SIGNATURE header of at least 16 KiB end to end. This provides headroom for the full payload and extension metadata: base64(gzip(preparedTransaction)) alone is typically 6–7 KiB, and encoding the complete PaymentPayload commonly produces a header exceeding 8 KiB. Deployments that cannot accept a 16 KiB PAYMENT-SIGNATURE header SHOULD use a transport, such as MCP, that does not carry the PaymentPayload in an HTTP header.`

phdargen:
> > On "any facilitator can relay it": looks possible. Canton has a `CanExecuteAsAnyParty` user right, and the execute authorization path checks `executeAs == submitterInfo.actAs` with no hosting check, so a facilitator can execute a payer's tx given that grant on its participant, without hosting the payer. Source: https://github.com/digital-asset/canton/blob/main/community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/ledger/api/auth/services/InteractiveSubmissionServiceAuthorization.scala
> 
> OK, please confirm with the Canton team that a client hosted on participant A can submit a transaction via a facilitator operating through participant B, where no relationship between the client and Participant B is required.
> 
> My current understanding of the flow is:
> 
> * Participant A prepares the transaction using the client’s contract state
> * The client signs the complete prepared transaction
> * The facilitator sends the prepared transaction and signature to Participant B’s ExecuteSubmission
> * Participant B must have canExecuteAsAnyParty API rights
> * Participant B forwards the transaction without hosting or signing on behalf of the client
> * Participant A, along with any other required confirming participants, validates and confirms the transaction
> * Participant B receives the final completion, allowing the facilitator to return settlement success
> * Participant B pays the traffic cost of submitting the transaction

Have you confirmed this with the Canton team @Denend?

Denend:
> > To sum up - I think the spec is ready. The current shape is gzip-inline + the three MUST security checks, self-contained (no trim) so any facilitator can relay without a Scan view.
> > On the trim: we left it out on purpose (after gzip it saves ~1KB not 6, and it adds a round-roll failure mode + turns "any facilitator" into "any facilitator with a Scan view"). Happy to revisit it later as an optimization, but it doesn't block merge or implementation. JFYI.
> > If you disagree with that take, let me know, otherwise I think this is good to merge.
> 
> OK agree its probably not worth the added complexity to trim, dont think we could reliably land under the 8kb budget either way.
> 
> * Prepared transaction: ~22 KB - 27 KB
> * gzip(Prepared transaction): 4.4 - 5.4 KB
> * base64(gzip(Prepared transaction)): ~5.9 KB - 7.2 KB
> * PaymentPayload JSON: ~6.6–7.9 KB (assuming minimal resource + accepted metadata, no extensions)
> * Final base64-encoded header: ~8.8–10.5 KB
> 
> Even if we trim, it would typically exceed and leave little to no headroom for additional extension metadata.
> 
> Don't really see a way around it, lets at least add sth like this to the spec: `HTTP servers and intermediaries supporting this scheme SHOULD accept a PAYMENT-SIGNATURE header of at least 16 KiB end to end. This provides headroom for the full payload and extension metadata: base64(gzip(preparedTransaction)) alone is typically 6–7 KiB, and encoding the complete PaymentPayload commonly produces a header exceeding 8 KiB. Deployments that cannot accept a 16 KiB PAYMENT-SIGNATURE header SHOULD use a transport, such as MCP, that does not carry the PaymentPayload in an HTTP header.`

on it right now, could you please stay here so we can try to get it done today:)

Denend:
> Have you confirmed this with the Canton team @Denend?

Confirmed against the Canton sources, and it's the flow our facilitator already runs. All 8 steps are correct:

1. **A prepares from the client's ACS** — prepare needs only a `readAs` claim, no hosting gate. [1][4]
2. **Client signs the prepared-tx hash** with its own key (standard external signing). [5]
3. **Facilitator forwards the prepared tx + signature to B's `ExecuteSubmission`.** [4]
4. **B needs `CanExecuteAsAnyParty`** (or `CanExecuteAs(party)`) — execute checks `executeAs == submitterInfo.actAs` as a pure claims check, no hosting/topology lookup. [1][2][3]
5. **B forwards without hosting or signing for the client**; it never holds the client's key. [1][6]
6. **A confirms** (Confirmation permission, default threshold 1). Submission permission isn't available for external parties, so B neither can nor needs to confirm. [7]
7. **B, as submitter, receives the command completion (`updateId`)**, so the facilitator returns success, no stakeholder visibility needed. [4]
8. **B pays the traffic for the submission** (the "sponsored gas"): traffic is charged to the sender of each submission request. Confirmation responses (A's) are separately reimbursed under CIP-104, but the submission is B's cost. [8]

And yes, no relationship between the client and B is required: B only needs the `CanExecuteAsAnyParty` grant on its own participant; there's no `PartyToParticipant` mapping tying the client to B.

**Refs:**
- [1] [InteractiveSubmissionServiceAuthorization.scala](https://github.com/digital-asset/canton/blob/main/community/ledger/ledger-api-core/src/main/scala/com/digitalasset/canton/ledger/api/auth/services/InteractiveSubmissionServiceAuthorization.scala) (prepare = readAs only; execute = `executeAs == actAs`, no hosting check)
- [2] [Claims.scala](https://github.com/digital-asset/canton/blob/main/community/base/src/main/scala/com/digitalasset/canton/auth/Claims.scala) (`canExecuteAs`, pure claims check)
- [3] [UserRight.scala](https://github.com/digital-asset/canton/blob/main/community/base/src/main/scala/com/digitalasset/canton/user/UserRight.scala) (`CanExecuteAsAnyParty`)
- [4] [Local and external parties (PPN/EPN/CPN)](https://docs.canton.network/overview/reference/external-party)
- [5] [External Signing: Submitting Transactions](https://docs.canton.network/appdev/deep-dives/external-signing-transactions)
- [6] [External Signing](https://docs.canton.network/appdev/deep-dives/external-signing) ("submission can happen through any node, not just one that hosts the party")
- [7] [Multi-Hosting and Resilience](https://docs.canton.network/appdev/deep-dives/multi-hosting) (Confirmation permission; threshold 1; Submission only for local parties)
- [8] [Canton traffic management](https://github.com/digital-asset/canton/blob/main/docs-open/src/sphinx/overview/explanations/canton/traffic-management.rst) (traffic charged to the sender of each submission request; accounted per participant)

Denend:
> HTTP servers and intermediaries supporting this scheme SHOULD accept a PAYMENT-SIGNATURE header of at least 16 KiB end to end. This provides headroom for the full payload and extension metadata: base64(gzip(preparedTransaction)) alone is typically 6–7 KiB, and encoding the complete PaymentPayload commonly produces a header exceeding 8 KiB. Deployments that cannot accept a 16 KiB PAYMENT-SIGNATURE header SHOULD use a transport, such as MCP, that does not carry the PaymentPayload in an HTTP header.

Changed the Scheme as requested @phdargen

phdargen (approved):
Alright let's get this in @Denend  🚀

Denend:
@phdargen thanks again for the review. Now that the exact-canton scheme is merged, here's our plan so it lands as the reference implementation:

1. TypeScript @x402/canton mechanism package with tests and a runnable example.
2. Our production facilitator listed in the facilitators directory.

We already run this end to end on Canton mainnet with live merchants and a working facilitator, so we're positioned to ship these quickly.

Denend:
@phdargen, we have completed both follow-ups as promised:

• Added the TypeScript @x402/canton mechanism package, including tests and a runnable example: https://github.com/x402-foundation/x402/pull/3242

• Added our production facilitator to the facilitators directory: https://github.com/x402-foundation/x402/pull/3243

Could you please take a look when you have a chance? We would appreciate your review and any feedback

PR #2635, opened 2026-06-15
fix(evm): clarify client signer shape

### LINKED ISSUE #2631: @x402/evm: example passes a viem WalletClient, but signer.address is read (undefined) → misleading `Address "undefined" is invalid`
Package: `@x402/evm` (v2.15.0), used via `@x402/fetch`.

**Summary:** Following the documented example with a viem `WalletClient` as the signer fails, because the EIP-3009 signing path reads `signer.address` — which is `undefined` on a `WalletClient` (the address lives at `signer.account.address`). The result is a misleading `Address "undefined" is invalid`, and the `@x402/fetch` wrapper masks it behind a generic `Failed to create payment payload`.

**Root cause:** the signer is expected to expose `.address` (and `.signTypedData`) — a viem `LocalAccount` has these, but a `WalletClient` does not (it has `.account.address`).

**Fix / workaround:** pass a viem `LocalAccount` (e.g. `privateKeyToAccount(...)`), or wrap with `toClientEvmSigner(account, publicClient)`.

**Suggested improvements:**
- Update the docstring/example to show a `LocalAccount`, not a `WalletClient`.
- Throw a clearer error (e.g. "signer must expose `.address` — pass a LocalAccount, not a WalletClient") instead of letting `Address "undefined" is invalid` bubble up from the typed-data call.

Found while an autonomous agent hand-rolled an x402 v2 `exact` / `eip155:8453` payment — an agent following the example verbatim gets stuck here. Thanks for x402!

go165:
## Summary
- allow `toClientEvmSigner` to compose from viem WalletClient-style `account.address`
- validate missing signer address with an actionable error before payment signing
- update EVM signer docs and add unit coverage for WalletClient-style input

Fixes #2631.

## Validation
- `node .../prettier@3.5.2/.../prettier.cjs --write typescript/packages/mechanisms/evm/src/signer.ts typescript/packages/mechanisms/evm/test/unit/signer.test.ts`
- `git diff --check`
- Blocked: `CI=true corepack pnpm install --filter @x402/evm...` failed fetching tarballs from the configured npm mirror (`https://mirrors.tencentyun.com/npm/...`) with `TypeError: fetch failed`.
- Blocked: direct Vitest startup fails before tests run because Rollup optional native package `@rollup/rollup-linux-x64-gnu` is missing from the existing install.

### DIFF
--- a/typescript/packages/mechanisms/evm/src/signer.ts
+++ b/typescript/packages/mechanisms/evm/src/signer.ts
@@ -3,13 +3,9 @@ import type { Log } from "viem";
 /**
  * ClientEvmSigner - Used by x402 clients to sign payment authorizations.
  *
- * Typically a viem WalletClient extended with publicActions:
+ * Typically a viem LocalAccount:
  * ```typescript
- * const client = createWalletClient({
- *   account: privateKeyToAccount('0x...'),
- *   chain: baseSepolia,
- *   transport: http(),
- * }).extend(publicActions);
+ * const account = privateKeyToAccount('0x...');
  * ```
  *
  * Or composed via `toClientEvmSigner(account, publicClient)`.
@@ -107,13 +103,11 @@ export type FacilitatorEvmSigner = {
  * Use this when your signer (e.g., `privateKeyToAccount`) doesn't have
  * `readContract`. The `publicClient` provides the on-chain read capability.
  *
- * Alternatively, use a WalletClient extended with publicActions directly:
+ * Alternatively, use a local account with an explicit public client:
  * ```typescript
- * const signer = createWalletClient({
- *   account: privateKeyToAccount('0x...'),
- *   chain: baseSepolia,
- *   transport: http(),
- * }).extend(publicActions);
+ * const account = privateKeyToAccount('0x...');
+ * const publicClient = createPublicClient({ chain: baseSepolia, transport: http() });
+ * const signer = toClientEvmSigner(account, publicClient);
  * ```
  *
  * @param signer - A signer with `address` and `signTypedData` (and optionally `readContract`)


typescript/packages/mechanisms/evm/src/signer.ts
```diff
@@ -3,16 +3,15 @@ import type { Log } from "viem";
...
- *   chain: baseSepolia,
- *   transport: http(),
- * }).extend(publicActions);
+ * const account = privateKeyToAccount('0x...');
+ * const signer = toClientEvmSigner(account, publicClient);
  * ```
  *
- * Or composed via `toClientEvmSigner(account, publicClient)`.
+ * viem WalletClient objects expose the address at `account.address`, so pass
+ * them through `toClientEvmSigner(walletClient)` before registration.
```
phdargen:
```suggestion
 * Typically a viem LocalAccount:
 * ```typescript
 * const account = privateKeyToAccount('0x...');
 * ```
 *
 * Or composed via `toClientEvmSigner(account, publicClient)`.
```

typescript/packages/mechanisms/evm/src/signer.ts
```diff
@@ -57,6 +56,12 @@ export type ClientEvmSigner = {
   estimateFeesPerGas?(): Promise<{ maxFeePerGas: bigint; maxPriorityFeePerGas: bigint }>;
 };
 
+export type ClientEvmSignerLike = Omit<ClientEvmSigner, "address" | "readContract"> & {
+  readonly address?: `0x${string}`;
+  readonly account?: { readonly address?: `0x${string}` } | null;
+  readContract?: ClientEvmSigner["readContract"];
+};
+
```
phdargen: please revert all the code changes, these are not needed

Only the inaccurate doc strings should be fixed

phdargen (changes requested):

phdargen (approved):
Thanks @go165

PR #2693, opened 2026-06-24
feat(svm): server-provided recent blockhash in the exact 402 challenge

lgalabru:
## Description

Let the resource server embed a recent blockhash in the `exact` 402 challenge so the client can build its payment transaction without its own RPC round-trip and against a blockhash the settling RPC has already observed.

- utils: `resolveBlockhash(rpc, requirements)` — prefer `extra.recentBlockhash` (+ `extra.lastValidBlockHeight`) from the challenge, fall back to `getLatestBlockhash`. The `exact` client uses it instead of an unconditional RPC call.
- exact/server: `ExactSvmScheme` takes an optional `{ rpcUrl }`. When set, `enhancePaymentRequirements` fetches a fresh blockhash and emits `extra.recentBlockhash` + `extra.lastValidBlockHeight` alongside `feePayer` (best-effort — omitted on RPC failure so the client falls back). Threaded through `registerExactSvmScheme({ rpcUrl })`.
- spec: document the optional `extra.recentBlockhash` / `extra.lastValidBlockHeight` fields and the client's use of them in scheme_exact_svm.md.

## Tests

  The change adds a dedicated unit test, test/unit/server.blockhash.test.ts, covering the server's blockhash
  injection in the 402 challenge:
  - When rpcUrl is set, enhancePaymentRequirements emits extra.recentBlockhash + extra.lastValidBlockHeight
  alongside feePayer.
  - On RPC failure the fields are omitted (best-effort) so the client falls back to getLatestBlockhash via
  resolveBlockhash.

  Ran from typescript/packages/mechanisms/svm:

```console
  $ npx vitest run
   ✓ test/unit/server.blockhash.test.ts (2 tests)
```

## Checklist

### DIFF
--- a/specs/schemes/exact/scheme_exact_svm.md
+++ b/specs/schemes/exact/scheme_exact_svm.md
@@ -31,8 +31,8 @@ How transaction fees are sponsored—and how the sponsoring party evaluates risk
 The protocol flow for `exact` on Solana is client-driven.
 
 1. Client makes a request to a Resource Server.
-2. Resource Server responds with a payment required signal containing `PaymentRequired`. The `extra` field contains a `feePayer`, identifying the sponsor.
-3. Client creates a transaction that effects a payment of an asset to the merchant for a specified amount.
+2. Resource Server responds with a payment required signal containing `PaymentRequired`. The `extra` field contains a `feePayer`, identifying the sponsor, and MAY contain a `recentBlockhash` for the Client to build against.
+3. Client creates a transaction that effects a payment of an asset to the merchant for a specified amount. If `extra.recentBlockhash` is present the Client uses it as the transaction lifetime; otherwise the Client fetches a recent blockhash itself.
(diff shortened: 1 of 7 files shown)


wnjoon:
@lgalabru 

Thanks for this TS implementation. I’m going to work on the Go SVM exact parity for the server-provided recent blockhash behavior in a separate PR (#2731).

I’ll keep the Go change scoped to the same shape:
- let the Go exact client prefer `extra.recentBlockhash` when present, with the current RPC fetch as fallback
- let the Go exact server optionally fetch and emit `extra.recentBlockhash` + `extra.lastValidBlockHeight`
- preserve existing constructor behavior and make server enrichment best-effort on RPC failure
- add focused Go tests and a changelog fragment

phdargen (approved):
Thanks @lgalabru, LGTM 🚀

PR #2697, opened 2026-06-25
docs(svm): add `upto` SVM scheme specification

lgalabru:
## Description

Adds the **SVM profile of the `upto` scheme** — `specs/schemes/upto/scheme_upto_svm.md` — the per-network companion to the network-agnostic [`scheme_upto.md`](specs/schemes/upto/scheme_upto.md) and the EVM profile [`scheme_upto_evm.md`](specs/schemes/upto/scheme_upto_evm.md).

`upto` lets a client authorize a **maximum** while the server settles for **actual** usage (`actual ≤ max`). On SVM this is realized via two profiles: a normative `payment-channel` profile (escrow deposit = ceiling, single operator-signed voucher, `settle_and_finalize` + refund) and an optional future `permit` profile. The spec maps the five core `upto` requirements onto SVM mechanisms and specifies the wire format, phases, error codes, and security properties.

Authored per the repo's [`authoring-specs`](.agents/skills/authoring-specs/SKILL.md) guidance: v2 transport/headers, CAIP-2 networks, atomic units, scheme data in `extra` (e.g. `feePayer`, `channelProgram`), and references to the core `x402-specification-v2.md` types.

## Tests

Documentation-only change (no code). N/A.

## Checklist

### DIFF
--- a/specs/schemes/upto/scheme_upto_svm.md
+++ b/specs/schemes/upto/scheme_upto_svm.md
@@ -14,73 +14,61 @@
 resource is consumed. Same target use cases as the generic spec: LLM token
 billing, per-byte metering, dynamic compute pricing.
 
-The hard part on Solana: a normal signed transfer commits to an *exact* amount
+A normal signed transfer commits to an *exact* amount
 and exact instruction data, so the server cannot lower the amount after the
 client signs without invalidating the signature. `upto` therefore requires an
 authorization that commits the client to a **ceiling** and lets the
-**facilitator** choose the actual amount at settlement. SVM offers two
-mechanisms that decouple "client authorizes a max" from "facilitator executes
-the actual," expressed here as two **profiles**.
+**operator** choose the actual amount at settlement. On SVM this is realized with
+an on-chain **payment channel**: the client escrows the ceiling, and the operator
+settles the actual amount against it with a single signed voucher.
+
+The **operator** is the party that co-signs the channel `open` as fee payer and
+submits settlement — the **server itself, or a facilitator it delegates to**
+(advertised as `extra.feePayer`). Nothing in `upto` requires a *separate*
+facilitator: a server can self-facilitate, running verify and settle directly.
+"Server/facilitator" below denotes whichever fills the operator role.
 
 ## 2. Mapping the five core requirements to SVM
 
-| Requirement (generic spec) | EVM mechanism | SVM mechanism (this spec) |
-|---|---|---|
-| Single-use authorization | Permit2 nonce | `payment-channel`: `finalize` makes the channel terminal (`ChannelStatus::Finalized`). `permit`: a one-shot nonce PDA consumed at settle. |
-| Time-bound validity (`validAfter`, `deadline`) | Permit2 `deadline` + witness `validAfter` | Authorization `validAfter` + `expiresAt`; `expiresAt` is also signed into the on-chain voucher / permit message. |
-| Recipient binding | Permit2 witness `to` | `payment-channel`: `channel.payee` + `distribution_hash` fixed at open. `permit`: `payTo` signed into the permit message and enforced by the program. |
-| Maximum amount enforcement | `permitted.amount` ceiling | `payment-channel`: on-chain `deposit` ceiling, `cumulative_amount ≤ deposit`. `permit`: `maxAmount` signed into the permit message, `actual ≤ maxAmount` enforced by the program. |
-| Phase-dependent amount semantics | `amount` = max at verify, actual at settle | Identical. `amount` in `PaymentRequirements` is the max during verification and the actual charge during settlement. |
+| Requirement (generic spec) | SVM mechanism |
+|---|---|
+| Single-use authorization | `finalize` makes the channel terminal (`ChannelStatus::Finalized`). |
+| Time-bound validity (`validAfter`, `expiresAt`) | Authorization `validAfter` + `expiresAt`; `expiresAt` is also signed into the on-chain voucher. |
+| Recipient binding | `channel.payee` + `distribution_hash` fixed at open. |
+| Maximum amount enforcement | On-chain `deposit` ceiling, `cumulative_amount ≤ deposit`. |
+| Phase-dependent amount semantics | `amount` in `PaymentRequirements` is the max during verification and the actual charge during settlement. |
 
-The facilitator MUST always verify against the client-signed ceiling, never
-against the settlement-time `amount`.
+The server/facilitator MUST always verify against the client-signed ceiling,
+never against the settlement-time `amount`.
 
-## 3. Profiles
+## 3. Payment-channel profile
 
-A server advertises one or more profiles in `extra.profiles`. The client picks
-one and signals it back in the payload's `profile` field.
-
-### 3.1 `payment-channel` (normative, v1)
-
-Backed by the on-chain payment-channels program (a `pay-kit`-controlled program;
-program id `CHNLxYvVA28MJP9PrFuDXccuoGXAx7jBacfLEkahyGsX` on the current
-deployment). The escrow **deposit is the ceiling**; a single signed voucher
-settles the actual amount; `finalize` closes the channel and refunds the unused
+v1 defines a single profile, `payment-channel`, carried in the payload's
+`profile` field. It is backed by the on-chain payment-channels program
+(advertised via `extra.channelProgram`; a canonical deployment is assumed when
+omitted). The escrow **deposit is the ceiling**; a single signed voucher settles
+the actual amount; `finalize` closes the channel and refunds the unused
 remainder to the payer in the same transaction.
 
-Strengths: every requirement is enforced on-chain by a program we control. The
-facilitator cannot overcharge (capped by `deposit`), cannot redirect funds
+Strengths: every requirement is enforced on-chain by the program. The
+operator cannot overcharge (capped by `deposit`), cannot redirect funds
 (`channel.payee` / `distribution_hash` are fixed at open), and cannot replay
-(channel is terminal after `finalize`).
+(channel is terminal after `finalize`). Conversely, because the channel's
+`authorizedSigner` is the operator (and the operator is the `settle_and_finalize`
+merchant), the operator can settle and finalize **at any time** — collecting the
+metered amount and refunding the client's unused deposit — without the client's
+cooperation. The channel-account rent is fronted by the operator: the `open`
+instruction's `rentPayer` account (the operator's own key) funds the PDA and
+escrow-ATA rent, and `finalize` **refunds it to that `rentPayer`** — so the
+client moves only stablecoin and never needs SOL. The operator can finalize a
+stale channel at will, reclaiming the rent it fronted; its only exposure is the
+cheap finalize transaction.
 
 Cost: the client locks `max` in escrow for the lifetime of the request, and the
-flow needs two on-chain transactions (open, then settle-and-finalize). Channel
-rent is reclaimed at finalize.
-
-### 3.2 `permit` (optional, v2 — requires a new program)
-
-Backed by an Ed25519-signed one-shot delegated transfer — the closest analog to
-EVM's `permitWitnessTransferFrom`. The client `approve`s a program-owned
-delegate on its token account (reusable across requests, like a one-time Permit2
-approval), then signs an off-chain **permit message** committing to a ceiling and
-a witness. At settlement the facilitator submits one transaction; a small
-on-chain program verifies the Ed25519 signature (via the Ed25519 precompile +
-instructions sysvar, the same pattern the payment-channels program already uses
-for vouchers), checks the witness, enforces `actual ≤ maxAmount`, transfers the
-actual amount to `payTo`, and consumes a nonce PDA for replay protection.
-
-Strengths: no escrow lockup; a single settlement transaction; reusable approval.
-Matches EVM ergonomics.
-
-Cost: requires a new, audited on-chain program. Until that program ships and is
-audited, `payment-channel` is the only normative profile.
-
-> Multi-delegator is explicitly **not** an `upto` backend. Its `FixedDelegation`
-> is a standing, multi-pull cap that binds the *delegatee* (operator), not the
-> final `payTo`, and offers no on-chain single-use guarantee. It is the right
-> primitive for `session`/streaming, not for one-shot `upto`. Using it here
-> would downgrade the recipient-binding and single-use requirements to
-> "trust the facilitator," defeating the scheme.
+flow needs two on-chain transactions (open, then settle-and-finalize). The
+channel-account rent is funded by the operator (the `open` `rentPayer` account)
+and reclaimed at finalize; the client supplies only the stablecoin deposit and
+never needs SOL.
 
 ## 4. Wire format
 
@@ -98,7 +86,7 @@ to `exact`. The core `PaymentRequirements`, `PaymentPayload`, and
 | `scheme` | string | ✓ | `"upto"` |
 | `network` | string | ✓ | CAIP-2, e.g. `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp` (mainnet), `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1` (devnet) |
 | `amount` | string | ✓ | **Phase-dependent**: max authorized at verification; actual charge at settlement. Base units. |
-| `asset` | string | ✓ | SPL mint address (or `"SOL"` — discouraged for `upto`; stablecoins recommended) |
+| `asset` | string | ✓ | SPL mint address (e.g. USDC `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`) |
 | `payTo` | string | ✓ | Base58 recipient |
 | `maxTimeoutSeconds` | number | ✓ | Completion window; also the basis for the authorization `expiresAt` |
 | `extra` | object | ✓ | See below |
@@ -107,14 +95,14 @@ to `exact`. The core `PaymentRequirements`, `PaymentPayload`, and
 
 | Field | Type | Required | Notes |
 |---|---|---|---|
-| `profiles` | string[] | ✓ | Subset of `["payment-channel","permit"]`, in server preference order |
+| `profiles` | string[] | ✓ | `["payment-channel"]` (the only v1 profile) |
 | `feePayer` | string | ✓ | Base58 operator key that sponsors fees (co-signs the setup transaction as fee payer) and settles |
-| `channelProgram` | string | – | Channel/permit program id; defaults to the canonical deployment |
+| `channelProgram` | string | – | Channel program id; defaults to the canonical deployment |
 | `decimals` | number | ✓ | Token decimals |
 | `tokenProgram` | string | ✓ | `Tokenkeg…` or `TokenzQ…` (Token-2022); the client SHOULD verify it against the on-chain mint owner |
 | `recentBlockhash` | string | – | Pre-fetched blockhash so the client can build setup transactions without an extra RPC round-trip |
 | `validAfter` | number | – | Earliest activation time (Unix seconds); default = now |
-| `splits` | `{recipient,bps}[]` | – | `payment-channel` only; distributed at finalize |
+| `splits` | `{recipient,bps}[]` | – | Distribution splits sealed at open; distributed at finalize |
 
 ### 4.2 `UptoPayload` (in `PAYMENT-SIGNATURE.payload`)
 
@@ -122,21 +110,21 @@ Common fields:
 
 | Field | Type | Notes |
 |---|---|---|
-| `profile` | string | `"payment-channel"` or `"permit"` |
+| `profile` | string | `"payment-channel"` (the only v1 profile) |
 | `from` | string | Payer wallet (base58) |
 | `maxAmount` | string | The signed ceiling (base units). MUST equal verification-phase `amount`. |
 | `expiresAt` | number | Deadline (Unix seconds); signed into the on-chain message |
 | `validAfter` | number | Activation time (Unix seconds) |
 | `nonce` | string | Unique per authorization |
 
-`payment-channel` profile adds:
+Plus the channel fields:
 
 | Field | Type | Notes |
 |---|---|---|
 | `channelId` | string | Channel PDA (base58) |
 | `deposit` | string | On-chain escrow = the ceiling; MUST equal `maxAmount` |
-| `authorizedSigner` | string | The **operator/facilitator** key (base58). The operator — not the client — signs the single settlement voucher (see §5 Phase 2). |
-| `openTransaction` | string | Base64 signed `open` transaction for the facilitator to broadcast if the channel is not yet open (pull-style). Omitted if the client already broadcast `open` (push-style) and `signature` is set. |
+| `authorizedSigner` | string | The **operator** key (base58) — the server's or facilitator's pubkey, equal to `extra.feePayer`. The operator — not the client — signs the single settlement voucher (see §5 Phase 2); as the channel `authorizedSigner`/merchant it alone can settle and finalize the channel. |
+| `openTransaction` | string | Base64 signed `open` transaction for the server/facilitator to broadcast if the channel is not yet open (pull-style). Omitted if the client already broadcast `open` (push-style) and `signature` is set. |
 | `signature` | string | Base58 signature of the broadcast `open` transaction (push) |
 
 > The voucher is **not** carried in the payload. Because the actual amount is
@@ -144,28 +132,7 @@ Common fields:
 > on-chain `deposit` ceiling plus the fixed `payee`, the operator (set as the
 > channel's `authorizedSigner` at open) signs the single voucher for the metered
 > amount at settlement. This keeps `upto` a single HTTP round-trip with a
-> handler-determined amount, matching the EVM trust model where the server fills
-> in `actual ≤ ceiling`.
-
-`permit` profile adds:
-
-| Field | Type | Notes |
-|---|---|---|
-| `tokenAccount` | string | Payer ATA being delegated (base58) |
-| `approveTransaction` | string | Base64 signed `approve` transaction for the facilitator to broadcast if no sufficient delegation exists |
-| `permitSignature` | string | Base58 Ed25519 signature by `from` over the permit message |
-
-Permit message (Borsh, signed by `from`):
-
-```
-nonce_pda: [u8;32]          // PDA(["upto", from, mint, nonce])
-mint:      [u8;32]
-pay_to:    [u8;32]
-facilitator:[u8;32]
-max_amount: u64 (le)
-valid_after: i64 (le)
-expires_at: i64 (le)
-```
+> handler-determined amount: the server fills in `actual ≤ ceiling`.
 
 ### 4.3 `SettlementResponse` (in `PAYMENT-RESPONSE`)
 
@@ -180,38 +147,34 @@ expires_at: i64 (le)
 
 ## 5. Phases
 
-### Phase 1 — Setup (gas/approval)
+### Phase 1 — Setup
 
-- `payment-channel`: the client builds an `open` transaction depositing
-  `maxAmount`. Push: the client broadcasts it and sends `signature`. Pull: the
-  client sends `openTransaction` and the facilitator broadcasts it (the
-  facilitator may co-sign as fee payer, matching `exact`'s fee-sponsorship).
-- `permit`: the client ensures a program-owned delegate is approved on
-  `tokenAccount` for at least `maxAmount`. If absent, the facilitator returns
-  `412 Precondition Failed` with `errorReason = APPROVAL_REQUIRED` and the
-  required `approve` transaction, mirroring EVM's `PERMIT2_ALLOWANCE_REQUIRED`.
+The client builds an `open` transaction depositing `maxAmount`, naming the
+operator as the open's `rentPayer` so the operator funds the channel rent (the
+client supplies only the stablecoin deposit). Push: the client broadcasts it and
+sends `signature`. Pull: the client sends `openTransaction` and the
+server/facilitator broadcasts it, co-signing as fee payer **and** as `rentPayer`
+(one operator signature covers both, matching `exact`'s fee-sponsorship).
 
 ### Phase 2 — Authorization signature
 
-- `payment-channel` (normative v1): the client's signature on the `open`
-  transaction **is** the authorization — it commits the `deposit` ceiling, the
-  `payee`, and the `mint`, with `authorizedSigner` set to the **operator**. The
-  client does not sign a voucher. After metering, the operator signs the single
-  voucher for `cumulativeAmount = actual` (Ed25519 over
-  `channel_id ‖ cumulative_amount_le ‖ expires_at_le`) and settles it. The
-  client is protected by the on-chain ceiling (the operator cannot exceed
-  `deposit`) and the fixed `payee` (the operator cannot redirect) — the same
-  trust model as EVM `upto`, in a single round-trip.
-- `permit`: sign the permit message with `from`.
+The client's signature on the `open` transaction **is** the authorization — it
+commits the `deposit` ceiling, the `payee`, and the `mint`, with
+`authorizedSigner` set to the **operator**. The client does not sign a voucher.
+After metering, the operator signs the single voucher for
+`cumulativeAmount = actual` (Ed25519 over
+`channel_id ‖ cumulative_amount_le ‖ expires_at_le`) and settles it. The client
+is protected by the on-chain ceiling (the operator cannot exceed `deposit`) and
+the fixed `payee` (the operator cannot redirect), in a single round-trip.
 
 ### Phase 3 — Verification (before serving the resource)
 
-The facilitator MUST, in order:
+The server/facilitator MUST, in order:
 
 1. Confirm `payload.maxAmount` equals verification-phase `requirements.amount`.
 2. Confirm `network`, `asset` (mint), `tokenProgram`, and `payTo` match the requirements.
-3. Confirm `feePayer` in `extra` is this server's key.
-4. `payment-channel`: confirm the channel exists (or the `openTransaction` is valid and broadcastable), `channel.deposit ≥ maxAmount`, `channel.payee` and `distribution_hash` match `payTo`/`splits`, `channel.status == Open`, and `channel.mint == asset`. `permit`: recover `permitSignature`, confirm signer is `from`; confirm the delegate allowance and the payer token balance both cover `maxAmount`.
+3. Confirm `feePayer` in `extra` is the operator's own key (the server's, or the facilitator's it delegates to).
+4. Confirm the channel exists (or the `openTransaction` is valid and broadcastable), `channel.deposit ≥ maxAmount`, `channel.payee` and `distribution_hash` match `payTo`/`splits`, `channel.status == Open`, `channel.mint == asset`, **`channel.authorizedSigner == operator`** (so the operator alone can settle and finalize), and the open's **`rentPayer == operator`** (the operator funds and reclaims the rent it co-signs for). An `open` naming any other authorized signer or rentPayer MUST be rejected before co-signing.
 5. Validate `validAfter ≤ now ≤ expiresAt`.
 6. Simulate the settlement instruction(s).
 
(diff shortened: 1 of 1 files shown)


gitteri (approved):

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,267 @@
...
+for vouchers), checks the witness, enforces `actual ≤ maxAmount`, transfers the
+actual amount to `payTo`, and consumes a nonce PDA for replay protection.
+
+Strengths: no escrow lockup; a single settlement transaction; reusable approval.
+Matches EVM ergonomics.
+
+Cost: requires a new, audited on-chain program. Until that program ships and is
+audited, `payment-channel` is the only normative profile.
+
+> Multi-delegator is explicitly **not** an `upto` backend. Its `FixedDelegation`
```
notorious-d-e-v: This seems like extra text.

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,267 @@
...
+[`x402-specification-v2.md`](../../x402-specification-v2.md).
+
+### 4.1 `PaymentRequirements` (in `PAYMENT-REQUIRED.accepts[]`)
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `scheme` | string | ✓ | `"upto"` |
+| `network` | string | ✓ | CAIP-2, e.g. `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp` (mainnet), `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1` (devnet) |
+| `amount` | string | ✓ | **Phase-dependent**: max authorized at verification; actual charge at settlement. Base units. |
+| `asset` | string | ✓ | SPL mint address (or `"SOL"` — discouraged for `upto`; stablecoins recommended) |
```
notorious-d-e-v: Why not use `So11111111111111111111111111111111111111111` for `SOL`? 

Also imo there's no need to discourage it if it functions the same way.

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,267 @@
...
+| `payTo` | string | ✓ | Base58 recipient |
+| `maxTimeoutSeconds` | number | ✓ | Completion window; also the basis for the authorization `expiresAt` |
+| `extra` | object | ✓ | See below |
+
+`extra`:
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `profiles` | string[] | ✓ | Subset of `["payment-channel","permit"]`, in server preference order |
+| `feePayer` | string | ✓ | Base58 operator key that sponsors fees (co-signs the setup transaction as fee payer) and settles |
```
notorious-d-e-v: Is the operator able to call settle and finalize, or have some other escape hatch, in order to reclaim the channel rent?

Otherwise a malicious actor will spoof the client and/or server in order to grief the server/facilitator.

Haven't dug into the payment channel program so I'm not certain.

edit: as I continue reading, what I'm understanding is that the `open` transaction authorizes the operator to settle/finalize/reclaim their rent and the client can't void it. Just want to make sure that's true.

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,267 @@
...
+- `permit`: sign the permit message with `from`.
+
+### Phase 3 — Verification (before serving the resource)
+
+The facilitator MUST, in order:
+
+1. Confirm `payload.maxAmount` equals verification-phase `requirements.amount`.
+2. Confirm `network`, `asset` (mint), `tokenProgram`, and `payTo` match the requirements.
+3. Confirm `feePayer` in `extra` is this server's key.
+4. `payment-channel`: confirm the channel exists (or the `openTransaction` is valid and broadcastable), `channel.deposit ≥ maxAmount`, `channel.payee` and `distribution_hash` match `payTo`/`splits`, `channel.status == Open`, and `channel.mint == asset`. `permit`: recover `permitSignature`, confirm signer is `from`; confirm the delegate allowance and the payer token balance both cover `maxAmount`.
```
notorious-d-e-v: Would add that channel's `authorizedSigner` should match the operator's pubkey so that they can settle/finalize/reclaim rent.

So if the operator is:
- the server then the `authorizedSigner` should match the server's pubkey.
- the facilitator then the `authorizedSigner` should match the facilitator's pubkey.

notorious-d-e-v (commented):
Nice work @lgalabru !

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,267 @@
...
+  monotonic `SettlementWatermarks.settled`. `permit`: one-shot nonce PDA.
+- **Time-bounded.** `validAfter`/`expiresAt` checked off-chain at verify and
+  signed into the on-chain message so a stale authorization cannot settle.
+- **Trust model.** As in the generic spec, the client trusts the server to meter
+  honestly within the ceiling. Everything above the ceiling, the destination,
+  and replay are enforced cryptographically/on-chain.
+
+## 8. Out of scope
+
+Multi-settlement streaming and recurring auto-pay are served by the `session`
```
notorious-d-e-v: I think `session` is being named `batch-settlement`

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,267 @@
+# SVM `upto` Scheme: Usage-Based Payment Authorization on Solana
```
notorious-d-e-v, line 1: In general I'd find/replace all instances of `facilitator` with `server/facilitator` and make sure that the server can carry out all these processes without depending on a facilitator.

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Clarified.

notorious-d-e-v:
Some more review after reading the [payment-channels program](https://github.com/solana-foundation/payment-channels/tree/main/program/payment_channels) with the help of trusty AI.

The overall direction makes sense, but a few security/flow claims don’t match the program semantics yet.

1. **The delegated operator cannot finalize unless it is also `payTo`.**

The spec collapses `authorizedSigner`, `feePayer`/operator, and the `settle_and_finalize` merchant into one role. In the program these are distinct: vouchers are checked against `channel.authorized_signer`, but `settleAndFinalize` requires a `merchant` signer and rejects unless `merchant == channel.payee`.

So in the delegated-facilitator case where `feePayer != payTo`, the facilitator can sign vouchers but cannot finalize/refund/close without the payee. The “no operator griefing” guarantee only holds if `feePayer == payTo`, or the spec needs to describe the weaker model.

Program reference: https://github.com/solana-foundation/payment-channels/blob/ef149467b457cad60249503794654f80c0f22ae4/program/payment_channels/src/instructions/settle_and_finalize.rs#L83-L119

2. **`channel.deposit >= maxAmount` should be equality for x402 `upto`.**

In the EVM `upto` profile, the client-signed ceiling is `permit2Authorization.permitted.amount`: settlement can be lower, but not higher. The SVM equivalent cap is `channel.deposit`, because the program only rejects vouchers where `cumulative_amount > deposit`, and `topUp` can raise the deposit while the channel is open.

The verifier should require:

`payload.maxAmount == verification requirements.amount == payload.deposit == on-chain channel.deposit`

Allowing `channel.deposit >= maxAmount` makes the x402 ceiling advisory rather than enforced.

Program references:
- Voucher cap: https://github.com/solana-foundation/payment-channels/blob/ef149467b457cad60249503794654f80c0f22ae4/program/payment_channels/src/instructions/helpers/voucher.rs#L76-L89
- Top-up raises deposit: https://github.com/solana-foundation/payment-channels/blob/ef149467b457cad60249503794654f80c0f22ae4/program/payment_channels/src/instructions/top_up.rs#L18-L22

3. **The time bounds are not client-bound as described.**

The spec says the client’s `open` signature is the authorization, but the voucher payload is only `channel_id || cumulative_amount || expires_at`, and under this spec that voucher is signed by the operator/authorizedSigner, not the client. That means the client does not cryptographically bind `validAfter` or `expiresAt`; they are off-chain policy unless added to the client-signed/open commitment somehow.

Program reference: https://github.com/solana-foundation/payment-channels/blob/ef149467b457cad60249503794654f80c0f22ae4/program/payment_channels/src/instructions/mod.rs#L18-L39

4. **`settleAndFinalize`/`finalize` do not move funds or close the channel by themselves.**

The spec says the flow is “open, then settle-and-finalize” and that finalize refunds/closes in the same transaction. In the program, `settleAndFinalize` only applies the optional voucher and flips status to `Finalized`; `finalize` only moves `Closing -> Finalized`. The actual splits, payer refund, escrow close, and channel tombstone happen in `distribute`. The spec should require `distribute` as part of settlement/cleanup, possibly bundled in the same transaction.

5. **Zero-amount settlement needs the no-voucher path.**

A `cumulative_amount = 0` voucher is invalid because voucher watermarks must be strictly monotonic and `settled` starts at `0`. The `0`-amount case should explicitly use `settleAndFinalize(has_voucher = 0)` plus `distribute`, rather than “a voucher for actual = 0”.

lgalabru:
Thanks @notorious-d-e-v — great review, all five are spot-on and grounded in the program. Addressed in cf94afc23:

1. **Roles / `merchant == channel.payee`.** Correct — `settle_and_finalize` requires the merchant signer to equal `channel.payee`, so the operator can only finalize because it *is* the payee. The spec now states the operator is the channel `payee` (alongside `authorizedSigner` + fee payer + `rentPayer`), and that the x402 `payTo` is realized either as that payee (when the server self-facilitates) or as a **program-enforced distribution split fixed at `open`** (when a separate facilitator is the payee). That resolves the delegated case **without a program change**: the facilitator is the payee/operator, the real merchant is a locked split recipient it cannot redirect or shortchange, and the facilitator can keep a fee via the payee remainder.

2. **`deposit == maxAmount`.** Changed the verifier from `≥` to strict `==`, with your reasoning noted (`topUp` can raise an open channel's deposit, so `≥` leaves the ceiling advisory).

3. **Time bounds are not client-bound.** Clarified: `expiresAt` is the *operator's* voucher value, enforced on-chain (settle rejected once `now ≥ expiresAt`); `validAfter` is off-chain verify-time policy only. The client signs only `open`, so neither is a client commitment — they bound when a metered settlement may land.

4. **`distribute` moves funds.** Corrected throughout: `settle_and_finalize`/`finalize` only advance *status*; `distribute` pays the splits, refunds `deposit − actual`, returns the fronted rent, and tombstones the channel (the two can be bundled in one tx).

5. **Zero-amount no-voucher path.** Now explicit: `settle_and_finalize(has_voucher = 0)` + `distribute`, since a `cumulative = 0` voucher is non-monotonic and invalid.

(The earlier inline nits — `server/facilitator` wording, USDC example, `authorizedSigner == operator`, `batch-settlement` naming — were folded into the previous revision.)

notorious-d-e-v:
Recent changes look good to me @lgalabru! 

Highlighting a couple of points that stood out while reviewing the most recent updates.

## Payment Recipient in merchant-sponsored flow versus facilitator-sponsored flow

Regarding merchant-sponsored flow (case 1) versus facilitator-sponsored flow (case 2) -- here's what I'm understanding. 

**Case 1: Merchant Self-Facilitates**
The merchant is all operator roles:
```
feePayer = merchant
rentPayer = merchant
payee = merchant
authorizedSigner = merchant
payTo = merchant
```

Flow:
1. Client opens a channel and escrows `maxAmount`.
2. Merchant pays tx fees and rent, and is the channel `payee`.
3. Merchant serves the request.
4. Merchant meters actual usage.
5. Merchant signs the voucher as `authorizedSigner`.
6. Merchant calls `settleAndFinalize`.
7. Merchant calls/bundles `distribute` in the same transaction.
8. Funds go to merchant; unused deposit goes back to client; account rent goes back to merchant.

**Case 2: Merchant Uses Facilitator**
The facilitator becomes the channel operator:
```
feePayer = facilitator
rentPayer = facilitator
payee = facilitator
authorizedSigner = facilitator
payTo = merchant
```

Since the Solana program requires `settleAndFinalize` to be signed by `channel.payee`, the facilitator must be `payee` if it needs to finalize without merchant signing.

So the merchant’s revenue address, `payTo`, cannot be the `channel payee` in this delegated model. Instead, the spec says `payTo` is included in the sealed distribution splits at channel open.

Flow:
1. Client opens a channel and escrows `maxAmount`.
2. Channel `payee` is facilitator, not merchant.
3. **Merchant’s `payTo` is locked into `distribution_hash` as a split recipient.**
4. Facilitator pays fees/rent and is `authorizedSigner`.
5. Merchant serves/meters the request.
6. Facilitator signs voucher and calls `settleAndFinalize`.
7. Facilitator calls/bundles `distribute` in the same transaction.
8. Solana Program pays the merchant’s locked split to `payTo`; unused deposit goes back to client; rent goes back to facilitator; facilitator may keep its fee as payee remainder.

It makes sense to me now but I'm writing it out in case it helps others in understanding.

I think it's important to call out that the merchant won't be the direct recipient of the funds in this scenario. Instead it is part of payment splitting that happens at the Solana Program level.


## Time bounds are not client-bound

As the Solana Program currently stands, there is nothing enforcing the client (buyer) preference for `validAfter` and `expiresAt`, in contrast to the evm version of `upto`.

Worth bringing up for the maintainers.

At the moment we are just trusting that the merchant/facilitator will respect it off-chain.


## Other Items

A couple of phrases still say “distributed at finalize” / “open, then settle-and-finalize” where “distribute” is the actual movement/close step. Minor wording, but worth tightening.

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+| `extra` | object | ✓ | See below |
+
+`extra`:
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `profiles` | string[] | ✓ | `["payment-channel"]` (the only v1 profile) |
+| `feePayer` | string | ✓ | Base58 operator key that sponsors fees (co-signs the setup transaction as fee payer) and settles |
+| `channelProgram` | string | – | Channel program id; defaults to the canonical deployment |
+| `decimals` | number | ✓ | Token decimals |
```
phdargen: please remove `decimals`, seems unused. If it is needed, client shouldn't trust server declared input but get it from onchain state

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+| `maxTimeoutSeconds` | number | ✓ | Completion window; also the basis for the authorization `expiresAt` |
+| `extra` | object | ✓ | See below |
+
+`extra`:
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `profiles` | string[] | ✓ | `["payment-channel"]` (the only v1 profile) |
+| `feePayer` | string | ✓ | Base58 operator key that sponsors fees (co-signs the setup transaction as fee payer) and settles |
+| `channelProgram` | string | – | Channel program id; defaults to the canonical deployment |
```
phdargen: please remove `channelProgram`, should always use canonical deployment

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+| Phase-dependent amount semantics | `amount` in `PaymentRequirements` is the max during verification and the actual charge during settlement. |
+
+The server/facilitator MUST always verify against the client-signed ceiling,
+never against the settlement-time `amount`.
+
+## 3. Payment-channel profile
+
+v1 defines a single profile, `payment-channel`, carried in the payload's
+`profile` field. It is backed by the on-chain payment-channels program
+(advertised via `extra.channelProgram`; a canonical deployment is assumed when
```
phdargen: please provide references/docs for the `channelProgram`

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+> the reference rejects an open whose `payTo` is not the operator.
+
+## 4. Wire format
+
+`upto` reuses the x402 v2 transport: a `402` response carries `PAYMENT-REQUIRED`;
+the paid retry carries `PAYMENT-SIGNATURE`; the response carries
+`PAYMENT-RESPONSE`. Only the `scheme` value and the payload shape change relative
+to `exact`. The core `PaymentRequirements`, `PaymentPayload`, and
+`SettlementResponse` types are defined in
+[`x402-specification-v2.md`](../../x402-specification-v2.md).
```
phdargen: can be removed

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+| `maxAmount` | string | The signed ceiling (base units). MUST equal verification-phase `amount`. |
+| `expiresAt` | number | Deadline (Unix seconds); signed into the on-chain message |
+| `validAfter` | number | Activation time (Unix seconds) |
+| `nonce` | string | Unique per authorization |
+
+Plus the channel fields:
+
+| Field | Type | Notes |
+|---|---|---|
+| `channelId` | string | Channel PDA (base58) |
```
phdargen: How is `channelId` derived? Does the client know it before the channel is opened, ie in a pull style transaction?

If a channel is closed, can it be reopened?

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+> handler-determined amount: the server fills in `actual ≤ ceiling`.
+
+### 4.3 `SettlementResponse` (in `PAYMENT-RESPONSE`)
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `success` | boolean | ✓ | |
+| `errorReason` | string | – | Omitted on success |
+| `payer` | string | – | `from` |
+| `transaction` | string | ✓ | Settle (or settle-and-finalize) signature; empty string if the actual charge is `0` |
```
phdargen: This should show the txHash of the transaction moving `amount` to `payTo`. Even if the charge is `0`, there needs to be a tx to close the channel, so this shouldnt be left empty

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+| `amount` | string | ✓ | Actual base units charged (may be `0`) |
+
+## 5. Phases
+
+### Phase 1 — Setup
+
+The client builds an `open` transaction depositing `maxAmount`, naming the
+operator as the open's `rentPayer` so the operator funds the channel rent (the
+client supplies only the stablecoin deposit). Push: the client broadcasts it and
+sends `signature`. Pull: the client sends `openTransaction` and the
```
phdargen: It is my understanding the rent is paid at channel open? If thats the case, its unclear to me how the operator could possibly pay the rent in a push tx, which the client broadcasts. Just naming the operator as `rentPayer` is supposedly not enough, guess it would require a operator signature to consent?

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+| `asset` | string | ✓ | SPL mint address (e.g. USDC `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`) |
+| `payTo` | string | ✓ | Base58 recipient. Realized on-chain as the channel `payee` when the server self-facilitates (`payTo == operator`), or as a program-enforced distribution split when a separate facilitator is the `payee`. |
+| `maxTimeoutSeconds` | number | ✓ | Completion window; also the basis for the authorization `expiresAt` |
+| `extra` | object | ✓ | See below |
+
+`extra`:
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `profiles` | string[] | ✓ | `["payment-channel"]` (the only v1 profile) |
```
phdargen: Please align with notation established for evm: `profiles` -> `assetTransferMethod` and use a single string instead of an array. A given `assetTransferMethod` entry serves as discriminator for the remaining `extra` fields.
https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md

```suggestion
| `assetTransferMethod ` | string | ✓ | `"payment-channel"` |
```

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+  (the `payTo`/`splits`) are fixed at open and re-checked by the program at
+  `distribute`, so the operator cannot redirect or shortchange `payTo`.
+- **No replay.** Terminal `ChannelStatus::Finalized` plus monotonic
+  `SettlementWatermarks.settled`.
+- **No operator griefing.** The operator is the channel `payee` (so it is the
+  `settle_and_finalize` merchant, which the program requires to equal
+  `channel.payee`) **and** the voucher `authorizedSigner`, so it can always
+  settle the metered amount, refund the client's unused deposit, and — via
+  `distribute` — close the channel and reclaim the rent it fronted, all without
+  the client. The client never needs SOL and cannot strand the operator's funds;
```
phdargen: I think in a push tx, the client does need SOL for gas + possibly rent

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+   when no token movement occurred.
+
+## 6. Error codes
+
+Standard x402 codes apply. Scheme-specific:
+
+- `invalid_upto_svm_payload_settlement_exceeds_amount` — actual > signed ceiling.
+- `CHANNEL_REQUIRED` (with `412`) — no open channel and no broadcastable `openTransaction`.
+
+## 7. Security properties
```
phdargen: Is there any escape hatch for the client to withdraw from the escrow in case the server is unresponsive/inactive and never closes the channel?

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+
+## 5. Phases
+
+### Phase 1 — Setup
+
+The client builds an `open` transaction depositing `maxAmount`, naming the
+operator as the open's `rentPayer` so the operator funds the channel rent (the
+client supplies only the stablecoin deposit). Push: the client broadcasts it and
+sends `signature`. Pull: the client sends `openTransaction` and the
+server/facilitator broadcasts it, co-signing as fee payer **and** as `rentPayer`
```
phdargen: For a pull tx it is not specified when exactly the `openTransaction` is broadcast. As stated below, verify/ checks if it is broadcastable and settle/ closes the channel, but when is it opened?

specs/schemes/upto/scheme_upto_svm.md
```diff
@@ -0,0 +1,268 @@
...
+
+## 7. Security properties
+
+- **No overcharge.** Capped by the on-chain `deposit`.
+- **No redirection.** `channel.payee` (the operator) and `distribution_hash`
+  (the `payTo`/`splits`) are fixed at open and re-checked by the program at
+  `distribute`, so the operator cannot redirect or shortchange `payTo`.
+- **No replay.** Terminal `ChannelStatus::Finalized` plus monotonic
+  `SettlementWatermarks.settled`.
+- **No operator griefing.** The operator is the channel `payee` (so it is the
```
phdargen: If a facilitator pays the rent, does it need to remember which channels it opened, so it can reclaim those which have not been closed after a while?

phdargen:
> **Case 2: Merchant Uses Facilitator** The facilitator becomes the channel operator:
> 
> ```
> feePayer = facilitator
> rentPayer = facilitator
> payee = facilitator
> authorizedSigner = facilitator
> payTo = merchant
> ```
> 
> Since the Solana program requires `settleAndFinalize` to be signed by `channel.payee`, the facilitator must be `payee` if it needs to finalize without merchant signing.
> 
> So the merchant’s revenue address, `payTo`, cannot be the `channel payee` in this delegated model. Instead, the spec says `payTo` is included in the sealed distribution splits at channel open.
> 
> Flow:
> 
> 1. Client opens a channel and escrows `maxAmount`.
> 2. Channel `payee` is facilitator, not merchant.
> 3. **Merchant’s `payTo` is locked into `distribution_hash` as a split recipient.**
> 4. Facilitator pays fees/rent and is `authorizedSigner`.
> 5. Merchant serves/meters the request.
> 6. Facilitator signs voucher and calls `settleAndFinalize`.
> 7. Facilitator calls/bundles `distribute` in the same transaction.
> 8. Solana Program pays the merchant’s locked split to `payTo`; unused deposit goes back to client; rent goes back to facilitator; facilitator may keep its fee as payee remainder.
> 

Thanks for the summary @notorious-d-e-v!

I think case 2 is currently underspecified and needs some more work. 

If the facilitator can close the channel without the server signing, it could act against the server and eg refund the full balance to the client. If this can't be avoided, this trust assumption should at least be acknowledged in the spec.

There is also an authentication gap. Facilitator verify/ + settle/ calls are unauthenticated http requests in general. How does the facilitator know if the entity calling settle/ is authorized to do so, ie is the merchant owning `payTo`? Everyone with the channelId could just call settle/ with any `amount`, eg the client with `amount=0` to force a refund.

phdargen:
Hi @lgalabru, thanks a lot for putting this together! Made a first pass through the specs and left a few comments/questions above. 

We generally target a gasless experience for both client and server for all operations and the point of the facilitator is to abstract away web3 complexities such that it becomes a choice: "Do I use an external facilitator for a simpler experience or do I self-facilitate and take on the web3 complexities like rpc config, wallet management and gas costs"?

For the client payload, I would thus consider restricting to pull tx only. This would simplify the specs and I don't see a strong case for the push flow

For the server side, the facilitator sponsored flow still needs some work that should imo not be left to a v2 implementation

phdargen:
As a first impression, I am not sure if payment-channels are the best fit for the upto scheme in terms of UX as this approach requires 2 txs (channel open and close) doubling settle latency + cost and seems a bit heavy. Have any alternatives been considered? 

Will also discuss with the team

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed with https://github.com/x402-foundation/x402/pull/2697/commits/d67e9d2cba9f601803b5a1749e2405f762a6cfc8

specs/schemes/upto/scheme_upto_svm.md
lgalabru: I kept `channelProgram` because it is consumed by downstream roles: the client uses it to derive/build the channel open transaction, and the server/facilitator verifies settlement/distribution instructions against it. 

I tightened the security boundary so it is discovery, not trust: implementations must reject program ids outside their supported set. Addressed with https://github.com/x402-foundation/x402/pull/2697/commits/ffb83b0cc117e20bb3c8e4c2c2bceab1868668ab

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed with https://github.com/x402-foundation/x402/commit/606c49d906a05adcc75faeda995efe5832f92e5a

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed with https://github.com/x402-foundation/x402/commit/a76757249ceae27be903f63418d5fb5dccc5704f

lgalabru: On reopening: a closed/tombstoned channel cannot be reopened at the same `channelId`.

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed with https://github.com/x402-foundation/x402/commit/9a9d45042e84bb9489963491ff1f627845d002b5

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed with https://github.com/x402-foundation/x402/commit/9a9d45042e84bb9489963491ff1f627845d002b5

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed with https://github.com/x402-foundation/x402/commit/8b870dd9c94df368ecb5bdc32d6b75c4bfc424a6

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Addressed by simply removing all the references to push / client-broadcast flow - this flow is relevant for other agentic protocols, but should not be a focus here.

The spec now only describes a partially signed `openTransaction` that the server/facilitator signs as fee payer and `rentPayer`, then broadcasts during verification. 

Relevant commits: https://github.com/x402-foundation/x402/commit/a76757249ceae27be903f63418d5fb5dccc5704f https://github.com/x402-foundation/x402/commit/9a9d45042e84bb9489963491ff1f627845d002b5 https://github.com/x402-foundation/x402/commit/6cc2a50ac0accb51f43a3ae572826d9073b306bc

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Yes. The payer can call `requestClose`, which moves the channel from `Open` to `Closing` and starts the grace period. If the server/operator remains inactive, anyone can call `finalize` after `grace_period`, then the payer can call `withdrawPayer` to recover `deposit - settled`. If no settlement happened, `settled = 0`, so the payer recovers the full deposit. This is delayed by `grace_period`, not immediate, and clients should verify that value in `openTransaction` before signing.

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Clarified with https://github.com/x402-foundation/x402/commit/6cc2a50ac0accb51f43a3ae572826d9073b306bc

specs/schemes/upto/scheme_upto_svm.md
lgalabru: Clarified with https://github.com/x402-foundation/x402/commit/ecd84177f86e85b4f35bc2150ce3447ee75d94da.

The program stores `rentPayer`, so rent can only be returned to the account that funded it. A rent sponsor should keep or reconstruct an index of sponsored channels, for example from `Opened` events or channel accounts where `rentPayer` equals the operator, so it can close stale channels and reclaim rent. This is operational bookkeeping, not a trust assumption.

lgalabru:
Thanks for the review, @phdargen!

To clarify the intent: the program has been designed for a gasless client experience. I believe the program has all the knobs we need; the main challenge for this spec is exposing them at the right x402 abstraction layer.

For example, I removed the push/pull terminology because I do not think it is a useful distinction in the context of x402. 

@notoriousdev both case 1 and case 2 are natively supported, via native split support.
So concretely we should have: 

Case 1: 

```json
{
  "scheme": "upto",
  "amount": "10000",
  "asset": "<mint>",
  "payTo": "<merchant/operator>",
  "extra": {
    "channelProgram": "CHNLxYvVA28MJP9PrFuDXccuoGXAx7jBacfLEkahyGsX",
    "recentBlockhash": "<cached>",
  }
}
```

Case 2

```json
{
  "scheme": "upto",
  "amount": "10000",
  "asset": "<mint>",
  "payTo": "<merchant>",
  "extra": {
    "facilitatorAddress": "<operator>",
    "facilitatorFee": 0,
    "channelProgram": "CHNLxYvVA28MJP9PrFuDXccuoGXAx7jBacfLEkahyGsX",
    "recentBlockhash": "<cached>",
  }
}
```

Based on the spec returned by the server, the client constructs the channel/open transaction with the corresponding split configuration.

Happy to unpack the underlying mechanics if useful, but my instinct is that detailed payment-channel program behavior is out of scope for this scheme spec. The other scheme specs generally define the x402 wire contract and validation expectations rather than walking through every implementation detail.

On the two-transactions/fees concern, I do not think that should be a blocker on Solana. The client remains gasless/stablecoin-only, fees are low, and the latency roadmap is moving toward shorter confirmation times (200ms in a few weeks, 50ms by Q3), so the extra transaction round trip should be acceptable for this flow.

lgalabru:
@notorious-d-e-v thanks for the feedback!
I clarified Case 1 and Case 2 with https://github.com/x402-foundation/x402/pull/2697/commits/a580a2d83d9c9780bde4d130bb0664e6c53dd2d0.

With these changes, `payTo` would be a `feePayer` by default. Is that the right level of abstraction?

gitteri:
> As a first impression, I am not sure if payment-channels are the best fit for the upto scheme in terms of UX as this approach requires 2 txs (channel open and close) doubling settle latency + cost and seems a bit heavy. Have any alternatives been considered?
> 
> Will also discuss with the team

Hey @phdargen, thank you for the review and feedback. As employees of the Solana Foundation and Solana integration owners (Ludo & myself are codeowners for @x402-foundation/svm) we should be involved in these discussions and should be leading the Solana implementations with our ecosystem

specs/schemes/upto/scheme_upto_svm.md
phdargen: Think its sufficient to define the canonical `channelProgram` in the specs and as a constant in the SDK. Neither EVM upto or batch-settlement put the respective canonical contract address in `extra`

specs/schemes/upto/scheme_upto_svm.md
phdargen: Found this reference in the batch-settlement thread: https://github.com/solana-foundation/payment-channels/tree/main

a link to this reference would be useful to have in the spec

specs/schemes/upto/scheme_upto_svm.md
phdargen: Might be fine to drop entirely, unless we plan to add a second method or make it optional defaulting to ''payment-channel"

specs/schemes/upto/scheme_upto_svm.md
phdargen: OK could you please add this info to the spec?

The `grace_period` should probably not left for clients to choose or they could set it to ~0 and withdraw before the server had the chance to settle. Suggest to have a server defined `extra.withdrawDelay` (aligning with batch-settlement EVM convention) that the client must use and facilitator verifies before broadcasting openTransaction

phdargen:
> Hey @phdargen, thank you for the review and feedback. As employees of the Solana Foundation and Solana integration owners (Ludo & myself are codeowners for @x402-foundation/svm) we should be involved in these discussions and should be leading the Solana implementations with our ecosystem

Agreed and your initiative is much appreciated. Was just sharing my initial thoughts as constructive feedback and, given the importance of this addition, wanted to get some more eyes from our team on this proposal

phdargen:
The `operator == extra.facilitatorAddress` flow as mapped out in the spec still has the authentication gap I highlighted in my earlier comment and puts an unnecessary trust assumption on the facilitator. The facilitator should not be channel `payee` or `authorizedSigner`.
I think we can still achieve the desired gasless client + server experience, where the facilitator sponsors both channel open and close as I will outline below.

I also think we can simplify the spec substantially by dropping the 'operator' concept including self-facilitation vs external facilitator distinction and just use the x402 standard roles:
- **client**: signs payment authorization
- **server**: provides resource, receives payment
- **facilitator**: verifies and settles payment, sponsors tx fee

A server may also act as facilitator (self-facilitate), this is implied and always true for all schemes and networks.

Now to my suggested flow: 

**Channel open** (unchanged)
- client builds an signs `OpenTransaction`
- facilitator cosigns as `feePayer` + rent payer and broadcasts

**Channel close**
- server has the authority on the `amount` to be settled. Thus it should sign the voucher, which also serves as authentication for the otherwise unauthenticated settle/ http request. A dedicated signing key, dubbed `receiverAuthorizer` aligning with the batch-settlement convention, should be used for this, such that servers do not need to expose their `payTo` private key in their deployment. `receiverAuthorizer` doesn't need to hold any funds
- server builds and signs the `settleAndFinalize` tx. Only the `payee` can call `settleAndFinalize`, so `receiverAuthorizer` should be named `payee`. However, the payment should go to `payTo`. This is achieved by setting `payTo` as `recipient` that receives a share of 100%.  0% go to the `payee=receiverAuthorizer`
- server calls settle/ with a payload including voucher + `settleAndFinalize`
- facilitator cosigns `settleAndFinalize` as `feePayer` and broadcasts along with the `distribute` tx

Example 402:
```
    {
      "scheme": "upto",
      "amount": "5000000",
      "payTo": "<Server cold wallet>", # receives funds
      "maxTimeoutSeconds": 300,
      "extra": {
        "feePayer": "<facilitator>", # sponsors fees
        "receiverAuthorizer": "<Server hot wallet>",  # signs voucher + settleAndFinalize
        "withdrawDelay": 3600,
      }
    }
```

Mapping of x402 wire fields to channel program:

```
payee  = extra.receiverAuthorizer # zero-share payee (signs settleAndFinalize), server hot wallet
authorized_signer = extra.receiverAuthorizer # voucher signer, server hot wallet
rent_payer  = extra.feePayer # pays rent + tx fees, facilitator
recipients  = [{ recipient: payTo, bps: 10000 }] # 100% to payTo (server cold wallet); payee remainder = 0
grace_period = extra.withdrawDelay  
```

A self-facilitating server may use the same private key for `payTo`, `receiverAuthorizer` and `feePayer` but there is no need to treat this differently. And for better operational security it is recommended to use different keys for each role, even if one entity acts as both server and facilitator

phdargen:
> The `operator == extra.facilitatorAddress` flow as mapped out in the spec still has the authentication gap I highlighted in my earlier comment and puts an unnecessary trust assumption on the facilitator. The facilitator should not be channel `payee` or `authorizedSigner`. I think we can still achieve the desired gasless client + server experience, where the facilitator sponsors both channel open and close as I will outline below.
> 
> I also think we can simplify the spec substantially by dropping the 'operator' concept including self-facilitation vs external facilitator distinction and just use the x402 standard roles:
> 
> * **client**: signs payment authorization
> * **server**: provides resource, receives payment
> * **facilitator**: verifies and settles payment, sponsors tx fee
> 
> A server may also act as facilitator (self-facilitate), this is implied and always true for all schemes and networks.
> 
> Now to my suggested flow:
> 
> **Channel open**
> 
> * client builds an signs `OpenTransaction`
> * facilitator cosigns as `feePayer` + rent payer and broadcasts
> 
> **Channel close**
> 
> * server has the authority on the `amount` to be settled. Thus it should sign the voucher, which also serves as authentication for the otherwise unauthenticated settle/ http request. A dedicated signing key, dubbed `receiverAuthorizer` aligning with the batch-settlement convention, should be used for this, such that servers do not need to expose their `payTo` private key in their deployment. `receiverAuthorizer` doesn't need to hold any funds
> * server builds and signs the `settleAndFinalize` tx. Only the `payee` can call `settleAndFinalize`, so `receiverAuthorizer` should be named `payee`. However, the payment should go to `payTo`. This is achieved by setting `payTo` as `recipient` that receives a share of 100%.  0% go to the `payee=receiverAuthorizer`
> * server calls settle/ with a payload including voucher + `settleAndFinalize`
> * facilitator cosigns `settleAndFinalize` as `feePayer` and broadcasts along with the `distribute` tx
> 
> Example 402:
> 
> ```
>     {
>       "scheme": "upto",
>       "network": "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp",
>       "amount": "5000000",
>       "asset": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
>       "payTo": "<Server cold wallet>", # receives funds
>       "maxTimeoutSeconds": 300,
>       "extra": {
>         "feePayer": "<facilitator>", # sponsors fees
>         "receiverAuthorizer": "<Server hot wallet>",  # signs voucher + settleAndFinalize
>         "withdrawDelay": 3600,
>       }
>     }
> ```
> 
> Mapping of x402 wire fields to channel program:
> 
> ```
> payee  = extra.receiverAuthorizer # zero-share payee (signs settleAndFinalize), server hot wallet
> authorized_signer = extra.receiverAuthorizer # voucher signer, server hot wallet
> rent_payer  = extra.feePayer # pays rent + tx fees, facilitator
> recipients  = [{ recipient: payTo, bps: 10000 }] # 100% to payTo (server cold wallet); payee remainder = 0
> grace_period = extra.withdrawDelay  
> ```
> 
> A self-facilitating server may use the same private key for `payTo`, `receiverAuthorizer` and `feePayer` but there is no need to treat this differently. And for better operational security it is recommended to use different keys for each role, even if one entity acts as both server and facilitator

The flow I proposed above would require a RPC for the server to fetch a fresh blockhash for the `settleAndFinalize` tx, what isn't ideal. As an optional convenience for servers that do not want to worry about a RPC, we could consider a flow with the facilitator as zero-share payee. Server remains the voucher signer, which would serve as authentication gating the settle/ call. 

The facilitator could not settle any nonzero amount without the server's voucher but it could call `settleAndFinalize(has_voucher=0)` leaving a residual trust assumption. Given that there is no incentive for the facilitator to do so, I think this would be acceptable

```
payee  = extra.feePayer # zero-share payee (signs settleAndFinalize), facilitator
authorized_signer = extra.receiverAuthorizer # voucher signer, server hot wallet
rent_payer  = extra.feePayer # pays rent + tx fees, facilitator
recipients  = [{ recipient: payTo, bps: 10000 }] # 100% to payTo (server cold wallet); payee remainder = 0
grace_period = extra.withdrawDelay  
```

phdargen (changes requested):

phdargen (approved):
Thanks @lgalabru, LGTM 🚀

PR #2698, opened 2026-06-25
docs(svm): add `batch-settlement` SVM scheme specification

lgalabru:
## Summary

- Adds the SVM profile for long-lived `batch-settlement` payment channels.
- Specifies deposit/top-up validation, cumulative vouchers, batched claim/distribution, replay handling, and lifecycle recovery.
- Makes payer-signed `request_close` the interoperable refund path: the facilitator sponsors initiation and finalizes after a 15-minute to 30-day grace period. Authenticated cooperative close remains an optional fast path.
- Aligns setup with the current SVM `upto` profile: protocol-default `authorization` flow, verified token-program binding, Memo/block-height hints, settlement-readiness checks, and post-confirmation state binding.

## Validation

- All 16 JSON examples parse.
- Local Markdown links resolve.
- `git diff --check upstream/main...HEAD`.

## Checklist

- [x] Commits are SSH-signed
- [x] Documentation-only change; no changelog fragment

### DIFF
--- a/specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
+++ b/specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
@@ -28,10 +28,11 @@ The x402 roles map to the payment-channel program as follows:
 - **Server**: resource provider; receives funds at `payTo`; owns per-channel
   offchain state, including the accepted cumulative watermark and latest
   voucher.
-- **Receiver authorizer**: server-controlled Ed25519 key advertised as
+- **Receiver authorizer**: optional server-controlled Ed25519 key advertised as
   `extra.receiverAuthorizer`. It authenticates server-approved cooperative
-  refund requests to the facilitator. It is an offchain x402 role and is not
-  recorded in the payment-channel account.
+  close requests to the facilitator. It is an offchain x402 role and is not
+  recorded in the payment-channel account. It is not required for unilateral
+  refund or rent recovery.
 - **Facilitator / sponsor**: account advertised as `extra.feePayer`; sponsors
   transaction fees and channel rent. It is the transaction fee payer and channel
   `rent_payer`, and occupies the channel `payee` lifecycle-authority seat with a
@@ -66,7 +67,7 @@ claim promptly.
 | Monotonic amount | Server-owned offchain watermark plus onchain `settled < maxClaimableAmount <= deposit` at redemption. |
 | Batched redemption | One `settle` per channel, packed transaction-size permitting; `distribute` pays settled deltas. |
 | Recipient binding | `distribution_hash` fixed at `open` sends funds to `payTo`; program re-checks it at `distribute`. |
-| Recovery of unused deposit | Server-authorized `settle_and_seal` + `distribute` for full cooperative close; `request_close` / `seal` / `withdraw_payer` for payer-forced close. |
+| Recovery of unused deposit | Client-signed `request_close`, followed after the grace period by permissionless `seal` + `distribute`; an optional server authorization can replace this with immediate cooperative `settle_and_seal` + `distribute`. |
 
 ## 3. Payment-Channel Method
 
@@ -114,16 +115,31 @@ to rediscover the channels it sponsored as specified in section 6. Token
 payouts and the return of unused escrow are completed by `distribute` and are
 not delayed by the later `reclaim` of PDA rent.
 
-For an x402 cooperative refund, the facilitator MUST additionally require a
-valid refund authorization from `extra.receiverAuthorizer`. That signature
-authenticates the server's HTTP request; it does not replace the facilitator's
-onchain `payee` signature or the client's voucher signature. The facilitator
-MUST bind `receiverAuthorizer` to the authenticated resource server and `payTo`
-through trusted configuration or an authenticated registration established no
-later than channel opening. It MUST NOT trust a receiver-authorizer key merely
-because it appears in `PaymentRequirements` or a settlement request. The
-trusted receiver authorizer is fixed for the channel lifetime; changing it
-requires cooperatively closing the channel and opening a new one.
+For an x402 immediate cooperative close, the facilitator MUST authenticate the
+resource server and bind that identity to `payTo` and the channel. It MAY do so
+with a valid `CloseAuthorization` signed by `extra.receiverAuthorizer`, or with
+facilitator-specific out-of-band authentication established no later than the
+channel's first deposit. Neither mechanism replaces the facilitator's onchain
+`payee` signature or the client's voucher signature. The facilitator MUST NOT
+trust a receiver-authorizer key or claimed server identity merely because it
+appears in `PaymentRequirements` or a settlement request. Any trusted binding
+is fixed for the channel lifetime; changing it requires closing the channel and
+opening a new one.
+
+For out-of-band authentication, the facilitator MUST record the authenticated
+server principal that submitted or sponsored the first deposit together with
+the derived `channelId` and verified `payTo`. It MUST require that same principal
+on an immediate cooperative-close request. This mechanism is facilitator-local;
+the interoperable wire fallback remains payer-signed forced close.
+
+The facilitator MUST accept a refund without a receiver authorizer. In that
+case, the client supplies a payer-signed `request_close` transaction, the
+facilitator adds its fee-payer signature and broadcasts it, and the facilitator
+or another permissionless crank finalizes the channel after the grace period.
+The facilitator MAY use the immediate cooperative path when it receives the
+server's latest valid voucher through an authenticated request under that
+binding. Out-of-band authentication is an optimization and is not required for
+interoperability.
 
 If the client invokes `request_close`, the channel enters `Closing` and regular
 `settle` is no longer available. Only the `payee` can apply a final voucher with
@@ -155,17 +171,22 @@ and `SettlementResponse` types are defined in
 
 | Field | Type | Required | Notes |
 |---|---|---|---|
+| `paymentFlow` | string | yes | MUST be `"escrow"`, because the deposit settles before the resource handler and later vouchers authorize asynchronous claims. |
 | `feePayer` | string | yes | Base58 sponsor key set as channel `rent_payer` and zero-share `payee`. Co-signs setup/top-up transactions as transaction fee payer and signs channel lifecycle transactions. |
-| `receiverAuthorizer` | string | yes | Base58 server-controlled Ed25519 key that authenticates cooperative refund requests to the facilitator. It is not a payment-channel account field. |
-| `withdrawDelay` | number | yes | Server-defined forced-close grace period in seconds. The client MUST encode this exact value as the program `grace_period`; the verifier MUST reject any other value. |
+| `receiverAuthorizer` | string | no | Base58 server-controlled Ed25519 key that authenticates an optional immediate cooperative close to the facilitator. It is not a payment-channel account field. |
+| `withdrawDelay` | number | yes | Forced-close grace period in seconds. MUST be an integer from `900` through `2592000` (15 minutes through 30 days), MUST be `>= maxTimeoutSeconds`, and MUST be encoded exactly as the program `grace_period`. |
+| `tokenProgram` | string | yes | SPL Token (`Tokenkeg...`) or Token-2022 (`TokenzQ...`) program that owns `asset`. The client and facilitator MUST verify it against the onchain mint owner. |
+| `memo` | string | no | Seller-defined UTF-8 payment reference for the setup transaction's Memo instruction. Maximum 256 bytes. |
 | `recentBlockhash` | string | no | Pre-fetched blockhash the client MAY use to build an `open` or `top_up` transaction without an RPC round trip. The client MUST refresh it if it is no longer valid. |
+| `lastValidBlockHeight` | string | no | Last block height at which `recentBlockhash` is valid, as a decimal string. Informational; MAY be ignored. Ignored when `recentBlockhash` is absent. |
 | `recentSlot` | number | no | Recent slot the client MAY use as `channelConfig.openSlot` when it does not fetch its own slot. The program still enforces the open-slot window. |
 | `channelState` | object | no | Corrective-only server channel snapshot for cumulative amount resynchronization. |
 | `voucherState` | object | no | Corrective-only signed voucher proof for cumulative amount resynchronization. |
 
-`recentBlockhash` and `recentSlot` are transaction-construction hints only. They
-are not persistent channel configuration and are not included in the voucher
-message. A client MAY ignore either hint and obtain a fresher value from an RPC.
+`recentBlockhash`, `lastValidBlockHeight`, and `recentSlot` are
+transaction-construction hints only. They are not persistent channel
+configuration and are not included in the voucher message. A client MAY ignore
+the hints and obtain fresher values from an RPC.
 
 The x402 wire format does not expose program-specific split arrays. The client
 derives the payment-channel accounts and distribution from the x402 fields:
@@ -196,10 +217,14 @@ Example:
   "payTo": "<server-receiver>",
   "maxTimeoutSeconds": 300,
   "extra": {
+    "paymentFlow": "escrow",
     "feePayer": "<facilitator-fee-payer>",
-    "receiverAuthorizer": "<server-refund-authorizer>",
+    "receiverAuthorizer": "<server-close-authorizer>",
     "withdrawDelay": 3600,
+    "tokenProgram": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
+    "memo": "invoice-123",
     "recentBlockhash": "<recent-blockhash>",
+    "lastValidBlockHeight": "321000000",
     "recentSlot": 341000000
   }
 }
@@ -217,11 +242,12 @@ may expose those program fields with language-specific casing.
 | x402 wire field | Payment-channels program notation |
 |---|---|
 | `PaymentRequirements.extra.feePayer` | `Channel.rent_payer` and zero-share `Channel.payee` |
-| `PaymentRequirements.extra.receiverAuthorizer` | No program field; trusted server key for offchain refund authorization |
+| `PaymentRequirements.extra.receiverAuthorizer` | No program field; optional trusted server key for offchain close authorization |
+| `PaymentRequirements.extra.tokenProgram` | `open.token_program` / `top_up.token_program`; MUST equal the onchain owner of `Channel.mint` |
 | `channelConfig.payer` | `Channel.payer` |
 | `channelConfig.payerAuthorizer` | `Channel.authorized_signer` |
 | `channelConfig.receiver` | Sole `DistributionEntry.recipient` with `bps = 10000`; the `Channel.payee` implicit remainder is always zero |
-| `channelConfig.receiverAuthorizer` | No program field; MUST equal `PaymentRequirements.extra.receiverAuthorizer` |
+| `channelConfig.receiverAuthorizer` | No program field; when supplied, MUST equal `PaymentRequirements.extra.receiverAuthorizer` |
 | `channelConfig.token` | `Channel.mint` |
 | `channelConfig.withdrawDelay` | `Channel.grace_period` |
 | `channelConfig.salt` | `Channel.salt` |
@@ -241,15 +267,15 @@ may expose those program fields with language-specific casing.
 | `payer` | string | Client wallet and channel payer. MUST NOT equal `feePayer`, because the program requires distinct payer and payee accounts. |
 | `payerAuthorizer` | string | Client-controlled Ed25519 voucher signer; maps to channel `authorized_signer`. MAY equal `payer` but MUST NOT equal `feePayer`. |
 | `receiver` | string | MUST equal `payTo`. |
-| `receiverAuthorizer` | string | MUST equal `extra.receiverAuthorizer`. This key authenticates cooperative refund requests offchain and is not a channel PDA seed or program account field. |
+| `receiverAuthorizer` | string | Optional. When supplied, MUST equal `extra.receiverAuthorizer`. This key authenticates cooperative close requests offchain and is not a channel PDA seed or program account field. |
 | `token` | string | MUST equal `asset`; maps to channel `mint`. |
 | `withdrawDelay` | number | MUST equal `extra.withdrawDelay`; maps to channel `grace_period`. |
 | `salt` | string | Decimal `u64` channel salt. |
 | `openSlot` | number | `u64` slot encoded in `open` and used as a channel PDA seed. |
 
-The client MUST determine the token program from the onchain owner of
-`channelConfig.token`; it MUST NOT accept an unverified token-program value from
-the server.
+The client and facilitator MUST confirm `extra.tokenProgram` equals the onchain
+owner of `channelConfig.token`; neither role may trust the server-provided value
+without that check.
 
 `channelId` is the program-derived address:
 
@@ -299,17 +325,17 @@ The signed message is exactly:
 This is the payment-channels program voucher layout (`VOUCHER_MAGIC`,
 `channel_id`, `cumulative_amount`, `expires_at`), 50 bytes total.
 
-`RefundAuthorization`:
+`CloseAuthorization`:
 
 | Field | Type | Notes |
 |---|---|---|
 | `validBefore` | number | Integer Unix seconds. The server MUST set it no later than its current time plus `maxTimeoutSeconds`; the facilitator MUST require `now < validBefore <= now + maxTimeoutSeconds`. |
-| `signature` | string | Base58 Ed25519 signature by `extra.receiverAuthorizer` over the digest below. |
+| `signature` | string | Base58 Ed25519 signature by the receiver authorizer bound to the server through `extra.receiverAuthorizer` or a trusted out-of-band registration. |
 
 The receiver authorizer signs the SHA-256 digest of this exact byte sequence:
 
 ```text
-UTF8("x402:batch-settlement:svm:refund:v1") || 0x00 ||
+UTF8("x402:batch-settlement:svm:close:v1") || 0x00 ||
 u16(len(network)).le || UTF8(network) ||
 programId[32] ||
 feePayer[32] ||
@@ -339,9 +365,10 @@ program deployment, facilitator, channel, voucher watermark, or expiry.
 ### 4.3 Client `PaymentPayload` Variants
 
 The client payload is a tagged union on `payload.type`. `/verify` accepts all
-three variants. `/settle` accepts `deposit` directly; the server enriches
-`refund` with a `RefundAuthorization` before forwarding it as specified in
-section 4.5. A `voucher` is accepted offchain by the server.
+three variants. `/settle` accepts `deposit` and `refund` directly. The server
+MAY enrich `refund` with its latest accepted voucher and, when needed, a
+`CloseAuthorization` before forwarding it as specified in section 4.5. A
+`voucher` is accepted offchain by the server.
 
 **`deposit`** opens a channel or tops up an existing channel and authorizes the
 current request:
@@ -365,10 +392,13 @@ current request:
     "payTo": "<server-receiver>",
     "maxTimeoutSeconds": 300,
     "extra": {
+      "paymentFlow": "escrow",
       "feePayer": "<facilitator-fee-payer>",
-      "receiverAuthorizer": "<server-refund-authorizer>",
+      "receiverAuthorizer": "<server-close-authorizer>",
       "withdrawDelay": 3600,
+      "tokenProgram": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
       "recentBlockhash": "<recent-blockhash>",
+      "lastValidBlockHeight": "321000000",
       "recentSlot": 341000000
     }
   },
@@ -378,7 +408,7 @@ current request:
       "payer": "<client-wallet>",
       "payerAuthorizer": "<client-voucher-signer>",
       "receiver": "<server-receiver>",
-      "receiverAuthorizer": "<server-refund-authorizer>",
+      "receiverAuthorizer": "<server-close-authorizer>",
       "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
       "withdrawDelay": 3600,
       "salt": "42",
@@ -418,10 +448,13 @@ transaction:
     "payTo": "<server-receiver>",
     "maxTimeoutSeconds": 300,
     "extra": {
+      "paymentFlow": "escrow",
       "feePayer": "<facilitator-fee-payer>",
-      "receiverAuthorizer": "<server-refund-authorizer>",
+      "receiverAuthorizer": "<server-close-authorizer>",
       "withdrawDelay": 3600,
+      "tokenProgram": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
       "recentBlockhash": "<recent-blockhash>",
+      "lastValidBlockHeight": "321000000",
       "recentSlot": 341000000
     }
   },
@@ -431,7 +464,7 @@ transaction:
       "payer": "<client-wallet>",
       "payerAuthorizer": "<client-voucher-signer>",
       "receiver": "<server-receiver>",
-      "receiverAuthorizer": "<server-refund-authorizer>",
+      "receiverAuthorizer": "<server-close-authorizer>",
       "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
       "withdrawDelay": 3600,
       "salt": "42",
@@ -447,17 +480,21 @@ transaction:
 }
 ```
 
-**`refund`** requests immediate cooperative closure and the return of all
-unspent escrow. It is a payment operation, not a paid resource request, so the
-application resource handler MUST be bypassed. The request MUST NOT contain an
-`amount`: the payment-channel program supports only a full refund of
-`deposit - settled`, and the closed channel cannot be reused.
+**`refund`** is retained as the cross-binding wire discriminator for a channel
+close. The client MUST provide a payer-signed `request_close` transaction. This
+is a payment operation, not a paid resource request, so the application
+resource handler MUST be bypassed. The request MUST NOT contain an `amount`:
+the payment-channel program supports only the return of all unused escrow, and
+the closed channel cannot be reused. A client MAY also provide a voucher as a
+hint for an immediate cooperative close, but the facilitator MUST NOT apply it
+unless an authenticated server confirms it is the latest accepted voucher.
 
 | Field | Type | Notes |
 |---|---|---|
 | `type` | string | `"refund"` |
 | `channelConfig` | `ChannelConfig` | Full channel configuration. |
-| `voucher` | `BatchVoucher` | Zero-charge voucher whose `maxClaimableAmount` equals the server's current `chargedCumulativeAmount`. |
+| `transaction` | string | Base64 client-signed `request_close` transaction. The transaction MUST set `extra.feePayer` as its Solana fee payer so the facilitator can co-sign and broadcast it. |
+| `voucher` | `BatchVoucher` | Optional latest accepted voucher. It is used only when supplied or confirmed by an authenticated server for immediate cooperative close. |
 
 ```json
 {
@@ -470,10 +507,12 @@ application resource handler MUST be bypassed. The request MUST NOT contain an
     "payTo": "<server-receiver>",
     "maxTimeoutSeconds": 300,
     "extra": {
+      "paymentFlow": "escrow",
       "feePayer": "<facilitator-fee-payer>",
-      "receiverAuthorizer": "<server-refund-authorizer>",
       "withdrawDelay": 3600,
+      "tokenProgram": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
       "recentBlockhash": "<recent-blockhash>",
+      "lastValidBlockHeight": "321000000",
       "recentSlot": 341000000
     }
   },
@@ -483,18 +522,12 @@ application resource handler MUST be bypassed. The request MUST NOT contain an
       "payer": "<client-wallet>",
       "payerAuthorizer": "<client-voucher-signer>",
       "receiver": "<server-receiver>",
-      "receiverAuthorizer": "<server-refund-authorizer>",
       "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
       "withdrawDelay": 3600,
       "salt": "42",
       "openSlot": 341000000
     },
-    "voucher": {
-      "channelId": "<channel-pda>",
-      "maxClaimableAmount": "5000",
-      "expiresAt": 0,
-      "signature": "<base58-ed25519-zero-charge-voucher-signature>"
-    }
+    "transaction": "<base64-client-signed-request-close-transaction>"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Phase 2 — Steady state (per request)
+
+The client increments the cumulative total by the per-request `amount`, signs a
+new `BatchVoucher`, and sends a `voucher` payload. No on-chain transaction. The
+operator accepts the voucher off-chain (Phase 3) and serves immediately.
+
+### Phase 3 — Voucher acceptance (before serving)
+
+For every voucher the operator MUST, via `core::session::accept_voucher`:
```
notorious-d-e-v: We may want to keep function names outside of the spec as it seems more like an implementation detail.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+   confirm `signer` equals the channel's `authorizedSigner`.
+2. Confirm `expiresAt > now` (reject expired vouchers).
+3. Enforce **monotonicity**: `cumulativeAmount` ≥ the stored watermark.
+4. Enforce the **deposit cap**: `cumulativeAmount ≤ channel.deposit`.
+5. Enforce `minVoucherDelta` (if set) on the increment over the previous
+   cumulative.
+6. **Idempotent replay**: a voucher equal to the current watermark is accepted as
+   a no-op (the same request retried), not double-charged.
+7. Atomically advance the stored cumulative (compare-and-set).
+
```
notorious-d-e-v: Does it make sense to add another step that the `operator` SHOULD add some buffer to `gracePeriodSeconds` to make sure that they have enough time to settle before forced close?

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+The client builds an `open` transaction depositing `depositAmount` (≥
+`extra.minimumDeposit`), with `authorizedSigner = payer` (client-voucher model),
+`payee = payTo`, and `distributionSplits` matching `extra.distributionSplits`.
+The client sends a `deposit` payload carrying the base64 client-signed
+transaction plus the first cumulative voucher.
+
+The operator MUST validate the `open` transaction before co-signing as fee payer
+(the SOL-drain guard, reused from `upto`): exactly one instruction, the expected
+program id and `open` discriminator, and accounts/args matching `channelConfig`.
+It then co-signs, broadcasts, confirms, binds the on-chain channel into its
```
notorious-d-e-v: **`ChannelStore` attribution: state belongs to the resource server, not the facilitator**

Phase 1 says the operator "binds the on-chain channel into its `ChannelStore`" and Phase 3 has it own the off-chain watermark. The word "operator" is ambiguous — read as *facilitator*, this conflicts with x402's rule that the facilitator carries no state for core scheme operations.

The EVM profile settles this explicitly. From **§ Server: State & Forwarding**:

> "The server is the sole owner of per-channel state."
>
> "The server must maintain per-channel state, keyed by channel ID" — including `chargedCumulativeAmount` ("Actual accumulated cost for this channel"), `signedMaxClaimable` ("`maxClaimableAmount` from the latest client-signed voucher"), and `signature` ("Client's voucher signature for the latest `signedMaxClaimable`").

The facilitator there stays stateless — it exposes the standard `/verify`, `/settle`, `/supported`, and the server mirrors on-chain values from the facilitator's response (`balance`, `totalClaimed`, etc.) after each request. The watermark and latest voucher live on the **server**; the facilitator just reads chain and submits transactions.

The SVM profile should make the same assignment, because the mechanics already allow it:

- **Serve-time acceptance** (Phase 3) is the only step that needs the off-chain watermark — a resource-server billing concern, and servers are permitted to be stateful.
- **Settlement** (Phase 4) needs only the *latest cumulative voucher*, which is self-contained (`channel_id ‖ cumulative_amount ‖ expires_at`). With on-chain `SettlementWatermarks.settled` + `deposit`, the facilitator can verify and `settle` **statelessly** — no `ChannelStore` required.


**Suggested edits:**
- Phase 1/3: attribute the watermark/`ChannelStore` to the **resource server**, mirroring EVM's "The server is the sole owner of per-channel state"; note the facilitator settles statelessly from the latest voucher + on-chain state.
- §7 Security properties: add facilitator-agnostic settlement (latest cumulative voucher + permissionless `settle` to the fixed `payee`).
- §8: clarify the durable `ChannelStore` follow-up is server-side state.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+commit round-trip and is deferred.
+
+## 3. Profile
+
+A server advertises supported profiles in `extra.profiles`. v1 defines a single
+normative profile:
+
+### 3.1 `payment-channel` (normative, v1)
+
+Backed by the on-chain payment-channels program (a `pay-kit`-controlled program;
```
notorious-d-e-v: Can we verify/cite this deployment? The referenced `payment-channels` repo declares only the local/test id `CQAyft83tN1w2bRofB5PZ79eVDU2xZUVo43LU1qL4zRg` and says production program ids are explicit deployments. If `CHNLx...` is real, the spec should point to the deployment/build artifact; otherwise readers may derive PDAs or submit instructions against the wrong program.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+[`x402-specification-v2.md`](../../x402-specification-v2.md).
+
+### 4.1 `PaymentRequirements` (in `PAYMENT-REQUIRED.accepts[]`)
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `scheme` | string | ✓ | `"batch-settlement"` |
+| `network` | string | ✓ | CAIP-2, e.g. `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp` (mainnet), `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1` (devnet) |
+| `amount` | string | ✓ | Per-request price, base units (fixed in v1) |
+| `asset` | string | ✓ | SPL mint address (or a known symbol like `"USDC"`) |
```
notorious-d-e-v: It may be better if we specify `asset` as the concrete SPL/Token-2022 mint pubkey, not a symbol like `"USDC"`.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `success` | boolean | ✓ | |
+| `errorReason` | string | – | Omitted on success |
+| `payer` | string | – | Channel `payer` |
+| `transaction` | string | ✓ | On-chain signature; **empty `""` for an off-chain voucher acceptance** (the common case) |
+| `network` | string | ✓ | CAIP-2 |
+| `amount` | string | ✓ | Amount moved on-chain (`""` for voucher-only) |
+| `chargedAmount` | string | – | The per-request charge committed off-chain |
```
notorious-d-e-v: `chargedAmount` and `channelState` should be nested under `extra`, matching the core `SettleResponse` shape and the EVM batch-settlement profile (`extra.chargedAmount`, `extra.channelState`). 

Also, core marks `amount` optional; voucher-only responses can use `amount: ""` by convention, but making it required here diverges from the shared type.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Phase 1 — Open (once per channel)
+
+The client builds an `open` transaction depositing `depositAmount` (≥
+`extra.minimumDeposit`), with `authorizedSigner = payer` (client-voucher model),
+`payee = payTo`, and `distributionSplits` matching `extra.distributionSplits`.
+The client sends a `deposit` payload carrying the base64 client-signed
+transaction plus the first cumulative voucher.
+
+The operator MUST validate the `open` transaction before co-signing as fee payer
```
notorious-d-e-v: This SOL-drain guard is too narrow for a server fee-payer flow. Validating the expected payment-channel instruction is not enough; the operator should validate the full compiled transaction, including ALT resolution, allowed instruction/account/signer set, and that the fee payer is never used as an authority/source/writable account except for fees. Otherwise a malicious client can make the server co-sign more than just fee payment.

I think it's well fleshed out in the `exact` scheme and implementation so we can borrow from there in order to make this spec more explicit.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+For every voucher the operator MUST, via `core::session::accept_voucher`:
+
+1. Verify the Ed25519 `signature` over the 48-byte message against `signer`, and
+   confirm `signer` equals the channel's `authorizedSigner`.
+2. Confirm `expiresAt > now` (reject expired vouchers).
+3. Enforce **monotonicity**: `cumulativeAmount` ≥ the stored watermark.
+4. Enforce the **deposit cap**: `cumulativeAmount ≤ channel.deposit`.
+5. Enforce `minVoucherDelta` (if set) on the increment over the previous
+   cumulative.
+6. **Idempotent replay**: a voucher equal to the current watermark is accepted as
```
notorious-d-e-v: Allowing `cumulativeAmount == stored watermark` as an idempotent no-op is unsafe at the HTTP layer unless it is tied to a request id/cached response. 

The signed voucher is only `{channelId, cumulativeAmount, expiresAt}`, so the same voucher can be replayed for a later request and buy another response without increasing the cumulative amount. 

For fixed-price v1, steady-state requests should require `previous + amount`; equality should only be accepted for a true retry of the same request.

Otherwise we may want to explicitly state that a voucher equal to the current watermark is okay as long as the server is keeping state on what it's already seen and tying it to a request id/cached response

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+- **No overcharge.** Each voucher is capped on-chain by `deposit`; `settle` draws
+  the voucher's exact cumulative and can never exceed it.
+- **No redirection.** `channel.payee` / `distribution_hash` are fixed at `open`.
+- **No replay / no rollback.** Off-chain monotonic watermark plus on-chain
+  monotonic `SettlementWatermarks.settled`; an old voucher cannot settle after a
+  newer one.
+- **No fee-payer drain.** The operator validates the `open` transaction (single
+  instruction, correct program/discriminator/accounts) before co-signing as fee
+  payer.
+- **Time-bounded.** `expiresAt` is checked off-chain at acceptance and signed into
```
notorious-d-e-v: The expiry rule needs to account for asynchronous redemption. 

The on-chain program re-checks `expires_at` when `settle` / `settleAndFinalize` executes, not only when the HTTP request is accepted. 

So the server must require `expiresAt == 0` or a TTL that remains valid through the expected settlement window; otherwise it can serve a request for a voucher that expires before batch redemption.

This is somewhat related to the gracePeriodSeconds buffer comment above as well.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,263 @@
...
+
+### Phase 4 — Batched settlement (operator-driven, out of band)
+
+The operator redeems accumulated vouchers asynchronously — **not** in the request
+path:
+
+- `settle_batch(channel_ids)`: for each channel, build `[ed25519_verify(latest
+  voucher), settle]` from the stored highest voucher and pack as many channels as
+  fit per transaction (tx-size-bounded; the implementation caps at
+  `MAX_CHANNELS_PER_SETTLE_TX = 3`, and logs any channel dropped from a batch —
```
notorious-d-e-v: Is `MAX_CHANNELS_PER_SETTLE_TX = 3` defined in a referenced implementation? I couldn’t find that constant in the payment-channels program. If this is an SDK policy, cite it; otherwise this should probably be phrased generically as transaction-size-bounded packing.

notorious-d-e-v (commented):
Nice work @lgalabru!

notorious-d-e-v:
For reference the `payment-channels` program is here: https://github.com/solana-foundation/payment-channels/tree/main

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: With recent optimizations, I believe we can bump to 4/tx.
Post `simd-0296`, we should be able to bump to 16/tx.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Done in 9b377a8b — dropped the SDK function names (`accept_voucher`, `settle_batch`); kept the on-chain instruction names (`settle`/`distribute`/`settle_and_finalize`).

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Added in 9b377a8b — Phase 4 now SHOULDs the operator to settle with a buffer before forced-close and advertise a `gracePeriodSeconds` that covers its batch cadence. Ties into the expiry window below.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Good call — fixed in 9b377a8b. Phases 1/3 and the §2 row now attribute the watermark/`ChannelStore` to the **resource server**; a separate facilitator stays stateless for this verification.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Resolved — the hardcoded id is gone; the program is referenced via `extra.channelProgram` (a canonical deployment assumed when omitted). Thanks for the catch.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Done in 9b377a8b — `asset` is now the concrete SPL/Token-2022 mint pubkey (USDC `EPjFW…` as the example), not a symbol.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Done in 9b377a8b — `chargedAmount`/`channelState` are nested under `extra` and `amount` is now optional, matching the core `SettleResponse` and the EVM batch companion.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Agreed — tightened in 9b377a8b. Phase 1 now validates the full compiled `open` tx (ALT resolution, allowed-instruction-set only, fee payer never an authority/source/writable except for fees), pointing at the `exact` SVM scheme.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Great catch — fixed in 9b377a8b. The `cumulativeAmount` is the per-request nonce: a new request requires a strict increment (`watermark + amount`); an equal voucher is only an idempotent retry that replays the response cached at that cumulative — never a fresh serve, so a replay can't buy a new response.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Verified + fixed in 9b377a8b. Confirmed the program re-checks `expires_at` at `settle` too (not only `settleAndFinalize`), so Phase 3 now requires `expiresAt == 0` or a TTL that outlasts the redemption window (grace + buffer).

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: Genericized in 9b377a8b — phrased as transaction-size-bounded (a few per tx today, more as the runtime allows) rather than a hardcoded constant.

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,442 @@
...
+| `voucher` | object | New cumulative `BatchVoucher`. |
+
+**`refund`** - cooperative close; the application resource handler is bypassed:
+
+| Field | Type | Notes |
+|---|---|---|
+| `type` | string | `"refund"` |
+| `channelId` | string | Channel PDA (base58). |
+| `voucher` | object | Zero-charge voucher whose `cumulativeAmount` equals the server's current watermark. |
+| `amount` | string | Optional requested refund amount; omitted means full unspent remainder. |
```
phdargen: I dont think the program supports partial refunds

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,442 @@
...
+   - `cumulativeAmount == watermark`: idempotent retry only. Return the cached
+     response for `(channelId, cumulativeAmount)` and do not execute the
+     resource handler again.
+   - `cumulativeAmount > watermark`: fresh charge.
+     `cumulativeAmount == watermark + PaymentRequirements.amount` and the
+     increment is at least `minVoucherDelta` when set.
+7. For a fresh charge, atomically advance the stored watermark, store the latest
+   voucher, serve the resource, and return `PAYMENT-RESPONSE` with
+   `transaction == ""`, `extra.commitmentId`, `extra.chargedAmount`, and
+   `extra.channelState`.
```
phdargen: EVM requires per-channel serialization, handler execution, then state commit only on success. The current SVM order would charge failed requests

phdargen:
Thanks a lot for putting this together @lgalabru, overall looks great! 

Given the complexity of the scheme, all /verify, /settle payloads + responses should be explicitly defined, including the async server initiated `claim` and `settle` actions. Wire `claim` can map to program `settle`; wire `settle` can map to `distribute`. Also think it useful to show examples for each.

Suggested wire format aligning even more closely with EVM convention:


| x402 wire field                        | `payment-channels` program notation                                                                                       |
| -------------------------------------- | ------------------------------------------------------------------------------------------------------------------------- |
| `channelConfig.payer`                  | `Channel.payer`                                                                                     |
| `channelConfig.payerAuthorizer`        | `Channel.authorized_signer`                                                              |
| `channelConfig.receiver`               | `Channel.payee` when equal to `receiverAuthorizer`; otherwise the sole `DistributionEntry.recipient` with `bps = 10000`  |
| `channelConfig.receiverAuthorizer`     | `Channel.payee`                                                                                        |
| `channelConfig.token`                  | `Channel.mint`                                                                                          |
| `channelConfig.withdrawDelay`          | `Channel.grace_period`                                                                             |
| `channelConfig.salt`                   | `Channel.salt`                                                                           |
| `channelConfig.openSlot`               | `Channel.open_slot`                                           |
| `voucher.channelId`                    | `VoucherArgs.channel_id`                                                                           |
| `voucher.maxClaimableAmount`           | `VoucherArgs.cumulative_amount`                                               |
| `voucher.expiresAt`                    | `VoucherArgs.expires_at`; `0` disables voucher expiry                                                                    |
| `channelState.balance`                 | `Channel.deposit`                                                   |
| `channelState.totalClaimed`            | `Channel.settled`                                                                                                        |
| `channelState.withdrawRequestedAt`     | `Channel.closure_started_at`                                                                                             |
| `channelState.chargedCumulativeAmount` |  offchain cumulative charge                                                                |

## Settlement payload `type` → program instruction

Each wire payload `type`names the *effect* of the payload, not the SVM program instruction it invokes.

| Wire payload `type` | Effect | EVM contract call | SVM program instruction |
| --- | --- | --- | --- |
| `deposit` | Open a channel or add escrow | ERC-3009 / Permit2 deposit collector | `open` / `topUp` |
| `voucher` | Offchain authorization; no onchain transaction | — | — |
| `claim` | Advance accounting; no funds move to the receiver | `claimWithSignature` | `settle` |
| `settle` | Move earned funds to the receiver | `settle(receiver, token)` | `distribute` |
| `refund` | Cooperative withdraw; return unspent escrow to the payer | `claimWithSignature` + `refundWithSignature` | `settle_and_seal` + `distribute` |

# 402 PaymentRequirements

```json
{
  "scheme": "batch-settlement",
  "amount": "1000",
  "payTo": "<server-receiver>",
  "maxTimeoutSeconds": 300,
  "extra": {
    "feePayer": "<facilitator-fee-payer>",
    "receiverAuthorizer": "<receiver-authorizer>",
    "withdrawDelay": 3600
  }
}
```



# Deposit Client PaymentPayload

```json
{
  "x402Version": 2,
  "accepted": {
    "scheme": "batch-settlement",
    "amount": "1000",
    "payTo": "<server-receiver>",
    "maxTimeoutSeconds": 300,
    "extra": {
      "feePayer": "<facilitator-fee-payer>",
      "receiverAuthorizer": "<receiver-authorizer>",
      "withdrawDelay": 3600
    }
  },
  "payload": {
    "type": "deposit",
    "channelConfig": {
      "payer": "<client-wallet>",
      "payerAuthorizer": "<client-voucher-signer>",
      "receiver": "<server-receiver>",
      "receiverAuthorizer": "<receiver-authorizer>",
      "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
      "withdrawDelay": 3600,
      "salt": "42",
      "openSlot": "341000000"
    },
    "voucher": {
      "channelId": "<channel-pda>",
      "maxClaimableAmount": "1000",
      "expiresAt": 0,
      "signature": "<base58-ed25519-signature>"
    },
    "deposit": {
      "amount": "100000",
      "transaction": "<base64-client-signed-open-or-top-up-transaction>"
    }
  }
}
```



# Voucher Client PaymentPayload

```json
{
  "x402Version": 2,
  "accepted": {
    "scheme": "batch-settlement",
    "amount": "1000",
    "payTo": "<server-receiver>",
    "maxTimeoutSeconds": 300,
    "extra": {
      "feePayer": "<facilitator-fee-payer>",
      "receiverAuthorizer": "<receiver-authorizer>",
      "withdrawDelay": 3600
    }
  },
  "payload": {
    "type": "voucher",
    "channelConfig": {
      "payer": "<client-wallet>",
      "payerAuthorizer": "<client-voucher-signer>",
      "receiver": "<server-receiver>",
      "receiverAuthorizer": "<receiver-authorizer>",
      "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
      "withdrawDelay": 3600,
      "salt": "42",
      "openSlot": "341000000"
    },
    "voucher": {
      "channelId": "<channel-pda>",
      "maxClaimableAmount": "5000",
      "expiresAt": 0,
      "signature": "<base58-ed25519-signature>"
    }
  }
}
```



# Refund Client PaymentPayload

```json
{
  "x402Version": 2,
  "accepted": {
    "scheme": "batch-settlement",
    "amount": "1000",
    "payTo": "<server-receiver>",
    "maxTimeoutSeconds": 300,
    "extra": {
      "feePayer": "<facilitator-fee-payer>",
      "receiverAuthorizer": "<receiver-authorizer>",
      "withdrawDelay": 3600
    }
  },
  "payload": {
    "type": "refund",
    "channelConfig": {
      "payer": "<client-wallet>",
      "payerAuthorizer": "<client-voucher-signer>",
      "receiver": "<server-receiver>",
      "receiverAuthorizer": "<receiver-authorizer>",
      "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
      "withdrawDelay": 3600,
      "salt": "42",
      "openSlot": "341000000"
    },
    "voucher": {
      "channelId": "<channel-pda>",
      "maxClaimableAmount": "3200",
      "expiresAt": 0,
      "signature": "<base58-ed25519-zero-charge-voucher-signature>"
    }
  }
}
```



# Voucher PaymentResponse

```json
{
  "success": true,
  "transaction": "",
  "payer": "<client-wallet>",
  "amount": "",
  "extra": {
    "chargedAmount": "700",
    "channelState": {
      "channelId": "<channel-pda>",
      "balance": "100000",
      "totalClaimed": "3200",
      "withdrawRequestedAt": 0,
      "chargedCumulativeAmount": "3900"
    }
  }
}
```



# Refund Server Settle Payload

```json
{
  "type": "refund",
  "channelConfig": {
    "payer": "<client-wallet>",
    "payerAuthorizer": "<client-voucher-signer>",
    "receiver": "<server-receiver>",
    "receiverAuthorizer": "<receiver-authorizer>",
    "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
    "withdrawDelay": 3600,
    "salt": "42",
    "openSlot": "341000000"
  },
  "voucher": {
    "channelId": "<channel-pda>",
    "maxClaimableAmount": "3200",
    "expiresAt": 0,
    "signature": "<base58-ed25519-zero-charge-voucher-signature>"
  },
  "transaction": "<base64-receiver-authorizer-signed-settle-and-seal-and-distribute-transaction>"
}
```



# Facilitator VerifyResponse

```json
{
  "isValid": true,
  "payer": "<client-wallet>",
  "extra": {
    "channelId": "<channel-pda>",
    "balance": "1000000",
    "totalClaimed": "500000",
    "withdrawRequestedAt": 0
  }
}
```



# Claim Server Settle Payload

```json
{
  "type": "claim",
  "claims": [
    {
      "voucher": {
        "channelConfig": {
          "payer": "<client-wallet-1>",
          "payerAuthorizer": "<client-voucher-signer-1>",
          "receiver": "<server-receiver>",
          "receiverAuthorizer": "<receiver-authorizer>",
          "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
          "withdrawDelay": 3600,
          "salt": "42",
          "openSlot": "341000000"
        },
        "channelId": "<channel-pda-1>",
        "maxClaimableAmount": "5000",
        "expiresAt": 0
      },
      "signature": "<base58-ed25519-voucher-signature-1>"
    },
    {
      "voucher": {
        "channelConfig": {
          "payer": "<client-wallet-2>",
          "payerAuthorizer": "<client-voucher-signer-2>",
          "receiver": "<server-receiver>",
          "receiverAuthorizer": "<receiver-authorizer>",
          "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
          "withdrawDelay": 3600,
          "salt": "43",
          "openSlot": "341000050"
        },
        "channelId": "<channel-pda-2>",
        "maxClaimableAmount": "7000",
        "expiresAt": 0
      },
      "signature": "<base58-ed25519-voucher-signature-2>"
    }
  ]
}
```



# Settle Server Settle Payload

```json
{
  "type": "settle",
  "channels": [
    {
      "channelId": "<channel-pda-1>",
      "channelConfig": {
        "payer": "<client-wallet-1>",
        "payerAuthorizer": "<client-voucher-signer-1>",
        "receiver": "<server-receiver>",
        "receiverAuthorizer": "<receiver-authorizer>",
        "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
        "withdrawDelay": 3600,
        "salt": "42",
        "openSlot": "341000000"
      }
    },
    {
      "channelId": "<channel-pda-2>",
      "channelConfig": {
        "payer": "<client-wallet-2>",
        "payerAuthorizer": "<client-voucher-signer-2>",
        "receiver": "<server-receiver>",
        "receiverAuthorizer": "<receiver-authorizer>",
        "token": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
        "withdrawDelay": 3600,
        "salt": "43",
        "openSlot": "341000050"
      }
    }
  ]
}
```



# Claim SettlementResponse

```json
{
  "success": true,
  "transaction": "<base58-transaction-signature>",
  "amount": ""
}
```



# Settle SettlementResponse

```json
{
  "success": true,
  "transaction": "<base58-transaction-signature>",
  "amount": "5000"
}
```



# Deposit SettlementResponse

```json
{
  "success": true,
  "transaction": "<base58-transaction-signature>",
  "payer": "<client-wallet>",
  "amount": "100000",
  "extra": {
    "channelState": {
      "channelId": "<channel-pda>",
      "balance": "100000",
      "totalClaimed": "3200",
      "withdrawRequestedAt": 0
    }
  }
}
```



# Refund SettlementResponse

```json
{
  "success": true,
  "transaction": "<base58-transaction-signature>",
  "payer": "<client-wallet>",
  "amount": "96800",
  "extra": {
    "channelState": {
      "channelId": "<channel-pda>",
      "balance": "0",
      "totalClaimed": "3200",
      "withdrawRequestedAt": 0
    }
  }
}
```



# Facilitator SupportedResponse

```json
{
  "kinds": [
    {
      "x402Version": 2,
      "scheme": "batch-settlement",
      "extra": {
        "feePayer": "<facilitator-fee-payer>",
\      }
    }
  ],
  "extensions": [],
  "signers": {
    "solana:*": [
      "<facilitator-fee-payer>"
    ]
  }
}
```



# Corrective 402 PaymentRequired

```json
{
  "x402Version": 2,
  "error": "invalid_batch_settlement_svm_cumulative_amount_mismatch",
  "accepts": [
    {
      "scheme": "batch-settlement",
      "amount": "1000",
      "payTo": "<server-receiver>",
      "maxTimeoutSeconds": 300,
      "extra": {
        "feePayer": "<facilitator-fee-payer>",
        "receiverAuthorizer": "<receiver-authorizer>",
        "withdrawDelay": 3600,
        "channelState": {
          "channelId": "<channel-pda>",
          "balance": "100000",
          "totalClaimed": "500",
          "withdrawRequestedAt": 0,
          "chargedCumulativeAmount": "3200"
        },
        "voucherState": {
          "signedMaxClaimable": "3200",
          "expiresAt": 0,
          "signature": "<base58-ed25519-voucher-signature>"
        }
      }
    }
  ]
}
```

phdargen:
On EVM, channelId = EIP712Hash(ChannelConfig) and every field of ChannelConfig is either in the 402 response (receiver, receiverAuthorizer, token, withdrawDelay) or client-chosen and stable (payer, payerAuthorizer, salt). So a client with zero local state recomputes channelId purely from the 402 + its own params and reads channels(channelId) for balance/totalClaimed
https://github.com/x402-foundation/x402/blob/ea2cd8177d0b694945fcd3c69e962ad880e8cab0/specs/schemes/batch-settlement/scheme_batch_settlement_evm.md?plain=1#L535-L548

On SVM the channelID depends on a dynamic `open_slot`, is there any way for the client to recover its open channelIDs for a given server from onchain state alone when its offchain state is lost?

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+| Monotonic amount | Server-owned offchain watermark plus onchain `settled < maxClaimableAmount <= deposit` at redemption. |
+| Batched redemption | One `settle` per channel, packed transaction-size permitting; `distribute` pays settled deltas. |
+| Recipient binding | `distribution_hash` fixed at `open` sends funds to `payTo`; program re-checks it at `distribute`. |
+| Refund of unused deposit | `settle_and_seal` + `distribute` for cooperative close; `request_close` / `seal` / `withdraw_payer` for payer-forced close. |
+
+## 3. Payment-Channel Method
+
+SVM `batch-settlement` defines a single payment method backed by the
+payment-channels program. Because there is only one method, the wire format does
+not include `extra.profiles` or `extra.assetTransferMethod`.
```
phdargen:
```suggestion
not include `extra.assetTransferMethod`.
```

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+program as both `Channel.rent_payer` and zero-share `Channel.payee`. Rent returns
+to that address during the final `distribute` fast path or a later `reclaim`.
+Because the sponsor holds the `payee` lifecycle authority, it can seal and clean
+up an abandoned channel without a client or server signature. It MUST be able
+to rediscover the channels it sponsored as specified in section 6. Token
+payouts and client refunds are completed by `distribute` and are not delayed by
+the later `reclaim` of PDA rent.
+
+If the client invokes `request_close`, only the `payee` can cooperatively
+`settle_and_seal` during the grace period. The server therefore depends on the
```
phdargen: is it not possible to call `settle` during grace period?

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+| Claim | `claims[].signature` | string | Base58 Ed25519 voucher signature. |
+| Settle | `type` | string | `"settle"` |
+| Settle | `channels` | array | One or more channels to distribute. |
+| Settle | `channels[].channelId` | string | Channel PDA. |
+| Settle | `channels[].channelConfig` | `ChannelConfig` | Full configuration used to derive accounts, validate the channel, and reconstruct the distribution. |
+
+The server forwards a client-authored `deposit` unchanged. It MAY settle
+`voucher` locally by storing the commitment and producing the response in
+section 4.4. Before forwarding `refund`, the server MUST build the cooperative
+close transaction and add `transaction`. The facilitator MUST validate the
```
phdargen: the server shouldn't build the `settle_and_seal`, the facilitator can and should do it on its own

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+        "maxClaimableAmount": "7000",
+        "expiresAt": 0
+      },
+      "signature": "<base58-ed25519-voucher-signature-2>"
+    }
+  ]
+}
+```
+
+The facilitator MUST process every `claims[]` entry or fail the request; it
```
phdargen: Should we limit the `claims[]` array size? How many can be batched in a single tx?

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+6. Enforce replay protection and the per-request ceiling:
+   - A previously accepted `(channelId, maxClaimableAmount)` is an idempotent
+     retry. Return its cached response and do not execute the resource handler
+     again.
+   - Any other `maxClaimableAmount <= signedMaxClaimable` is stale and MUST be
+     rejected.
+   - A fresh voucher MUST have `maxClaimableAmount ==
+     chargedCumulativeAmount + PaymentRequirements.amount`.
+7. Execute the resource handler and determine `chargedAmount`, which MUST be
+   greater than zero and less than or equal to `PaymentRequirements.amount`.
```
phdargen: The program doesnt support dynamic prices, so `chargedAmount` is not to be determined but always equal to `PaymentRequirements.amount`

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `success` | boolean | yes | Whether acceptance or settlement succeeded. |
+| `errorReason` | string | no | Omitted on success. |
+| `payer` | string | no | Channel `payer`. |
+| `transaction` | string | yes | Onchain signature for deposit/top-up/refund/claim/settle operations; empty string for offchain voucher acceptance. |
+| `network` | string | yes | CAIP-2 network identifier. |
+| `amount` | string | no | Amount moved onchain; empty for voucher acceptance and `claim`. |
+| `extra.commitmentId` | string | no | MUST be non-empty for voucher acceptance, e.g. `channelId:maxClaimableAmount`. |
```
phdargen, line 536: `commitmentId` is redundant, please drop

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
```diff
@@ -0,0 +1,1116 @@
...
+  "amount": "",
+  "extra": {
+    "commitmentId": "<channel-pda>:5000",
+    "chargedAmount": "700",
+    "channelState": {
+      "channelId": "<channel-pda>",
+      "balance": "100000",
+      "totalClaimed": "3200",
+      "withdrawRequestedAt": 0,
+      "chargedCumulativeAmount": "4700"
```
phdargen:
```suggestion
    "chargedAmount": "700",
    "channelState": {
      "channelId": "<channel-pda>",
      "balance": "100000",
      "totalClaimed": "3200",
      "withdrawRequestedAt": 0,
      "chargedCumulativeAmount": "4700"
```

specs/schemes/batch-settlement/scheme_batch_settlement_svm.md
lgalabru: The network-agnostic batch-settlement spec requires a non-empty commitment identifier on success, so we need to keep this field here: https://github.com/x402-foundation/x402/blob/main/specs/schemes/batch-settlement/scheme_batch_settlement.md#commitment-identifier

CarsonRoscoe (commented):
Hey @lgalabru, thank you for the spec! I think it's nearly ready to land,

Just some final comments:

1. The corrective 402 flow is missing the client-side signature-verification instruction. 4.6 shows the `voucherState` JSON shape, but does not explain what the client does with it. The EVM sibling spec explicitely states here `"The client verifies the voucher signature before adopting chargeCumulativeAmount and retrying"`. Trivial to self-verify and confirm that the server is sharing their voucher with them, so they can trust this was the prior consented state.

2. The `withdrawDelay` has a range of 900s-2,592,000s. In the `upto` spec, which uses the same contract, it says `"MUST be an integer greater than zero"`. If I am right that these use the same contract, that means this range is not enforced on the contract side, and instead of a facilitator expectation. Would clarify that facilitator's must enforce this.

3. The voucher-expiry `settlement buffer` is unquantified. 5.3.4 states "the server MUST require either `expiresAt == 0 or expiresAt >= now + withdrawDelay + a settlement buffer"`. Nothing defines a buffer's magnitude in the spec, so a client has no way to compute an `expiresAt` guaranteed to pass without guessing conservatively. When compared with `upto`, their analogous rule ties the bound to an advertised `maxChannelLifetimeSecs` facilitator value, but this spec has no equivalent term

lgalabru:
@CarsonRoscoe thanks for the review — all three points addressed in 1c98208 (+ 2a0b716):

1. **Corrective 402 client verification** — §4.6 now instructs the client to verify `voucherState.signature` against its own `payerAuthorizer` key (reconstructing the 50-byte voucher message of §4.2 from the derived `channelId`, `signedMaxClaimable`, and `expiresAt`), and to reject the snapshot when the signature fails or `chargedCumulativeAmount > signedMaxClaimable`, before adopting the new cumulative base and retrying — mirroring the EVM sibling's rule.

2. **`withdrawDelay` range** — you're right: the payment-channels program accepts any positive `grace_period`. The spec now states explicitly that the 900–2,592,000 range is an x402 conformance bound, not a program constraint, and that verifying facilitators MUST enforce it.

3. **Settlement buffer** — quantified via a new optional `extra.settlementBufferSeconds` (default `60`). The Phase 3 bound is now `expiresAt == 0 || expiresAt >= now + withdrawDelay + settlementBufferSeconds`, and enforcers MUST NOT require a larger margin than the advertised value — so a client can always compute a passing `expiresAt` from the 402 alone, no guessing.

Also folded in offband feedback from Philippe: the scheme now resolves to the protocol-default `authorization` payment flow (`/verify` checks the voucher → work → `/settle` commits it) instead of `escrow`. The deposit `open`/`top_up` broadcasts in the post-handler `/settle`, matching the EVM binding's trust model, and `extra.paymentFlow` is omitted from the wire. The reference implementation is updated accordingly in #3164.

lgalabru:
Follow-up in a14ec63: non-expiring vouchers are now normative. Voucher expiry added a second clock the server had to beat on top of the forced-close grace period, and a nonzero `expiresAt` could make an accepted voucher unredeemable while the channel was still open — after the resource had been served. The client MUST sign `expiresAt = 0`, the server and facilitator MUST reject nonzero values (`invalid_batch_settlement_svm_voucher_expiry`), and `extra.settlementBufferSeconds` is dropped. The field stays in the 50-byte signed message (always zero) for program compatibility. Implementation updated in #3164 (13536887c).

phdargen (approved):
Thanks @lgalabru, LGTM 🚀

PR #2707, opened 2026-06-26
Hardened Hedera facilitator verify

### LINKED ISSUE #2701: @x402/hedera: verify() passes unsigned/wrong-key txs and unassociated HTS transfers (wrongPrivateKey, unsignedTx, htsUnassociated)
## Package / area

- **Package:** `@x402/hedera` (`typescript/packages/mechanisms/hedera`)
- **Scheme:** `exact` facilitator (`ExactHederaScheme`)
- **Spec:** [`specs/schemes/exact/scheme_exact_hedera.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_hedera.md) §6

---

## Summary

The Hedera exact facilitator `verify()` path does not reliably reject payments that will fail on-chain at settlement. An exploit harness against a facilitator using `createHederaPreflightTransfer()` reports three cases as **VULNERABILITY** (verify passes, settle fails on consensus):

| Exploit | Verify | Settle error | Verdict |
|---------|--------|--------------|---------|
| `wrongPrivateKey` | `isValid: true` | `INVALID_SIGNATURE` | VULNERABILITY |
| `unsignedTx` | `isValid: true` | `INVALID_SIGNATURE` | VULNERABILITY |
| `htsUnassociated` | `isValid: true` | `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT` | VULNERABILITY |

Cases that **do** work today: `requirementsMismatch`, `receiverMismatch` (blocked by transfer-semantics checks), `insufficientBalance` (blocked by preflight).

---

## Impact

A facilitator that trusts `verify()` alone may:

1. **Approve invalid payer signatures** — unsigned or wrong-key transactions pass verify; the facilitator only learns they are invalid after adding its fee-payer signature and submitting (wasted fees / node round-trips).
2. **Approve HTS transfers to unassociated receivers** — preflight returns `{ ok: true }` when `payTo` cannot receive the token; settlement fails with `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT`.

Per  [`specs/schemes/exact/scheme_exact_hedera.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_hedera.md) §6, facilitators SHOULD pre-check balance and association and MUST enforce security-critical constraints before sponsoring transactions. Missing payer-signature verification is a gap against that intent.

---

## Root cause

Two independent bugs:

### 1. No payer signature verification in `verify()`

[`ExactHederaScheme.verify()`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts#L91) and [`inspectHederaTransaction()`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/hedera/src/utils.ts#L82) only decode and validate **transfer semantics** (amount, `payTo`, fee payer, sums). Nothing checks that the inferred payer actually signed the frozen transaction body.

`createHederaPreflightTransfer()` cannot fix this — it only queries balance/association, not signatures.

### 2. Broken HTS association preflight heuristic

[`createHederaPreflightTransfer()`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/hedera/src/preflight.ts#L44) relied on consensus-node `AccountInfo.tokenRelationships`, which Hedera nodes no longer reliably return (absent from `AccountInfoJson`). The `maxAuto` / `currentAuto` slot math ran on empty or partial data and incorrectly returned `{ ok: true }` for unassociated `payTo` accounts.

---

## Reproduction

Run an exploit harness against a facilitator wired like the e2e TypeScript facilitator:

```ts
preflightTransfer: createHederaPreflightTransfer(buildHederaClient),
// no verifyPayerSignature hook
```

Example harness output (abbreviated):

```
wrongPrivateKey  → verify: { isValid: true }  → settle: INVALID_SIGNATURE  → VULNERABILITY
unsignedTx       → verify: { isValid: true }  → settle: INVALID_SIGNATURE  → VULNERABILITY
htsUnassociated  → verify: { isValid: true }  → settle: TOKEN_NOT_ASSOCIATED_TO_ACCOUNT → VULNERABILITY
insufficientBalance → verify: preflight_failed → PREFLIGHT_CAUGHT ✓
```

Facilitator setup reference: [`e2e/facilitators/typescript/index.ts`](https://github.com/x402-foundation/x402/blob/main/e2e/facilitators/typescript/index.ts)

---

## Why existing unit tests did not catch this

[`test/unit/preflight.test.ts`](https://github.com/x402-foundation/x402/blob/main/e2e/facilitators/typescript/index.ts) (prior version):

- Mocked `@hiero-ledger/sdk` with fakes whose `tokenRelationships` used string keys, unlike the real `Map<TokenId, …>` keyed by object identity — tests passed against an idealized shape.
- Covered happy paths and explicit “no slots” rejection only; no case matching on-chain `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT` after a false-positive preflight.
- Had no signature concept — `wrongPrivateKey` / `unsignedTx` require verify-level checks, not preflight.

Per [Testing](https://github.com/x402-foundation/x402/blob/main/typescript/CONTRIBUTING.md#testing): unit tests should mock external I/O; mirror-node responses must be mocked for preflight tests, and signature verification needs separate coverage in `verify()`.

---

## Expected behavior

1. **`verify()` MUST reject** transactions where the inferred payer did not sign the frozen body (`invalid_exact_hedera_payload_signature_invalid`).
2. **Preflight SHOULD reject** HTS transfers when `payTo` is not associated and has no remaining auto-association capacity, using a reliable data source (Mirror Node REST API).
3. Facilitator examples and e2e facilitator SHOULD wire both hooks by default.

---

### 1. Payer signature verification (fail-closed)

- Add optional `verifyPayerSignature` to `FacilitatorHederaSigner`.
- Add `createHederaVerifyPayerSignature(buildClient)` — `AccountInfoQuery(payer).key` + `PublicKey.verifyTransaction(tx)` (support `KeyList`/threshold keys).
- Wire into `ExactHederaScheme.verify()` before preflight; reject on failure or throw.

### 2. Mirror Node–backed preflight

- Rewrite `createHederaPreflightTransfer(config?)` to use Mirror Node REST:
  - HBAR: `GET /api/v1/accounts/{payer}`
  - HTS balance: `GET /api/v1/accounts/{payer}/tokens?token.id={asset}`
  - HTS association: same for `payTo`; if missing, check `max_automatic_token_associations` and page `/tokens` counting `automatic_association: true`
- Breaking change: factory no longer takes `buildClient`; use `{ mirrorNodeUrl? }` override for private networks.

### 3. Tests

- `preflight.test.ts`: mock `fetch` with realistic mirror-node JSON.
- New `signature.test.ts`: correct key / wrong key / unsigned tx.
- `facilitator.test.ts`: verify-level signature rejection.

### 4. Spec

- Update `scheme_exact_hedera.md` §6: signature verification MUST; mirror-node pre-check guidance.

### 5. Call sites

- [`e2e/facilitators/typescript/index.ts`](https://github.com/x402-foundation/x402/blob/main/e2e/facilitators/typescript/index.ts)
- [`examples/typescript/facilitator/advanced/all_networks.ts`](https://github.com/x402-foundation/x402/blob/main/examples/typescript/facilitator/advanced/all_networks.ts)
- [`typescript/site/app/facilitator/index.ts`](https://github.com/x402-foundation/x402/blob/main/typescript/site/app/facilitator/index.ts)

```ts
preflightTransfer: createHederaPreflightTransfer(),
verifyPayerSignature: createHederaVerifyPayerSignature(buildHederaClient),
```

---

## Acceptance criteria

- [ ] `wrongPrivateKey` and `unsignedTx` → `verify.isValid === false`, reason `invalid_exact_hedera_payload_signature_invalid`
- [ ] `htsUnassociated` → `verify.isValid === false`, reason `invalid_exact_hedera_payload_preflight_failed` with `pay_to_not_associated`
- [ ] All `@x402/hedera` unit tests pass (`pnpm --filter @x402/hedera test`)
- [ ] Spec §6 updated
- [ ] Changeset added for `@x402/hedera` (minor: security fix + breaking preflight factory signature)

---

## References

- [`typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts)
- [`typescript/packages/mechanisms/hedera/src/preflight.ts`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/hedera/src/preflight.ts)
- [`typescript/packages/mechanisms/hedera/src/signer.ts`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/hedera/src/signer.ts)
- [`specs/schemes/exact/scheme_exact_hedera.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_hedera.md)
- [TypeScript contributing guide — Testing](https://github.com/x402-foundation/x402/blob/main/typescript/CONTRIBUTING.md#testing)

---

phdargen:
## Description

Hardened the exact Hedera facilitator `verify()` path so a payment that passes verification is the same one that succeeds at settlement (security fix for unsigned / wrong-key payloads and unassociated recipients).

`verify()` now (1) confirms the inferred payer actually signed the frozen transaction body by fetching the payer's onchain account key and checking the signature — including KeyList/threshold accounts — and (2) pre-checks balance and token association against the Hedera Mirror Node REST API, which is the reliable data source (consensus-node token data is no longer dependable). Both run unconditionally and fail closed.

Breaking change to `FacilitatorHederaSigner`: `verifyPayerSignature` is a new required capability and `preflightTransfer` is now required (was optional). Use the new `createHederaVerifyPayerSignature(buildClient)` default and `createHederaPreflightTransfer(config?)`, whose signature changed from `(buildClient)` to an optional `{ mirrorNodeUrl? }`.

Closes https://github.com/x402-foundation/x402/issues/2701

## Checklist

### DIFF
--- a/e2e/facilitators/typescript/index.ts
+++ b/e2e/facilitators/typescript/index.ts
@@ -232,7 +232,7 @@ if (process.env.HEDERA_ACCOUNT_ID && process.env.HEDERA_PRIVATE_KEY) {
       buildHederaClient,
       hederaKey,
     ),
-    verifyPayerSignature: createHederaVerifyPayerSignature(buildHederaClient),
+    verifyPayerSignature: createHederaVerifyPayerSignature(),
     preflightTransfer: createHederaPreflightTransfer(),
   });
   console.info(`Hedera Facilitator account: ${hederaAccountId}`);
--- a/e2e/facilitators/typescript/index.ts
+++ b/e2e/facilitators/typescript/index.ts
@@ -49,6 +49,7 @@ import {
   createHederaClient,
   createHederaPreflightTransfer,
   createHederaSignAndSubmitTransaction,
+  createHederaVerifyPayerSignature,
   toFacilitatorHederaSigner,
 } from "@x402/hedera";
 import { ExactHederaScheme } from "@x402/hedera/exact/facilitator";
@@ -231,7 +232,8 @@ if (process.env.HEDERA_ACCOUNT_ID && process.env.HEDERA_PRIVATE_KEY) {
       buildHederaClient,
       hederaKey,
     ),
-    preflightTransfer: createHederaPreflightTransfer(buildHederaClient),
+    verifyPayerSignature: createHederaVerifyPayerSignature(buildHederaClient),
+    preflightTransfer: createHederaPreflightTransfer(),
   });
   console.info(`Hedera Facilitator account: ${hederaAccountId}`);
 }
--- a/e2e/pnpm-lock.yaml
+++ b/e2e/pnpm-lock.yaml
@@ -123,7 +123,7 @@ importers:
         version: 1.2.6
       '@signinwithethereum/siwe':
         specifier: ^4.1.0
-        version: 4.1.0(viem@2.48.11(bufferutil@4.0.9)(typescript@5.8.3)(utf-8-validate@5.0.10)(zod@3.25.71))
+        version: 4.1.0(ethers@6.16.0(bufferutil@4.0.9)(utf-8-validate@5.0.10))(viem@2.48.11(bufferutil@4.0.9)(typescript@5.8.3)(utf-8-validate@5.0.10)(zod@3.25.71))
       '@x402/core':
         specifier: workspace:~
         version: link:../core
@@ -1421,9 +1421,12 @@ importers:
 
   ../typescript/packages/mechanisms/hedera:
     dependencies:
+      '@hiero-ledger/proto':
+        specifier: 2.31.0
+        version: 2.31.0(ansi-regex@6.2.2)(ansi-styles@6.2.3)(debug@4.4.1)(protobufjs@8.0.1)(strip-ansi@7.1.2)
       '@hiero-ledger/sdk':
-        specifier: 2.80.0
-        version: 2.80.0(bn.js@5.2.1)(react-native@0.85.2(@babel/core@7.28.4)(@types/react@19.2.2)(bufferutil@4.0.9)(react@19.2.6)(utf-8-validate@5.0.10))
+        specifier: 2.85.0
+        version: 2.85.0(bn.js@5.2.3)(bufferutil@4.0.9)(react-native@0.85.2(@babel/core@7.28.4)(@types/react@19.2.2)(bufferutil@4.0.9)(react@19.2.6)(utf-8-validate@5.0.10))(utf-8-validate@5.0.10)
       '@x402/core':
         specifier: workspace:~
         version: link:../../core
@@ -2803,6 +2806,9 @@ importers:
 
 packages:
 
+  '@adraffy/ens-normalize@1.10.1':
+    resolution: {integrity: sha512-96Z2IP3mYmF1Xg2cDm8f1gWGf/HUVedQ3FMifV4kG/PQ4yEP51xDtRAEfhVNt5f/uzpNkZHwWQuUcu6D6K+Ekw==}
+
   '@adraffy/ens-normalize@1.11.0':
     resolution: {integrity: sha512-/3DDPKHqqIqxUULp8yP4zODUY1i+2xvVWsv8A79xGWdCAG+8sb0hRh0Rk2QyOJUnnbyPUAZYcpBuRe3nS2OIUg==}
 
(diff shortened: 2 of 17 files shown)


phdargen:
cc @piotrswierzy

piotrswierzy:
@phdargen 
Thank you, will review it!

piotrswierzy:
## Review — Hardened Hedera facilitator verify

Nice fix. 

### 1. (Medium) Signature check uses a *paid* query while preflight is free — and runs first

`createHederaVerifyPayerSignature` (`signer.ts` ~L300) fetches the payer key via `AccountInfoQuery`, a **paid consensus-node query** billed to the operator, while `preflightTransfer` uses the **free** Mirror Node REST API. The `scheme.ts` comment says signature runs before preflight "to avoid wasted Mirror Node calls on bad signatures" — but Mirror Node calls are free and `AccountInfoQuery` is the expensive one, so the ordering pays to save nothing, and an attacker spraying unsigned/wrong-key payloads forces a paid query on every attempt.

The Mirror Node `/accounts/{id}` response already includes the account `key`. Sourcing the key there makes the **entire** verify path free, single-sourced, and free of any operator-funding requirement. The only trade-off is Mirror Node lag on a just-rotated key — but preflight already trusts the Mirror Node for balances, which churn far more than keys.

Per the Hedera docs, the Mirror Node REST API (Get Account by Alias, ID, or EVM Address) is the recommended source for account info:
https://docs.hedera.com/native/accounts/get-info

> - **Cost-effective and scalable:** Mirror node providers offer paid plans with a large number of queries included. The Hedera-hosted mirror node offers free queries with specific throttles for testing. While some SDK queries are currently free, these are subject to change in the future.
> - **Performance:** Mirror nodes don't burden consensus nodes, allowing them to focus on processing transactions...
> - **Historical data:** Mirror nodes store complete transaction history and balance snapshots.

**Suggestion:** fetch the payer key from the Mirror Node (`GET /api/v1/accounts/{payer}` → `key`) and drop `AccountInfoQuery`; or, if the consensus query is deliberate for zero-lag key freshness, reorder so the free preflight gates the paid query and fix the comment.

### 2. (Medium) Breaking change is under-documented in the changeset

The `minor` bump is consistent with repo convention (cf. the EVM ERC-6492 fix, labeled `[Breaking for facilitator implementers …]` under `### Minor Changes`) — so `minor` is fine. What's missing is matching that precedent's discoverability + migration guidance. This release is breaking three ways:
- `FacilitatorHederaSigner` now **requires** `verifyPayerSignature` and `preflightTransfer` (was optional) → compile breakage for direct implementers, and a runtime **"reject 100% of payments"** on upgrade if `verifyPayerSignature` isn't wired (fail-closed).
- `createHederaPreflightTransfer(buildClient)` → `(config?)` is a **silent** change: an old caller passing the client factory puts a function where a config object is expected, so `config.mirrorNodeUrl` is `undefined` and it silently falls back to the public Mirror Node — no compile or runtime error.

**Suggestion:** keep `minor`, but prefix the changeset with `**[Breaking for facilitator implementers]**` and add explicit migration steps for the two now-required capabilities and the `createHederaPreflightTransfer` signature change (calling out the silent fallback).

### 3. (Low) "passes verify ⇒ succeeds at settle" is overstated

The changeset claims a verified payment "is the same one that succeeds at settlement," but preflight only checks payer balance + payTo association. States that still pass `verify()` and fail at `settle()`: token paused (`TOKEN_IS_PAUSED`), relationship frozen (`ACCOUNT_FROZEN_FOR_TOKEN`), KYC not granted, custom-fee tokens, and **transaction expiry** (`TRANSACTION_EXPIRED` — a frozen tx has a ~120s valid window; verify does no expiry check). All reasonable to leave out of scope — just soften the wording to "closes the unsigned/wrong-key and unassociated-recipient gaps." (Note this PR also *adds* the signature MUST to spec §6 — it's tightening the spec, not just implementing it.)

### 4. (Low) `inferPayer` checks only the first debited account — spec §4 allows multiple senders

`scheme.ts` L318 `inferPayer` returns the first negative entry, and only that account is passed to `verifyPayerSignature` and `preflightTransfer`. But spec §4 says "transfer value from the client's **account(s)**" and §5 only restricts positive receivers — multi-sender is spec-legal. A two-sender payload gets only the first sender's signature/balance verified; a payload missing the second sender's signature passes `verify()` and burns the facilitator fee at `settle()` (fails closed at consensus, so not fund-loss, but it reintroduces the verify⇒settle asymmetry this PR removes).

**Suggestion:** rejecting multi-sender would violate §4, so verify **all** debited accounts (signature + preflight per debited account). If multi-sender is intended to be unsupported, narrow §4 explicitly rather than assuming it in code.

### 5. (Low, pre-existing) §6 replay-protection MUST is still unimplemented

Worth flagging since this PR rewrites §6: the spec says the transaction **MUST NOT** have been previously submitted/observed, but neither old nor new `verify()` performs any replay/idempotency check. On-chain `DUPLICATE_TRANSACTION` is the backstop (no double-spend), but `verify()` can still return `isValid: true` for a payload that fails at settle as a duplicate. Out of scope to fix here — either file a follow-up for a best-effort idempotency cache (tx-id seen-set with TTL ≥ valid window), or downgrade the §6 wording to SHOULD.

### 6. (Nits)
- `keySignsTransaction` (`signer.ts` L271): `key.threshold ?? keys.length` keeps `threshold === 0` (→ always true). Invalid on-chain, but a `> 0` guard removes the foot-gun.
- `preflight.ts` URL interpolation of `payer`/`payTo`/`asset` without `encodeURIComponent` — safe today (regex-validated), cheap to harden defensively.
- The transaction is `Transaction.fromBytes`-decoded 3× per payment (inspect, verify-sig, settle), and `settle()` re-runs `verify()`. Harmless; only worth deduping if profiled.

phdargen:
Thanks for the feedback @piotrswierzy! Addressed your comments in the last commit, also bumped the @hiero-ledger/sdk to latest

piotrswierzy:
Thanks @phdargen — re-reviewed 25ce3a4. This resolves all the code-level findings/

  One small note on the `ProtobufEncoded` key path (not blocking): reconstructing the KeyList via the internal
  `Key._fromProtobufKey` is the right call IMO — there's no public API to rebuild a threshold Key from protobuf, and
  reusing the SDK's tested path is safer than hand-rolling the KeyList recursion. It's fail-closed and CI-guarded by the
  round-trip test. Two cheap hardenings worth considering:

  1. Guard the internal so a future SDK rename surfaces a diagnosable message rather than a generic signature_invalid:
```  
  const fromProto = (Key as unknown as { _fromProtobufKey?: (k: proto.IKey) => Key })._fromProtobufKey;
  if (typeof fromProto !== "function") {
    return { ok: false, reason: "signature_unverifiable", message: "SDK _fromProtobufKey unavailable" };
  }
```
  2. A one-line comment at the `@hiero-ledger/proto` pin noting it must stay in lockstep with the SDK and that
  `_fromProtobufKey` should be re-checked on any SDK bump.

  LGTM to merge🚀

phdargen:
> Thanks @phdargen — re-reviewed [25ce3a4](https://github.com/x402-foundation/x402/commit/25ce3a42bcea43091a7a24abe70893742ed48137). This resolves all the code-level findings/
> 
> One small note on the `ProtobufEncoded` key path (not blocking): reconstructing the KeyList via the internal `Key._fromProtobufKey` is the right call IMO — there's no public API to rebuild a threshold Key from protobuf, and reusing the SDK's tested path is safer than hand-rolling the KeyList recursion. It's fail-closed and CI-guarded by the round-trip test. Two cheap hardenings worth considering:
> 
> 1. Guard the internal so a future SDK rename surfaces a diagnosable message rather than a generic signature_invalid:
> 
> ```
>   const fromProto = (Key as unknown as { _fromProtobufKey?: (k: proto.IKey) => Key })._fromProtobufKey;
>   if (typeof fromProto !== "function") {
>     return { ok: false, reason: "signature_unverifiable", message: "SDK _fromProtobufKey unavailable" };
>   }
> ```
> 
> 2. A one-line comment at the `@hiero-ledger/proto` pin noting it must stay in lockstep with the SDK and that
>    `_fromProtobufKey` should be re-checked on any SDK bump.
> 
> LGTM to merge🚀

Thanks for reviewing @piotrswierzy, implemented your suggestions in the last commit. 
Should be good to go

CarsonRoscoe (approved):

piotrswierzy (approved):

Issue #2699, opened 2026-06-25
Proposal: a "Signed Allowance Draw" asset-transfer method for `exact`-svm (hot-wallet-free agent payments on Solana)


alexanxin:
## The problem

`exact`-svm settles with a `TransferChecked` from the payer's own token account. That's fine for a person paying once. It's a poor fit for an autonomous agent, which then has to hold a hot wallet for every task. If the agent key leaks, whatever sits in that wallet is gone, and every top-up wants a fresh approval.

EVM already has an answer to "let something spend up to a cap without handing it the keys": Permit2, surfaced through the `upto` scheme. Solana has no Permit2. What it does have is an official allowance primitive, the Subscriptions & Allowances program (`De1egAFMkMWZSN5rYXRj9CAdheBamobVNubTsi9avR44`), shipped as the `@solana/subscriptions` package. We want to wire it into `exact`-svm.

## What we're proposing

A second asset-transfer method for `exact`-svm, sitting next to `TransferChecked`. Call it Signed Allowance Draw.

The user signs an on-chain allowance once. That's a `FixedDelegation` letting an agent signing key pull up to a cap of one SPL token from the user's account until an expiry. From then on, each `exact` payment draws the exact `PaymentRequirements.amount` from that allowance via the program's `transferFixed`, straight to `ATA(payTo)`. The facilitator co-signs gas as `extra.feePayer` and submits, the same way it does today.

The point of it: the money never lands in a hot wallet. The agent key can only ever move up to the signed cap, only to the address it signs, only until the allowance expires. The principal stays in the user's account the whole time. How the user signs that allowance is their wallet's business, a hardware tap, a software prompt, whatever it is. The scheme only cares that a signature happened.

## It's still `exact`

None of this bends the scheme's guarantees:

- The draw equals `PaymentRequirements.amount` exactly. The cap is a separate cumulative ceiling the program enforces across draws.
- The recipient is baked into the `transferFixed` instruction the agent key signs, so the facilitator can't redirect it.
- The facilitator is the `feePayer` and nothing else. Never the source of funds, never the draw authority.
- Each payment is a single-use, partially-signed v0 transaction on a recent blockhash.

[...]

PR #2717, opened 2026-06-26
exact-svm: allowance-backed payments via the Subscriptions program

alexanxin:
Enables hot-wallet-free agent payments on `exact`-svm using the officially-shipped Subscriptions & Allowances program (`De1egAFMkMWZSN5rYXRj9CAdheBamobVNubTsi9avR44`, `@solana/subscriptions`). Discussion in #2699.

A user signs an on-chain allowance once. Each `exact` payment draws `PaymentRequirements.amount` via `transferFixed`, which CPIs the token transfer to `ATA(payTo)`. The principal stays in the user's account, so an agent's spender key only ever moves funds within the signed cap and expiry.

Since `exact`-svm is outcome-based and already accepts program-mediated (CPI) transfers (§1.1), this is not a new asset-transfer method. It is two Sponsor Acceptance Policy changes:

- §2.2.2: allowlist the Subscriptions program, so allowance draws reach verification like the existing smart-wallet programs.
- §2.4: one allowance-specific MUST (the sponsor confirms the allowance resolves and covers the draw before signing, fail-closed, so `verify()` predicts `settle()`) plus settlement-receipt binding guidance (bind the executed net-balance delta, delegation as provenance).

Spec-only. A `python/x402` mechanism and a devnet-proven reference follow. The verify negatives (`allowance_missing` / `allowance_expired` / `allowance_under_cap`) are a fourth case for the #2701 rail-agnostic `verify()` matrix; receipt binding follows #2666. Forward-compatible with a per-payment `upto` axis if the program gains a recipient-bound variable draw (solana-foundation/subscriptions#181).

cc @DrVelvetFog for the svm conformance cross-check.

Drafted with AI assistance and reviewed before submission.


DrVelvetFog:
Cross-checked the spec against the svm conformance set. Both the verify-side and settle-side points we converged on are in the text and hold:

- The allowance MUST is fail-closed, at the same status as the §1 outcome check, and you've stated the `verify()`-predicts-`settle()` invariant explicitly with the three negatives named. That's exactly the shape that slots into the rail-agnostic matrix on #2701.
- Receipt binding is to the executed `transferFixed`'s net-balance delta to `ATA(payTo)` with the delegation as provenance, not the delegation itself. Same finalized-settlement rule the exact-Sui and svm-`upto` vectors already use, and it stays independently recomputable from the recipient ATA's post-balances.

Two refinements so the conformance artifact is complete:

1. The named negatives — `allowance_missing` / `allowance_expired` / `allowance_under_cap` — cover existence, expiry, and cap. But the MUST also requires `delegator`, `spender`, and `mint` to match the draw, and there's no named negative for the case where the allowance resolves, is active, and is funded, yet binds the wrong delegator / spender / mint. That mismatch wants its own fail-closed code (`allowance_mismatch` or similar) alongside the other three — otherwise an implementation that only checks the named set passes a real-but-wrong-authorization allowance. It's the same precondition class as recipient-not-provisioned on the #2701 matrix: a resolvable artifact that doesn't authorize *this* draw.

2. Binding is a SHOULD here. Worth confirming that matches the normative strength the exact-Sui and svm-`upto` receipt-binding vectors carry, so the matrix stays uniform on the settle side — SHOULD in one mechanism and MUST in another is the kind of seam a conformance suite trips on.

Forward-compat looks right: modeling the draw as authorized-envelope-then-executed-draw means `transferFixed` now and `transferUpTo` (subscriptions#181) later bind the same way — one vector, no rewrite — provided #181 keeps the executed delta recomputable from on-chain post-balances, which is the thread we left open there.

alexanxin:
Both folded in.

1. Added `allowance_mismatch` as the fourth named negative, a resolvable, active, funded allowance whose delegator, spender, or mint doesn't match the draw. Same precondition class as recipient-not-provisioned: resolvable but doesn't authorize this draw, so an implementation can't pass a real-but-wrong-authorization allowance by checking only existence, expiry, and cap.

2. Elevated binding from SHOULD to MUST and aligned the wording with #2666: bind the facilitator's verified net-balance result to `ATA(payTo)`, recomputable from the recipient ATA's post-balances, keyed to the executed settlement id in `PAYMENT-RESPONSE`, delegation as provenance. That matches the normative strength the other `exact` receipt-binding vectors carry, so the settle side stays uniform.

Pushed. And #181 stays the open thread for keeping the executed delta recomputable from on-chain post-balances on the `transferUpTo` side.

DrVelvetFog:
Both look right in the text now.

`allowance_mismatch` as the fourth named negative closes the gap — a resolvable, active, funded allowance that binds the wrong delegator/spender/mint now fails closed, so an implementation can't pass a real-but-wrong-authorization allowance by checking only existence/expiry/cap. That's the four-case precondition set the #2701 rail-agnostic matrix wants: missing / expired / under-cap / mismatch, same class as recipient-not-provisioned.

Elevating binding to MUST aligned with #2666 makes the settle side uniform — exact-Sui, svm-`upto`, and allowance-backed exact-svm now carry the same normative strength (bind the verified net-balance result to `ATA(payTo)`, recomputable from post-balances, delegation as provenance, keyed to the executed settlement id). No SHOULD/MUST seam for a conformance suite to trip on.

That's the spec side complete from where I'm standing. The two open threads stay as we left them: the `python/x402` mechanism PR, and keeping the executed delta recomputable from on-chain post-balances on the `transferUpTo` side (subscriptions#181) so `transferFixed`-now and `transferUpTo`-later bind through one vector with no rewrite. Glad to cross-check the mechanism PR against the svm conformance set when it's up.

alexanxin:
Appreciate the thorough cross-check, Tony. The spec is better for it. The `python/x402` mechanism is the next piece, and I'll tag you for the conformance pass when it's up. We'll also keep the executed delta recomputable from on-chain post-balances on the `transferUpTo` side (subscriptions#181) so `transferFixed`-now and `transferUpTo`-later bind through one vector.

PR #2720, opened 2026-06-27
fix(py): return 500 (not empty 402) on unexpected settlement error in FastAPI middleware

### LINKED ISSUE #2603: FastAPI payment middleware silently swallows unexpected exceptions as an empty-body 402 (no logging)

kakedashi3:
## What

The FastAPI middleware's settlement `except Exception` returned an empty-body `402`, so an internal failure (RPC outage, bug) was indistinguishable from a genuine "payment required" — an autonomous client may retry and **pay again** into a broken endpoint. It now logs the exception and returns `500`. The genuine settlement-failure path (`if not settle_result.success`) still returns `402` unchanged.

Closes #2603.

## Changes
- `fastapi.py`: module logger + `logger.exception(...)` and `500` in the post-verification settlement catch-all.
- Test: `test_settlement_unexpected_exception_returns_500`.
- Towncrier changelog fragment.

## Scope
- Scoped to FastAPI per the issue. The same empty-402 catch-all also exists in `flask.py` (WSGI) — happy to include the parity fix here or as a follow-up (asked on the issue thread).
- Settle-path idempotency (raised on the issue) is out of scope for this PR.

## Verification
`uv run pytest tests/unit/http/middleware/test_fastapi.py` → 32 passed; `ruff check` clean.

---
AI-assisted (Claude Code), reviewed before submitting.

### DIFF
--- a/python/x402/changelog.d/2603.bugfix.md
+++ b/python/x402/changelog.d/2603.bugfix.md
@@ -1,0 +1,1 @@
+FastAPI middleware now returns `500` (and logs the exception) when the settlement path raises an unexpected error, instead of a silent empty-body `402` that could prompt a paying client to retry.
--- a/python/x402/http/middleware/fastapi.py
+++ b/python/x402/http/middleware/fastapi.py
@@ -6,6 +6,7 @@ Provides payment-gated route protection for FastAPI applications.
 from __future__ import annotations
 
 import asyncio
+import logging
 from collections.abc import Awaitable, Callable
 from typing import TYPE_CHECKING, Any
 
@@ -48,6 +49,8 @@ from ._bazaar_utils import (
 from ._bazaar_utils import (
     validate_bazaar_extensions as _validate_bazaar_extensions,
 )
+
+logger = logging.getLogger(__name__)
 
 # ============================================================================
 # FastAPI Adapter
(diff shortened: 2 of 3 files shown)


kakedashi3:
Closing as a duplicate of #2622 — @moped2110 (who filed the issue) already has an open fix for this, which I missed before opening this PR. Apologies for the noise. #2622 looks complete and just pending review; if a changelog fragment would help it land, happy to hand mine over.

PR #2721, opened 2026-06-28
fix(python/flask): return 500 (not empty 402) on unexpected settlement error

kakedashi3:
## What

The Flask WSGI middleware's settlement catch-all returns an empty-body `402 Payment Required` for *any* unexpected exception, with no logging:

```python
except Exception:
    # Settlement error - return empty body with 402
    start_response("402 Payment Required", [("Content-Type", "application/json")])
    return [json.dumps({}).encode("utf-8")]
```

An internal failure (RPC outage, DB error, a bug) is therefore indistinguishable from a genuine "payment required" — a paying client/agent is told to retry and pay again into a broken endpoint, and operators get no signal at all.

This is the same bug as #2603 (FastAPI), which is being fixed for FastAPI in #2622. This PR applies the identical fix to the **Flask WSGI** middleware, which #2622 does not touch.

## Changes
- `flask.py`: add a module logger; the settlement catch-all now `logger.exception(...)` and returns **500** (`{"error": "internal server error during settlement"}`). The genuine settlement-failure path (`settle_result.success is False`) still returns `402` unchanged.
- Test: `test_settlement_unexpected_exception_returns_500`.
- Towncrier changelog fragment.

## Verification
`uv run pytest tests/unit/http/middleware/test_flask.py` → 40 passed; `ruff check` clean.

---
AI-assisted (Claude Code), reviewed before submitting.


kakedashi3:
Updating this to match @moped2110's #2622 instead of the 500 approach. On reflection the `402 + PAYMENT-RESPONSE` (success=False, error_reason) + `logger.exception` shape is the right call for parity: the original hazard was the *empty* 402 (indistinguishable from "pay me"), and a structured settlement-failure body removes that while staying consistent with the TS and Go cores — which route unexpected settle errors through the 402 settlement-failure path — and with the FastAPI fix in #2622. A flask-only 500 would re-introduce exactly the cross-language inconsistency this parity work is trying to remove.

So I'll re-push this as the flask mirror of #2622: catch-all → `402 + PAYMENT-RESPONSE` via the settlement-headers helper (success=False + error_reason), plus `logger.exception` so the failure isn't silent. I'll keep `error_reason` generic and leave the raw exception text in the log only, matching the caveat raised on #2622.

@moped2110 — happy to do the flask port in this shape to mirror yours, or if you'd rather fold flask into #2622 I'll close this. Whichever keeps it one consistent pass.

moped2110:
Go for it — keep #2721 open and do the Flask mirror in the #2622 shape. Two focused PRs (FastAPI here, Flask there), same 402 + PAYMENT-RESPONSE (success=false, error_reason) + logger.exception, reads cleaner for review than folding both into one and keeps each framework's diff independently reviewable. Parity's preserved either way, and it keeps #2622 from expanding mid-review.
Agreed on keeping error_reason generic with the raw exception in the log only — the safer call. I'll align #2622 to the same (generic reason, detail in the log) so the two mirror exactly; that also closes the caveat I raised there. Happy to cross-review the Flask PR whenever it's up.

kakedashi3:
Done — pushed the Flask mirror (`7567de7f`): the settle catch-all now logs the exception and returns `402 + PAYMENT-RESPONSE` (success=false) via `_create_settlement_headers`, same shape as #2622. Client-facing `error_reason` is generic (`unexpected_settlement_error`); the raw exception detail stays in the log only. Test asserts 402 + PAYMENT-RESPONSE header + that the raw detail is logged but not leaked. Ready for cross-review whenever you have a moment, @moped2110.

moped2110:
Cross-reviewed 7567de7f — structure LGTM and mirrors #2622 exactly: catch-all logs the exception, returns 402 + PAYMENT-RESPONSE (success=false) via _create_settlement_headers, generic client-facing reason, raw detail log-only. Nice touch asserting the raw detail is logged-but-not-leaked.
One thing on the error_reason value: let's use the canonical unexpected_settle_error instead of unexpected_settlement_error. It's the spec string from the ErrorReasons enum in types/verify/x402Specs.ts, enforced via z.enum(ErrorReasons) on SettleResponseSchema.errorReason — unexpected_settlement_error isn't in that enum, so a client validating the response against the spec schema would reject it. #2622 already uses unexpected_settle_error, so this also keeps both PRs in parity. Everything else is good to go.

kakedashi3:
Good catch — fixed in `6347aa00`. Switched `error_reason` to the canonical `unexpected_settle_error` from the `ErrorReasons` enum (was `unexpected_settlement_error`, which isn't in `z.enum(ErrorReasons)` so a spec-validating client would've rejected the `PAYMENT-RESPONSE`). Now matches #2622 and the enum exactly. Tests still green (40 passed). Thanks for the cross-review, @moped2110.

phdargen:
Thanks for the contribution @kakedashi3, looks good! Could you please fix the format/lint issue, then we can get this in

kakedashi3:
Fixed, thanks @phdargen.

Two things were wrong:

- `ruff format` — an `assert` in `test_flask.py` had been reverted to the pre-`ruff format` style. Reformatted; `ruff check` and `ruff format --check` are both clean now.
- `uv.lock` — the branch carried a large re-resolved lock that also broke `test-python` (the SVM extra failed to resolve, so 8 test modules errored on `No module named 'solana.rpc.api'`). This PR adds no dependencies, so the lock had no business changing; it's now untouched and `test-python` is green.

The diff is down to the three files it should touch: `flask.py`, its test, and the changelog fragment.

phdargen (approved):

Issue #2679, opened 2026-06-22
CDP facilitator verify returns 400 'invalid paymentPayload' for one resource, settles for another (same account/keys)


GTCC777:
## Summary
The CDP facilitator's `verify` returns **400 `invalid paymentPayload`** for one x402 resource, while an **identical** payment to a different resource on the **same CDP account and same API keys** settles successfully.

## Setup
- x402-gated API endpoints on Vercel
- `@coinbase/x402` v2.1.0, `@x402/core` v2.14.0 (server); `@x402/fetch` payer client
- USDC settlement on Base (`eip155:8453`) via the CDP facilitator

## Works vs. fails (everything identical except the resource URL)
| | Resource | Result |
|---|---|---|
| ✅ settles (200) | `https://cryptopulse-xi-five.vercel.app/api/research-brief` | OK |
| ❌ 400 | `https://signalpulse-peach.vercel.app/api/scan/crypto` | `invalid paymentPayload` |

## Exact error (failing resource)
```
Facilitator verify failed (400): {
  "errorMessage": "'paymentPayload' is invalid: must match one of [x402V2Pay...]",
  "errorLink": "https://docs.cdp.coinbase.com/api-reference/v2/errors#invalid-request"
}
```
Recent correlationIds: `a0fe20061efe6545-IAD`, `a0fe0359bc261655-IAD`, `a0fdee8ffb02a712-IAD`

## Both resources use identical PaymentRequirements
```json
{
  "scheme": "exact",
  "network": "eip155:8453",
  "asset": "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
  "payTo": "0x50ab2018c06c6E4eAA9BA52057Eb55eD284912fc",
  "amount": "500000",
  "maxTimeoutSeconds": 300,
  "extra": { "name": "USD Coin", "version": "2" }
}
```
Same CDP account, same `CDP_API_KEY_ID` / `CDP_API_KEY_SECRET`, same package versions, same payer client. The advertised `accepts[0]` is byte-identical between the two resources.

## Question
Why does `verify` reject the payload for `signalpulse-peach.vercel.app/api/scan/crypto` when an identical request to `cryptopulse-xi-five.vercel.app/api/research-brief` succeeds? Is there any account-side, resource-side, or rate/quota state tied to a specific resource/origin that would cause this? The correlationIds above should show the exact server-side rejection reason.

GTCC777:
[...]

PR #2722, opened 2026-06-28
fix: warn at startup when ResourceInfo description/tags exceed CDP facilitator limits

chopmob-cloud:
## Problem

The CDP facilitator returns a generic `400 'paymentPayload' is invalid: must match one of [x402V2Pay...]` when `ResourceInfo.description` is too long or `ResourceInfo.tags` has more than 8 items. The error does not name the offending field, so developers face a long debugging cycle (report: #2679).

Root cause (diagnosed by @GTCC777 in #2679): the CDP facilitator enforces an undocumented payload-size limit on the `ResourceInfo` block echoed in the client `PaymentPayload`. Exceeding it produces the same top-level pattern-mismatch error as any other payload schema failure.

Empirical limits from #2679:
- **tags**: 14 items fail; 8 items pass
- **description**: ~450 characters with 14 tags fails; concise description passes

## Changes

**`typescript/packages/core/src/http/x402HTTPResourceServer.ts`**

Add JSDoc on `RouteConfig.description` and `RouteConfig.tags` documenting the empirical CDP limits and linking to #2679. Surfaces in IDE tooling at the point where the developer sets the field.

**`typescript/packages/extensions/src/bazaar/startupValidation.ts`**

Add `validateResourceInfoLimits(routes)` that warns at middleware startup if a route exceeds the empirical limits. Fires at the same startup phase as the existing `validateBazaarRouteExtensions` check, before any payment request is attempted.

Example warning:
```
x402: Route "GET /api/scan/crypto" declares 14 tags. The CDP facilitator silently rejects
payloads with more than 8 tags (returns generic 400 "invalid paymentPayload" -- not a
tag-count error). Trim to 8 or fewer. See: https://github.com/x402-foundation/x402/issues/2679
```

## Notes

- Constants (`CDP_MAX_TAGS = 8`, `CDP_MAX_DESCRIPTION_LENGTH = 200`) are empirical; can be updated if Coinbase publishes the enforced limits.
- No breaking changes: warn-only, does not throw.
- Python middleware equivalent can follow in a separate PR if maintainers want parity.

### DIFF
--- a/typescript/packages/core/src/http/x402HTTPResourceServer.ts
+++ b/typescript/packages/core/src/http/x402HTTPResourceServer.ts
@@ -144,9 +144,24 @@ export interface RouteConfig {
 
   // HTTP-specific metadata
   resource?: string;
+  /**
+   * Human-readable description of this resource.
+   *
+   * Keep concise: the CDP facilitator enforces an undocumented payload-size limit and returns a
+   * generic 400 "invalid paymentPayload" (not a field-specific error) when the combined
+   * ResourceInfo payload is too large. Descriptions over ~200 characters have been observed to
+   * trigger this. See: https://github.com/x402-foundation/x402/issues/2679
+   */
   description?: string;
   mimeType?: string;
   serviceName?: string;
+  /**
+   * Searchable tags for this resource.
+   *
+   * The CDP facilitator silently rejects payloads with more than 8 tags, returning a generic
+   * 400 "invalid paymentPayload" error rather than a tag-count-specific message. Keep to 8 or
+   * fewer. See: https://github.com/x402-foundation/x402/issues/2679
+   */
   tags?: string[];
   iconUrl?: string;
   customPaywallHtml?: string;
--- a/typescript/packages/extensions/src/bazaar/startupValidation.ts
+++ b/typescript/packages/extensions/src/bazaar/startupValidation.ts
@@ -13,8 +13,14 @@ export { checkIfBazaarNeeded } from "@x402/core/server";
 
 /**
  * Empirical limits observed for the CDP facilitator (https://github.com/x402-foundation/x402/issues/2679).
- * The CDP facilitator returns a generic 400 "invalid paymentPayload" when ResourceInfo
- * fields exceed these values, with no field-specific error message.
+ *
+ * The facilitator behaves like a combined ResourceInfo payload-size budget rather than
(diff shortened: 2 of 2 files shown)


GTCC777:
Author of #2679 here — thanks for upstreaming this, it'll save people a real debugging cycle.

One empirical refinement from my original repro that's worth baking into the warning copy: the limit behaves like a **combined `ResourceInfo` payload-size budget**, not two independent caps. In my case a **~450-char description settled fine on its own**; it only tripped the generic `'paymentPayload' is invalid` when paired with **14 tags**. Trimming tags to 8 fixed it *without* shortening the description. So `CDP_MAX_DESCRIPTION_LENGTH = 200` is a safe conservative default, but teams running ≤8 tags can usually carry longer descriptions — the tag count is the sharper lever.

Suggestion: the warning could frame exceeding *either* threshold as a heuristic for "you're near the budget," and note that the most reliable single fix is `tags ≤ 8` first. The warn-only, non-throwing approach is exactly right given the limits are still undocumented. 👍

chopmob-cloud:
Thanks @GTCC777, that is a sharp distinction and I have folded it in.

The doc comment and both warnings now frame the thresholds as a single combined `ResourceInfo` payload budget rather than two independent caps, note that a long description can settle on its own, and surface `tags <= 8` as the most reliable single fix. The 200-char / 8-tag values stay as conservative heuristics for being near the budget, and the approach remains warn-only (non-throwing) given the limits are still undocumented.

Commits are signed and verified now as well. Appreciate the precise repro.

GTCC777:
Looks great — the combined-budget framing reads exactly right, and keeping the 200-char/8-tag values as conservative "near the budget" heuristics is the pragmatic call while the limits stay undocumented. Thanks for folding it in so cleanly. 👍

phdargen:
Thanks for the contribution, however this is a facilitator-specific implementation detail and should not be handled at the x402 SDK level. The CDP facilitator is a Coinbase product, other facilitators might have their own policies. 

The CDP team updated their docs and will improve error surfacing, also see
https://github.com/x402-foundation/x402/issues/2789#issuecomment-4915799162

chopmob-cloud:
Completely fair, and I agree it belongs at the facilitator rather than in the multi-facilitator SDK. The intent was only to spare developers the unnamed 400 debugging cycle from #2679, so CDP updating the docs and improving error surfacing is the right fix and a better home for it. Thanks for the quick look, and thanks @GTCC777 for the original repro.

PR #2723, opened 2026-06-28
spec(exact): verify() must be fail-closed relative to settle() -- three-case negative matrix

chopmob-cloud:
## What

Adds a normative  fail-closed section to .

## Why

#2701 identified that the Hedera  passes unsigned and wrong-key transactions because it validates transfer semantics but skips payer-signature verification and recipient-association preflight. #2707 fixes the Hedera implementation. But the invariant itself is not stated anywhere in the top-level  spec, so every new non-EVM mechanism can silently reproduce the same class of bug.

This PR adds the cross-mechanism normative text so the property is owned at the spec level, not rediscovered per-mechanism.

## Changes

 gains a new section:

- **Invariant**:  MUST return  for every condition that would cause  to fail at consensus.
- **Three mandatory negative cases** every  mechanism MUST handle:
  1. Payer signature invalid (wrong key, unsigned body)
  2. Recipient cannot receive the asset (unassociated token, inactive account)
  3. Payer balance insufficient
- **Consequence table** showing what happens for each missing check (verify passes, settle fails).
- **Error reason conventions** (, ) for per-mechanism docs to adopt.

## Relationship to other work

- Complements #2707 (Hedera implementation fix for the same three cases).
- Complements #2666 (settlement-receipt conformance vectors on the settle() side).
- No code changes; spec-only.

## Acceptance criteria

- [ ]  states that  MUST be fail-closed for the three cases
- [ ] Error reason conventions are documented so per-mechanism docs can adopt them
- [ ] No changes to implementation files

### DIFF
--- a/specs/schemes/exact/scheme_exact.md
+++ b/specs/schemes/exact/scheme_exact.md
@@ -39,3 +39,54 @@ While implementation details vary by network, facilitators MUST enforce security
 - Simulation verification: SHOULD simulate via emulation during `/verify` to confirm expected balance changes.
 
 Network-specific rules are in per-network documents: `scheme_exact_svm.md` (Solana), `scheme_exact_stellar.md` (Stellar), `scheme_exact_evm.md` (EVM), `scheme_exact_sui.md` (SUI), `scheme_exact_ton.md` (TON).
+
+## `verify()` must be fail-closed relative to `settle()`
+
+`verify()` is the facilitator's last opportunity to reject a payment before it sponsors fees and submits the transaction to consensus. It MUST return `isValid: false` for every condition that would cause `settle()` to fail on-chain.
+
+A `verify()` that validates transfer semantics (amounts, destinations, scheme version) but skips security-critical pre-conditions is unsafe: the facilitator sponsors fees and network round-trips for transactions the ledger will reject.
+
+### Three negative cases every `exact` mechanism MUST handle
+
+These apply to all networks regardless of signing scheme:
+
+**1. Payer signature invalid**
+
+`verify()` MUST reject transactions where the inferred payer did not produce a valid signature over the frozen transaction body. This includes:
+
+- A transaction signed with the wrong private key
+- An unsigned (serialized but never signed) transaction body
+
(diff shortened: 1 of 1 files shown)


DrVelvetFog:
This belongs at the `exact`-spec level — it's the verify-side complement to the settle-side conformance in #2666, and stating it once is what stops each new non-EVM rail from silently reproducing the #2701 class.

The MUST-vs-SHOULD split across the three cases is the right call, and it's worth stating *why* in the text, because it changes how each case is tested: a payer signature over the frozen body is an **immutable** precondition — verify() is a true fail-closed predictor of settle() there, so MUST is correct. Recipient-provisioning and balance are **point-in-time** — a payer can drain funds or close an account between verify and settle (TOCTOU), so a verify-pass is a best-effort assertion about verify-time state, not a settle guarantee. That's exactly why they're SHOULD, and a conformance suite needs to know the difference: case 1 is a deterministic reject, cases 2–3 are verify-time-state rejects. Making that rationale explicit keeps the "fail-closed predictor" framing honest about what verify() can and can't promise.

One extensibility note: the three cases are the right canonical set for transaction-signed `exact`, but they don't fully cover mechanisms where the authorization artifact is *separable* from the transaction. The allowance-backed exact-svm mechanism in #2717 carries a fourth precondition class — `allowance_mismatch`: an allowance that resolves, is active, and is funded, yet binds the wrong delegator/spender/mint, so it doesn't authorize *this* draw. That isn't reducible to "signature invalid" (the signature is valid) or to recipient/balance. Suggest the section state that the three are the mandatory floor and that mechanisms MAY add precondition rows for their authorization model, so #2717's `allowance_*` and a future rail both map onto the same matrix instead of forking it.

On the error-reason conventions: keeping them rail-agnostic at the reason-code level (not the mechanism-specific failure string) is what lets `htsUnassociated`, `allowance_mismatch`, and an EVM revert all land on the same matrix rows. That's the verify-side half of the conformance artifact I offered to build on #2701 — per-mechanism verify() negative vectors + a standalone checker, the way #2666 makes the settle-side binding a test rather than prose. Happy to target this section's case list directly so "MUST be fail-closed" ships as something a facilitator can run, not just read.

(Procedural: the bot flagged unsigned commits — the repo won't review until they're verified. Rebase + sign clears it; hit the same gate on #2666.)

PR #2744, opened 2026-06-30
fix(evm): emit spec-compliant authorization_value_mismatch error for cross-SDK parity

### LINKED ISSUE #2742: TS ErrorReasons enum rejects a spec-documented error code emitted by the Python/Go facilitators (authorization_value_mismatch)
## Summary

The error reason `invalid_exact_evm_payload_authorization_value_mismatch` is defined
in the spec text and emitted by both the Python and Go facilitators, but it is **not**
present in the TypeScript `ErrorReasons` enum that validates facilitator responses.
Because `error` / `invalidReason` / `errorReason` are validated with
`z.enum(ErrorReasons)`, a response carrying this (documented) reason fails schema
validation on the TS side — a cross-language interop break for a legitimate, spec-listed
error code.

## Where it's defined (uses `_mismatch`)

- **Spec:** `specs/x402-specification-v2.md`, error-reason registry —
  `invalid_exact_evm_payload_authorization_value_mismatch`: *"Payment amount does not
  exactly match the required amount."*
- **Python:** `python/x402/mechanisms/evm/constants.py` —
  `ERR_AUTHORIZATION_VALUE_MISMATCH = "invalid_exact_evm_payload_authorization_value_mismatch"`
- **Go:** `go/mechanisms/evm/exact/v1/facilitator/errors.go` and
  `go/mechanisms/evm/exact/facilitator/errors.go` —
  `ErrAuthorizationValueMismatch = "invalid_exact_evm_payload_authorization_value_mismatch"`

## Where it's missing

- **TypeScript:** `typescript/packages/legacy/x402/src/types/verify/x402Specs.ts` —
  the `ErrorReasons` array contains `invalid_exact_evm_payload_authorization_value`
  but **not** the `..._value_mismatch` variant. `ErrorReasons` backs:
  - `x402ResponseSchema.error: z.enum(ErrorReasons)`
  - `VerifyResponseSchema.invalidReason: z.enum(ErrorReasons)`
  - `SettleResponseSchema.errorReason: z.enum(ErrorReasons)`

## Impact

A facilitator built on the Python or Go reference code returns
`invalid_exact_evm_payload_authorization_value_mismatch` (per the spec) for an
amount mismatch. A TS client that parses that response with the schemas above throws
a `ZodError`, since the value isn't in the enum. So a spec-documented, reference-emitted
error reason is unparseable by the reference TS client.

## Reproduction (TypeScript)

```ts
import { SettleResponseSchema } from "x402/types/verify";

SettleResponseSchema.parse({
  success: false,
  errorReason: "invalid_exact_evm_payload_authorization_value_mismatch",
  transaction: "0x0000000000000000000000000000000000000000000000000000000000000000",
});
// -> ZodError: Invalid enum value ... 'invalid_exact_evm_payload_authorization_value_mismatch'
```

## Notes / history

`git log -S 'authorization_value_mismatch'` on `x402Specs.ts` returns no commits — the
`_mismatch` form was never in the TS enum; the sibling `..._authorization_value` entry
has been there since the v2 development commit (`#705`). So this is a long-standing gap,
not a recent regression.

## Suggested fix

Add `invalid_exact_evm_payload_authorization_value_mismatch` to the `ErrorReasons`
array in `x402Specs.ts`. It's also worth clarifying whether the existing
`invalid_exact_evm_payload_authorization_value` entry is a distinct reason or whether it
was meant to be the `_mismatch` one — the Go `v1` facilitator defines both as separate
constants, so the intent there isn't obvious from the outside.

### Related minor inconsistency (optional)

`go/mechanisms/evm/exact/facilitator/errors.go` defines
`ErrInvalidAuthorizationValue = "invalid_exact_evm_authorization_value"` — missing the
`_payload_` segment present in every other spelling. Possibly worth aligning in the same
pass.

DrVelvetFog:
Closes #2742.

## Problem

The `@x402/evm` facilitator and the Go facilitator emit **different error codes for the identical check** (authorization value ≠ required amount):

- **Go** [`go/mechanisms/evm/exact/facilitator/eip3009.go:62`] → `invalid_exact_evm_payload_authorization_value_mismatch` (present in the spec error registry)
- **TypeScript** [`typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts:205`] → `invalid_exact_evm_authorization_value` — **not** in the spec registry

The TS string also violates the `errors.ts` header contract ("These strings must be character-for-character identical to the Go constants ... to maintain cross-SDK parity"). Go's own `ErrInvalidAuthorizationValue` (the non-`mismatch` string) is defined but never emitted, so `…_authorization_value_mismatch` is the only correct reason for this condition.

Separately — the case in the linked issue — the legacy `x402` package validates facilitator responses with `z.enum(ErrorReasons)`, and that enum was missing the documented reason. A TypeScript client therefore rejected an otherwise-valid response from the Python/Go facilitators as a schema error.

## Fix

- `@x402/evm`: emit `ErrAuthorizationValueMismatch` from the EIP-3009 value check; add the constant (kept `ErrInvalidAuthorizationValue`, still referenced by the v1 path).
- legacy `x402`: add `invalid_exact_evm_payload_authorization_value_mismatch` to `ErrorReasons`.
- Tighten the existing exact-facilitator unit test to assert the emitted reason (it previously only checked `isValid === false`).
- Changeset (patch for both packages).

## Verification

- `@x402/evm` unit suite: **728/728 pass** (`pnpm test` after building deps).
- Prettier clean on all changed files.

---

*Disclosure: drafted with AI assistance (Claude Code). I verified the diff, the cross-SDK behavior, and the tests.*

### DIFF
--- a/typescript/.changeset/evm-authorization-value-mismatch-parity.md
+++ b/typescript/.changeset/evm-authorization-value-mismatch-parity.md
@@ -1,0 +1,6 @@
+---
+'@x402/evm': patch
+'x402': patch
+---
+
+Align the exact EVM authorization-value-mismatch error code across SDKs. The `@x402/evm` facilitator now emits the spec-documented `invalid_exact_evm_payload_authorization_value_mismatch` reason when an authorization value does not match the required amount, matching the Go facilitator (the previous `invalid_exact_evm_authorization_value` string was not in the spec error registry). The legacy `x402` `ErrorReasons` enum now accepts this reason so responses from the Python/Go facilitators no longer fail TypeScript schema validation.
--- a/typescript/packages/legacy/x402/src/types/verify/x402Specs.ts
+++ b/typescript/packages/legacy/x402/src/types/verify/x402Specs.ts
@@ -17,6 +17,7 @@ export const ErrorReasons = [
   "invalid_exact_evm_payload_authorization_valid_after",
   "invalid_exact_evm_payload_authorization_valid_before",
   "invalid_exact_evm_payload_authorization_value",
+  "invalid_exact_evm_payload_authorization_value_mismatch",
   "invalid_exact_evm_payload_signature",
   "invalid_exact_evm_payload_undeployed_smart_wallet",
   "invalid_exact_evm_payload_recipient_mismatch",
--- a/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts
+++ b/typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts
@@ -205,7 +205,7 @@ export async function verifyEIP3009(
   if (BigInt(eip3009Payload.authorization.value) !== BigInt(requirements.amount)) {
     return {
       isValid: false,
-      invalidReason: Errors.ErrInvalidAuthorizationValue,
+      invalidReason: Errors.ErrAuthorizationValueMismatch,
       payer,
     };
   }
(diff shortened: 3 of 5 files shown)


phdargen (approved):
Thanks for the contribution @DrVelvetFog, LGTM 🚀

PR #2544, opened 2026-06-02
spec(extensions): add vcx-v1 + reference implementation

johnnwilliams27:
## Summary

Adds VCX (Verifiable Credential Exchange) — a new identity extension for
x402 v2 — and its reference TypeScript implementation. Closes the
proposal opened on #2400.

VCX adds a `VCX` HTTP header carrying a three-layer identity envelope
(Principal VC + Agent DID + Payment Source) and integrates via the
standard `onPaymentRequired` / `onProtectedRequest` v2 lifecycle hooks.
Verifiers run a four-step verification (principal credential JWS +
revocation + trust list + hash pinning → agent delegation conditions →
envelope-to-payment binding → settlement passthrough).

Resolves the comment-thread items raised by @danielnorkin, @seritalien
(Vauban), and @chopmob-cloud:

- `vcx_present` observability flag + issuer DID (claim-free passive-
  observer telemetry)
- JCS canonicalization (RFC 8785) defined inline in the spec
- `envelopeDigest` (`sha256:<hex>` over JCS-canonical bytes) suitable
  for facilitator settle-time receipt embedding
- §16.1 audit-time binding skeleton, with the full normative
  facilitator-receipt shape deferred to a companion PR

## What's added

**Spec** (`specs/extensions/vcx.md`, ~55KB): normative v1.0 covering
§1-§19.

**Implementation** (`typescript/packages/extensions/src/vcx/`, 24 source
files): subpath of `@x402/extensions` (minor bump 2.14.0 → 2.15.0). New
runtime deps: `did-jwt`, `did-jwt-vc`, `did-resolver`, `canonicalize`.
Public surface re-exported from package root + via the new
`@x402/extensions/vcx` subpath.

**Tests** (`typescript/packages/extensions/test/vcx-*.test.ts`, 12 files
+ 1 test-utils): ~70 cases covering envelope shape, delegation
enforcement, revocation, trust-list resolution, hash pinning, did:web
TLS hardening, SD-JWT VC end-to-end.

**Changeset**: `@x402/extensions` minor.

## v1.0 spec MUSTs implemented

§6 envelope shape, §6.4 envelopeDigest, §7.1 did:key, §7.2 did:web with
full TLS hardening (HTTPS-only, cert validation, no cross-host redirect),
§8.1 inline + §8.2 well-known JWS trust list + didDocumentHash pinning,
§9.1-§9.4 four-step verification, §10 vc-embedded delegation, §11.1
[...]

Issue #2400, opened 2026-05-20
Proposal: VCX (Verified Credential Extension)


johnnwilliams27:
### **Proposal: VCX - Principal Identity Extension for x402**
Type: Standards-Track Extension

Contributing party: PayPal Holdings, Inc.

### **The gap**
x402 solves how machines and humans pay for things on the internet. It doesn't yet solve how an agent can provide representation of the principal they are operating on behalf of. This solves a fundamentally different problem than intent mandates.

That gap is fine for permissionless, low-stakes flows. It's a wall for everything else: regulated commerce with AML/KYC obligations, premium APIs that need to know their counterparty, autonomous agents acting on behalf of a human or business principal. 

Implementers facing these use cases are already shipping ad-hoc identity sidecars: custom headers, out-of-band attestation endpoints, API keys layered onto payment. Each one re-introduces the friction x402 was designed to eliminate. Each one is incompatible with the next.
The longer this goes unaddressed, the more fragmented the identity layer becomes — and the harder it gets to land a single standard later.

### **What VCX is**
An optional identity extension to x402, delivered through the v2 Extension architecture. A new `IDENTITY` header rides alongside the payment, carrying a verifiable envelope with three layers:

- Principal — a W3C Verifiable Credential from a trusted issuer, attesting to the KYC'd human or business behind the transaction
- Agent — a DID identifying the software actor (e.g., an AI agent), cryptographically delegated by the principal
- Payment Source — the wallet, account, or instrument, bound to the payment's sender field

The three layers are linked. Verification happens in a single HTTP roundtrip, alongside payment settlement. Servers and clients that don't need identity are unaffected.

### **Why this matters**
- It unlocks the use cases x402 can't reach today. Regulated APIs, deferred settlement, premium content gating, agent-to-merchant commerce; all of these need a verifiable identity link, and right now they're building around the gap rather than through it.
- It's payment-method-agnostic. Because VCX binds to a generic payment-source identifier and lets x402's own settlement step prove control, the same envelope works whether the underlying payment is an on-chain stablecoin, a balance transfer, ACH, or a card. Identity stops being a per-rail problem.
- It's issuer-agnostic. Any W3C VC issuer can serve as a VCX issuer. Resource servers declare which issuers they trust. There's no central registry and no gatekeeper. A DeFi protocol can accept credentials from PayPal, Cloudflare, Google, Coinbase, etc. simultaneously. A compliance-tight service can restrict to one.
[...]

PR #2747, opened 2026-06-30
spec(extensions): add VCX (Verifiable Credential Exchange) v1.0

johnnwilliams27:
## Summary

Adds the normative **VCX (Verifiable Credential Exchange) v1.0** spec — a new identity extension for x402 v2. Closes the proposal opened on #2400.

This is the **spec half** of #2544, split out so the normative document can be reviewed independently of the TypeScript reference implementation. The implementation lands in companion PR #2748.

VCX adds a `VCX` HTTP header carrying a three-layer identity envelope (Principal VC + Agent DID + Payment Source) and integrates via the standard `onPaymentRequired` / `onProtectedRequest` v2 lifecycle hooks. Verifiers run a four-step verification (principal credential JWS + revocation + trust list + hash pinning → agent delegation conditions → envelope-to-payment binding → settlement passthrough).

## What's in this PR

**Spec only** — `specs/extensions/vcx.md` (~50KB): normative v1.0 covering §1–§19.

## v1.0 normative MUSTs

§6 envelope shape, §6.4 envelopeDigest, §7.1 did:key, §7.2 did:web with full TLS hardening (HTTPS-only, cert validation, no cross-host redirect), §8.1 inline + §8.2 well-known JWS trust list + didDocumentHash pinning, §9.1–§9.4 four-step verification, §10 vc-embedded delegation, §11.1 short-lived + §11.2 Bitstring Status List v1.0 revocation, §12 all three disclosure tiers, §13.1 transactionId uniqueness, §13.2 envelope-to-payment binding, §13.4 fail-closed on unknown delegation conditions, §17 JCS canonicalization (RFC 8785).

## v1.1 carve-outs (explicit non-goals per §19)

§7.4 DID Method Registry beyond did:key/did:web, §8.3 ETSI Trusted List, §13.6 OHTTP verifier-side support, §16 onAfterSettle settlement correlation, §16.1 facilitator-receipt normative shape (skeleton only), settlement-attestation + composite-trust-query profiles.

## Refs

- Companion implementation PR: #2748
- Original combined PR: #2544
- Discussion: #2400
- Comment-period contributors: @danielnorkin, @seritalien (Vauban), @chopmob-cloud — thank you for the early feedback that shaped vcx_present, JCS framing, and the audit-time-binding section

### DIFF
--- a/specs/extensions/vcx.md
+++ b/specs/extensions/vcx.md
@@ -1,0 +1,1067 @@
+# VCX: Verifiable Credential Exchange for x402
+
+**An x402 v2 Extension Specification**
+
+> **Status:** Draft.
+
+The keywords MUST, MUST NOT, REQUIRED, SHALL, SHALL NOT, SHOULD, SHOULD NOT,
+RECOMMENDED, MAY, and OPTIONAL in this document are to be interpreted as
+described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) and
+[RFC 8174](https://www.rfc-editor.org/rfc/rfc8174).
+
+---
+
+## 1. Summary
+
+VCX (Verifiable Credential Exchange) is an optional identity extension for the
+[x402](https://github.com/x402-foundation/x402) HTTP payment protocol,
+implemented per the x402 v2 Extensions architecture. It introduces:
+
+- A single new HTTP header, `VCX`, carrying a three-layer identity envelope
+  (Principal / Agent / Payment Source).
+- Identity requirements emitted by the resource server's `onPaymentRequired`
+  hook, declared via `declareVCXExtension({...})` in route configuration.
+- A four-step verification procedure executed by the resource server's
+  `onProtectedRequest` hook.
+- Three selective-disclosure tiers (issuer-only, selective claims, full
+  disclosure) with concrete cryptographic backing.
+- A defined credential-revocation contract with two profiles (short-lived TTL
+  or W3C Bitstring Status List v1.0).
+- A trust-list distribution model supporting three transports.
+
+VCX is fully backwards-compatible with x402 v2 core: x402 clients and servers
+that do not implement VCX continue to interoperate normally. Resources that
+require VCX simply gate access on extension success in addition to payment
+settlement.
+
+## 2. Motivation
+
+x402 v2 settles the question of *whether* payment occurred. Many resource
+servers also need to settle the question of *who* paid: regulated financial
+services, KYC-gated content, jurisdictional content controls, large-value
+compliance reporting, and agent-mediated transactions where the principal's
+liability differs from the operating wallet.
+
+Existing identity solutions are payment-method-specific (custodial APIs,
+OAuth flows, Plaid) and do not compose with x402. Bolting identity onto x402's
+core `X-PAYMENT` header would force every implementation to handle an
+optional concern in mandatory code paths.
+
+The x402 v2 Extensions architecture is the natural home for this concern.
+Extensions are inherently optional, register through the standard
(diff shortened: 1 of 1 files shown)


johnnwilliams27:
@phdargen split the spec out per request

chopmob-cloud:
Good to see VCX converge on RFC 8785 JCS. That is the shared canonical floor our open reference layer already builds on, so a VCX identity envelope and our content-addressed references compose cleanly and recompute offline. Happy to develop the interoperability.

Donk338:
§19 defers "settlement-attestation + composite-trust-query profiles", and the §16 `onAfterSettle` correlation / §16.1 facilitator-receipt shape, to v1.1. I want to put operational data behind keeping those carve-outs shaped for one specific evidence class — the one a buyer-side gate can't get anywhere else: whether the resource actually delivered after settlement.

The §9.1–§9.4 four-step verification and the §13.2 envelope-to-payment binding are all pre-settlement. They establish who the principal is and that the envelope is bound to this payment. In a protocol with no native refund path, the buyer's dominant residual risk sits on the other side of settlement.

We run a trust registry over the x402 ecosystem — continuous liveness/challenge probing plus paid delivery spot-checks, where we send real USDC on Base to listed sellers and record what comes back. State as of 2026-07-26:

- 2,462 registered services, 1,420,241 liveness/challenge probes
- 272 spot-checks with on-chain settlement against 102 services, $4.56 USDC spent, 2026-07-07 → 2026-07-26
- of the 181 that resolved to a verdict: 138 delivered, 43 did not

The 43 non-deliveries concentrate in 5 services out of the 96 with a resolved settled check. The part that matters for this spec: **all five scored 1.0 liveness, 1.0 challenge validity, and 1.0 price honesty.** Every pre-settlement signal was clean.

| failure shape | checks | services | what the buyer sees |
|---|---|---|---|
| HTTP 402 after on-chain settlement | 22 | 3 | valid paywall, payment lands, still refused |
| HTTP 502 | 11 | 1 | upstream dead after payment |
| HTTP 200, body is not the purchased resource | 10 | 1 | invisible to status-code checking |

The last row motivates the ask. A correct VCX envelope, a valid 402, an honest price and a 200 response can all coexist with the buyer getting nothing. And one of the five is otherwise a legitimately verified service (score 87.4) that failed a single paid delivery — so this is a reliability signal, not only a fraud one.

Two requests, both scoped to what §19 already defers:

1. **§16 / §16.1** — when the `onAfterSettle` correlation and facilitator-receipt shape get drafted, leave room for a *delivery outcome* on the receipt, not just settlement confirmation. Settlement landing on-chain and the resource being served are different events, and only the second is what the buyer paid for.

2. **composite-trust-query profile** — if that lands in v1.1, the evidence taxonomy it would reference (the trust-provider extension in #2300 is converging on `behavioral | regulatory | self-attested | third-party | cryptographic | observational`) has no member meaning "measured post-settlement delivery". Every existing member characterises the payer or the transaction context. I'd argue for an explicit `delivery` class rather than folding it into `behavioral`: behavioural evidence is about the payer, delivery evidence is about the seller, and a gate wants them separable.

Happy to contribute spec text for either, and to publish the underlying spot-check dataset in whatever shape is useful for the comment period.

Methodology, so the numbers are checkable: "settled" means an on-chain USDC tx hash is recorded against the check; spot-checks against our own services are excluded; 91 of the 272 checks did not resolve to a delivered/not-delivered verdict and are excluded from the 138/43 split rather than counted as failures.

Our conforming implementation of the #2300 `TrustQuery` interface is at `https://aegis.borisinc.com/trust/evaluate` (returns `x402-trust-evaluation-v0.1`).

Issue #2178, opened 2026-05-02
Observer interest: open-source x402 conformance testing harness — 25 tests already shipped


msaleme:
Hi x402 Foundation maintainers and contributors,

Following the [Linux Foundation announcement](https://www.coinbase.com/blog/coinbase-and-cloudflare-will-launch-x402-foundation) and the launch of the x402 Foundation as the neutral home for the x402 standard, I want to surface an existing open-source artifact that may be relevant to the conformance testing work the Foundation will scope.

## What exists

`agent-security-harness` ([github.com/msaleme/red-team-blue-team-agent-fabric](https://github.com/msaleme/red-team-blue-team-agent-fabric), Apache 2.0, on PyPI) ships a **25-test x402 security testing module** that has been live since 2026-03-22. The module covers seven attack categories the harness identified as primary surface area for x402 implementations:

| Test ID | Category | Coverage |
|---|---|---|
| X4-001-003 | Payment Challenge Validation | Missing headers, malformed auth, currency mismatch |
| X4-004-006 | Recipient Address Manipulation | Dynamic payTo routing attacks (V2), address spoofing, invalid addresses |
| X4-007-010 | Session Token Security | Token fabrication, expiry bypass, sensitive data leakage |
| X4-011-013 | Spending Limit Exploitation | Rate limit bypass, underpayment, budget exhaustion |
| X4-014-016 | Facilitator Trust | Fake facilitator injection, verification bypass, unreachable facilitator |
| X4-017-018 | Information Disclosure | Leaked keys in 402 response, stack traces in errors |
| X4-019-020 | Cross-Chain Confusion | Wrong network, wrong token type (EURC vs USDC) |

Run shape:
```bash
pip install agent-security-harness
agent-security test x402 --url https://your-x402-endpoint.com
```

## What makes it relevant to conformance testing

Three attributes specifically relevant to standards-body work:

1. **Vendor-neutral.** The harness is authored independently and ships under Apache 2.0 (code) / MIT-0 (skill bundle). No facilitator, issuer, or wallet vendor is privileged in the test design.

2. **CSG mapping.** Each test links to the Constitutional Self-Governance mechanism that should catch it (Hard Constraints, Harm Test, Twelve Numbers, Falsification Requirement). The mapping makes it possible to evaluate whether an x402 implementation's defensive posture is structurally sufficient, not just whether it passes the test.

[...]

PR #2776, opened 2026-07-03
conformance: x402 core conformance-vector set + stdlib runner (#2178)

msaleme:
## Summary

An open, dependency-free **x402 conformance-vector set** — delivering the open-source conformance harness proposed in #2178.

A protocol with governance but no conformance vectors is a promise, not a guarantee: two implementations can each claim x402 compliance and still diverge on the same request. This set lets an implementer *prove* conformance.

## What's here (`conformance/`)

- **52 core conformance vectors** (`vectors/x402-core/`) — each a normative x402 requirement plus the request that exercises it (payment-challenge structure, recipient/amount integrity, session security, spending limits, facilitator trust, info disclosure, cross-chain confusion, replay/double-spend, authorization bypass, settlement finality, protocol abuse).
- **`run_conformance.py`** — stdlib-only runner (no dependencies): `--validate` (offline schema check) and `--url` (replay each vector against your endpoint and report the response against each requirement).
- **`schema.json`** — the vector schema.

## Design

Vectors are **data, not a reference implementation** — you test *your* x402 code against them. They're generated from the Apache-2.0 [`agent-security-harness`](https://pypi.org/project/agent-security-harness/) so they stay faithful to a maintained suite (provenance in each vector's `source`).

Roadmap (in the README): optional machine-checkable `assert` blocks for auto-judging deterministic vectors, and an extension conformance set alongside the relevant extension spec.

## Why it matters

While building the harness these vectors come from, adversarial review of its *own reference verifier* caught three fail-open defects **in that harness code** (a credential-release check trusting a self-asserted flag, a scope check that failed open on an omitted field, and an SSRF blocklist that missed most of the RFC1918 `172.16/12` range) — each fixed and guarded by a negative test before release. Fail-open is easy to ship even in code written to catch it; that is exactly why machine-checkable conformance vectors are worth having.

Apache-2.0, matching this repo. Happy to align vectors to your preferred layout/schema.

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -1,0 +1,75 @@
+<!--
+Copyright 2026 Michael K. Saleme
+SPDX-License-Identifier: Apache-2.0
+-->
+# x402 Conformance Vectors
+
+An open, dependency-free conformance-vector set for x402 implementations —
+delivering the open-source conformance harness proposed in **#2178**.
+
+A protocol with governance but no conformance vectors is a promise, not a
+guarantee: two implementations can each claim x402 compliance and still diverge
+on the same request. These vectors let an implementer *prove* conformance.
+
+## What a vector is
+
+A vector is **data**, not a reference implementation — a normative x402
+requirement plus the request that exercises it:
+
+```json
+{
+  "id": "X4-001",
+  "title": "402 Payment Challenge Headers Present",
+  "category": "payment_challenge",
+  "requirement": "HC-1: Payment protocol must return complete challenge",
+  "method": "GET",
+  "request": { "path": "/" },
+  "normative": "MUST",
+  "expected": "A conformant x402 implementation satisfies the stated requirement; a divergent response is a finding."
+}
+```
+
+The schema is `schema.json`. This set carries **52 vectors** across the x402
+core conformance surface: payment-challenge structure, recipient/amount
+integrity, session security, spending limits, facilitator trust, information
+disclosure, cross-chain confusion, identity verification, replay / double-spend,
+authorization bypass, settlement finality, and protocol abuse.
+
+## Running it against your implementation
+
+```bash
+# validate the vector set (offline, no network)
+python conformance/run_conformance.py --validate
+
+# replay every vector against your x402 endpoint and report the response
+python conformance/run_conformance.py --url https://your-x402-endpoint --report report.json
+```
+
+The runner is stdlib-only (no dependencies). It replays each vector's request
+and pairs the normative requirement with your implementation's observed
+response, so conformance can be judged per requirement in CI or review.
+
+## Scope and roadmap
+
+- **This set: x402 core (52 vectors).** Vectors are generated from the
+  open-source [`agent-security-harness`](https://pypi.org/project/agent-security-harness/)
+  (Apache-2.0) so they stay faithful to a maintained test suite; provenance is
+  recorded in each vector's `source` field.
+- **Planned:** optional machine-checkable `assert` blocks (status / header /
+  body predicates) so the runner can auto-judge deterministic vectors, and an
+  x402 **extension** conformance set (e.g. request-integrity / spend-governance)
+  contributed alongside the relevant extension spec.
+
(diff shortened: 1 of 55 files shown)


ahouramasoud1369-alt:
Verified

msaleme:
Checking in on this one. Current state from my side:

- `check-verified-commits` is passing. Both commits show as verified, so the signing request from the automated comment above is resolved.
- `labeler` is passing.
- The only red check is `Vercel`, reporting `Authorization required to deploy`. That is a deploy authorization on the Coinbase Vercel team, which I do not think a fork contributor can satisfy.

No reviews yet, so I wanted to confirm the PR is not waiting on something from me. Happy to rebase, split it into smaller pieces, or narrow the scope if that would make it easier to review.

msaleme:
@CarsonRoscoe @phdargen apologies for the direct tag. I'm trying to get this routed rather than
continue bumping it.

I found a possible routing gap: `conformance/` is not matched by `.github/CODEOWNERS`, and there is
no catch-all rule. That may explain why no automatic reviewer was requested.

Could one of you point me to the appropriate reviewer or owning team for this path? If a CODEOWNERS
entry would help, I'm happy to submit that separately.

One other question: does the failing Vercel check block merge? It reports `Authorization required
to deploy` against the Coinbase team, which does not appear actionable by a fork contributor. The
remaining checks are green and both commits are verified.

If PR size is the obstacle, I can split the runner, schema, README, and a representative vector set
from the remaining vectors. Otherwise, I would prefer to keep the complete conformance set
together.

shibutatsu (commented):
Thanks for putting this together. I checked the current PR head and ran the offline validator successfully. The corpus is useful staging material, but I see two issues that need to be separated before it can arbitrate x402 core conformance.

First, the live runner does not execute most of the vector semantics. It only consumes `path`, `headers`, and `body`; fields such as `variants`, `n_requests`, `concurrent_requests`, or attacker inputs are ignored. For example, X4-002, X4-011, and X4-043 all reduce to one ordinary GET request, and response bodies are not recorded even for disclosure checks. The report therefore cannot support the stated requirement for those vectors. I would either encode concrete executable request cases plus observation/assertion rules, or mark these entries as metadata-only and exclude them from `--url` replay until that exists.

Second, the normative source needs to be explicit per vector. Requirements such as OATR presence, a fixed manifest TTL, session-token controls, or application rate limits do not appear to be x402 core requirements. Marking them `MUST` under an x402-core directory can make an otherwise conforming implementation look non-conformant. A pinned x402 spec version and exact section or requirement reference should be required for core vectors; harness-specific hardening checks can live in a separately named optional security profile.

Once those boundaries are explicit, the dependency-free format and reproducible data layout could be a solid base for shared vectors.

ahouramasoud1369-alt (approved):

msaleme:
You are right on both points. I checked the full corpus rather than only the examples.

At the current head, `replay()` consumes only `request.path`, `request.headers`, and `request.body`. Forty-six of the 52 vectors contain none of those fields. They therefore fall back to `/`: 45 issue a bare GET and X4-024 issues a body-less POST. The runner records only status and headers, so disclosure-body expectations cannot be evaluated.

`--validate` currently checks that `request` is a dict, but not that it contains an executable request or assertion rules. Those vectors can therefore validate despite not being meaningfully replayable.

You are also right on normativity. The current schema has no required pinned x402 specification version or exact section reference. The `source` field identifies provenance; it does not establish normative authority. Treating these as `MUST` core-conformance vectors is not supportable on that basis.

I will narrow the PR accordingly:

1. Require pinned spec version and exact requirement reference for `x402-core`.
2. Move non-core hardening checks to a separately named optional profile.
3. Mark vectors without concrete request and assertion semantics as metadata-only, excluding them from `--url` replay.
4. Record response bodies where a vector's assertion requires them.

This reduces the scope of the PR, but makes its conformance claims defensible. Thank you for the careful review.

ahouramasoud1369-alt:
[]()

PR #2791, opened 2026-07-05
fix(examples/go): make the echo server + http client pair work from a fresh clone

farid-moradi:
## Description

Following the echo server example README on a fresh clone hits three problems:

- The README says `cp .env-example .env` but the file doesn't exist
- Neither `servers/echo` nor `clients/http` builds: the SDK's dependency bumps were never propagated to the example modules, so `go build` fails until `go mod tidy` is run
- In `clients/http`, the payment details block re-does the header lookup that `extractPaymentResponse` already does internally, and parse errors are silently swallowed (`if err == nil`)

This adds the missing `.env-example` (same format as `servers/custom`), re-tidies both modules under go1.24.1 (the pinned toolchain), and reworks the payment response handling to call the helper once and return parse errors.

Other `examples/go` modules probably have the same stale go.mod problem.

## Tests

No tests in these example dirs. Local verification:

- Reproduced the fresh-clone build failure on go1.24.1 and go1.26.4; both examples build after
- Ran the pair end to end on Base Sepolia via `https://x402.org/facilitator`: payment settled, tx `0x95b3357c39e0e43a940f8c117dd29d22aa15e3ef22cad6e074d1f8c60a75895f`
- `gofmt -l` and `go build ./...` clean in both dirs

## Checklist

- [ ] I added a changelog fragment for user-facing changes (skipped: examples are not published as a package)


phdargen (approved):
Thanks for the contribution @farid-moradi, LGTM 🚀

PR #2792, opened 2026-07-05
specs: add Payment Preflight Record extension (pre-signature record + fail-closed coverage)

bluetieroperations-create:
## Summary

This PR adds `specs/extensions/extension-payment-preflight-record.md`: an optional, composable extension standardizing the record a client agent assembles **between receiving a 402 challenge and signing the payment authorization**. It defines no new scheme and modifies no settlement flow — it standardizes the pre-signature half that every documented agent-payment failure was missing: an independent check between the decision and the damage, with the signature attaching to assembled evidence rather than to a raw transaction the agent was steered into.

## What it specifies

- **A four-block record** (`payment` / `risk` / `authority` / `execution`) with separation of powers: a verdict-only risk oracle that holds nothing, a state holder for budget/idempotency/rollback, an expiring record-bound human confirmation for the HOLD path, and the client signing last under deterministic rules.
- **Machine-readable coverage declaration with fail-closed evaluation**: the oracle declares an envelope of typed predicates over the record's own fields; the consumer evaluates membership locally as a pure function; everything undecidable (missing fields, unknown predicates, malformed ranges, unvalidated versions) reads as unchecked, and unchecked is STOP-equivalent. No executable patterns cross the wire.
- **Bounded, fail-closed HOLD**: a HOLD is a pre-authorization with a TTL; an unanswered escalation resolves to decline, not a lingering pending state. Escalation depth is operator policy keyed to amount, and the human always adjudicates the server-authored record on a surface the agent's context cannot reach.
- **Dispute semantics** (server-authored confirmation records, separate resolution surface, signed receipts) and **calibration guidance** (pre-registered verdicts scored against outcomes, with outcome *provenance* as a first-class axis so self-reported outcomes can't launder into a headline number).

## Why an x402 extension

The 402 challenge's `accepts` entry is the natural binding point: the record commits to the *challenged* payment, closing a steering-attack class. Structurally this is the sibling of the Offer and Receipt extension — offer (server commits to terms) / preflight record (client justifies accepting them) / receipt (proof of exchange): before, during, after. The pre-signature slot in `specs/extensions/` is currently empty.

## Positioning vs adjacent work

| Work | Relationship |
| --- | --- |
| Offer and Receipt (this repo) | structural sibling; composes |
| Mastercard Verifiable Intent | complementary — VI credentials ride in the authority block; VI's scope statement excludes confirmation receipts, budget state, risk data |
| AP2 Agent Authorization Framework | overlaps the authority block only, as static mandates vs this record's live state; AP2 v0.1's open-ended Risk Payload was removed in v0.2 — the `risk` block is a candidate schema for the vacated slot |
| OWASP AI Agent Security (2026) | prescribes exactly this shape as guidance, with no schema; this record is a candidate schema for the requirement |

## Reference implementation

A reference evaluator (signing rules + coverage-envelope membership) exists with adversarial scenario coverage, mutation-tested and independently audited before this submission — the audit found and closed real bypass classes (multi-predicate dispatch smuggling, unbounded ranges that cover everything, substring scope matching), which are now called out as normative fail-closed requirements. Happy to contribute it alongside the spec if useful.

## Status

Draft v0.1 — RFC. This is submitted expecting review here to change it; the version history is ready for that. The schema was shaped in the open by public design review from a dozen-plus agents and builders (credited in the spec's Acknowledgements) plus a live exercise of the risk-oracle role on Base.

### DIFF
--- a/specs/extensions/extension-payment-preflight-record.md
+++ b/specs/extensions/extension-payment-preflight-record.md
@@ -70,7 +70,7 @@ data is harmless; unknown *rules* are not.
 | `risk.price_anomaly` | object | no | risk oracle | e.g. quote vs the payee's own historical median |
 | `risk.sanctions` | object | no | risk oracle | list identity + version pin + hit flag |
 | `risk.coverage` | object | yes | risk oracle | the declared envelope (§5) |
-| `risk.oracle` | object | yes | risk oracle | name, receipt/forecast identifiers |
+| `risk.oracle` | object | yes | risk oracle | name, receipt/forecast id, and the verdict's binding commitment (§5.6) |
 | `authority.policy_basis` | `auto\|human_confirmed` | yes | state holder | which rule permitted this |
 | `authority.budget` | object | yes | state holder | `{cap, spent, remaining, window}` — live state, not a static mandate |
 | `authority.signer_scope` | object | yes | state holder | `{key_id, action_classes[], scope_terminates_on?}` |
@@ -97,7 +97,9 @@ Example (abbreviated):
               { "field": "payment.network", "in": ["eip155:8453"] },
               { "field": "payment.quote_usd", "range": { "gt": 0, "lte": 10000 } },
               { "field": "payment.payee_address", "format": "eth-address" } ] },
-            "oracle": { "name": "…", "receipt_id": "…" } },
+            "oracle": { "name": "…", "receipt_id": "…",
+              "binding": { "record_id": "rec-9f2c01", "payee_address": "0xABC…",
+                           "asset": "USDC", "network": "eip155:8453", "quote_usd": 4.50 } } },
   "authority": { "policy_basis": "human_confirmed",
                  "budget": { "cap": 25, "spent": 5, "remaining": 20, "window": "daily" },
                  "signer_scope": { "key_id": "k1", "action_classes": ["api-purchase"] },
@@ -153,6 +155,21 @@ freshest-stale check has expired as outside coverage for that dimension. This ta
 the Still OS claim-vocabulary specification's per-resolver freshness kinds, which named the
 problem precisely.
 
+**5.6 Verdict binding (Normative).** Coverage answers *does this verdict's envelope contain
+this action* — not *was this verdict computed for this action*. They are different questions,
+and satisfying only the first lets a verdict computed for one challenge be replayed against any
+other action inside the same envelope: two payments to different payees both satisfy
+`payment.network in ["eip155:8453"]`, so an envelope-only check passes a GO the oracle issued
+for the wrong payee. Binding closes this. The oracle's signed receipt (`risk.oracle.receipt_id`)
+MUST commit to the payment it was computed for — at minimum `record_id` and the binding tuple
+(`payment.payee_address`, `payment.asset`, `payment.network`, `payment.quote_usd`); the
+`risk.oracle.binding` object echoes that committed tuple for local inspection. The consumer MUST
+verify BOTH that every echoed value equals the record's own field AND that the receipt attests
+the echo — the echo is not client-assertable, its authority is the oracle's signature over the
+receipt. A verdict whose receipt does not bind, or binds values other than the record carries,
+is OUTSIDE coverage (§3.5) — STOP-equivalent. Envelope membership is necessary, not sufficient:
+a verdict the oracle never issued for *this* challenge is unchecked.
+
 # 6. Signing Rules (Normative)
 
 The client MUST NOT sign the payment authorization unless ALL of the following hold:
@@ -161,12 +178,15 @@ The client MUST NOT sign the payment authorization unless ALL of the following h
 2. `risk.verdict != STOP`; a HOLD verdict additionally requires `authority.human_confirmation`
    present, `status: approved`, unexpired, and bound to THIS `record_id`;
 3. the payment falls INSIDE `risk.coverage` under §5.2;
-4. `payment.quote_usd` is a finite positive number ≤ `authority.budget.remaining`;
-5. `execution.idempotency_key` has never been used;
-6. `authority.signer_scope.action_classes` contains `payment.action_class` exactly
+4. the verdict is bound to THIS payment — `risk.oracle.binding` commits (through the oracle's
+   receipt) to this `record_id` and the `payment.*` binding tuple, and every committed value
+   equals the record's own (§5.6); a verdict not bound to this specific challenge is unchecked;
+5. `payment.quote_usd` is a finite positive number ≤ `authority.budget.remaining`;
+6. `execution.idempotency_key` has never been used;
+7. `authority.signer_scope.action_classes` contains `payment.action_class` exactly
    (array membership — a scalar scope MUST fail);
-7. `execution.rollback.route` is populated (even if `none` — the agent must know);
-8. every rule above evaluated without error; any evaluation failure is a refusal, not an
+8. `execution.rollback.route` is populated (even if `none` — the agent must know);
+9. every rule above evaluated without error; any evaluation failure is a refusal, not an
    exception for the caller to interpret.
 
 A confirmation is a pre-authorization that expires — not a standing power of attorney.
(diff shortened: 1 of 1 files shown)


AmitabhainArunachala:
This is a strong spec — the separation of powers (verdict-only oracle, state holder, confirmation service, client signs last) and the consumer-side fail-closed coverage evaluation (§5.2) are the right architectural choices. The "unchecked is STOP-equivalent" principle is the single most important normative rule here; without it, every coverage gap becomes an implicit pass.

We run a live x402 seller endpoint with a 22-check fail-closed receipt verification matrix (10 paid routes on Base mainnet, CDP facilitator, `exact` scheme). Our operational experience maps directly onto this spec and may be a useful datapoint.

**1. Coverage declaration → check manifest interop.** Our 22 checks fall into four groups that align with your four-block record:

- **Binding checks** (6): resource URL hash, payee address == 402 challenge `payTo`, amount == challenge price, asset/network match, facilitator URL match, response hash binding. These are the concrete predicates your §5 `dimensions[]` would declare over `payment.*`.
- **Settlement truth checks** (8): payer ≠ payee (sybil/wash defense), on-chain tx confirmation, EIP-3009 nonce one-time-use, `transferWithAuthorization` domain match, block confirmation depth, gas/fee reasonableness. These are the "money actually moved, from someone else, once" predicates.
- **Delivery checks** (4): HTTP 200 after payment, response body SHA-256 match, content-type match, latency within tolerance. These are the "goods were delivered" predicates your §7 dispute semantics needs as input.
- **Identity/authority checks** (4): payee ERC-8004 agent ID resolution, payer wallet nonce tracking, idempotency key uniqueness, DPoP `jti` anti-replay. These overlap your `authority.signer_scope` and `execution.idempotency_key`.

The practical observation: **your `coverage.dimensions[]` is the general case of what our check matrix is a concrete instance of.** A preflight record that carries `risk.coverage` dimensions matching our binding + settlement checks would let a consumer evaluate our entire binding layer locally as a pure function — exactly §5.2's design goal.

**2. A concrete gap to flag: `risk.coverage` doesn't bind the verdict to the specific 402 challenge.** Your `payment.*` fields are populated from the 402 `accepts` entry (§9), which is correct. But §5.2's coverage evaluation checks whether the *action* is inside the envelope, not whether the *verdict was computed for this specific challenge*. A risk oracle could return a GO verdict computed against challenge A, and the consumer could evaluate it against challenge B's `payment.*` fields and pass coverage. Consider adding a normative requirement that `risk.oracle.receipt_id` or a challenge hash be bound into the coverage check — so the consumer verifies the verdict was computed for *this* challenge, not just *a* challenge in the same envelope.

**3. The §8 calibration provenance tiers are the hardest part to get right in practice.** We've seen the self-reported-outcome problem directly: our own self-pay test cycles (18 rounds, same-wallet payer == payee) would score as "successful transactions" under any naive volume metric. Your four-tier provenance hierarchy (chain-attested > third-party-resolver > counterparty-corroborated > self-report, segregated and discounted) is the right defense. The operational question is whether oracle calibration data is itself an x402-payable resource (so the calibration receipts are verifiable the same way as payment receipts) or whether it lives outside the payment rail.

**4. Interop offer.** If you produce a preflight record fixture (even a synthetic HOLD verdict with a coverage declaration), we'll run it through our verification endpoint and post the full check-by-check result publicly. Our `/mandate/verify` and `/receipt/verify` routes already evaluate AP2 mandates and x402 settlement receipts against the 22-check matrix — mapping your preflight record's `risk.coverage` dimensions into our predicate evaluator would be a straightforward adapter. One fixture JSON is all we need.

bluetieroperations-create:
Thanks — this is exactly the kind of review that sharpens a spec, and the mapping is right.

**1. Interop mapping.** Agreed: `coverage.dimensions[]` is the general case, your 22 checks a concrete instance. Your four groups line up — binding + settlement are predicates over `payment.*` (settlement is also where `execution.settlement_status` gets its truth), delivery checks are the §7 dispute inputs, and identity/authority overlaps `authority.signer_scope` + `execution.idempotency_key`. A record declaring your binding + settlement predicates in `risk.coverage` is exactly what §5.2 wants evaluated as a pure function.

**2. The challenge-binding gap — fixed.** You're right, and it's real: envelope membership answered "does this verdict apply to an action *like* this," not "was it computed for *this* challenge" — a GO from challenge A rides into challenge B in the same envelope. Added **§5.6 Verdict binding (Normative)**: the oracle's receipt MUST commit to `record_id` + the binding tuple (`payee_address`, `asset`, `network`, `quote_usd`); `risk.oracle.binding` echoes it; the consumer MUST verify both that the echo equals the record **and** that the receipt attests it — unbound/mis-bound fails closed (§3.5). It's now signing rule §6.4 + a §12 security bullet. Credited you in §14.

**3. Calibration provenance — on-rail.** Our lean: calibration data lives on the payment rail, verifiable the same way as payment receipts. The oracle is itself an x402-payable resource (§9) and §8 already requires pre-registered, hash-bound receipts — so a calibration receipt is just another x402 receipt, and the same fail-closed verification you run on settlement receipts applies. That's what keeps the grade-your-own-homework problem (your 18 self-pay rounds, payer == payee — exactly §8's self-report tier) from re-entering through the calibration side door. Off-rail calibration would reintroduce the unverifiable trust assumption §8 exists to remove.

**4. Interop — fixture below.** Taking you up on it: a synthetic HOLD record with a coverage declaration and the new §5.6 binding — valid against the draft, HOLD-not-yet-signable (confirmation pending), price 5× over the payee's own median. If you run it through `/receipt/verify` (or `/mandate/verify`) and post the check-by-check result, that's a strong public datapoint on the dimensions↔predicate adapter. One ask: flag any check where the record's shape doesn't hand your evaluator what it needs — that's spec feedback.

```json
{
  "record_id": "rec-8b41f0a2",
  "created_at": "2026-07-13T18:22:05Z",
  "payment": {
    "rail": "x402",
    "network": "eip155:8453",
    "asset": "USDC",
    "payee_address": "0x9A7fEd3C4b2109fE8A6D5c4B3a2109fEd3C4b210",
    "payee_descriptor": "https://api.example-oracle.xyz/.well-known/x402",
    "resource": "per-call inference: forecast/v1",
    "quote_usd": 4.5,
    "action_class": "api-purchase"
  },
  "risk": {
    "verdict": "HOLD",
    "counterparty_reputation": 0.5,
    "price_anomaly": {
      "quote_usd": 4.5,
      "payee_median_usd": 0.9,
      "ratio": 5.0,
      "flag": "above_median"
    },
    "sanctions": { "list": "OFAC-SDN", "pinned": "2026-07-12", "hit": false },
    "coverage": {
      "version": 1,
      "dimensions": [
        { "field": "payment.rail", "in": ["x402"] },
        { "field": "payment.network", "in": ["eip155:8453"] },
        { "field": "payment.asset", "in": ["USDC"] },
        { "field": "payment.quote_usd", "range": { "gt": 0, "lte": 10000 } },
        { "field": "payment.payee_address", "format": "eth-address" }
      ]
    },
    "oracle": {
      "name": "blackwall-x402-preflight",
      "receipt_id": "bw_rcpt_01J8Q2ZK3M7YV5N9AB4C6D8E0F",
      "binding": {
        "record_id": "rec-8b41f0a2",
        "payee_address": "0x9A7fEd3C4b2109fE8A6D5c4B3a2109fEd3C4b210",
        "asset": "USDC",
        "network": "eip155:8453",
        "quote_usd": 4.5
      }
    }
  },
  "authority": {
    "policy_basis": "human_confirmed",
    "budget": { "cap": 50, "spent": 12.5, "remaining": 37.5, "window": "daily" },
    "signer_scope": { "key_id": "agent-key-3", "action_classes": ["api-purchase"] },
    "instruction_epoch": "2026-07-13T09:00:00Z",
    "human_confirmation": {
      "id": "cfm_7c1d0e",
      "status": "pending",
      "record_id": "rec-8b41f0a2",
      "expires_at": "2026-07-13T18:32:05Z"
    },
    "hold_policy": { "ttl_seconds": 600, "assurance_tier": "confirm_fact_sheet", "decline_on_expiry": true }
  },
  "execution": {
    "idempotency_key": "idem-8b41f0a2",
    "settlement_status": "unsigned",
    "rollback": { "route": "refund", "contact": "payee /.well-known/x402" }
  }
}
```

bluetieroperations-create:
@AmitabhainArunachala — no rush, just floating this back up: the synthetic HOLD fixture (with the new §5.6 binding) is in the reply above whenever you get a chance to run it through your 22-check matrix. Happy to adjust the record shape if anything doesn't map cleanly to your evaluator's predicates — that's useful spec feedback either way. Thanks again for the sharp review.

Issue #1020, opened 2026-01-23
Automatically create an ATA for the payTo address on Solana


matthewmueller:
**Problem:** Transactions currently fail on Solana if you're trying to pay a seller who hasn't setup their associated token account (ATA) with the asset/mint address they're offering.

In the x402 v2 API, the asset/mint address has been abstracted out a bit, so you can just write `price: "$0.01"` and it infers the mint address (such a great idea), but this also means that it's harder to check if the `payTo` address has set up their ATA with the appropriate mint address.

From the seller's perspective, they've already told you how they want to get paid and the buyer is trying to pay you, but the transaction is failing because the seller hasn't quite setup their account just right.

**Proposal:** Could the x402 libraries do this for you? I'm assuming this would happen on the x402 server, but also maybe in the facilitator. Would love to get your thoughts!

ukstv:
@matthewmueller That's by design. Creating ATA on the facilitator/server side opens an attack vector for draining their funds effectively.

1. Attacker sets up as a "seller" with a `payTo` address they control.
2. Facilitator/server creates the ATA for them, spending SOL on that.
3. Attacker closes the ATA via `closeAccount` instruction, receiving the rent SOL back to themselves.
4. Repeat with different mint addresses or new wallets.

This is both a theoretical and practical attack, seen in the wild. The current behavior (requiring sellers to have their ATA set up) is the correct defense here.

matthewmueller:
Hey @ukstv, thanks for the very helpful explanation. What if the server created the ATA? Then the attacker (set up as a seller) would end up paying the SOL to rent the ATA?

It's a bit of a bummer because then sellers also need to have a nominal amount of SOL to setup ATAs, even if they just want to transact in stablecoins.

But I still think it'd be a UX improvement if the `ExactSvmScheme` signer or the `x402ResourceServer` server set this up for the seller.

notorious-d-e-v:
Think this is better supported as a:

- resource server endpoint by facilitators, where the facilitator is technically the merchant, and the merchant is the client/buyer. The client/buyer calls it one time and pays via stables (or whatever currency the facilitator asks for). In this case it wouldn't require SOL and follows the typical x402 flow.

- an extension where the facilitator advertises supporting this and creates it on behalf of the merchant, but perhaps requires that the merchant already has some type of payment system with the facilitator where the facilitator will draw down credits from the merchant

[...]

PR #2798, opened 2026-07-06
feat(svm): idempotent create-ATA for first-time recipients in exact scheme (v1+v2)

### LINKED ISSUE #2395: Add idempotent ATA creation before SVM TransferChecked payments
## Summary

`@x402/svm`'s exact SVM client schemes build payment transactions with `TransferChecked`, but do not create the recipient associated token account (ATA) first.

If the recipient has never held the token before, the destination ATA may not exist yet, causing the payment transaction to fail with an error like `InstructionError: [2, InvalidAccountData]`.

## Expected behavior

The SVM exact payment transaction should include an idempotent associated token account creation instruction before `TransferChecked`, using the payment recipient as the ATA owner and the payment mint/token program.

Because the instruction is idempotent, it should be safe when the ATA already exists and should allow first-time token recipients to be paid without pre-creating their ATA out of band.

## Affected area

This appears to apply to both current exact SVM client paths:

- `ExactSvmScheme.createPaymentPayload`
- `ExactSvmSchemeV1.createPaymentPayload`

kakedashi3:
Closes #2395

## Problem

The exact SVM client builds `TransferChecked` against the recipient's associated token account without ensuring it exists. The first payment to any `payTo`/mint whose ATA has never been created fails with `InstructionError: [.., InvalidAccountData]`.

As discussed in #2395, this cannot be fixed client-side alone: the facilitator's static path pins `instructions[2]` to `TransferChecked` and only exempts Lighthouse/Memo beyond it, so a prepended create-ATA trips `no_transfer_instruction`. This PR makes the coordinated client + facilitator change, for both v2 and v1.

## Design (as converged in the issue thread)

**Client (v1 + v2):** when the destination ATA does not exist (one `getAccountInfo` check), prepend a `CreateAssociatedTokenIdempotent` funded by the facilitator fee payer via `createNoopSigner`. The repeat-payment layout is byte-identical to today — the new instruction only appears for first-time recipients, and the idempotent discriminator makes a lost race with a concurrent creation harmless.

**Facilitator (v1 + v2) static path:** accept exactly one optional create-ATA at index 2, shifting `TransferChecked` to index 3 (instruction-count bounds raised by one only when it is present). Because the fee payer funds the rent in the gasless flow, every field is pinned — an unscoped allowance would let a client make the facilitator rent-fund arbitrary `(owner, mint)` ATAs (credit @DrVelvetFog for surfacing this):

| pin | value |
|---|---|
| program | ATA program |
| data | exactly `0x01` (idempotent; legacy `Create` is empty-data + 7 accounts) |
| accounts | exactly 6 |
| `account[0]` funder | `requirements.extra.feePayer` |
| `account[2]` owner | `payTo` |
| `account[3]` mint | `asset` |
| `account[4]` | System Program |
| `account[5]` token program | the transfer's token program (ATA derivation seed — keeps the check sound for Token-2022) |
| `account[1]` ata | the `TransferChecked` destination, itself required to equal the ATA derived from (`payTo`, token program, mint) |

The byte-level shape follows @chopmob-cloud's spec in the thread; their byte-validated positive/negative fixture set (`pos-01`, `neg-01-legacy-create`, `neg-02-tampered-destination`, `neg-03-sender-funder`) is offered under Apache-2.0 and will be added to this PR.

## Changes

- `src/exact/createAta.ts` (new): shared static validator for the optional create-ATA instruction
- `src/exact/client/scheme.ts`, `src/exact/v1/client/scheme.ts`: conditional idempotent create-ATA
- `src/exact/facilitator/scheme.ts`, `src/exact/v1/facilitator/scheme.ts`: optional index-2 create-ATA with full pinning
- `test/unit/createAta.test.ts` (new): validator pins (incl. legacy `Create`, funder≠feePayer, Token-2022), client layout with/without existing ATA, facilitator accept + funder-pin reject end-to-end
- changeset: `@x402/svm` minor

`pnpm build` clean, 206/206 tests pass, eslint clean.

## Open questions for maintainers

1. **Path 2 interplay:** create-ATA pin failures are currently *not* in `LAYOUT_RECOVERABLE_REASONS`, so they hard-fail rather than falling through to smart-wallet simulation. Smart wallets wrap transfers in CPI (their top-level ix is the wallet program, not the ATA program), so this shouldn't affect them — but happy to move them into the recoverable set if you prefer.
2. Whether the client should *always* include the idempotent create (deterministic layout, +1 RPC saved) instead of conditionally — conditional was chosen to keep the common path byte-identical.

## Notes

- Draft until maintainers confirm the intended shape (per the thread); @chopmob-cloud will land the byte fixtures here, and @DrVelvetFog offered to review the vectors.
- AI disclosure: implemented with AI assistance (Claude); design from the #2395 thread, all code reviewed and tested locally by me.

### DIFF
--- a/typescript/.changeset/svm-idempotent-ata-create.md
+++ b/typescript/.changeset/svm-idempotent-ata-create.md
@@ -1,0 +1,5 @@
+---
+'@x402/svm': minor
+---
+
+Support first-time token recipients in the exact SVM scheme (#2395). When the destination associated token account does not exist yet, the client (v1 and v2) now prepends a `CreateAssociatedTokenIdempotent` instruction funded by the facilitator fee payer — only when the ATA is actually missing, so the repeat-payment layout is unchanged. The facilitator static path (v1 and v2) accepts exactly one optional create-ATA at index 2 (TransferChecked shifts to index 3) and pins every field of it — idempotent discriminator byte, six-account layout, funder == fee payer, owner == payTo, mint == asset, system program, token program == the transfer's, and derived ATA == the transfer destination — so the only account the facilitator can be made to rent-fund is the one it is being paid into. New error reasons: `invalid_exact_svm_payload_create_ata_{not_idempotent,account_count,funder_mismatch,owner_mismatch,mint_mismatch,system_program_mismatch,token_program_mismatch,destination_mismatch}`.
--- a/typescript/packages/mechanisms/svm/src/exact/client/scheme.ts
+++ b/typescript/packages/mechanisms/svm/src/exact/client/scheme.ts
@@ -5,11 +5,13 @@ import {
 import { TOKEN_PROGRAM_ADDRESS } from "@solana-program/token";
 import {
   findAssociatedTokenPda,
+  getCreateAssociatedTokenIdempotentInstruction,
   getTransferCheckedInstruction,
   TOKEN_2022_PROGRAM_ADDRESS,
 } from "@solana-program/token-2022";
 import {
   appendTransactionMessageInstructions,
+  createNoopSigner,
   createTransactionMessage,
   getBase64EncodedWireTransaction,
   partiallySignTransactionMessageWithSigners,
@@ -18,6 +20,7 @@ import {
   setTransactionMessageFeePayer,
   setTransactionMessageLifetimeUsingBlockhash,
   type Address,
+  type Instruction,
 } from "@solana/kit";
 import type { PaymentPayload, PaymentRequirements, SchemeNetworkClient } from "@x402/core/types";
 import {
@@ -108,6 +111,27 @@ export class ExactSvmScheme implements SchemeNetworkClient {
       throw new Error("feePayer is required in paymentRequirements.extra for SVM transactions");
     }
 
+    // If the recipient has never held this token, the destination ATA does not
+    // exist yet and TransferChecked would fail with InvalidAccountData (#2395).
+    // Prepend an idempotent create funded by the facilitator fee payer — only
+    // when the ATA is actually missing, so the common repeat-payment layout is
+    // unchanged. The facilitator statically pins every field of this
+    // instruction, and the idempotent discriminator makes a lost race with a
+    // concurrent creation harmless.
+    const prefixInstructions: Instruction[] = [];
+    const destinationInfo = await rpc.getAccountInfo(destinationATA, { encoding: "base64" }).send();
+    if (!destinationInfo.value) {
+      prefixInstructions.push(
+        getCreateAssociatedTokenIdempotentInstruction({
+          payer: createNoopSigner(feePayer),
+          ata: destinationATA,
+          owner: paymentRequirements.payTo as Address,
+          mint: paymentRequirements.asset as Address,
+          tokenProgram: tokenProgramAddress,
+        }),
+      );
+    }
+
     const { value: latestBlockhash } = await rpc.getLatestBlockhash().send();
 
     const sellerMemo = paymentRequirements.extra?.memo as string | undefined;
@@ -140,7 +164,7 @@ export class ExactSvmScheme implements SchemeNetworkClient {
           getSetComputeUnitLimitInstruction({ units: DEFAULT_COMPUTE_UNIT_LIMIT }),
           tx,
         ),
-      tx => appendTransactionMessageInstructions([transferIx, memoIx], tx),
+      tx => appendTransactionMessageInstructions([...prefixInstructions, transferIx, memoIx], tx),
       tx => setTransactionMessageLifetimeUsingBlockhash(latestBlockhash, tx),
     );
 
--- a/typescript/packages/mechanisms/svm/src/exact/createAta.ts
+++ b/typescript/packages/mechanisms/svm/src/exact/createAta.ts
@@ -1,0 +1,116 @@
+import { TOKEN_PROGRAM_ADDRESS } from "@solana-program/token";
+import {
+  ASSOCIATED_TOKEN_PROGRAM_ADDRESS,
+  TOKEN_2022_PROGRAM_ADDRESS,
+} from "@solana-program/token-2022";
+
+/** The System Program, pinned as `account[4]` of a create-ATA instruction. */
+export const SYSTEM_PROGRAM_ADDRESS = "11111111111111111111111111111111";
+
+/**
+ * Single-byte instruction discriminator for CreateAssociatedTokenIdempotent.
+ * The legacy non-idempotent Create carries empty data (and a 7-account layout
+ * including the rent sysvar), so pinning the data to exactly this byte is what
+ * rejects a non-idempotent create being slipped into the same slot.
+ */
+export const CREATE_ATA_IDEMPOTENT_DISCRIMINATOR = 1;
+
+/**
+ * Minimal structural view of a decompiled instruction, shared by the v1 and v2
+ * facilitator schemes (whose concrete instruction types differ only in generics).
+ */
+export type DecompiledInstructionLike = {
+  programAddress: { toString(): string };
+  accounts?: ReadonlyArray<{ address: { toString(): string } }>;
+  data?: Readonly<Uint8Array>;
+};
+
+/**
+ * Whether an instruction targets the Associated Token Account program.
+ *
+ * @param instruction - A decompiled top-level instruction
+ * @returns True when the instruction's program is the ATA program
+ */
+export function isAssociatedTokenProgramInstruction(
+  instruction: DecompiledInstructionLike,
+): boolean {
+  return instruction.programAddress.toString() === ASSOCIATED_TOKEN_PROGRAM_ADDRESS.toString();
+}
+
+/**
+ * Accounts of a validated CreateAssociatedTokenIdempotent instruction that the
+ * caller must additionally pin against the parsed TransferChecked instruction:
+ * `ata` must equal the transfer destination (which the static path already
+ * requires to be the ATA derived from payTo/tokenProgram/mint), and
+ * `tokenProgram` must equal the transfer's token program. The token program is
+ * an ATA derivation seed, so without that last pin a Token-2022 create could
+ * point the rent at a different derived address than the one being paid.
+ */
+export type CreateAtaAccounts = {
+  ata: string;
+  tokenProgram: string;
+};
+
+/**
+ * Statically validate an optional CreateAssociatedTokenIdempotent instruction
+ * at index 2 of an exact-SVM payment transaction (see #2395).
+ *
+ * The fee payer (facilitator) funds the rent in the gasless flow, so every
+ * degree of freedom must be pinned or a client could make the facilitator
+ * rent-fund an arbitrary (owner, mint) ATA unrelated to the payment:
+ *
+ * - data == exactly the single idempotent discriminator byte `0x01`
+ * - exactly six accounts: [funder, ata, owner, mint, systemProgram, tokenProgram]
+ * - funder  == requirements.extra.feePayer
+ * - owner   == requirements.payTo
+ * - mint    == requirements.asset
+ * - systemProgram == the System Program
+ * - tokenProgram  ∈ {Token, Token-2022}
+ *
+ * @param instruction - The decompiled instruction at index 2
+ * @param expected - Pinned values from the payment requirements
+ * @param expected.payTo - The payment recipient (ATA owner)
+ * @param expected.asset - The payment mint
+ * @param expected.feePayer - The facilitator fee payer funding the rent
+ * @returns The `{ ata, tokenProgram }` accounts on success, or an
+ *   `invalidReason` string when any pin fails
+ */
+export function validateCreateAtaIdempotentInstruction(
+  instruction: DecompiledInstructionLike,
+  expected: { payTo: string; asset: string; feePayer: string },
+): CreateAtaAccounts | { invalidReason: string } {
+  const data = instruction.data;
+  if (!data || data.length !== 1 || data[0] !== CREATE_ATA_IDEMPOTENT_DISCRIMINATOR) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_not_idempotent" };
+  }
+
+  const accounts = instruction.accounts ?? [];
+  if (accounts.length !== 6) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_account_count" };
+  }
+
+  const [funder, ata, owner, mint, systemProgram, tokenProgram] = accounts.map(account =>
+    account.address.toString(),
+  );
+
+  if (funder !== expected.feePayer) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_funder_mismatch" };
+  }
+  if (owner !== expected.payTo) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_owner_mismatch" };
+  }
+  if (mint !== expected.asset) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_mint_mismatch" };
+  }
+  if (systemProgram !== SYSTEM_PROGRAM_ADDRESS) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_system_program_mismatch" };
+  }
+  if (
+    tokenProgram !== TOKEN_PROGRAM_ADDRESS.toString() &&
+    tokenProgram !== TOKEN_2022_PROGRAM_ADDRESS.toString()
+  ) {
+    return { invalidReason: "invalid_exact_svm_payload_create_ata_token_program_mismatch" };
+  }
+
+  return { ata, tokenProgram };
+}
--- a/typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
@@ -37,6 +37,10 @@ import {
   getTokenPayerFromTransaction,
   transactionMessageHash,
 } from "../../utils";
+import {
+  isAssociatedTokenProgramInstruction,
+  validateCreateAtaIdempotentInstruction,
+} from "../createAta";
 import { verifySmartWalletTransaction, verifyPostSettlement } from "./smartWalletVerification";
 
 /**
@@ -554,15 +558,19 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
     const decompiled = decompileTransactionMessage(compiled);
     const instructions = decompiled.instructions ?? [];
 
-    // Allow 3-7 instructions:
+    // Allow 3-7 instructions (8 when an idempotent create-ATA is present):
     // - 3 instructions: ComputeLimit + ComputePrice + TransferChecked
     // - 4 instructions: ComputeLimit + ComputePrice + TransferChecked + Lighthouse or Memo
     // - 5 instructions: ComputeLimit + ComputePrice + TransferChecked + Lighthouse + Lighthouse or Memo
     // - 6 instructions: ComputeLimit + ComputePrice + TransferChecked + Lighthouse + Lighthouse + Memo
     // - 7 instructions: + a third wallet-injected Lighthouse (Phantom, see #2097)
+    // An optional CreateAssociatedTokenIdempotent at index 2 (first payment to a
+    // payTo/mint whose ATA does not exist yet, see #2395) shifts TransferChecked
+    // to index 3 and raises each bound by one.
     // See: https://github.com/x402-foundation/x402/issues/828
     //  and: https://github.com/x402-foundation/x402/issues/2097
-    if (instructions.length < 3 || instructions.length > 7) {
+    //  and: https://github.com/x402-foundation/x402/issues/2395
+    if (instructions.length < 3 || instructions.length > 8) {
       return {
         isValid: false,
         invalidReason: "invalid_exact_svm_payload_transaction_instructions_length",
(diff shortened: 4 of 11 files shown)


DrVelvetFog:
Read through the implementation — the validator pins match what we landed on in #2395: data pinned to the single `0x01` idempotent byte (rejecting both the empty-data legacy `Create` and an explicit `0x00`), the six-account layout, and funder/owner/mint/systemProgram all pinned with `tokenProgram ∈ {Token, Token-2022}`. Good that the token-program pin is explicit rather than assumed — it's load-bearing, since the ATA derivation seeds on the token program, so without it a Token-2022 create could point the rent at a different derived address than the one being paid. The `neg-02` recomputation path — derive the ATA from owner/mint/tokenProgram and require it to equal the transfer destination rather than trusting `account[1]` — is what actually closes the rent-redirect, so that's the one I most want a fixture to exercise.

On the open question — conditional include vs. always include — I'd keep it conditional, as you have it. From the operator side the deciding factor is rollout across facilitators at mixed update states: with the conditional path, repeat payments (the overwhelming common case) keep the classic layout — the compute-budget pair, TransferChecked, and memo — so a facilitator that hasn't merged this yet still validates them exactly as before; only a first payment to a fresh ATA carries the create and needs the updated static path. Always-include inverts that: every payment grows the index-2 instruction, so every facilitator has to update in lockstep or all payments break — a flag-day dependency I'd avoid on a decentralized operator set. The one `getAccountInfo` only fires on the first-payment path, so steady-state cost is unchanged, and since the create is idempotent a stale "missing" read that races a concurrent creation is a no-op rather than a failure — so conditional buys the backward-compat with no correctness cost. Right call.

I'll run the vector byte-verification as soon as @chopmob-cloud's fixtures land — recompute the derived-ATA equality for `neg-02`, and confirm `neg-01` (empty-data legacy create) and `neg-03` (sender-as-funder) reject against the validator's actual pins, so we know the set exercises the recomputation path and not just field equality.

chopmob-cloud:
Fixtures below, dropped straight on the PR as you preferred, so the byte-verification can run against the actual set. One positive plus three negatives, deterministic keypairs so they reproduce byte-for-byte. I added `account[4]` == System Program to the assertion list, so the set now spells out every pin in the table.

Each negative rejects on a distinct pin, so passing the whole set requires a validator to enforce all of them, not only the obvious data-and-length checks:

- `neg-01-legacy-create` rejects the legacy non-idempotent `Create` (empty data, 7 accounts).
- `neg-02-tampered-destination` puts a wrong address in `account[1]`, so it only passes if the validator recomputes the ATA from `payTo` + token program + mint and requires that to equal the transfer destination, rather than trusting `account[1]`.
- `neg-03-sender-funder` is an otherwise-correct idempotent create funded by the sender instead of the fee payer, locking the funder pin.

<details><summary><code>ata-create-fixtures.json</code> (1 ACCEPT + 3 REJECT)</summary>

```json
{
  "SPDX-License-Identifier": "Apache-2.0",
  "copyright": "Copyright 2026 Christopher Hopley / AlgoVoi (chopmob-cloud). Contributed under Apache-2.0.",
  "context": {
    "payTo": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
    "payment_mint": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
    "transfer_token_program": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
    "fee_payer": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
    "transfer_destination": "DNDTCnZkNk358qDFZd9unHtnrc73SsXcpVWtwJJMrR4B",
    "system_program": "11111111111111111111111111111111"
  },
  "assertions": [
    "program == ATA program",
    "data == 0x01",
    "accounts length == 6",
    "account[1] == transfer_destination",
    "account[2] == payTo",
    "account[3] == payment_mint",
    "account[5] == transfer_token_program",
    "account[0] == fee_payer",
    "account[4] == System Program"
  ],
  "fixtures": [
    {
      "id": "fixture-pos-01",
      "description": "correct idempotent create, all seven assertions hold",
      "instruction": {
        "program_id": "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
        "data_b64": "AQ==",
        "data_hex": "01",
        "accounts": [
          {
            "pubkey": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
            "is_signer": true,
            "is_writable": true
          },
          {
            "pubkey": "DNDTCnZkNk358qDFZd9unHtnrc73SsXcpVWtwJJMrR4B",
            "is_signer": false,
            "is_writable": true
          },
          {
            "pubkey": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "11111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
            "is_signer": false,
            "is_writable": false
          }
        ]
      },
      "expected_verdict": "ACCEPT",
      "failed_assertions": []
    },
    {
      "id": "fixture-neg-01-legacy-create",
      "description": "legacy non-idempotent Create (empty instruction data)",
      "instruction": {
        "program_id": "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
        "data_b64": "",
        "data_hex": "",
        "accounts": [
          {
            "pubkey": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
            "is_signer": true,
            "is_writable": true
          },
          {
            "pubkey": "DNDTCnZkNk358qDFZd9unHtnrc73SsXcpVWtwJJMrR4B",
            "is_signer": false,
            "is_writable": true
          },
          {
            "pubkey": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "11111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "SysvarRent111111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          }
        ]
      },
      "expected_verdict": "REJECT",
      "failed_assertions": [
        "data == 0x01",
        "accounts length == 6"
      ]
    },
    {
      "id": "fixture-neg-02-tampered-destination",
      "description": "account[1] tampered to a different ATA, not the derived TransferChecked destination",
      "instruction": {
        "program_id": "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
        "data_b64": "AQ==",
        "data_hex": "01",
        "accounts": [
          {
            "pubkey": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
            "is_signer": true,
            "is_writable": true
          },
          {
            "pubkey": "FHPASu6WrzXmbm5NbAQy9BxwX5naUHKn5z8ycurGoSX",
            "is_signer": false,
            "is_writable": true
          },
          {
            "pubkey": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "11111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
            "is_signer": false,
            "is_writable": false
          }
        ]
      },
      "expected_verdict": "REJECT",
      "failed_assertions": [
        "account[1] == transfer_destination"
      ]
    },
    {
      "id": "fixture-neg-03-sender-funder",
      "description": "otherwise-correct idempotent create funded by the sender, not the fee payer",
      "instruction": {
        "program_id": "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
        "data_b64": "AQ==",
        "data_hex": "01",
        "accounts": [
          {
            "pubkey": "9hSR6S7WPtxmTojgo6GG3k4yDPecgJY292j7xrsUGWBu",
            "is_signer": true,
            "is_writable": true
          },
          {
            "pubkey": "DNDTCnZkNk358qDFZd9unHtnrc73SsXcpVWtwJJMrR4B",
            "is_signer": false,
            "is_writable": true
          },
          {
            "pubkey": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "11111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
            "is_signer": false,
            "is_writable": false
          }
        ]
      },
      "expected_verdict": "REJECT",
      "failed_assertions": [
        "account[0] == fee_payer"
      ]
    }
  ]
}
```
</details>

<details><summary><code>verify.py</code> (dependency-free; recomputes each verdict from the pins)</summary>

```python
#!/usr/bin/env python3
# SPDX-License-Identifier: Apache-2.0
# Copyright 2026 Christopher Hopley / AlgoVoi (chopmob-cloud). Contributed under Apache-2.0.
"""Dependency-free verifier for the x402 create-ATA facilitator conformance fixtures.

Recomputes the facilitator's index-2 accept/reject verdict for each fixture purely
from the pinned assertions, and checks it against the fixture's declared verdict.
The fixtures are data, not a reference implementation: run your own facilitator's
static-path check against them, or run this to confirm the set is internally sound.

    python3 verify.py            # verifies ata-create-fixtures.json in this dir
"""
import json, sys, base64

ATA_PROGRAM = "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL"
SYSTEM_PROGRAM = "11111111111111111111111111111111"

def evaluate(instr, ctx):
    """Return the ordered list of pin assertions this create-ATA instruction fails."""
    accts = [a["pubkey"] for a in instr["accounts"]]
    def at(i): return accts[i] if i < len(accts) else None
    fails = []
    if instr["program_id"] != ATA_PROGRAM:            fails.append("program == ATA program")
    if instr.get("data_hex", "") != "01":             fails.append("data == 0x01")
    if len(accts) != 6:                               fails.append("accounts length == 6")
    if at(0) != ctx["fee_payer"]:                     fails.append("account[0] == fee_payer")
    if at(1) != ctx["transfer_destination"]:          fails.append("account[1] == transfer_destination")
    if at(2) != ctx["payTo"]:                         fails.append("account[2] == payTo")
    if at(3) != ctx["payment_mint"]:                  fails.append("account[3] == payment_mint")
    if at(4) != SYSTEM_PROGRAM:                       fails.append("account[4] == System Program")
    if at(5) != ctx["transfer_token_program"]:        fails.append("account[5] == transfer_token_program")
    return fails

def main(path="ata-create-fixtures.json"):
    with open(path, "rb") as fh:
        d = json.loads(fh.read().decode("utf-8"))
    ctx = d["context"]
    ok = True
    for fx in d["fixtures"]:
        instr = fx["instruction"]
        fails = evaluate(instr, ctx)
        verdict = "ACCEPT" if not fails else "REJECT"
        vok = verdict == fx["expected_verdict"]
        aok = set(fails) == set(fx.get("failed_assertions", []))
        # data_b64 must decode to data_hex when both are present
        bok = True
        if "data_b64" in instr:
            bok = base64.b64decode(instr["data_b64"]).hex() == instr.get("data_hex", "")
        good = vok and aok and bok
        ok = ok and good
        print(f"  {'PASS' if good else 'FAIL'}  {fx['id']}: {verdict}"
              + ("" if good else f"  (expected {fx['expected_verdict']}, fails={fails})"))
    print("OK" if ok else "MISMATCH")
    return 0 if ok else 1

if __name__ == "__main__":
    sys.exit(main(*sys.argv[1:]))
```
</details>

Apache-2.0, each file carries its own SPDX identifier and copyright, so they can land in the test suite directly or I can open them as a follow-up against the SVM test dir alongside `createAta.test.ts` if that is cleaner.

On your conditional-vs-always question: conditional is the correct choice. What settled it for us is that the conditional path leaves the repeat-payment layout byte-identical, so a facilitator that has not shipped this yet keeps validating the common case exactly as before; the single `getAccountInfo` only fires on a first payment to a fresh ATA. It is the same choice our own fix made.

DrVelvetFog:
Ran the byte-verification I said I would. The part that matters: I re-derived the destination rather than trusting the declared `transfer_destination` — `findProgramAddress([payTo, tokenProgram, mint], ATA program)` with the `ProgramDerivedAddress` salt gives `DNDTCnZk…JJMrR4B` (bump 255, off-curve), a byte-for-byte match to `pos-01`. So `pos-01` is a genuine accept under a validator that *recomputes* the destination, not one that trusts `account[1]` — which is the equality that actually closes the rent-redirect, so good to see it hold against a fresh derivation and not just a field match.

Rejects, run against a port of `validateCreateAtaIdempotentInstruction` plus the two facilitator caller pins, each failing on exactly one pin and mapping to the real reason codes:

- `neg-01` → `…_create_ata_not_idempotent` (empty data ≠ single `0x01`)
- `neg-03` → `…_create_ata_funder_mismatch` (`account[0]` is the sender, not `feePayer`)
- `neg-02` → `…_create_ata_destination_mismatch`

One note on `neg-02` for whoever wires these into the suite: it doesn't fail anything *inside* `validateCreateAtaIdempotentInstruction` — that function never reads `account[1]`; it returns `{ata, tokenProgram}` and the reject happens one level up at `createAtaAccounts.ata !== destATA`. So it exercises the transitive guard (`ata == transfer dest == derived ATA`) exactly as intended, but only against the full facilitator path — point it at the validator in isolation and it reads as a false accept. Worth a comment on the case so it lands on the right unit.

Set's sound: 1 accept + 3 rejects, each negative isolating a distinct pin. 👍 on the vectors.

**One gap worth closing.** Every fixture in the set is SPL Token, so the token-program *seed* — the pin we flagged as load-bearing for Token-2022 — isn't actually driven yet. Since the ATA derives on `(owner, token program, mint)`, the same owner+mint lands on a different address under each program (for the pair below, using PYUSD: `5AJ5Pm…41y` under Token-2022 vs `6v1NJ9v…w17` under SPL). I byte-verified a Token-2022 companion that covers both directions — reproduces under your `verify.py` and against the facilitator port:

- `pos-02-token2022` — legit Token-2022 payment, `account[5]` = Token-2022, `account[1]` the Token-2022-derived ATA → **ACCEPT**. Catches the likelier regression: a validator that hardcodes SPL Token in the derivation would silently reject *every* Token-2022 payment, and nothing in the current set guards that.
- `neg-04-cross-program-create` — transfer under Token-2022 but the create names SPL Token in `account[5]`, `account[1]` left as the correct Token-2022 destination so only the token-program pin fails → **REJECT** (`…_create_ata_token_program_mismatch`). On-chain this shape rent-funds the SPL-derived ATA instead of the paid one.

Mint is PYUSD (`2b1kV6Dk…24GXo`), a live Token-2022 mint on mainnet — owner-verified via `getAccountInfo` — so the destination reproduces from `(payTo, Token-2022 program, mint)`. Dropping it inline below under Apache-2.0. @chopmob-cloud — happy to fold it into your set or open it as a follow-up against the SVM test dir alongside `createAta.test.ts`, whichever you'd prefer for placement.

<details><summary><code>ata-create-fixtures-token2022.json</code> (1 ACCEPT + 1 REJECT, Token-2022)</summary>

```json
{
  "SPDX-License-Identifier": "Apache-2.0",
  "copyright": "Copyright 2026 Antoni Jagodka (DrVelvetFog). Contributed under Apache-2.0.",
  "note": "Token-2022 companion to chopmob-cloud's ata-create-fixtures.json (#2798). Exercises the token-program derivation seed in both directions. payment_mint is PYUSD, a live Token-2022 mint on Solana mainnet; transfer_destination is the ATA derived under Token-2022 and reproduces from (payTo, Token-2022 program, mint).",
  "context": {
    "payTo": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
    "payment_mint": "2b1kV6DkPAnxd5ixfnxCpjxmKwqjjaYmCZfHsFu24GXo",
    "transfer_token_program": "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb",
    "fee_payer": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
    "transfer_destination": "5AJ5PmLBPTY9JkKPSkSdNrghd1iqoWh1U55g3ByVG41y",
    "system_program": "11111111111111111111111111111111"
  },
  "assertions": [
    "program == ATA program",
    "data == 0x01",
    "accounts length == 6",
    "account[1] == transfer_destination",
    "account[2] == payTo",
    "account[3] == payment_mint",
    "account[5] == transfer_token_program",
    "account[0] == fee_payer",
    "account[4] == System Program"
  ],
  "fixtures": [
    {
      "id": "fixture-pos-02-token2022",
      "description": "legit Token-2022 payment (PYUSD): create-ATA derived under Token-2022 (account[5]=Token-2022), account[1] the Token-2022 ATA. Accepts only if the validator derives the destination parametrically on the token program rather than hardcoding SPL Token.",
      "instruction": {
        "program_id": "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
        "data_b64": "AQ==",
        "data_hex": "01",
        "accounts": [
          {
            "pubkey": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
            "is_signer": true,
            "is_writable": true
          },
          {
            "pubkey": "5AJ5PmLBPTY9JkKPSkSdNrghd1iqoWh1U55g3ByVG41y",
            "is_signer": false,
            "is_writable": true
          },
          {
            "pubkey": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "2b1kV6DkPAnxd5ixfnxCpjxmKwqjjaYmCZfHsFu24GXo",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "11111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb",
            "is_signer": false,
            "is_writable": false
          }
        ]
      },
      "expected_verdict": "ACCEPT",
      "failed_assertions": []
    },
    {
      "id": "fixture-neg-04-cross-program-create",
      "description": "transfer is under Token-2022 (PYUSD) but the create-ATA names SPL Token in account[5]. account[1] is left as the correct Token-2022 destination so only the token-program pin fails; on-chain this would rent-fund the SPL-derived ATA (6v1NJ9vKgoG2v7BJXjcZhBtdsMk33bSYsUX51DV3Jw17) instead of the paid Token-2022 ATA (5AJ5PmLBPTY9JkKPSkSdNrghd1iqoWh1U55g3ByVG41y).",
      "instruction": {
        "program_id": "ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL",
        "data_b64": "AQ==",
        "data_hex": "01",
        "accounts": [
          {
            "pubkey": "AKnL4NNf3DGWZJS6cPknBuEGnVsV4A4m5tgebLHaRSZ9",
            "is_signer": true,
            "is_writable": true
          },
          {
            "pubkey": "5AJ5PmLBPTY9JkKPSkSdNrghd1iqoWh1U55g3ByVG41y",
            "is_signer": false,
            "is_writable": true
          },
          {
            "pubkey": "GyGKxMyg1p9SsHfm15MkNUu1u9TN2JtTspcdmrtGUdse",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "2b1kV6DkPAnxd5ixfnxCpjxmKwqjjaYmCZfHsFu24GXo",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "11111111111111111111111111111111",
            "is_signer": false,
            "is_writable": false
          },
          {
            "pubkey": "TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA",
            "is_signer": false,
            "is_writable": false
          }
        ]
      },
      "expected_verdict": "REJECT",
      "failed_assertions": [
        "account[5] == transfer_token_program"
      ]
    }
  ]
}
```
</details>

kakedashi3:
Both fixture sets are now wired into the suite (ca1cdd6d) — that settles the placement question: the JSON lands as `test/fixtures/ata-create-fixtures.json` + `ata-create-fixtures-token2022.json` (SPDX headers and copyright lines kept verbatim), with a data-driven runner at `test/unit/createAtaFixtures.test.ts` alongside `createAta.test.ts`.

**All six fixtures pass against the implementation unchanged** — 212/212 tests green. Each REJECT is additionally asserted against its exact reason code, so the mapping @DrVelvetFog verified (`neg-01 → …not_idempotent`, `neg-02 → …destination_mismatch`, `neg-03 → …funder_mismatch`, `neg-04 → …token_program_mismatch`) is now enforced by CI.

@DrVelvetFog two things taken directly from your byte-verification:
- the runner evaluates every fixture against the **full static-path pin set** (shared validator + the facilitator's two caller pins), with a comment on exactly the trap you flagged — `neg-02` passing the isolated validator is intentional, since `validateCreateAtaIdempotentInstruction` never reads `account[1]` and the reject lives at the caller's `ata == transfer destination` pin;
- `pos-02-token2022`/`neg-04` were exactly the missing coverage — the existing unit tests pinned `account[5]` by equality but nothing drove the derivation seed in both directions. Folded into the set as your separate companion file rather than merged, so the attribution and the PYUSD context note stay self-contained.

And thanks for the operator-side rationale on conditional-vs-always — flag-day coupling across independently updated facilitators is the argument that settles it. I've marked open question 2 resolved in favor of conditional on that basis; question 1 (whether `create_ata_*` rejections should be Path-2-recoverable) remains the one point where maintainer input is still needed before undrafting.

phdargen:
Thanks for the contribution @kakedashi3, unfortunately this approach would introduce a facilitator griefing attack vector as discussed in https://github.com/x402-foundation/x402/issues/1020

kakedashi3:
Thanks @phdargen — understood, and agreed after re-reading #1020. For the record of this thread: the pin set here closes the *redirect* vector (the facilitator can only be made to fund the exact ATA it is being paid into) but not the *reclaim* vector documented in #1020 — the payTo owner can empty and `closeAccount` the very ATA the facilitator just rent-funded, pocket the rent, and repeat with fresh (wallet, mint) pairs. Since buyer and seller can be the same party in that loop, per-payment scoping can't price it out; it's inherent to facilitator-funded creation, no matter how tightly pinned. That prior art was sitting in the issue tracker since January and none of us in the #2395 thread searched for it — that one's on me as the implementer.

Closing this PR in deference. Two threads worth separating out of the wreckage:

1. **The griefing-free subset**: a client-only change that *detects* the missing destination ATA and fails fast with an explicit error (instead of the opaque `InstructionError: [.., InvalidAccountData]` at settle time), pointing the seller at self-provisioning. No facilitator change, nobody sponsors rent. Happy to open that as a small PR if you'd take it.
2. **Recipient self-provisioning as the sanctioned path** — @chopmob-cloud has published a reference + spec in #1020, and @notorious-d-e-v's facilitator-as-merchant endpoint idea also lives there. That's the right venue for it.

@DrVelvetFog @chopmob-cloud — thank you for the review and fixture work here. The validator/fixture methodology holds up even though the funding model doesn't; sorry for building on the wrong foundation.

DrVelvetFog:
No worries at all @kakedashi3 — redirect-vs-reclaim is the right read, and #1020 is prior art any of us could've walked past. The fail-fast client-only detection is the right salvage: fixes the actual UX bug with zero sponsorship surface and composes cleanly with recipient self-provisioning. Happy to take a look if you open it.

kakedashi3:
@DrVelvetFog It's up: #2802 — client-only detection, error names the ATA/payTo/mint and points at self-provisioning, `getAccountInfo` fetched in parallel with the blockhash so no latency is added. Review very welcome.

PR #2801, opened 2026-07-06
feat(xrpl): add exact scheme TypeScript reference implementation

aristotle-satoshi:
## Description

Adds the XRPL `exact` scheme **TypeScript reference implementation** (`@x402/xrpl`), implementing the spec merged in #2547 ([`specs/schemes/exact/scheme_exact_xrpl.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_xrpl.md)). This is the follow-up implementation PR requested in the spec review, structured per `CONTRIBUTING.md` (New Chains) and `.agents/skills/contributing/references/new-mechanism.md`: single-SDK reference implementation, v2-only, builder-pattern registration, no changes to `@x402/core` or other packages.

### What's included

- **`typescript/packages/mechanisms/xrpl`** — client, resource-server, and facilitator schemes for payer-signed XRPL `Payment` transactions:
  - Both asset transfer methods: `sequence` (default) and `ticketSequence` (`Sequence = 0` + `TicketSequence`), with the spec's resolution order (`accepted.extra` → `requirements.extra` → `"sequence"`), mismatch rejection, and per-method verification — current on-network account `Sequence` for `sequence`, validated ticket availability for `ticketSequence` (spec §7). A `createTickets()` utility submits `TicketCreate` and returns the created ticket sequences.
  - Facilitator verification per spec: envelope checks incl. `maxTimeoutSeconds` and `extra.areFeesSponsored=false` (§1), offline signature validation and signer-to-account authorization (§10) — the embedded `SigningPubKey` must be the account's configured regular key or its non-disabled master key pair (validated `account_info`, mirroring rippled's key ordering), with non-canonical signing keys and single-signed blobs carrying a `Signers` array rejected — destination/tag with fail-closed validation of malformed `destinationTag` (§4), NetworkID binding with omit-below-1024 rule (§5), XRP drops and IOU decimal-value amount rules with exact decimal arithmetic and the IOU SendMax policy (§6), `InvoiceID = SHA-256(invoiceId)` invoice binding with Memos rejected (§8), safety rejections — `Delegate`, `Memos`, `Paths`, `DeliverMin`, `tfPartialPayment`, both/neither `Amount`/`DeliverMax`, fee cap (§9), and simulation (§11; signature fields are stripped before `simulate` because the XRPL API only accepts unsigned transactions).
  - Settlement re-runs verification, submits with `failHard`, and succeeds only on a validated `tesSUCCESS` (spec Settlement).
  - The facilitator is keyless (`getSigners()` returns `[]`, `getExtra()` advertises `{ areFeesSponsored: false }`): the payer signs and pays the XRPL fee embedded in the signed transaction.
- **e2e** — full wiring for the `xrpl` family: harness networks/types/env plumbing, axios + fetch clients, express/fastify/hono/next servers (`GET /exact/xrpl`, Next `withX402` route), the TypeScript facilitator, and the mock facilitator.
- **Examples** — `xrpl` added to the client/server/facilitator `advanced/all_networks` examples (alphabetical by network prefix).
- **Release wiring** — `publish_npm_scoped_x402_xrpl.yml`, `publish_npm_scoped_x402_all.yml`, changeset `fixed` group, `prepare/tag-typescript-release.py`, root `test:integration` filter, and a changeset (package versioned at the current fixed-group version 2.17.0).

### Design decisions

- **Client follows the advertised method**: clients use the `assetTransferMethod` pinned in the payment requirements and default to `"sequence"`; resource servers offer `"ticketSequence"` by advertising it (optionally as a second `accepts` entry, as sanctioned by the spec). Core's `PaymentPayloadResult` does not let a scheme write `accepted.extra`, so the client never needs the non-default-method echo path; the facilitator still implements the full three-step resolution for payloads that do declare a method.
- **Deterministic server scheme**: `enhancePaymentRequirements` adds `areFeesSponsored: false` and validates extras, but never injects per-request values (requirements are rebuilt on every request, so a generated `invoiceId` would break `findMatchingRequirements`). Invoice binding activates when the resource configuration provides `extra.invoiceId`.
- **New dependency: `xrpl@^4.6.0`** (first use in the workspace). It provides the binary codec (decode), offline `verifySignature`, `hashSignedTx`, autofill, `simulate`, and `submitAndWait` in one maintained package; using the narrower `ripple-binary-codec`/`ripple-keypairs` would still require reimplementing autofill/submit/simulate plumbing. Ledger reads and submission are injectable for tests.
- No `extra.decimals`: IOU `amount` is the exact XRPL issued-currency `value` string per the merged spec, compared with exact decimal arithmetic.

## Tests

From `typescript/`:

- `pnpm format:check && pnpm lint:check && pnpm build && pnpm test` — pass (Node 22, pnpm 11.1.1).
- `pnpm --filter @x402/xrpl test` — 105 unit tests pass (offline; envelope, per-method sequencing, IOU rules, invoice binding, signer authorization, safety checks, tickets, simulation stripping, settlement outcomes, client/server behavior).
- `pnpm --filter @x402/xrpl test:integration` — passes (mocked-submission settlement round-trip, plus an env-gated live settlement test — `XRPL_PAYER_SEED`/`XRPL_LIVE_PAYTO`, skipped when unset — verified on XRPL Testnet: [`6AF0A4F6…`](https://testnet.xrpl.org/transactions/6AF0A4F630F2B43CA44544D5246B58260A01860A3CCBB331625229CD78696791), [`F8CF2183…`](https://testnet.xrpl.org/transactions/F8CF2183E5B238E164F5E60342986B22654F0ABD52FFE6A70DFAB1A98270A30A)).

E2E against **XRPL Testnet** (`xrpl:1`) with funded faucet accounts, from `e2e/`:

```
pnpm test --testnet --families=xrpl --versions=2
```

All **10/10 scenarios pass** (axios + fetch × express + fastify + hono + next, including the Next `withX402` route), each settling a real validated payment on testnet. Sample settled transactions: [`146F2681…`](https://testnet.xrpl.org/transactions/146F2681420EEBEEAF63E3DD5A595670BC7096D4C1AECC55C0B86D95691B852F), [`353AD2AD…`](https://testnet.xrpl.org/transactions/353AD2ADB36251EBE30E9CE4144C3F44EF4FB26641EF2FF1F18D1D24F519E3B5). Note for local runs: `e2e/servers/next` bundles workspace packages at build time, so rerun `pnpm build` in `e2e/servers/next` (or `./setup.sh`) after changing mechanism packages.

E2E CI wiring (`e2e_tests.yml` secrets / `ci-select-families.sh`) is left to maintainers, following the near/keeta precedent.

## Checklist

## AI assistance

Per the repo's AI-Assisted Contributions policy: this carries forward the XRPL implementation originally developed alongside the spec in #2547 (removed there when that PR became spec-only), updated with AI assistance to the merged spec (`assetTransferMethod`, `areFeesSponsored`, issued-currency `value` amounts) plus the e2e/example wiring. Design choices and spec compliance were reviewed by me, and the full flow was confirmed on XRPL Testnet with the transactions linked above.


aristotle-satoshi:
Thanks @phdargen for consolidating the XRPL work onto this PR.

Status update: rebased onto `main` and all 16 required checks are green (the Vercel check needs team authorization, as with #2547). Since the PR was opened I also tightened facilitator verification — the embedded `SigningPubKey` is now bound to the transaction's `Account` (regular key or non-disabled master key, mirroring rippled's ordering), non-canonical signing keys and stray `Signers` arrays are rejected, and a malformed `destinationTag` in the requirements now fails closed. Unit tests are at 105, plus an env-gated live settlement test against XRPL Testnet.

The full e2e matrix (10/10 scenarios, all client × server combinations) passes on XRPL Testnet, with settled transactions linked in the description.

Ready for review whenever you have a slot.

typescript/packages/mechanisms/xrpl/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,313 @@
...
+   */
+  private async getAvailableTicketSequence(network: Network): Promise<number> {
+    const account = this.signer.classicAddress;
+    const ticketSequence = this.options.getAvailableTicketSequence
+      ? await this.options.getAvailableTicketSequence(account, network)
+      : (await getXrplTicketSequences(account, network, this.options))[0];
+    if (ticketSequence === undefined) {
+      throw new Error(
+        `No available XRPL ticket for ${account}; create tickets with createTickets() before ticketSequence payments`,
+      );
```
phdargen, line 155: How about auto creating tickets when needed? Suggest to have a client policy of how much tickets to create when none is available defaulting to 1, so could be disabled or create more at once depending on client's preference

e2e/servers/express/test.config.json
```diff
@@ -197,6 +197,13 @@
       "requiresPayment": true,
       "protocolFamily": "near"
     },
+    {
+      "path": "/exact/xrpl",
```
phdargen: we will need tests for all assetTransfer here, similar to /exact/evm/eip3009 + /exact/evm/permit2:
```
/exact/xrpl/sequence
/exact/xrpl/ticketSequence
```

typescript/packages/mechanisms/xrpl/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,689 @@
...
+
+      return {
+        isValid: true,
+        payer,
+      };
+    } catch (error) {
+      return invalidVerify(
+        `invalid_exact_xrpl_payload_malformed: ${error instanceof Error ? error.message : String(error)}`,
+        payer,
+      );
```
phdargen, line 172: `invalidReason` should be a stable machine-readable code, while sanitized details go in `invalidMessage` (see EVM mechanism)
```suggestion
      return invalidVerify(
        "invalid_exact_xrpl_facilitator_error",
        payer,
        error instanceof Error ? error.message : String(error),
      );
```

typescript/packages/mechanisms/xrpl/src/exact/facilitator/scheme.ts
```diff
@@ -0,0 +1,689 @@
...
+    if (payload.accepted.extra?.invoiceId !== requirements.extra?.invoiceId) {
+      return "invalid_exact_xrpl_invoice_mismatch";
+    }
+    if (payload.accepted.extra?.destinationTag !== requirements.extra?.destinationTag) {
+      return "invalid_exact_xrpl_destination_tag_mismatch";
+    }
+    if (
+      requirements.asset !== "XRP" &&
+      payload.accepted.extra?.issuer !== requirements.extra?.issuer
+    ) {
```
phdargen: If both requirement and accepted IOU issuers are absent, undefined !== undefined is false and envelope validation passes. Please first require a valid classic issuer, then compare it with accepted.extra.issuer. This should return a stable `invalid_exact_xrpl_iou_issuer_missing` rather than failing later as an issuer mismatch

typescript/packages/mechanisms/xrpl/src/exact/client/scheme.ts
```diff
@@ -0,0 +1,313 @@
...
+    const transaction: Payment = {
+      TransactionType: "Payment",
+      Account: this.signer.classicAddress,
+      Destination: requirements.payTo,
+      Amount: amount,
+      ...(invoiceId !== undefined ? { InvoiceID: invoiceIdToInvoiceIdField(invoiceId) } : {}),
+      ...(this.options.feeDrops !== undefined ? { Fee: this.options.feeDrops } : {}),
+      ...(lastLedgerSequence !== undefined ? { LastLedgerSequence: lastLedgerSequence } : {}),
+      ...(destinationTag !== undefined ? { DestinationTag: destinationTag } : {}),
+    };
```
phdargen, line 122: network ID should be included for custom networks
```suggestion
    const networkId = parseXrplNetworkId(requirements.network);
    const transaction: Payment = {
      TransactionType: "Payment",
         ...(networkId > 1024 ? { NetworkID: networkId } : {}),
      Account: this.signer.classicAddress,
      Destination: requirements.payTo,
      Amount: amount,
      ...(invoiceId !== undefined ? { InvoiceID: invoiceIdToInvoiceIdField(invoiceId) } : {}),
      ...(this.options.feeDrops !== undefined ? { Fee: this.options.feeDrops } : {}),
      ...(lastLedgerSequence !== undefined ? { LastLedgerSequence: lastLedgerSequence } : {}),
      ...(destinationTag !== undefined ? { DestinationTag: destinationTag } : {}),
    };
```

typescript/packages/mechanisms/xrpl/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,164 @@
...
+        throw new Error(`Invalid money format: ${money}`);
+      }
+      return money;
+    }
+
+    const cleanMoney = money.trim().replace(/^\$/, "");
+    if (!/^\d+(\.\d+)?$/.test(cleanMoney)) {
+      throw new Error(`Invalid money format: ${money}`);
+    }
+    return Number(cleanMoney);
```
phdargen: Please reuse core utils where appropriate, eg  `import { parseMoneyString } from "@x402/core/utils";`
```suggestion
 return parseMoneyString(money);
```

typescript/packages/mechanisms/xrpl/src/exact/server/scheme.ts
```diff
@@ -0,0 +1,164 @@
...
+  isIntegerString,
+  isValidDestinationTag,
+  isXrplAssetTransferMethod,
+  requireClassicAddress,
+} from "../../utils";
+
+/**
+ * XRPL server implementation for the exact payment scheme.
+ */
+export class ExactXrplScheme implements SchemeNetworkServer {
```
phdargen, line 23: Would recommend to define a default USD stablecoin for the moneyParser to resolve bare money values like "$0.1", see eg SVM:
https://github.com/x402-foundation/x402/blob/8b1abaeaef282e6307a2936b102c6d9223e61802/typescript/packages/mechanisms/svm/src/exact/server/scheme.ts#L131-L141

But if you don't want to pick a default for your ecosystem thats fine

phdargen:
Thanks for the contribution @aristotle-satoshi, made a first pass and looks great! Just a couple of nits that should be resolved quickly. 

I also tested e2e and examples successfully. Two more things I'd like to test are:
1.  non-native token payments (is there any faucet available?) 
2. ticketSequence payments (should be added to e2e, see comment above)

phdargen:
We would also be happy to include XRPL in the x402.org testnet facilitator, provided your team/Ripple could supply sufficient testnet funds.

You may add XRPL as an optional network here: https://github.com/x402-foundation/x402/blob/main/typescript/site/app/facilitator/index.ts
could also be done as a follow-up

typescript/packages/mechanisms/xrpl/src/exact/client/scheme.ts
aristotle-satoshi: Implemented in a656536e. The client now has a `ticketCreateCount` policy option exactly as suggested: when a `ticketSequence` payment finds no available ticket, it auto-creates tickets via a `TicketCreate` signed through the `ClientXrplSigner` abstraction. It defaults to 1, supports up to 250 per `TicketCreate` (the ledger's per-account cap), and `0` disables auto-creation for pre-provisioned setups. Available tickets are still reused first and skip the creation transaction entirely.

One note: this required widening `ClientXrplSigner.sign` from `Payment` to `SubmittableTransaction`, so custom signers can now receive a `TicketCreate`. Unit tests cover the default/configured/disabled policies and invalid counts.

e2e/servers/express/test.config.json
aristotle-satoshi: Added in 67e7313b, mirroring the EVM pattern: explicit `/exact/xrpl/sequence` and `/exact/xrpl/ticketSequence` endpoints on Express, Fastify, Hono, and Next (both the proxy middleware and `withX402`), each pinning `extra.assetTransferMethod` — 10 distinct XRPL endpoints in the minimized matrix.

Verified live on XRPL Testnet with issued-currency pricing (via the fixture from d4be4487): 10/10 scenarios passed in one combined run (Next 4/4, Express 2/2, Hono 2/2, Fastify 2/2), covering `sequence`, `ticketSequence`, axios, fetch, Next proxy, and `withX402`.

typescript/packages/mechanisms/xrpl/src/exact/facilitator/scheme.ts
aristotle-satoshi: Applied as suggested in df002369: the `verify()` catch-all now returns the stable `invalid_exact_xrpl_facilitator_error` code, with the raw diagnostic carried separately in `invalidMessage`. Added a unit test asserting the stable code plus a non-empty `invalidMessage`.

typescript/packages/mechanisms/xrpl/src/exact/facilitator/scheme.ts
aristotle-satoshi: Fixed in a2d8d365 exactly as described: both the requirement and accepted issuers now go through classic-address validation before the comparison, returning the stable `invalid_exact_xrpl_iou_issuer_missing` when either is absent or malformed; only two valid issuers reach the `invalid_exact_xrpl_iou_issuer_mismatch` comparison. Unit tests cover both-missing, invalid on either side, and valid-but-mismatched issuers.

typescript/packages/mechanisms/xrpl/src/exact/client/scheme.ts
aristotle-satoshi: Applied in 62a00389: the client now inserts `NetworkID` while building the payment for custom networks (`networkId > 1024`), per the suggestion. The existing prepared-transaction validation (client and facilitator) still rejects preparers that drop or replace it — both directions are now covered by client-side unit tests.

typescript/packages/mechanisms/xrpl/src/exact/server/scheme.ts
aristotle-satoshi: Applied as suggested in ba7b1ce9 — the scheme now reuses `parseMoneyString` from `@x402/core/utils`.

typescript/packages/mechanisms/xrpl/src/exact/server/scheme.ts
aristotle-satoshi: We'd prefer not to pick an ecosystem default in this PR: unlike SVM/EVM, where the SDK can map each network to a canonical USDC deployment, XRPL issued currencies are identified by (currency, issuer), and there is no issuer the SDK can safely assume across Mainnet, Testnet, and custom networks. Explicit `AssetAmount` pricing works today, and applications can register a `MoneyParser` when they want a network-specific RLUSD policy.

Happy to revisit an RLUSD default in a follow-up if a canonical per-network issuer mapping makes sense for the ecosystem.

aristotle-satoshi:
Thanks for the review @phdargen! All seven inline threads are addressed, one commit per topic:

- a656536e — automatic ticket creation policy (`ticketCreateCount`: default 1, up to 250, `0` disables)
- df002369 — stable `invalid_exact_xrpl_facilitator_error` code, diagnostics moved to `invalidMessage`
- a2d8d365 — fail-closed IOU issuer validation (`invalid_exact_xrpl_iou_issuer_missing`)
- 62a00389 — client inserts the custom-network `NetworkID` while building payments
- ba7b1ce9 — reuse `parseMoneyString` from `@x402/core/utils`
- 67e7313b — e2e coverage for both asset-transfer methods on every server (10 XRPL endpoints)
- d4be4487 — self-managed XRPL Testnet IOU fixture for real issued-currency e2e

On your two follow-up points:

1. **Non-native token payments** — there's no public IOU faucet on XRPL Testnet (the faucet only funds XRP), so d4be4487 adds a self-managed fixture instead: `pnpm xrpl:iou:setup` creates or reuses a dedicated Testnet issuer plus the payer/payee accounts, funds them through the XRP faucet, enables `DefaultRipple`, sets up the trust lines, and issues a fixture-only test `USD` (credentials persist to the git-ignored `e2e/.env`; the currency is explicitly not an SDK default). The full XRPL e2e matrix then runs against that issued currency — 10/10 scenarios passed on Testnet, with on-ledger balances matching (1 test USD per payment). `ticketSequence` e2e is covered per the thread reply above.

2. **x402.org testnet facilitator** — that would be great, and a follow-up after this PR lands sounds right. Testnet funding won't be a blocker on our side; once you're ready, let us know where to send the Testnet XRP and we'll open the follow-up PR adding XRPL as an optional network in `typescript/site/app/facilitator/index.ts`.

Unit tests are at 116 (from 105), the simulated settlement integration tests pass, and CI is green at d4be4487. Ready for another pass whenever you have a slot.

e2e/scripts/xrpl-iou-setup.ts
```diff
@@ -0,0 +1,393 @@
+/**
+ * Creates a reusable XRPL Testnet issued-currency fixture for e2e tests.
```
phdargen, line 2: This is great, thank you! Wish we had this for all the networks

phdargen:
Thanks for the update @aristotle-satoshi, the changes look all great!

Unfortunately, there is still a security vulnerability we will have to patch. 
The facilitator has a duplicate-settlement vulnerability: concurrent /settle calls with the same signed blob all pass verification and each submitAndWait resolves to the same validated tesSUCCESS for the single onchain tx hash, so the facilitator returns success: true to every caller. XRPL dedups the ledger effect (one payment lands), but submitAndWait is an idempotent read on a hash and gives no submitter-exclusive result. A malicious client could thus gain access to the server resource N times, while only paying once.

This needs a facilitator-side dedup guard keyed on the signed tx hash, mirroring the SVM SettlementCache mitigation.

Could you please amend the spec with a dedicated section similar to SVM:
https://github.com/x402-foundation/x402/blob/ac7a51b62ff706270303a84459ed49cc158e5493/specs/schemes/exact/scheme_exact_svm.md?plain=1#L345-L362

and then add a settlement cache following the SVM implementation here:
https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/svm/src/settlement-cache.ts

You can find a script I used for testing here: https://github.com/phdargen/x402/blob/xrl-vul/examples/typescript/clients/advanced/xrpl_race.ts with output:
```
Building and signing payment transaction once...
Firing 10 parallel requests with the same payment header...
Results:
--------
  Request 0: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 1: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 2: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 3: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 4: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 5: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 6: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 7: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 8: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}
  Request 9: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"7982B432BDDAD03AEE86EF313D62DEBD0DDBA6C8C09E357B28B30489376017B4","network":"xrpl:1"}

Time: 19868ms

VULNERABLE: 10/10 requests succeeded with the same payment.
```
After the fix, this should only succeed once, with the others failing with a `"duplicate_settlement"` error

aristotle-satoshi:
Thanks for catching this @phdargen — confirmed and fixed in `3bce520e`. You're right that `submitAndWait` is an idempotent read on the tx hash: with `failHard: true` it still *resolves* (not rejects) `tesSUCCESS` for an already-submitted blob, so every concurrent `/settle` sharing one signed blob was returning `success: true`.

The fix follows the SVM `SettlementCache` (`settlement-cache.ts`, wired into `settle()` synchronously between re-verification and submission — the winner submits, the rest get `errorReason: "duplicate_settlement"`), plus a new `Duplicate Settlement Mitigation (REQUIRED)` section in the spec. I keyed on the canonical **transaction hash** rather than the raw blob so it can't be dodged by re-encoding the same tx.

Two deliberate departures from the SVM version, both because XRPL's expiry model differs from Solana's:

1. **Retention window is per-entry, not a flat 120s.** SVM's 120s is safe because Solana's ~60–90s blockhash lifetime *protocol-bounds* the confirmation window. XRPL has no such bound: a tx stays landable until its `LastLedgerSequence`, which the scheme derives from `maxTimeoutSeconds` (uncapped, and every example here uses `600`). A flat 120s TTL would evict the entry while the tx is still landable, and a slow-to-validate duplicate (e.g. a minimum-fee tx during fee escalation) would re-pass verification and re-submit — reopening the exact race. So each entry is retained for `maxTimeoutSeconds * 1000 + 120s`, which covers the `LastLedgerSequence` horizon with margin.

2. **Spec scopes the guarantee to per-process and requires a shared store when scaled.** An in-process map only serializes duplicates within one process; two `/settle` calls hitting different replicas would each miss their local cache. The spec now requires atomic dedup across every process serving `/settle`, with the in-process map as the single-process reference implementation and a shared atomic store required for horizontally scaled facilitators. The README calls out the same boundary.

One behavioral note documented in the spec: an entry isn't cleared on a transient settlement failure, so a same-blob retry within the window is also rejected as `duplicate_settlement` — i.e. that error means "already seen," not "settled." Clearing on failure would reopen the window (the tx may still land), so this matches SVM's choice.

New unit tests cover: second settlement of the same blob rejected; **10 concurrent settles → exactly one success, nine `duplicate_settlement`** (your `xrpl_race.ts` scenario in-process); distinct transactions not blocked; dedup across two scheme instances sharing a cache; per-entry TTL honored (a duplicate at T+120s is still rejected while the tx is landable, evicted only after the window); and cross-instance blocking. XRPL package is green at 122 unit tests, and I rebased onto `main` so it's mergeable again.

Would you be able to re-run your `xrpl_race.ts` against `3bce520e`? It should now show 1/10 succeeding.

---

On the x402.org testnet facilitator inclusion — still happy to do that as a follow-up once this lands; testnet funding on our side isn't a blocker.

phdargen:
Thanks for the quick turnaround @aristotle-satoshi, changes look good and agree with the minor departures wrt SVM.
Confirmed with the script that the duplicates are blocked correctly:
```
Firing 10 parallel requests with the same payment header...
Results:
--------
  Request 0: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 1: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 2: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 3: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 4: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 5: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 6: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 7: HTTP 200 - {"report":{"weather":"sunny","temperature":70}} | payment: {"success":true,"payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"9611AB4EF77932920CA24C02DADDF9D90CB8080D3CADD8D6D53ED4FAF4AB0729","network":"xrpl:1"}
  Request 8: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}
  Request 9: HTTP 402 - {} | payment: {"success":false,"errorReason":"duplicate_settlement","payer":"rEufzrJvf5ZqcH2fNnzjugMc8SPa4Ln6Tx","transaction":"","network":"xrpl:1"}

Time: 21073ms

PROTECTED: Only 1/10 requests succeeded. 9 were correctly rejected.
```

phdargen (approved):
Happy to get this in @aristotle-satoshi, great work!

aristotle-satoshi:
> You may add XRPL as an optional network here: https://github.com/x402-foundation/x402/blob/main/typescript/site/app/facilitator/index.ts — could also be done as a follow-up

@phdargen The site facilitator registration is up in #2857 — `xrpl:1` behind an explicit `FACILITATOR_XRPL_ENABLED=true` opt-in. On the testnet-funds condition: enabling it needs no facilitator-side key or funds (the payer signs and pays fees; `getSigners()` is `[]`) — details in the PR description.

PR #613, opened 2025-11-07
Refactor SVM client and verification logic to remove create ATA instruction

notorious-d-e-v:
## Description

Refactor SVM client and verification logic to remove create ATA instruction handling

- Removed create associated token account (ATA) instruction logic from SVM client and tests.
- Updated transfer instruction handling to assume destination ATA exists.
- Adjusted verification functions to reflect changes in instruction expectations.
- Cleaned up related test cases to remove references to create ATA instructions.

## Tests

Updated existing tests.

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits

PR #2802, opened 2026-07-06
fix(svm): fail fast with a clear error when the destination ATA does not exist

kakedashi3:
Follow-up to #2798 / #2395, as offered there — the griefing-free subset.

## Problem

The first payment to a `payTo`/mint whose ATA has never been created fails at settle time with an opaque `InstructionError: [.., InvalidAccountData]`. Neither the buyer nor the seller learns what is wrong or who can fix it.

## Change (client-only, v1 + v2)

Check the destination ATA in `createPaymentPayload` and throw an explicit error when it does not exist:

```
Destination associated token account <ata> does not exist for payTo <payTo> and mint <mint> on <network>;
the transfer would fail on-chain with InvalidAccountData. The recipient must provision their own ATA
before they can accept payments for this asset (see x402-foundation/x402#1020).
```

- **No latency added**: the `getAccountInfo` is fetched via `Promise.all` alongside the existing `getLatestBlockhash`.
- **No instruction layout change, nobody sponsors rent**: this deliberately stays on the right side of the line drawn in #1020 / #613 (in-band creation removed; destination ATA is assumed to exist) — it just makes the failure of that assumption legible and points at the sanctioned fix (recipient self-provisioning).
- Repeat payments are byte-identical to today.

## Tests

4 new unit tests (v1/v2 × missing/existing ATA); 198/198 pass, lint clean, changeset (patch) included.

AI disclosure: implemented with AI assistance (Claude); reviewed and tested locally by me.


DrVelvetFog:
Took a look — this is the right shape, and it correctly handles the one thing that could've broken it. `destinationATA` is derived under the mint's actual token program (from `getCachedMintMetadata().programAddress`), so the existence check inspects the right address for both SPL and Token-2022 — the subtlety that would've quietly made it check the wrong ATA. Parallelizing `getAccountInfo` with `getLatestBlockhash` keeps it latency-free, and staying read-only (no instruction, no sponsored rent, repeat payments byte-identical) keeps it on the right side of #1020. The error naming ata/payTo/mint/network and pointing at self-provisioning is exactly the UX the opaque `InvalidAccountData` was missing.

One refinement, since it's the same fail-fast goal: `if (!destinationInfo.value)` only fires when *nothing* exists at the ATA. It won't catch an account existing there that isn't a usable token account — most concretely a lamport-only System account squatting the ATA address (the known ATA-blocking vector on Solana: anyone can prefund the PDA, making it System-owned). That passes the existence check but still fails `TransferChecked` at settle — the exact opaque failure this PR is making legible. Tightening the guard to also require `destinationInfo.value.owner === tokenProgramAddress` closes it, and doubles as the honest signal for the "someone blocked my ATA" case. Cheap to add; take it or leave it.

Otherwise this reads correct and well-scoped — fast turnaround on the salvage.

kakedashi3:
Taken — good catch on the squatted-PDA case. The guard now also requires `destinationInfo.value.owner === tokenProgramAddress` (the mint's actual program, so Token-2022 compares against the right owner too), with a distinct error that names the squatting owner and notes that the recipient's own idempotent create converts a prefunded system account into a usable token account. New unit test covers a System-owned squatter; 199/199 green. Thanks @DrVelvetFog.

DrVelvetFog:
Confirmed, the guard reads right. Nice that it does double duty too — keying on `owner !== tokenProgramAddress` also catches a token-program *mismatch* (an SPL-owned account where the mint's Token-2022, say), not just the System squatter, so "exists but unusable" is closed either way. 👍

PR #2810, opened 2026-07-07
fix(python/flask): use sync server in payment_middleware_from_config

kakedashi3:
## What

`flask_payment_middleware_from_config` (the Flask convenience factory) instantiates the **async** `x402ResourceServer` and passes it to `PaymentMiddleware`, which builds `x402HTTPResourceServerSync`. The sync HTTP server rejects a server whose `verify_payment` is a coroutine:

```python
# x402_http_server.py (x402HTTPResourceServerSync.__init__)
if verify_method and inspect.iscoroutinefunction(verify_method):
    raise TypeError("x402HTTPResourceServerSync requires a sync server ... "
                    "Use x402ResourceServerSync instead of x402ResourceServer ...")
```

So the factory raises `TypeError` at construction and is unusable — it fails 100% of the time.

## Fix

Use `x402ResourceServerSync` in the Flask factory, mirroring how the FastAPI factory correctly uses the async `x402ResourceServer` for its async `x402HTTPResourceServer`. One-line import/instantiation change; `register()` is defined on the shared base so scheme registration is unaffected.

The package's own `__init__.py` docstring already shows the correct sync usage (`server = x402ResourceServerSync(facilitator_client)`), confirming intent.

## Tests

Added `test_payment_middleware_from_config_builds_with_sync_server` — constructs via the factory and asserts it does not raise. Previously this raised `TypeError`; now passes. Full flask middleware suite: 40 passed. ruff clean.

AI disclosure: found via a systematic cross-SDK scan and implemented with AI assistance (Claude); reviewed and tested locally by me.

### DIFF
--- a/python/x402/changelog.d/2810.bugfix.md
+++ b/python/x402/changelog.d/2810.bugfix.md
@@ -1,0 +1,1 @@
+Fixed `flask_payment_middleware_from_config`, which raised `TypeError` at construction because it built the async `x402ResourceServer` for the sync Flask middleware; it now uses `x402ResourceServerSync`.
--- a/python/x402/http/middleware/flask.py
+++ b/python/x402/http/middleware/flask.py
@@ -555,9 +555,13 @@ def payment_middleware_from_config(
     Returns:
         PaymentMiddleware instance.
     """
-    from ...server import x402ResourceServer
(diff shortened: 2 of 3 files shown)


phdargen (approved):
Thanks for the contribution @kakedashi3, LGTM 🚀

Issue #2832, opened 2026-07-10
CDP facilitator `/verify` rejects spec-compliant x402 v2 `paymentPayload` (400 invalid_request) — regression since ~2026-07-08


meloliva14:
> **UPDATE (2026-07-12) — root cause isolated:** CDP's `x402V2PaymentPayload` schema rejects any payload whose **`resource.description` exceeds 500 characters** (character count, not bytes; `'a'*500` passes, `'a'*501` fails). `extensions` — including full bazaar blobs — and all other `resource` fields are accepted. **Seller fix without any library change: trim your resource description to ≤ 500 chars.** Full bisection matrix in [this comment](https://github.com/x402-foundation/x402/issues/2832#issuecomment-4949690187). Remaining asks for CDP: document the cap, and surface per-field errors instead of the misleading union-mismatch message below.

**Summary**
Since ~2026-07-08 the Coinbase CDP hosted facilitator's `/verify` (and `/settle`) endpoint rejects **spec-compliant** x402 v2 payment payloads with:

```
Facilitator verify failed (400): {
  "errorType": "invalid_request",
  "errorMessage": "'paymentPayload' is invalid: must match one of [x402V2PaymentPayload, x402V1PaymentPayload]. x402V1PaymentPayload requires 'scheme'",
  "errorLink": "https://docs.cdp.coinbase.com/api-reference/v2/errors#invalid-request"
}
```

This breaks **all** mainnet payments for services using the x402 Python library + the CDP facilitator. The exact same code + library version settled successfully through 2026-07-08, then every paid call began failing — with no change on our side.

**Environment**
- `x402` (python): **2.15.0** (also reproduced on 2.14.0)
- `cdp-sdk`: 1.47.1
- Network: Base mainnet (`eip155:8453`), USDC
- Facilitator: CDP hosted, via `cdp.x402.create_facilitator_config(...)`

**What the library sends**
[...]

PR #2837, opened 2026-07-12
feat(python): opt-in to omit optional PaymentPayload fields for strict facilitators

meloliva14:
## What

Adds an opt-in `omit_optional_payload_fields` flag to `FacilitatorConfig`. When enabled, the client strips the optional `resource` and `extensions` fields from the `PaymentPayload` before POSTing to `/verify` and `/settle`. Off by default — the full spec-compliant payload is sent unless you turn it on.

## Why

Some hosted facilitators validate the posted `paymentPayload` against a strict schema and reject requests that include the optional `resource`/`extensions` siblings, returning a 4xx that fails verify/settle even though the payment itself is valid. Context and a reproducible trace are in #2832.

This is safe because the signature is over the scheme's typed data (for `exact`/EVM, the EIP-3009 `transferWithAuthorization` domain), not the payload envelope — so dropping `resource`/`extensions` doesn't change what was signed or what the facilitator verifies.

## How

- New `FacilitatorConfig.omit_optional_payload_fields: bool = False`, plus a module constant `OPTIONAL_PAYMENT_PAYLOAD_FIELDS = ("resource", "extensions")`.
- Applied at the single `_build_request_body` chokepoint, so it covers both the async (`HTTPFacilitatorClient`) and sync (`HTTPFacilitatorClientSync`) clients, for both verify and settle.
- Also wired through the `dict`-style config path.
- No behavior change unless a caller opts in, and a no-op against facilitators that already accept the full payload.

## Tests

- `tests/unit/http/test_facilitator_client.py`: default keeps `resource`/`extensions`; the flag drops them while leaving `x402Version`/`payload`/`accepted` intact; covers async verify and sync settle.
- `uv run pytest tests/unit/http/test_facilitator_client.py` is green; `ruff check` / `ruff format` clean.

## Notes

- Categorized as a `feature` changelog fragment since it adds an opt-in option rather than changing default behavior — happy to recategorize as `bugfix` if you'd prefer.
- Fragment is named `2832.feature.md` after the linked issue; glad to rename to the PR number.
- The flag name and shape are flexible — if you'd rather this be an allowlist (e.g. `facilitator_payload_fields`) or live somewhere else, I'll adjust to match your design.


novadyne-hq:
Confirming this from production, in case it's useful for the review: we've been running the same drop-optional-siblings behavior (omitting `resource`/`extensions`) against the CDP hosted facilitator on live Base mainnet payments for a while, and it holds — `/verify` and `/settle` both return 200 and settlement is unchanged. Your "Why" section captures the reason exactly: the EIP-3009 `transferWithAuthorization` signature is over the typed-data domain, not the payload envelope, so dropping those two fields doesn't re-sign or invalidate anything.

One field note for anyone testing the flag: in our experience the strict-facilitator rejection surfaces at `/verify`, which is free and gasless — so you can validate the opt-in end-to-end without moving any USDC, which is handy for a pre-flight/canary check.

Agree with keeping it `Refs` rather than `Fixes` — the durable fix is still CDP's validation to settle — but this unblocks sellers in the meantime, which is what matters today. Thanks for turning the subclass into a first-class option.

meloliva14:
Appreciate you backing this with the production data. An independent seller running the same drop-optional-siblings shape on live mainnet, with `/verify` and `/settle` staying green, is exactly the corroboration a reviewer can lean on.

And +1 on the `/verify`-only point. That's the free, gasless probe I've been running as a canary too (unfunded key, so it never settles or skews metrics), which turns a silent facilitator-shape regression into a same-interval alarm without touching any USDC. Thanks again for the production read.

meloliva14:
Update: the root cause behind #2832 turned out narrower than this PR's original framing — CDP's schema rejects `resource.description` > 500 characters, nothing else (full bisection: https://github.com/x402-foundation/x402/issues/2832#issuecomment-4949690187). Affected sellers can fix without any library change by trimming their description.

Leaving this PR up as opt-in interop hardening — a client-side switch that keeps verify/settle working against any facilitator that's stricter than the spec, whatever the specific strictness — but with the framing correction on the record. If maintainers feel the narrow cause makes this unnecessary, no hard feelings closing it; the alternative worth considering in that case is better error surfacing when a facilitator rejects the payload envelope.

phdargen:
Thanks for the contribution and investigation @meloliva14. The root cause should be fixed instead of stripping spec compliant wire fields client side. The CDP team is aware and will implement better error surfacing soon

meloliva14:
Completely agree — the durable fix is CDP-side, and honestly the better error surfacing is the highest-leverage part (that union-mismatch message is what had the whole thread chasing the wrong cause for days). Glad the team's on it.

We took the same line on our end, for the record — trimmed the offending field back under the cap so we're spec-compliant with no stripping; the PR was only ever to spare other Python sellers the hand-rolling in the interim. Seller fix for anyone landing here from #2832: keep `resource.description` ≤ 500 chars, no library change. Thanks for the thoughtful review, @phdargen.

PR #2842, opened 2026-07-12
feat(x402-v2): Robinhood facilitator compliance, hardening, and CI

peaceofheaven777:
.


AmitabhainArunachala:
## Security review — from a live x402 facilitator operator

We run the **Agent Payment Receipt & Mandate Verifier** at https://x402.167-172-95-184.nip.io (10 paid routes on Base mainnet via CDP facilitator, 19 settled payments). Reviewed `payment.ts` + `index.ts` against our 22-check fail-closed mandate verification matrix. The hardening work (fail-fast key check, payTo pinning, idempotency cache, gas cap, structured logging) is solid. Four findings:

**1. No resource binding — payment is portable across resources**
`verifyEIP3009` verifies signature, nonce, balance, amount, payTo, and time window — but never checks that the payment is bound to the specific resource the buyer is purchasing. A valid payment for `/price` ($0.01) can be replayed against `/audit/x402` ($1.00) because the facilitator accepts any signed authorization whose amount ≥ the route price. The x402 spec expects the resource URL to be part of the payment binding.

*Fix:* include the resource URL (or a hash of it) in the EIP-712 typed data, or reject settlements where `requirements.resource` doesn't match a server-side allowlist for the signed amount.

**2. Amount check is conditional — overpayment is accepted**
```typescript
if (requirements?.amount != null) {
    const required = toBig(requirements.amount);
    if (a.value !== required) return { isValid: false, ... };
}
```
If `requirements.amount` is omitted (which is legal per the v2 wire format when the amount is implied by the resource), the facilitator will settle whatever `value` the payer signed. A payer who accidentally signs `$100.00` for a `$0.01` resource pays $100.00.

*Fix:* make `requirements.amount` mandatory in `/settle`, or derive it from a server-side price table keyed by resource URL, never from the client-supplied requirements.

**3. `validBefore` checked against wall clock but not against challenge TTL**
The time window check uses `now = Math.floor(Date.now() / 1000)`. This is correct for preventing expired authorizations. But there is no check that `validBefore` falls within the `maxTimeoutSeconds` window of the original 402 challenge. A payer could sign a `validBefore` far in the future, allowing the authorization to be settled long after the resource was delivered — a deferred-settlement attack vector if the facilitator's settle queue backs up.

*Fix:* reject if `validBefore > now + maxTimeoutSeconds` from the original challenge.

**4. `.v1.bak` files committed**
`server.ts.v1.bak` (1031 lines) and `index.ts.v1.bak` (99 lines) are in the PR. These expose the pre-v2 wire format and shouldn't ship. Remove or `.gitignore` them.

**Interop offer:** if you produce a sample `paymentPayload` + `paymentRequirements` fixture, I'll run it through our `/mandate/verify` endpoint (https://x402.167-172-95-184.nip.io/mandate/verify) and post the full 22-check verdict — binding, authority, scope, spend cap, recipient, expiry, nonce, resource binding, and receipt obligations — as a public interop datapoint.

We are not counting self-pay or proof-loop transactions as revenue. This is an operational peer review from a live x402 seller, not a sales pitch.

PR #2843, opened 2026-07-12
fix(core): return 400 for undecodable PAYMENT-SIGNATURE header

### LINKED ISSUE #2397: Malformed PAYMENT-SIGNATURE should return 400 instead of 402

kakedashi3:
Fixes #2397.

`extractPayment()` swallowed a `PAYMENT-SIGNATURE` decode failure with a `console.warn` and returned `null`, so a request carrying a corrupt payment header took the same path as one carrying no header at all and got `402 Payment Required`. A client cannot tell "you didn't pay" apart from "your payment header is broken", and retries the same broken header.

`specs/transports-v2/http.md` maps this case to 400:

| x402 Error      | HTTP Status | Description                               |
| --------------- | ----------- | ----------------------------------------- |
| Invalid Payment | 400         | Malformed payment payload or requirements |

A present-but-undecodable header now returns `400` with `{ "error": "invalid_payload" }` — the spec's error name for a malformed payload — and no facilitator `verify` call. An absent header is unchanged: still `402` with `PAYMENT-REQUIRED`.

Both decode failure paths are covered by tests: a non-base64 header, and a base64 header that isn't valid JSON.

Prior art: #2439 by @peterxing took the same shape (throw from `extractPayment`, catch at the call site). This keeps that shape, drops the now-redundant `console.warn`, and reuses the spec error name instead of introducing a new one.

AI-assisted: implementation and tests were drafted with an AI coding agent and reviewed before opening.

### DIFF
--- a/typescript/.changeset/malformed-payment-signature-400.md
+++ b/typescript/.changeset/malformed-payment-signature-400.md
@@ -1,0 +1,5 @@
+---
+"@x402/core": patch
+---
+
+Return `400` instead of `402` when a request carries a `PAYMENT-SIGNATURE` header that cannot be decoded. The header was previously treated as absent (the decode error was swallowed with a `console.warn`), so a client that sent a corrupt payment header got the same `402 Payment Required` as a client that sent none, and could not tell the two apart. The HTTP transport spec maps a malformed payment payload to `400`, so a present-but-undecodable header now yields `400` with body `{ "error": "invalid_payload" }` and no facilitator `verify` call.
--- a/typescript/packages/core/src/http/x402HTTPResourceServer.ts
+++ b/typescript/packages/core/src/http/x402HTTPResourceServer.ts
@@ -519,7 +519,20 @@ export class x402HTTPResourceServer {
     const paymentOptions = this.normalizePaymentOptions(routeConfig);
 
     // Check for payment header (v1 or v2)
-    const paymentPayload = this.extractPayment(adapter);
+    let paymentPayload: PaymentPayload | null;
+    try {
+      paymentPayload = this.extractPayment(adapter);
+    } catch {
+      // A present-but-undecodable header is a client error, not a missing payment.
+      return {
+        type: "payment-error",
+        response: {
+          status: 400,
+          headers: { "Content-Type": "application/json" },
+          body: { error: "invalid_payload" },
+        },
+      };
+    }
 
     // Create resource info, using config override if provided
     const resourceInfo = {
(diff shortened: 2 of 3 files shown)


AmitabhainArunachala:
Reviewed from a live x402 facilitator operator perspective — we run the Agent Payment Receipt & Mandate Verifier at https://x402.167-172-95-184.nip.io (10 paid routes on Base mainnet via CDP facilitator).

**This fix is correct and spec-compliant.** The spec table maps `Invalid Payment → 400 → Malformed payment payload`, and swallowing the decode error into a `console.warn` + `null` return was a real bug: it made a corrupt-header retry indistinguishable from a no-header first attempt.

**Operational confirmation:** we've seen external wallet-equipped agents hit our `/research/{topic}` and `/price/{coin}` routes with malformed `PAYMENT-SIGNATURE` headers (truncated base64, missing separator, wrong envelope version). Under the current behavior they get `402 Payment Required` back and retry the same broken header — a tight loop that burns their rate limit and never converges. Returning `400 { "error": "invalid_payload" }` gives the agent enough signal to inspect the header instead of re-sending it.

**One observation on the error body:** the spec says `400` for malformed payload but doesn't prescribe the body shape. `{ "error": "invalid_payload" }` is reasonable, but a slightly richer body would help agent-side debugging:

```json
{ "error": "invalid_payload", "header": "payment-signature", "reason": "base64_decode_failed" }
```

This is optional — the current body is sufficient to break the retry loop. Just noting it as a possible follow-up.

**Test coverage looks solid** — both the non-base64 case and the base64-but-not-JSON case are covered, and the `mockFacilitator.verifyCalls.length === 0` assertion confirms no wasted verify call.

LGTM. The commit is signed, the changeset is scoped, and the fix is minimal.

PR #2839, opened 2026-07-12
docs: add TWZRD to third-party extensions

twzrd-sol:
Adds a row for TWZRD to `docs/dev-tools/third-party-extensions.md`, following the pattern of the existing vendor entries (#1047, #1616).

**What it is:** [twzrd-x402-gate](https://www.npmjs.com/package/twzrd-x402-gate) is a buyer-side counterparty trust preflight for x402 clients. It registers on the official client's `onBeforePaymentCreation` lifecycle hook (or wraps fetch for clients that compose over it) and returns allow/warn/block on the selected payment requirement - including wash-flag refusal for flagged merchants - before any payload is constructed or signed. Zero runtime dependencies, MIT.

Docs-only change; no changelog fragment per CONTRIBUTING.md.

Disclosure: this PR was prepared with AI assistance and human-reviewed; the linked package and its hook integration against `x402Client.onBeforePaymentCreation` are shipped and tested.

PR #1616, opened 2026-03-15
feat(ecosystem): add OMATrust entry to follow up on offer-receipt extension

alftom:
Here you go:

---

## Description

Adds OMATrust as an ecosystem partner. OMATrust is a decentralized reputation layer (built by OMA3) that uses x402 signed offers and receipts as cryptographic proof-of-interaction in on-chain EAS attestations.

This PR adds:
- `partners-data/omatrust-offer-receipt/metadata.json` — single ecosystem entry under "Infrastructure & Tooling"
- `public/logos/oma3-icon.png` — OMA3 logo (256x256 PNG)

The entry links to the OMATrust developer docs, which cover both resource server integration (enabling offer-receipt signing) and client integration (submitting receipts as attestation proofs).

More context: https://docs.omatrust.org/integrations/x402/overview

## Tests

No code changes — this is a data-only addition (one JSON metadata file and one PNG). Verified the JSON structure matches existing entries (e.g., `cdp/metadata.json`). Logo renders at 120x120 on the ecosystem page; 256x256 source is sufficient.

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge)
- [x] I added a changelog fragment for user-facing changes — N/A, data-only ecosystem entry

PR #2566, opened 2026-06-06
Add TWZRD Agent Intel — Solana x402 trust layer MCP server

twzrd-sol:
TWZRD Agent Intel is a Solana-native MCP server providing trust scoring and x402 receipt generation for AI agents. Live at https://intel.twzrd.xyz — free preflight checks + paid signed V5 trust receipts via Streamable-HTTP.

This adds it to the third-party extensions listing under `docs/dev-tools/third-party-extensions.md`, alongside other ecosystem extensions like OMATrust and PEAC Protocol that also add receipt/trust semantics on top of x402.

PR #2845, opened 2026-07-13
docs: add TWZRD to third-party extensions

twzrd-sol:
Adds a row for TWZRD to `docs/dev-tools/third-party-extensions.md`, following the pattern of the existing vendor entries (#1047, #1616).

**What it is:** [twzrd-x402-gate](https://www.npmjs.com/package/twzrd-x402-gate) is a buyer-side counterparty trust preflight for x402 clients. It registers on the official client's `onBeforePaymentCreation` lifecycle hook (or wraps fetch for clients that compose over it) and returns allow/warn/block on the selected payment requirement - including wash-flag refusal for flagged merchants - before any payload is constructed or signed. Zero runtime dependencies, MIT.

Docs-only change; no changelog fragment per CONTRIBUTING.md.

Disclosure: this PR was prepared with AI assistance and human-reviewed; the linked package and its hook integration against `x402Client.onBeforePaymentCreation` are shipped and tested.

---
<sub>Resubmission of #2839 with a verified-signed commit. The original carried the identical docs change but an unsigned commit, which the `check-verified-commits` gate blocked; #2839 is being closed in favour of this one. Same author, same diff.</sub>


twzrd-sol:
Withdrawing this PR — we received the curation policy on #2566 ("at this point, we do not consider community submissions to this table") and this resubmission doesn't change that. The commit-signing fix (#2839 → #2845) addressed a bot gate, not the actual policy filter.

Appreciate the ecosystem work Coinbase has done here. We're on x402scan and Agentic.Market, and the real Coinbase integration path is through the AgentKit hook seam — which is separate from this directory.

Closing to respect the policy rather than collect the canned rejection.

PR #355, opened 2025-08-31
Feat : Add Starknet Support for x402 payments

kamalbuilds:
## Description

1. Added Starknet Networks to x402 payment system:
    - starknet (mainnet)
    - starknet-sepolia (testnet)
  2. Implemented Core Starknet Modules:
    - /src/shared/starknet/wallet.ts - Account management and signing
    - /src/shared/starknet/client.ts - RPC provider and network connectivity
    - /src/shared/starknet/usdc.ts - USDC token operations on Starknet
    - /src/shared/starknet/index.ts - Module exports
  3. Updated Type System:
    - Extended network types to include Starknet
    - Added Starknet signer and client types
    - Integrated with existing multi-network architecture
  4. Fixed All TypeScript Issues:
    - Resolved chain ID type conflicts (numeric vs string)
    - Updated middleware for Starknet-specific handling
    - Fixed payment requirement selectors
  5. Production-Ready Features:
    - Native account abstraction support
    - TypedData message signing
    - USDC transfers and approvals
    - Fee estimation
    - Full TypeScript typing

  📦 Package Integration:

  - Added starknet: ^7.6.4 dependency
  - Build system fully functional
  - All tests passing
  
## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits

PR #2849, opened 2026-07-14
specs(exact): propose Starknet exact scheme for x402 v2 (spec-only)

adipundir:
## Description

Proposes the `exact` payment scheme specification for **Starknet** (spec-only, PR 1 of the [New Chains workflow](https://github.com/x402-foundation/x402/blob/main/CONTRIBUTING.md#new-chains)). This follows the guidance given when the earlier v1-era Starknet implementation PR was closed ([#355 closing comment](https://github.com/x402-foundation/x402/pull/355#issuecomment-3860896938)): target v2 and agree on the spec first.

**Mechanism**: the client signs a [SNIP-9](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-9.md) Outside Execution ([SNIP-12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md) typed data) authorizing exactly one `transfer` call from its own account contract; the facilitator submits it via `execute_from_outside_v2` and pays all gas. This is the Starknet-native analog of EIP-3009: no ERC-20 approvals, no client gas, replay protection enforced on-chain by the account contract's single-use SNIP-9 nonce (no facilitator-side nonce storage).

Changes:
- Add `specs/schemes/exact/scheme_exact_starknet.md` — CAIP-2 networks (`starknet:SN_MAIN` / `starknet:SN_SEPOLIA`), payload structure, facilitator verification MUST-rules (canonical typed-data reconstruction, caller binding, time window, on-chain nonce check, exact-amount transfer intent, mandatory settlement simulation, facilitator safety), settlement logic, duplicate settlement mitigation.
- Add a Starknet block to the Critical Validation Requirements in `specs/schemes/exact/scheme_exact.md` and list the new per-network document.

Per the New Chains process, the reference implementation (`typescript/packages/mechanisms/starknet/`) follows in a subsequent PR after spec approval.

Disclosure: this spec was drafted with significant AI assistance and human-reviewed; all chain constants (SNIP-9 typed-data structure and entry points, CAIP-2 identifiers, Circle USDC addresses, selectors, magic values) were verified against primary sources (starknet-io/SNIPs, ChainAgnostic/namespaces, Circle developer docs).

## Tests

No code affected (spec-only).

## Checklist

- [x] ~I have formatted and linted my code~ (docs-only)
- [x] ~All new and existing tests pass~ (docs-only)
- [x] Changelog fragment not required (docs-only change)


specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+
+Source: [Circle USDC contract addresses](https://developers.circle.com/stablecoins/usdc-contract-addresses). Any ERC-20-compatible Starknet token exposing standard `transfer` and balance entry points can be used as `asset`.
+
+### Transport Header Mapping (HTTP v2)
+
+| Header | Direction | Content |
+| ------ | --------- | ------- |
+| `PAYMENT-REQUIRED` | Server → Client | Base64-encoded `PaymentRequired` |
+| `PAYMENT-SIGNATURE` | Client → Server | Base64-encoded `PaymentPayload` |
+| `PAYMENT-RESPONSE` | Server → Client | Base64-encoded `SettlementResponse` |
```
phdargen: please remove, scheme should remain transport agnostic

specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Error Codes
+
+`invalidReason` and `errorReason` MUST be stable enum tokens (below) so wire consumers can switch on them; any human-readable context MUST travel in the `invalidMessage` (`VerifyResponse`) / `errorMessage` (`SettleResponse`) fields defined by the canonical `@x402/core` types, never appended to the code. Standard v2 codes (`insufficient_funds`, `invalid_payload`, `invalid_payment_requirements`, `invalid_scheme`, `invalid_network`, `invalid_x402_version`, `invalid_transaction_state`, `unexpected_verify_error`, `unexpected_settle_error`) apply as usual. Starknet-specific values:
```
phdargen: Please dont refer to a specific SDK implementation (@x402/core is the reference TS implementation, there is also go/py) or its internal types

specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+- **Settlement atomicity**: `execute_from_outside_v2` executes the transfer in the same transaction that consumes the nonce; a revert rolls back both.
+
+### Default Assets
+
+| Token | Network | Address |
+| ----- | ------- | ------- |
+| USDC (Circle) | `starknet:SN_MAIN` | `0x033068f6539f8e6e6b131e6b2b814e6c34a5224bc66947c47dab9dfee93b35fb` |
+| USDC (Circle) | `starknet:SN_SEPOLIA` | `0x0512feac6339ff7889822cb5aa2a86c848e9d392bb0e3e237c008674feed8343` |
+
+Source: [Circle USDC contract addresses](https://developers.circle.com/stablecoins/usdc-contract-addresses). Any ERC-20-compatible Starknet token exposing standard `transfer` and balance entry points can be used as `asset`.
```
phdargen: This is a SDK level implementation choice, should not be part of the nominative spec

specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+- `payTo`: recipient's Starknet address.
+- `extra.areFeesSponsored`: whether the facilitator sponsors gas. This spec covers the sponsored flow only; when present it MUST be `true`, and `false` MUST be rejected (`invalid_payment_requirements`). When absent, the requirement is treated as sponsored — the only flow this spec defines — and MUST be accepted identically to `areFeesSponsored: true`. An unsponsored flow is planned for a follow-up.
+- `extra.caller` (optional): the address that will be the on-chain caller of `execute_from_outside_v2` at settlement — the facilitator's executor account, or its paymaster's forwarder contract (the address the payer's account observes as the SNIP-9 `Caller`). Binding a specific caller prevents third parties from submitting the execution (see Security Considerations). Facilitators that cannot guarantee the submitting address MUST NOT advertise `extra.caller`. Two client flows produce the `OutsideExecution` `Caller`:
```
phdargen: Let's align with the wire format used for other networks like SVM etc
```suggestion
- `extra.feePayer` (optional): the address that will be the on-chain caller of `execute_from_outside_v2` at settlement — the facilitator's executor account, or its paymaster's forwarder contract (the address the payer's account observes as the SNIP-9 `Caller`). Binding a specific caller prevents third parties from submitting the execution (see Security Considerations). Facilitators that cannot guarantee the submitting address MUST NOT advertise `extra.caller`. Two client flows produce the `OutsideExecution` `Caller`:
```

specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+  - **Paymaster (sponsored) flow**: the paymaster sets `Caller` to its own forwarder; the client does not choose it. When `extra.caller` is advertised the client MUST verify the paymaster-built `Caller` equals it and reject a mismatch, and the advertised value MUST be that forwarder. When it is absent, the resulting forwarder `Caller` is accepted by the facilitator under rule 4 when that forwarder is in its trusted configuration. In all cases the client MUST verify the paymaster-built message against its own intent before signing — exactly one `Call` with `To` = the requirements' `asset`, `Selector` = `sn_keccak("transfer")`, `Calldata` = `[payTo, amount_low, amount_high]` for the exact required amount, and time bounds per the Timeout Mapping — and MUST refuse to sign otherwise: the signature authorizes exactly what the message says, not what the client asked the paymaster to build.
```
phdargen: As written, the paymaster can't be an independent third party. It's effectively the facilitator, so having the client fetch/verify the paymaster-built `Caller` is an out-of-band client→facilitator communication. The client doesn't even know which facilitator the server uses, and we generally want to avoid any direct client→facilitator interaction, should always be proxied through server

specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+  - **Direct flow** (client builds the `OutsideExecution` itself, no paymaster): the client MUST set `Caller` to the advertised `extra.caller`, or to the SNIP-9 `ANY_CALLER` sentinel (`0x414e595f43414c4c4552`, the short string `ANY_CALLER`) when none is advertised.
```
phdargen: Contrasting this statement against the `Paymaster (sponsored) flow` framing above, makes it sound like this would be a non-sponsored flow

specs/schemes/exact/scheme_exact_starknet.md
```diff
@@ -0,0 +1,384 @@
...
+- The client never needs gas funds in the sponsored flow. Executor/paymaster selection beyond `extra.caller` is facilitator-local configuration and MUST NOT be required from the client-facing `PaymentRequirements`.
```
phdargen: Overall, I think this part can be simplified substantially. Since this flow is sponsored-only with a single-call rule, there's nothing a paymaster needs to inject into the message, so the client can always build the OutsideExecution itself. I'd drop the paymaster-built vs direct distinction (and ANY_CALLER) and make the server advertise a single non-optional extra.feePayer (taken from feePayer announced by facilitator /supported), defined as the address the client MUST set as the SNIP-9 Caller. Whether the facilitator then settles from its own account or routes through a forwarder/paymaster becomes a pure implementation detail that never touches the wire or the client

phdargen: When there is an non-optional `extra.feePayer`, we can also drop `extra.areFeesSponsored=true` as this is implied by the presence of a feePayer

phdargen:
Thanks for the contribution @adipundir, this looks great! 

The only thing to pin down is the `caller` assignment and paymaster vs direct flow that seems unnecessarily complex, would prefer if we could define a single canonical payload construction path

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Done, and this simplified things a lot — thanks for pushing on it. The spec now defines a single construction path: `extra.feePayer` is REQUIRED, the client always builds and signs the `OutsideExecution` itself, and both the paymaster-built vs direct distinction and `ANY_CALLER` are gone. Whether the facilitator settles from its own account or via a forwarder/paymaster is now stated as a facilitator-local detail that never reaches the wire or the client.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Renamed to `extra.feePayer`. I did not apply the suggestion verbatim since your later comment supersedes it: the field is now REQUIRED rather than optional.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Agreed. The client no longer fetches anything from a paymaster. The spec now states that every facilitator interaction is proxied through the resource server.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Fixed. With the two-flow framing removed the ambiguity is gone — the spec now states plainly that because `extra.feePayer` is required, every payment in this scheme is sponsored and the client never needs gas funds.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Dropped. A required `feePayer` implies it, as you noted.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Removed the HTTP header table.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Removed.

specs/schemes/exact/scheme_exact_starknet.md
adipundir: Removed the USDC address table. `asset` is now described generically as any SNIP-2 token exposing `transfer` and a balance getter, with no default or recommended token.

adipundir:
Thanks a lot for the thorough review @phdargen — the feedback was genuinely useful, and the suggestion to collapse this into one canonical construction path made the spec noticeably simpler. Net −7 lines while removing an entire branch of the flow.

All points are addressed in 2edeb50:

- **Single canonical path.** `extra.feePayer` is now REQUIRED and taken verbatim from the facilitator's `/supported` entry. The client builds and signs the `OutsideExecution` from the `PaymentRequirements` alone and sets `Caller` to that address. The paymaster-built vs direct distinction is gone.
- **No client→facilitator interaction.** Stated explicitly — every facilitator interaction is proxied through the resource server. How the sponsor settles behind `feePayer` (own account, forwarder, or SNIP-29 paymaster) is now called out as an implementation detail that never touches the wire.
- **`ANY_CALLER` removed** as an option, and explicitly forbidden as a `feePayer` value — it's a legal SNIP-9 value, so leaving it unmentioned would let a caller-bound authorization silently become a bearer one.
- **`extra.areFeesSponsored` dropped** — implied by a required `feePayer`.
- **Transport-agnostic** — HTTP header table removed.
- **No SDK coupling** — `@x402/core` type references replaced with behavioural wording; USDC address table removed and `asset` described generically via SNIP-2.

Individual replies are on each thread. Happy to adjust anything further.

phdargen (approved):
Thanks for prompt update @adipundir, LGTM 🚀

For the implementation you may find this skill useful: https://github.com/x402-foundation/x402/blob/main/.agents/skills/contributing/SKILL.md

PR #2852, opened 2026-07-14
fix stellar tx fee

### LINKED ISSUE #2764: Stellar exact-scheme facilitator should derive settlement fees from simulation, not the client's bid
## Problem

The current facilitator uses the client's fee bid as its settlement budget: it checks `clientFee <= this.maxTransactionFeeStroops` (configurable, defaults to 50,000 stroops) at verify time, then clamps the rebuilt transaction at `Math.min(clientFee, cap)` at settlement ([line 403](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts#L403)).

This model has two problems:

1. **The client's bid can go stale.** It reflects network conditions at simulation time; by settlement, Soroban resource fees may have shifted, making the bid stale relative to what the network actually requires.
2. **Raising the cap doesn't fix the underlying issue.** Even though `maxTransactionFeeStroops` is already configurable per-instance, any fixed number is still a static ceiling chosen once — it doesn't adapt as network load or protocol upgrades change resource pricing.

## Proposed fix

The facilitator already re-simulates the transaction at settlement for verification. That simulation response contains the actual current resource fee from the network. The facilitator should derive its settlement fee from that result — `simulationResourceFee + inclusion buffer (>= 100 stroops)` — rather than from the client's bid. This makes the fee self-adapting: it is always exactly what Stellar requires at the moment of submission, regardless of how much time passed since the client simulated or what the client put in the envelope.

`maxTransactionFeeStroops` should become a **safety circuit breaker**, not a budget: if the simulation-derived fee exceeds a high ceiling (e.g. 1,000,000 stroops), the facilitator aborts because the transaction looks abnormal. Under normal operation this ceiling never triggers; it isn't something operators need to tune.

This should be written compatibly with [stellar/x402-stellar#44](https://github.com/stellar/x402-stellar/issues/44) (a pending unsponsored client-paid-fees flow, where the inner fee stops being facilitator-paid).

## Spec gap

[`specs/schemes/exact/scheme_exact_stellar.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_stellar.md) has no fee section. It should document: (a) the client must include a nonzero inclusion fee (>= 100 stroops) — `assembleTransaction` double-counts the resource fee when the inclusion fee is exactly 0 (see also #1343 in `stellar/js-stellar-sdk`); (b) facilitators MUST derive their settlement fee from a fresh simulation at settle time, not from the client's bid.

## Key files

[`specs/schemes/exact/scheme_exact_stellar.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_stellar.md), [`typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts`](https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts) (`DEFAULT_MAX_TRANSACTION_FEE_STROOPS` at line 35, the `fee_exceeds_maximum` check at line 294, and the settlement clamp at line 403).

## Acceptance criteria

- [ ] Settlement fee derived from the simulation result at settle time (`simulationResourceFee + inclusion buffer`), not from `Math.min(clientFee, cap)`
- [ ] `maxTransactionFeeStroops` repurposed as a safety circuit breaker at a high ceiling; the `fee_exceeds_maximum` error becomes a signal of an abnormal transaction, not a routine tuning issue
- [ ] Spec gains a fee section: nonzero (>= 100 stroop) inclusion fee required on the client envelope, and facilitators must use simulation-derived fees at settlement
- [ ] `fee_exceeds_maximum` `VerifyResponse` populates the existing `invalidMessage` field with the simulation-derived fee and the ceiling so the operator can investigate (this error should be rare after the fix)

phdargen:
## Description

Derive Stellar exact settlement fees from settle-time simulation instead of the client bid. On SDK v16, TransactionBuilder.build() sets tx.fee = fee + sorobanData.resourceFee(), but the facilitator passed the client’s total transaction.fee (already inclusion + resource) as fee while also attaching the client’s sorobanData, so the resource fee was added twice. 

The fix rebuilds with BASE_FEE as inclusion-only and fresh sorobanData from simulation (tx.fee = BASE_FEE + minResourceFee). Verify drops the client-fee check, enforces a simulation-derived ceiling via maxTransactionFeeStroops, and settle reuses the same simulation.

Closes https://github.com/x402-foundation/x402/issues/2764

## Checklist

### DIFF
--- a/specs/schemes/exact/scheme_exact_stellar.md
+++ b/specs/schemes/exact/scheme_exact_stellar.md
@@ -34,7 +34,7 @@ The protocol flow for `exact` on Stellar is client-driven with facilitator-spons
 10. **Resource Server**, upon successful verification, forwards the payload to the facilitator's `/settle` endpoint.
     - NOTE: `/settle` MUST perform full verification independently and MUST NOT assume prior verification.
 11. **Facilitator** rebuilds the transaction with its own account as the source, preserving all operations and auth entries.
-12. **Facilitator** simulates the transaction to verify it succeeds and emits the expected transfer events.
+12. **Facilitator** simulates the transaction to verify it succeeds, emits the expected transfer events, and derives the settlement fee and fresh Soroban resource data (footprint and resource fee cap) from that simulation.
 13. **Facilitator** signs the rebuilt transaction with its own key and submits it to the Stellar network via RPC `sendTransaction`.
 14. **Facilitator** polls for transaction confirmation and responds with a `SettlementResponse` to the **Resource Server**.
 15. **Resource Server** grants the **Client** access to the resource in its response upon successful settlement.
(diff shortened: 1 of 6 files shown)


phdargen:
cc @marcelosalloum @oceans404

CarsonRoscoe (approved):

Issue #2848, opened 2026-07-13
Extension proposal: compliance-fields — optional tax/audit record composing with offer-receipt


wowlegend:
The merged `offer-receipt` extension gives x402 a clean, privacy-minimal proof of payment. What it deliberately doesn't give sellers is a record their accountant or tax authority can consume — and those requirements are dated and near-term (EU member-state e-invoicing mandates through 2026, EN 16931 EU-wide under ViDA; Japan QIS and Korea e-Tax already live; HK IRO s.51C 7-year record-keeping; 1099-DA gross-proceeds context in the US; MiCA Art 68(9) for CASP customers). The OpenAI Instant Checkout retreat this March was partly a tax-operations story — agent commerce dies without this layer.

I'd like to propose an optional `compliance-fields` extension that:

1. **Never touches the offer-receipt schema** — it binds to the receipt artifact by digest (`receiptDigest = keccak256(canonical JSON)`), so the base EIP-712 types stay fixed;
2. Is **micro-transaction-honest** — a MINIMAL tier isomorphic to the EU simplified-invoice content (VAT Directive Arts 220a/226b, ~5 fields, legally sufficient sub-€100), with EN 16931-aligned fields as an optional FULL tier;
3. **Reuses existing vocabularies** — EN 16931 business-term semantics, and the adjustment enum ACP and UCP have already converged on (`refund/return/credit/price_adjustment/dispute/cancellation` × `pending/completed/failed`) so records round-trip card-rail order objects;
4. Gives x402 a standard **corrective-record chain** — refund/correction records reference the original by digest (`refundOf`), matching ViDA's mandatory corrective-invoice reference, composing with `exact`, `auth-capture`, and `batch-settlement` refund flows;
5. Signs an **attestation over digests only** (fixed EIP-712 schema `{version, recordDigest, receiptDigest, issuedAt}`), so the record schema can evolve without breaking signatures — same forward-compatibility philosophy as offer-receipt §2.

Deliberately out of scope: tax computation, filing/transmission channels (Peppol APs, KSeF, PDPs), delivery proof (complementary to the SAR proposal in #1195 and the delivery-receipt attestation proposal in #2833 — either's digest can ride along as a reference), and identity verification (belongs to AP2/8004-layer mechanisms).

[...]

PR #2811, opened 2026-07-07
docs+spec: clarify signer authorization in offer-receipt extension (#2462)

alftom:
## Description

Addresses https://github.com/x402-foundation/x402/issues/2462

This PR makes two related changes:

**1. Spec: clarify signer authorization (§4.5.1, §5.5)**

- §4.5.1 now distinguishes signature validity from signer authorization and describes an attack vector where a valid signature alone is insufficient. Lists concrete authorization mechanisms (`payTo` signing, `did:web`, DNS TXT, external registries) and notes that mutable sources don't provide durable historical evidence after key rotation.
- §5.5 adds guidance that when receipts are verified outside the immediate payment session, authorization should be evaluated as of `issuedAt`, not at verification time. Key revocation is treated as prospective.

**2. Docs: expand offer-receipt guidance for sellers**

- `docs/extensions/offer-receipt.mdx` — Adds a "Signer Authorization Approaches" table comparing mechanisms by durability, updates key management guidance to address rotation workflows.
- `docs/getting-started/quickstart-for-sellers.mdx` — Adds step 6 pointing sellers to the Signed Offers & Receipts extension.
- `docs/dev-tools/third-party-extensions.md` — Updates OMATrust description to reflect signer-authorization focus.

## Tests

Documentation and specification changes only — no code changes. Verified markdown renders correctly and all internal links resolve.

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge)
- [x] I added a changelog fragment for user-facing changes (docs-only changes can skip)

PR #2853, opened 2026-07-14
specs: add compliance-fields extension — tax/audit record + verifier disqualification rules (#2848)

wowlegend:
Spec for the `compliance-fields` extension proposed in #2848 (issue-first per specs/CONTRIBUTING.md).

An optional, signed compliance record composing with `offer-receipt` by digest reference — never touching its EIP-712 types. Summary of what's normative:

- **Tiers are legal thresholds:** MINIMAL is isomorphic to the EU simplified-invoice content (VAT Directive Arts 220a/226b, legally sufficient sub-€100); FULL aligns to EN 16931. Tiers MUST NOT be mapped onto verification-confidence scales.
- **Canonicalization v1 = RFC 8785 (JCS)**, digest = keccak256, versioned **together** via `canonicalizationVersion` (changing either independently invalidates existing signatures). Byte-level cross-implementation conformance vectors are published with the reference implementation — including the integer-like-key ordering case that sort-then-stringify implementations get wrong.
- **Corrective chain:** `refundOf` + the ACP/UCP adjustment vocabulary, per ViDA's corrective-invoice reference; composes with `exact`, `auth-capture`, and `batch-settlement` refund flows.
- **Sequential numbering:** `seq`/`correctionSeq` carry Art 226(2)/BT-1 semantics — a completeness control. Normative independence rule: an issuer-attested sequence evidences ordering only; verifiers MUST NOT treat it as evidence that no record was omitted. Who supplies non-issuer observation is service-layer, out of scope.
- Deliberately out of scope: tax computation, filing/transmission channels, delivery proof (complementary to #1195 SAR and #2833 delivery-receipt — either's digest can ride along), identity.

Reference implementation: TypeScript, MIT (`tersign` on npm; conformance fixtures in-repo). Happy to iterate on placement/naming per maintainer preference.

Disclosure: drafted with AI assistance, human-reviewed and human-submitted.

---
**Executable form:** the extension's normative disqualifications (independence, completeness, existence, economic phase) are published as a two-sided conformance suite — stdlib verifier, live Bitcoin-anchored vectors: https://github.com/tersignhq/evidence-record-conformance

### DIFF
--- a/specs/extensions/compliance_fields.md
+++ b/specs/extensions/compliance_fields.md
@@ -1,0 +1,129 @@
+# Extension: `compliance-fields`
+
+## Summary
+
+The merged `offer-receipt` extension gives x402 a cryptographic proof of payment, and stays deliberately privacy-minimal. What it does not give sellers is a record a tax authority, accountant, or auditor can consume — and those requirements are dated and near-term: EU member-state structured e-invoicing through 2026 (EN 16931 EU-wide under ViDA), Japan's Qualified Invoice System and Korea e-Tax (live), Hong Kong IRO s.51C record-keeping (7-year retention), US 1099-DA gross-proceeds context, MiCA Art 68(9) for CASP customers.
+
+`compliance-fields` defines an OPTIONAL, signed **compliance record** that composes with the base receipt by digest reference. Sellers that ignore it lose nothing; sellers that emit it get machine-verifiable records their accountants and regulators already recognize.
+
+Design principles:
+
+1. **Never touches the offer-receipt schema.** The record binds to the receipt artifact by `receiptDigest`; the base EIP-712 types stay fixed.
+2. **Micro-transaction-honest.** A MINIMAL tier isomorphic to the EU simplified-invoice content (VAT Directive Arts 220a/226b — ~5 fields, legally sufficient sub-€100), with EN 16931-aligned fields as an optional FULL tier.
+3. **Vocabulary reuse, not invention.** EN 16931 business-term semantics; the adjustment enum ACP and UCP have converged on (`refund/return/credit/price_adjustment/dispute/cancellation` × `pending/completed/failed`).
+4. **Corrections are chained.** Refund/correction records reference the original by digest (`refundOf`), matching ViDA's mandatory corrective-invoice reference; composes with `exact`, `auth-capture`, and `batch-settlement` refund flows.
+5. **Attestation signs digests only** (fixed EIP-712 schema), so the record schema can evolve without breaking signatures — the same forward-compatibility approach as `offer-receipt`.
+
+---
+
+## `PaymentRequired`
+
+Servers MAY advertise support:
+
+```json
+{
+  "extensions": {
+    "compliance-fields": {
+      "info": {
+        "tiers": ["minimal", "full"],
+        "jurisdictions": ["EU-226b", "HK-51C"]
+      }
+    }
+  }
+}
+```
+
+## Response placement
+
+On success, alongside the base receipt over the v2 extension-response path:
+
+```
+extensions["compliance-fields"].info = {
+  record:      <ComplianceRecord>,   // JSON, schema below
+  attestation: <SignedArtifact>      // EIP-712 or JWS
+}
+```
+
+---
+
+## `ComplianceRecord` (v1)
+
+| Field | Type | Tier | Semantics (mandate basis) |
+|---|---|---|---|
+| `version` | int | M | `1` |
+| `canonicalizationVersion` | int | M | `1` (see Canonicalization below) |
+| `receiptDigest` | bytes32 hex | M | keccak256 of the canonicalized base receipt artifact (binding) |
+| `issuedAt` | int (unix s) | M | issue date (Art 226(1)/226b(a); EN 16931 BT-2) |
+| `seq` | int | F | sequential number per issuer (Art 226(2); EN 16931 BT-1) — assigned at counter-signature time by a ledger service when used |
+| `correctionSeq` | int | F | position within a record's correction chain; inherits `seq`'s semantics, so an omitted correction is detectable on the same basis as an omitted record |
+| `issuer.name` | string | M | supplier identity (Art 226b(b)) |
+| `issuer.taxId` | string | F* | VAT ID / JP T-number / HK BR no. (*M where the jurisdiction mandates it) |
+| `issuer.jurisdiction` | string | M | ISO 3166 + regime label |
+| `buyer.principal` | string | F | the signed principal behind the paying agent (deployer / AP2 mandate subject) — never merely a wallet |
+| `buyer.mandateRef` | bytes32 hex | F | hash of the AP2 Checkout/Payment Mandate, when the payment carried one |
+| `supply.description` | string | M | nature of goods/services (Art 226b(c); BT-153/BT-154 class) |
+| `lines[]` | array | F | description, quantity, unitPrice, net (Art 226(6)-(8)) |
+| `tax.scheme` | enum `none/vat/gst/jct/sales` | M | tax regime marker |
+| `tax.currency` | string | M | ISO 4217 |
+| `tax.amount` | string | M unless `none` | tax payable or data to compute it (Art 226b(d); JP QIS per-rate rule) |
+| `tax.breakdown[]` | array | F | per-rate `{rate, taxable, tax, category}` (Art 226(8)-(10)) |
+| `settlement.fiat` | object | F | `{amount, currency, source, asOf}` — fiat-equivalent valuation at issuance (1099-DA gross-proceeds context) |
+| `settlement.txHash` | string | F | on-chain settlement binding (ViDA "payment details") |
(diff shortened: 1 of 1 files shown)


wowlegend:
All commits on this PR are now signed and verified, and checks are green — the earlier `check-verified-commits` block is resolved.

Scope recap for review: this adds `specs/extensions/compliance-fields.md`, an optional extension composing with offer-receipt — no changes to core protocol, schemes, or SDKs. The normative section is three evaluator-side constraints (independence, completeness/existence bound, economic-phase separation) that the evidence-record discussions in #2887 and #2906 now reference; each is executable as a two-sided conformance vector (suite linked in the PR description).

@phdargen — requesting a core review when you get a chance (CODEOWNERS routes `/specs/` to `@x402-foundation/core`). Happy to split the normative section into a separate file or trim scope if that makes review easier.

wowlegend:
Updated the spec to be self-contained. The canonicalization conformance vectors are now inline as normative test vectors rather than links to fixtures in an implementation repo, and the reference-implementation link is gone — the extension now depends on no third-party service or repository, and one file is the whole diff.

The inlined vectors are worth the lines they cost. RFC 8785 orders keys by UTF-16 code unit, so `"1" < "10" < "2"`; a sort-then-stringify implementation whose runtime hoists integer-like keys into numeric order produces a different digest from the same record, and a vector set with only non-numeric keys cannot catch it.

One scope note, since #2811 merged on the 23rd. §4.5.1 now says applications verifying after key rotation SHOULD "preserve or reference temporally immutable authorization evidence... to confirm the key was authorized at issuance time" — which is the same property this extension's `seq`/`correctionSeq` completeness control depends on, and it makes the remaining delta here narrower than when this was filed. What §4.5.1 leaves open is who preserves that evidence: authorization evidence retained by the party that issued the artifact evidences its own record-keeping. That is the gap this extension's normative independence rule addresses, and it is stated as an evaluator-side constraint — no new field, no dependency on anyone's service.

Happy to trim scope further or split the normative section out if that makes review easier.

0rkz:
Ran the two inline vectors independently before commenting — both reproduce exactly. keccak256 over the UTF-8 bytes of the canonical forms, under two unrelated stacks (Python `eth_utils` and node `viem`):

```
{"a":1,"b":"x"}            → 0x84fc3d9faf736ddfdb9baab9973656bd8d9bd142f1dfff8aa513a774fddfdd04
{"1":"c","10":"a","2":"b"} → 0x426b770f81b8ad5e307bcfb767deb02f8d32cd340d81a946be88bb184857e81b
```

The key-ordering rule checks out as written. One gap worth closing before this lands: neither vector exercises number *serialization* — the only number in either is `1`, which every implementation emits identically — and the money-bearing members are where §3.2.2.3 actually bites. `tax.amount` is declared `string`, but `lines[].quantity` / `unitPrice` / `net` and `tax.breakdown[].rate` / `taxable` / `tax` carry no declared type. A producer emitting any of those as a JSON number inherits RFC 8785 §3.2.2.3 (ECMAScript `Number::toString`; worked samples in Appendix B) — exactly the divergence class the extension's own text warns about, and neither vector would catch it.

Two cheap fixes, either works: declare every money-bearing member a string (decimal or integer minor units) so numbers can't enter the digest, or add one vector with a numeric member (e.g. a `taxable` that round-trips differently under float re-serialization) so a non-conforming producer fails the vector instead of diverging silently.

*(Disclosure: I work on PayPerByte — we sign digests over canonical JSON on x402, so producer/verifier byte-agreement is the part I care about.)*

wowlegend:
Good catch — and it runs two members deeper than flagged. `settlement.fiat.amount` and the `adjustment` amounts were also undeclared, so eight money/rate/quantity members carried no type, not six.

Took your first option plus the boundary half of the second. Every money/rate/quantity member is now a decimal string (`^-?(0|[1-9][0-9]*)(\.[0-9]+)?$`); the only JSON numbers a record admits are the int-declared members, all exact integers; and a record containing a non-integer JSON number anywhere is non-conformant — verifiers MUST reject before digesting.

Decimal strings rather than integer minor units because `tax.amount` was already a string, the base x402 spec types every amount `string`, and EN 16931 terms are decimals — minor units would drag in an ISO 4217 exponent registry (JPY 0, BHD 3) for no added digest-safety, and rates and quantities have no minor unit at all.

On the vector: I inverted yours rather than adding a float case. With floats excluded at the type layer, the silent-divergence class that remains is coercion of numeric-looking *strings*, so the new vector pins `"1.10"` beside an exact integer — a stack that floats the string re-emits `1.1` and fails the digest. A float-bearing vector would instead canonize the ECMAScript serialization of a value the record now forbids, so your float case lands as the pinned MUST-reject rather than a digest row.

One point under your §3.2.2.3 cite worth making explicit: even two byte-perfect JCS stacks digest the nearest double, not the decimal the source system held — parse-time loss, not just serialization divergence. For records whose consumers are tax authorities that is wrong even when it's deterministic, which is why the fix is exclusion at the type layer rather than tighter serialization. Appendix B's samples and the §3.1 string RECOMMENDATION both point the same way. Pushed as `c715850`.

Disclosure: drafted with AI assistance, human-reviewed. — Tersign

wowlegend:
Added a fourth evaluator-side rule to the normative section (`55b6738`): **an independence claim MUST NOT be read as covering any fact the record does not itself commit to** — a settlement-scoped record carries no delivery independence, however independent the attestor. First proposed in the #2887 discussion (2026-07-27), where its sharpest property was stated up front: it is a rule this extension's own reference records fail-safe on deliberately. Two-sided conformance vectors for it are live in the suite (scope-within-commitments accepting, scope-overreach rejecting).

Disclosure: drafted with AI assistance, human-reviewed. — Tersign

wowlegend:
@phdargen — this has been open since 14 July and has never had a formal review request on it (my July ask was prose, which only produced a mention). Asking once more, with a decision path rather than just a bump.

Where it stands: one added file, `specs/extensions/compliance_fields.md`, +202 lines, nothing else in the diff. No changes to core protocol, schemes, or SDKs. Self-contained since 27 July — the conformance vectors are inline as normative test vectors, and it depends on no third-party repository or service.

It has had external review even without a formal one. @0rkz reproduced both inline canonicalization vectors under two unrelated stacks (Python `eth_utils`, node `viem`) and found a real gap while doing it: neither vector exercised number *serialization*, which is where RFC 8785 §3.2.2.3 actually bites on money-bearing members. That is fixed — every money/rate/quantity member is now a declared decimal string, and a record carrying a non-integer JSON number anywhere is non-conformant.

The substance is four evaluator-side constraints, all stated as MUST NOTs rather than new required fields: independence, completeness under an existence bound, economic-phase separation, and independence-scope-within-commitments. They constrain how a reader may *read* a record; they do not ask emitters to carry anything new.

Two ways to make this easier to land, and I am happy with either:

1. **Merge as an extension doc.** Nothing composes with it that is not already merged.
2. **Cut it down.** If four constraints is more than the extension slot should carry, say which one you would keep and I will reduce the PR to that, and take the rest to a discussion instead.

A close is also an answer — if evaluator-side constraints do not belong in `specs/extensions/`, that is worth knowing, and I will move the material to the evidence-record discussions where it is already referenced.

— Tersign

*Drafted with AI assistance.*

0rkz:
Re-verified at head `9f567d9b` (`9f567d9bd513a673c56d3323e02e593172916157`).

**Vectors.** All three inline canonicalization vectors in `compliance_fields.md` reproduce — the two from our 07-29 run plus the new third one — independently on two stacks (Python `pycryptodome` keccak, node `js-sha3`):

```
{"a":1,"b":"x"}                                   → 0x84fc3d9f…fddfdd04  MATCH
{"1":"c","10":"a","2":"b"}                        → 0x426b770f…4857e81b  MATCH
{"issuedAt":1735689600,"tax":{"amount":"1.10"}}   → 0x81086b58…4e78b250  MATCH
```

**Gap confirmed closed.** We flagged six money/rate/quantity members with no declared type (`lines[].quantity/unitPrice/net`, `tax.breakdown[].rate/taxable/tax`). At head, the Numbers section declares eight members decimal-string, matching `^-?(0|[1-9][0-9]*)(\.[0-9]+)?$`: our original six, `tax.amount` (already `string` before, now under the same regex), and `settlement.fiat.amount` — the one genuinely new member beyond what we flagged. The only JSON numbers a record admits are the seven int-declared fields, and any other JSON number anywhere is a hard MUST-reject before digesting; `adjustment` defers to the same rule by reference rather than naming its own fields.

As an interoperability consumer, the four evaluator-side rules read as self-contained — nothing in them depends on anything outside this diff.

wowlegend:
Noted and appreciated, @0rkz — a second independent reproduction on two stacks at head `9f567d9b`, plus confirmation the number-domain gap is closed, is exactly the kind of evidence a reviewer should be able to lean on. For the record alongside it: the four evaluator-side rules in this PR have since been confirmed composable from the consumer side by two further independent efforts that arrived at the same derived-not-declared relation without reference to this text (x402#3065). The text has not moved since 3 August; it is waiting on review, not on content.

— Tersign

*Drafted with AI assistance.*

PR #2861, opened 2026-07-15
docs: Specify exact Lightning on BIP-122

benthecarman:
## Description

Define how to do x402 using bitcoin lightning

## Tests

N/A

## Checklist

### DIFF
--- a/specs/schemes/exact/scheme_exact.md
+++ b/specs/schemes/exact/scheme_exact.md
@@ -54,4 +54,4 @@ While implementation details vary by network, facilitators MUST enforce security
 - Replay protection: the SNIP-9 nonce MUST be unused at verification; it is consumed on-chain at execution.
 - Simulation verification: MUST simulate the settlement and fail closed unless it shows exactly one asset `Transfer` from payer to `payTo` for the exact amount.
 
-Network-specific rules are in per-network documents: `scheme_exact_svm.md` (Solana), `scheme_exact_stellar.md` (Stellar), `scheme_exact_evm.md` (EVM), `scheme_exact_sui.md` (SUI), `scheme_exact_ton.md` (TON), `scheme_exact_starknet.md` (Starknet).
+Network-specific rules are in per-network documents: `scheme_exact_svm.md` (Solana), `scheme_exact_stellar.md` (Stellar), `scheme_exact_evm.md` (EVM), `scheme_exact_sui.md` (SUI), `scheme_exact_ton.md` (TON), `scheme_exact_starknet.md` (Starknet), `scheme_exact_bip122.md` (Bitcoin Lightning).
--- a/specs/schemes/exact/scheme_exact_bip122.md
+++ b/specs/schemes/exact/scheme_exact_bip122.md
@@ -1,0 +1,407 @@
+# Scheme: `exact` on Bitcoin Lightning (`BIP-122`)
+
+## Summary
+
+This document specifies the x402 `exact` payment scheme for Bitcoin Lightning
+networks in the CAIP-2 `bip122` namespace. The client pays a fresh BOLT11 invoice
+from the resource server and returns its 32-byte payment preimage. The facilitator
+verifies `SHA-256(preimage) == payment_hash` and requires the invoice signing key
+to match `payTo`. This check does not require access to the receiver's Lightning
+node. A client MUST NOT select this scheme unless its payer Lightning node returns
+the preimage after payment.
+
+The scheme supports only the `bolt11` asset transfer method and the `upfront`
+payment flow. The resource server processes the protected request only after the
+facilitator validates the proof and records the payment hash.
+
+This scheme targets x402 protocol version 2 and uses the core
+`PaymentRequirements`, `PaymentPayload`, and `SettlementResponse` types from
+[x402-specification-v2.md](../../x402-specification-v2.md).
+
+## Scheme and Networks
+
+- `scheme` MUST be `"exact"`.
+- `asset` MUST be `"BTC"`.
+- Mainnet MUST use
+  `bip122:000000000019d6689c085ae165831e93` and BOLT11 currency `bc`.
+- Bitcoin testnet MUST use
+  `bip122:000000000933ea01ad0ee984209779ba` and BOLT11 currency `tb`.
+- Messages on the wire MUST use one of these concrete network identifiers.
+
+## Asset Transfer Method and Payment Flow
+
+`"bolt11"` is the only supported asset transfer method.
+`extra.assetTransferMethod` MAY be omitted, in which case it defaults to
+`"bolt11"`. Any explicit value MUST be `"bolt11"`.
+
+`"upfront"` is the only supported payment flow because a Lightning payment settles
+before its preimage is available. The resource server MUST set
+`extra.paymentFlow` to `"upfront"` in every payment requirement. Clients and
+resource servers MUST reject any other value.
+
+The protocol sequence is:
+
+1. The resource server creates a fresh BOLT11 invoice and returns it in a payment
+   requirement.
+2. The client validates and pays the invoice, then constructs a payment payload
+   containing the preimage.
+3. The resource server sends the payload to the facilitator's `/settle` endpoint.
+   It MUST NOT call `/verify` for this flow.
+4. The facilitator validates the proof and atomically records the payment hash as
+   used.
+5. The resource server processes the protected request only after `/settle`
+   succeeds.
+
+## Terminology and Adapter Requirements
+
+- **Receiver**: The seller-side Lightning node that creates invoices.
+- **Payer**: The client-side Lightning node that pays an invoice.
+- **Payment hash**: The 32-byte value committed to by the BOLT11 invoice.
+- **Preimage**: The 32-byte secret whose SHA-256 digest is the payment hash.
+- **Replay store**: Storage that atomically records settled payment hashes and
+  persists across facilitator restarts.
+
+A payer adapter MUST return the payment preimage when it reports a payment as
+paid. A client MUST NOT use an adapter that cannot return the preimage.
+
+A receiver adapter MUST be able to create a fresh invoice for an exact
+millisatoshi amount. The resource server MUST have exclusive invoice-issuance
+authority for the receiver key in `payTo`; an untrusted party MUST NOT be able to
+create invoices signed by that key.
+
+## Amounts
+
+`amount` MUST be a decimal string that encodes a positive integer number of
+millisatoshis. It MUST NOT contain a sign, decimal point, exponent, separator, or
+unit suffix. The BOLT11 invoice MUST specify an integral millisatoshi amount.
+
+Examples:
+
+| Meaning | Wire `amount` |
+|---|---:|
+| 1 millisatoshi | `"1"` |
+| 1 satoshi | `"1000"` |
+| 21 satoshis | `"21000"` |
+| 1 bitcoin | `"100000000000"` |
+
+User-facing SDKs SHOULD parse satoshi inputs such as `"21"`, `"21 sat"`, and
+`"21 sats"` as `"21000"`. Parsers MUST use exact decimal arithmetic and reject
+negative values and sub-millisatoshi precision. They MUST also reject fiat or
+bitcoin inputs such as `$1`, `1 USD`, or `0.0001 BTC` unless the application has
+registered a conversion parser. The error SHOULD direct the caller to use
+satoshis or an explicit atomic `AssetAmount`.
+
+## `PaymentRequirements`
+
+The resource server MUST generate a fresh BOLT11 invoice for each payment
+challenge and place it in `extra.invoice`.
+
+The following deterministic example is a test vector, not a reusable challenge.
+Its validation time is Unix timestamp `1700000000`, equal to the invoice creation
+time.
+
+```json
+{
+  "scheme": "exact",
+  "network": "bip122:000000000019d6689c085ae165831e93",
+  "amount": "25000",
+  "asset": "BTC",
+  "payTo": "036360e856310ce5d294e8be33fc807077dc56ac80d95d9cd4ddbd21325eff73f7",
+  "maxTimeoutSeconds": 300,
+  "extra": {
+    "assetTransferMethod": "bolt11",
+    "paymentFlow": "upfront",
+    "invoice": "lnbc250n1pj48ugqpp54y3u9s8ylemsv8l3ewyzzu0klhujvuvmkl6llchq23vy8rzjsf0qsp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygsdq80q6rqvsxqzfvcqpjpal2l7zmrg46wpsxvv8aly29hzrjhvyxcrxxdm3r4ky8etpthh9p3r2ly8jvtlv6wprwvrm5t2zrxxmvpg57xhf24x2ngrd8smj8jtcp79fu42"
+  }
+}
+```
+
+The `extra` fields are:
+
+| Field | Required | Meaning |
+|---|---|---|
+| `extra.assetTransferMethod` | No | If present, MUST be `"bolt11"`; defaults to `"bolt11"`. |
+| `extra.paymentFlow` | Yes | MUST be `"upfront"`. |
+| `extra.invoice` | Yes | Fresh, signed BOLT11 invoice for `amount` on `network` that passes the checks below. |
+
+`maxTimeoutSeconds` MUST be a positive integer. The BOLT11 invoice expiry MUST
+equal `maxTimeoutSeconds` exactly.
+
+`payTo` MUST be the receiver node's valid compressed secp256k1 public key: exactly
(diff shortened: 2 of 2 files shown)


phdargen:
Thanks a lot for the contribution @benthecarman!

We are currently discussing if and how we want to support client prepayments in the x402 protocol that are likely required here, see https://github.com/x402-foundation/x402/pull/2520

Please also note that there is another Lightning proposal here: https://github.com/x402-foundation/x402/pull/1311
If you guys could compare notes and agree on a single draft, that would be ideal

benthecarman:
Hey yeah i saw that impl and this one #1873 but both seem to not even use the invoice preimage for proof of payment so didn't seem to actually use LN's properties.

javierpmateos:
Aligned with the preimage-based approach here ... verifying sha256(preimage) == payment_hash locally, with no node query, is what makes this a portable proof rather than a server-side lookup.

I opened a parallel draft before seeing this one (#2938; the spec PR is closed) and would rather consolidate than compete. Two things from that draft that might be worth folding in:

1. Requirements binding via description_hash. Setting the BOLT11 `h` field to the hash of the canonicalized payment requirements makes the invoice ... already signed by the recipient node key — a commitment to the full terms. That closes the invoice-substitution vector raised in #1311 cryptographically rather than relying on TLS intactness, and adds no new signature scheme to the protocol.

2. Reference implementation: TypeScript, verification rules as pure functions plus a spent-set for replay, LND behind a swappable interface, 15 end-to-end tests against real signed BOLT11 invoices ... https://github.com/javierpmateos/x402-lightning. Happy to point it at whatever this spec settles on, or to contribute test vectors.

On #2520: worth noting it places Lightning in the upfront scheme's payment-proof family, not under exact — all three current drafts (this one, #1311, and mine) are written as exact schemes. If that lands as written, consolidation probably wants to target the upfront structure with an assetTransferMethod discriminator rather than reconciling three exact-scheme documents.

benthecarman:
Updated this PR with the new upfront exact scheme

erikreppel:
This is an AI review generated on behalf of @erikreppel.

Thanks for putting together the preimage-based Lightning proposal. The basic mechanics and deterministic vector look sound: the BIP-122 identifiers are correct, and I independently verified the invoice checksum, timestamp, 300-second expiry, payment hash/preimage relationship, and recovered signer matching `payTo`.

I found the following issues that should be addressed before approval:

### 1. Blocking: bind the proof to the protected request

The spec makes `invoice` dynamic, permits `accepted.extra.invoice` to differ from the current challenge, and validates only signer, amount, network, and timeout. A paid invoice from request A can therefore authorize request B whenever both requests have the same payment terms: the server ignores the dynamic invoice while matching, and the facilitator validates the accepted invoice against those otherwise-identical terms.

This does not satisfy the current `exact` client-submitted proof requirement that a payment be bound to the request: https://github.com/x402-foundation/x402/blob/626df07e/specs/schemes/exact/scheme_exact.md#L44-L56

The spec currently requires an inline BOLT11 `d` field and rejects `h`: https://github.com/x402-foundation/x402/blob/f8542bef42cb7a4fa799baef0a58d9bf71511828/specs/schemes/exact/scheme_exact_bip122.md#L136-L158

Please define a signed, canonical request commitment—BOLT11 `h` is a natural option—including a unique challenge identifier and sufficient authoritative resource context. The client, server, and facilitator validation rules need to define exactly how that commitment is computed and checked, including on paid retries. A negative cross-route/same-price test vector would also be useful.

### 2. Blocking: define actual-amount semantics

The facilitator checks the invoice face amount, but the preimage commits only to the payment hash, not the actual `total_msat` received. BOLT11/BOLT4 permit payments above the invoice amount (up to twice the expected amount), so the checks here cannot prove an exact received value:

- https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#L328-L342
- https://github.com/lightning/bolts/blob/master/04-onion-routing.md#L370-L405

The current `exact` family requires client-submitted methods to state how underpayment and overpayment are handled, including any return path. Please choose and document one of these models:

- Require a trusted receiver receipt/query that includes the amount actually settled; or
- Define the invoice face value as the x402 settled amount, explicitly acknowledge that Lightning-level overpayment is not detectable from the preimage, and specify whether overpayment is accepted and whether any refund exists.

A client-side adapter check helps compliant clients but does not let the facilitator independently establish the actual received amount.

### 3. Blocking: include the network in the replay key

The replay section atomically stores only the payment hash: https://github.com/x402-foundation/x402/blob/f8542bef42cb7a4fa799baef0a58d9bf71511828/specs/schemes/exact/scheme_exact_bip122.md#L274-L285

The family spec requires a canonical consumption key combining the concrete CAIP-2 network and network payment identifier. Using only the hash can cause a mainnet payment and testnet payment using the same preimage to collide. Please define an unambiguous `network + payment_hash` key. Also define “fresh invoice” normatively as using a previously unused payment hash/preimage, rather than merely a different invoice string.

### 4. Important: avoid interpreting a bare price as satoshis

The proposed rule treats bare `"21"` as 21 sats while rejecting `$1` and `1 USD`: https://github.com/x402-foundation/x402/blob/f8542bef42cb7a4fa799baef0a58d9bf71511828/specs/schemes/exact/scheme_exact_bip122.md#L87-L92

That is inconsistent with the repository's generic `Money`/default-asset pricing conventions and can create severe multi-network pricing mismatches (for example, the same bare `"1"` meaning approximately $1 on one network but one satoshi here). I recommend requiring explicit atomic `AssetAmount` pricing (`{ asset: "BTC", amount: "21000" }`) by default, optionally accepting only qualified convenience forms such as `"21 sat"`/`"21 sats"`, and reserving other conversions for registered parsers.

One process note: the PR's latest commit predates the newer client-submitted payment-proof requirements now on `main`, which likely explains these semantic gaps even though there is no textual merge conflict. The earlier maintainer request to consolidate with #1311 also appears unresolved.

phdargen:
Agreed with @erikreppel on the request-binding gap and the related exact-family points.

On alignment with https://github.com/x402-foundation/x402/pull/1311: that draft uses a different mechanic. @benthecarman's approach proves payment with the preimage and a local `SHA-256(preimage) == payment_hash` check. https://github.com/x402-foundation/x402/pull/1311 has the client return the invoice string and the resource server query its Lightning node for settlement. That enables eg QR/second-device wallets that never see the preimage, but it needs a live receiver node at verify time and an invoice/spent-hash store.

Suggest to go with @benthecarman' approach as a simpler starting point.

[#1311](https://github.com/x402-foundation/x402/pull/1311) could be added later as a second method, e.g. `extra.assetTransferMethod = "invoice"`, with this one as `"preimage"` (default, may be omitted). Clients that cannot return a preimage would then select `"invoice"` when it exists

javierpmateos:
On the request-binding gap that @erikreppel raised and @phdargen agreed with: the `description_hash` binding I mentioned upthread is exactly this mechanism. Setting BOLT11 `h` to the hash of the canonicalized requirements makes the invoice ( already signed by the recipient node key ) a commitment to the terms, with no new signature scheme, no session, and no extra round trip. The current spec rejects `h` and requires inline `d`, which is what leaves the cross-request substitution open.

This may also be relevant beyond Lightning: in #3053 @alftom raised needing proof that the client is authorized to use the transaction, and @CarsonRoscoe noted that a bare transaction hash "lacks proving that the intent of that payment was for this endpoint," with SIWX considered but felt heavy-handed at scheme level. Lightning happens to have a native place to put that commitment. Rails whose payment artifact carries no signable field would still need something else, but where one exists, binding it seems lighter than a session layer.

On @erikreppel's "sufficient authoritative resource context": that part matters more than it first appears. In v2 `resource` lives in the `PaymentRequired` envelope and reaches the facilitator only through the client echo, so a facilitator attesting to *what* was paid for is attesting to a client claim unless the commitment covers it. The commitment should be over `{resource, requirements}`, not requirements alone ... my implementation does that, with substituted-resource and missing-resource cases as regression tests: https://github.com/javierpmateos/x402-lightning

Two implementation notes that are easy to get wrong:

-- The binding must anchor to the invoice's signed `h`, not to a convenience copy in `extra`. If verification is conditional on an unsigned field being present, stripping that field disables the check ... I had exactly that bug and it took an external review to catch.
-- Canonicalization has to be pinned ... RFC 8785 in my implementation. Worth versioning it together with the digest function as well, so the rule can evolve without invalidating existing commitments; the compliance-fields extension (#2853) does that with a `canonicalizationVersion` field and it seems like the right pattern here.

Happy to contribute the binding rules and test vectors here rather than maintain a parallel draft. My repo still declares `scheme: "upfront"` from the now-closed #2520; realigning it to `exact` + `extra.paymentFlow: "upfront"` per #3053/#3240 is a rename, and the binding is orthogonal to that packaging either way.

On @phdargen's method split: preimage as default with `"invoice"` as a second method for QR/second-device flows makes sense, and the binding applies to both ... either way the invoice needs to be bound to the request.

benthecarman:
Addressed the four points:

- Added request binding through the signed BOLT11 description hash, using RFC 8785 canonical JSON, which is already used by the repo’s offer/receipt extension. The server computes the expected hash from the actual request; the invoice’s unique payment hash identifies the challenge. Added same-price cross-route rejection vectors.

- Defined the invoice amount as the x402 settled amount. Overpayment is not detectable from the preimage, grants no extra credit, and has no protocol refund path.

- Changed the replay key to `network:payment_hash` and defined fresh invoices as requiring unused payment hashes. 
 
- Required explicit `AssetAmount` pricing by default, with optional qualified satoshi inputs. The JSON, BOLT11 signatures, and request-binding vectors pass validation.

javierpmateos:
Reviewed the binding. I think this is stronger than what I originally proposed.

Hashing the HTTP request itself (`{domain, method, url, bodyHash, headers}`) rather than the requirements object also covers the method and body. And putting the version directly in the signed domain tag (`"...:http:1"`) seems better than having a separate version field, since it can't get out of sync with what was actually signed.

A couple of things also look particularly important to me, since both were bugs in my own implementation: recomputing the expected digest from the actual request instead of trusting `accepted.extra.requestHash`, and making missing binding fields fail rather than silently disabling the check. Those are probably the two easiest ways for this to get accidentally weakened.

There is one point that may be worth clarifying in the text.

The Request Binding section says identical requests can share a request hash "while their invoices remain distinct challenges", and the requirements section says the signed payment hash identifies the challenge. But there doesn't seem to be a rule tying a presented preimage to the specific challenge that issued it.

The test vector:

`replace only requirements.extra.invoice with a fresh valid invoice for the same request and terms → settlement succeeds using the original accepted invoice`

makes sense for the intended paid-retry path. But it also means that two concurrently outstanding challenges for the same request and price are effectively interchangeable: a preimage paid against one outstanding challenge can settle a request issued under another, for the same resource and price.

The replay store still limits this to one successful claim, so this isn't necessarily a problem. But that protection is stateful, while stateless verification is one of the useful properties of the preimage model.

This may be completely fine by design ... whoever has the preimage did pay for that request at that price. I just think it would be useful to state explicitly whether challenge identity is meant to be part of the security model, so implementations don't diverge on whether they need to track issued payment hashes per challenge.

Happy to contribute test vectors for the concurrent-challenge case if useful.

benthecarman:
did some clarifications for @javierpmateos

javierpmateos:
That clears it up ... thanks. I agree that explicitly putting challenge identity out of scope ("servers need not track issued payment hashes per challenge") is the right approach for a stateless model.

The Settlement section also makes the distinction much clearer between stateless proof verification and stateful single-use enforcement, which was there before but only implicitly.

The three concurrent-challenge vectors cover the issue well. Nothing else from me on the binding.

phdargen:
Thanks for the update @benthecarman, looks good to me. 

In general, specs should remain transport-agnostic and this one is fairly http specific. Admittedly it's a bit tricky here because of the request binding, but imo we should at least define the wire type transport-agnostically. So I'd suggest to replace the http-only `extra.requestHeaders` with a more general shape that could also accommodate others like mcp, eg `extra.requestBindingProfile` + `extra.requestBindingParams`:
```
"requestBindingProfile": "http:1",
"requestHash": "c0c8112de741926ff0f2dd25bd2ca499aa5e94aa5e84f2ce7673cdac7d9078fe",
"requestBindingParams": { "headers": [] }
```
If the domain tag is derived as "x402:exact:bip122:bolt11:" + profile, then http:1 reproduces the current tag exactly and none of the existing test vectors change.

The main text should then avoid HTTP-specific language wherever it isn't essential, with one dedicated section for the per-profile binding rules covering both http:1 and mcp:1

vinlim:
@phdargen @benthecarman we're drafting a native on-chain Bitcoin method, provisionally named `onchain`, for `exact` on the same `bip122` mainnet identifier. Three questions about how it would work with `bolt11`.

We propose a client-submitted method with the `upfront` flow. The server issues a fresh, request-bound address in `payTo`, and the payer signs and broadcasts. The payload identifies the payment output with `txid` and `vout`. The merchant checks that the payment matches the request and has enough confirmations as adverised depth. Verification and settlement can run locally or delegate to external provider. It will always carry an explicit `extra.assetTransferMethod: "onchain"`, so `bolt11` stays as the default.


1. Bitcoin can produce signed but unbroadcast transactions. We propose payer broadcast to support ordinary send-payment wallet interfaces and reuse merchants’ existing payment monitoring. Is that an acceptable use of the [client-submitted family](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact.md#client-submitted-payment-proof)?
2. This spec currently puts amount for asset "BTC" in millisatoshis. The chain's unit is the satoshi, so an on-chain method would be 1000 apart from bolt11 under one scheme, network and asset. We can implement either. Which would be the prefered mechanism?
3. Should the on-chain method be added to scheme_exact_bip122.md, or documented in a sibling file with shared mechanism rules and method dispatch?

Thanks

javierpmateos:
@vinlim ... on the units question specifically: if `amount` means millisatoshis for `bolt11` but satoshis for `onchain`, while `scheme`, `network`, and `asset` stay the same, then there's a 1000x difference hidden behind the payment method.

That means a consumer can't interpret `amount` correctly without first checking the method.

@erikreppel raised a similar issue in point 4 above around parsing bare numbers.

Whichever unit we choose, I think it would be cleaner if it stayed consistent across methods on this network. And if there's a real reason the units need to differ, then I'd rather make the unit explicit in the payload instead of having consumers infer it from the method.

vinlim:
> @vinlim ... on the units question specifically: if `amount` means millisatoshis for `bolt11` but satoshis for `onchain`, while `scheme`, `network`, and `asset` stay the same, then there's a 1000x difference hidden behind the payment method.
> 
> That means a consumer can't interpret `amount` correctly without first checking the method.
> 
> @erikreppel raised a similar issue in point 4 above around parsing bare numbers.
> 
> Whichever unit we choose, I think it would be cleaner if it stayed consistent across methods on this network. And if there's a real reason the units need to differ, then I'd rather make the unit explicit in the payload instead of having consumers infer it from the method.

@javierpmateos we are aligned in the sense that we would also prefer it to be consistent across this network. Also, weare open to adopting millisatoshis for `onchain` as well. We would just like to hear from other contributors before settling on the shared unit.

benthecarman:
Addressed the transport feedback: replaced requestHeaders with requestBindingProfile and requestBindingParams, and defined http:1 and mcp:1 profiles. The existing HTTP invoice and hash vectors are unchanged. Added MCP vectors covering tool, argument, server, and metadata changes.

On units, I’d keep millisatoshis consistent across methods using the same network and BTC asset. Added that requirement; an onchain method would restrict amounts to multiples of 1,000.

I’d keep the onchain proposal separate so we can get this one in.

phdargen:
Thanks for the update @benthecarman, the transport changes look good!

On the units: networks consistently use the smallest atomic unit. Thus I think onchain BTC should be satoshis. Lightning could also use satoshis on the wire and convert to millisatoshis internally if needed, which is what I believe [MPP](https://mpp.dev/payment-methods/lightning/charge) is doing. Having units differ among assetTransferMethods also seems weird.

Related: Should Lightning have its own network id, so it stays distinct from native BTC? Then it could use millisatoshis 

cc @erikreppel

benthecarman:
> On the units: networks consistently use the smallest atomic unit. Thus I think onchain BTC should be satoshis. Lightning could also use satoshis on the wire and convert to millisatoshis internally if needed, which is what I believe [MPP](https://mpp.dev/payment-methods/lightning/charge) is doing. Having units differ among assetTransferMethods also seems weird.
> Related: Should Lightning have its own network id, so it stays distinct from native BTC? Then it could use millisatoshis

Yeah bitcoin using msats isn't elegant, if we can split them by network id, then i think that'd be ideal, msats for LN and sats for on-chain

erikreppel:
I'm inclined to agree with @phdargen

phdargen:
How about using eg `lightning:000000000019d6689c085ae165831e93` then?

benthecarman:
> How about using eg lightning:000000000019d6689c085ae165831e93 then?

According to codex said `lightning:` was too long for `CAIP-2`, used `lnbtc:` instead

phdargen (approved):
Sounds good, let's get this in

PR #2862, opened 2026-07-15
fix(svm): migrate SVM mechanism to async RPC for solana>=0.40 compatibility

TheCrazyGM:
## Summary

solana 0.40.0 removed the synchronous `Client` from `solana.rpc.api`, shipping only `AsyncClient` from `solana.rpc.async_api`. The x402 SVM mechanism hardcoded the sync import and crashed on any import of the svm module — even just accessing `NETWORK_CONFIGS` from `constants.py`.

This PR migrates the entire SVM signer/facilitator/client stack to use async RPC calls, with `try/except` imports for backward compatibility with solana<0.40.

### Mechanism (core fix)
- **signers.py**: `FacilitatorKeypairSigner.simulate_transaction`, `send_transaction`, `confirm_transaction` are now `async`; `time.sleep` → `asyncio.sleep`
- **signer.py**: `FacilitatorSvmSigner` protocol updated — RPC methods now `async`
- **mint_cache.py**: `get_cached_mint_metadata` now `async` (`await get_account_info`)
- **exact/client.py, exact/v1/client.py**: `create_payment_payload` now `async`; `await get_latest_blockhash` and `get_cached_mint_metadata`
- **exact/facilitator.py, exact/v1/facilitator.py**: `verify` and `settle` now `async`; `await` on signer RPC methods

### Framework integration
- **facilitator_base.py**: `_verify_v1/v2`, `_settle_v1/v2` detect async scheme methods and yield `scheme_verify`/`scheme_settle` commands for the caller to `await`
- **facilitator.py**: Async facilitator (`x402Facilitator`) awaits scheme coroutines; sync facilitator (`x402FacilitatorSync`) raises `TypeError` with clear message directing users to `x402Facilitator`

### Config & tests
- **pyproject.toml**: Added `asyncio_mode = "auto"` for pytest async test detection
- All SVM test files updated: `async def` test methods, `AsyncMock` for RPC client mocks

## Why

The solana-py package dropped the sync `Client` entirely in 0.40.0. Making methods async is the correct long-term fix since the x402 Python server already runs in an async framework (FastAPI/uvicorn). The `try/except` import pattern preserves backward compatibility with solana<0.40.

## Backward compatibility

```python
try:
    from solana.rpc.async_api import AsyncClient as SolanaClient
except ImportError:
    from solana.rpc.api import Client as SolanaClient  # fallback for <0.40
```

Works with both `solana>=0.40.0` (async-only) and `solana>=0.36.0,<0.40.0` (sync Client).

## Test results

1739 passed, 65 skipped, 0 failed.

## Breaking change note

SVM facilitator schemes (`ExactSvmFacilitatorScheme`, `ExactSvmSchemeV1`) are now async. Users calling `verify()`/`settle()` directly must use `await` or switch from `x402FacilitatorSync` to `x402Facilitator`. The sync facilitator raises a clear `TypeError` with migration guidance.


python/x402/mechanisms/svm/exact/client.py
```diff
@@ -83,7 +91,7 @@ def _get_client(self, network: str) -> SolanaClient:
         self._clients[caip2_network] = client
         return client
 
-    def create_payment_payload(
+    async def create_payment_payload(
```
lgalabru, line 96: This method is now async, but x402Client and x402ClientSync still call scheme methods without awaiting them. Both fail because the result is a coroutine instead of a payload. Please update the client driver to support async scheme methods, and make the sync client reject them clearly. Please cover both V1 and V2 with tests.

python/x402/facilitator.py
```diff
@@ -144,8 +144,11 @@ async def verify(
         result = None
         try:
             while True:
-                _, hook, ctx = gen.send(result)
-                result = await self._execute_hook(hook, ctx)
+                phase, hook, ctx = gen.send(result)
+                if phase == "scheme_verify":
+                    result = await hook
```
lgalabru: If this await raises, the exception skips the generator, so on_verify_failure never runs. Please send the exception back with gen.throw(). The same issue exists in settle.

lgalabru (changes requested):
Thanks @TheCrazyGM, this contribution is headed in the right direction.
Could you please review my feedback?

TheCrazyGM:
@lgalabru I'm on it, thanks for looking it over.

PR #2876, opened 2026-07-16
examples(python): demonstrate service metadata in the bazaar server example

rascal-3:
## Summary

The TypeScript advanced bazaar example demonstrates `serviceName` / `tags` on the route config, but the Python bazaar example doesn't — even though the Python SDK's `RouteConfig` fully supports `service_name` / `tags` / `icon_url`. In practice this metadata is the difference between a catalog listing that shows up as a bare domain and a named, tagged service page, and it's easy to miss since values that violate the indexer caps are soft-dropped silently (per the validation rules in `specs/extensions/bazaar.md`).

- add `service_name="Weather API"` / `tags=["weather", "api"]` to both routes in `examples/python/servers/bazaar/main.py`, mirroring the TS example
- README: short section on service metadata, the keep-it-identical-across-routes grouping behavior, and the ≤32-char / ≤5-tags / absolute-URL caps

Verified the end-to-end effect on a production FastAPI x402 service this week (settle carrying the declaration + `resource.serviceName` → enriched catalog listing); details in x402-foundation/x402#2112.

Refs #2112

### DIFF
--- a/examples/python/servers/bazaar/README.md
+++ b/examples/python/servers/bazaar/README.md
@@ -40,6 +40,8 @@ async def get_weather(city: str) -> dict:
 
 Note that the x402 route key uses `:city` (Express convention) while the FastAPI handler uses `{city}` (FastAPI convention). The x402 middleware handles the `:city` matching; FastAPI handles the `{city}` extraction for your handler.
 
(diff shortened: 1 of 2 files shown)


phdargen (approved):
Thanks for the contribution @rascal-3

PR #2884, opened 2026-07-17
fix(evm): auto-wrap any eth_account BaseAccount, not just LocalAccount

0xkurious:
## Problem

`_wrap_if_local_account` (duplicated in the `exact`, `upto`, and
`batch_settlement` EVM client schemes) only auto-wraps a signer in
`EthAccountSigner` when it is an instance of `eth_account.signers.local.LocalAccount`:

```python
if isinstance(signer, LocalAccount):
    return EthAccountSigner(signer)
return signer
```

Third-party wallet-provider adapters that extend `eth_account`'s abstract
`BaseAccount` directly — rather than being a raw-private-key `LocalAccount` —
fall through unwrapped. Coinbase AgentKit's `EvmWalletProvider.to_signer()` is
one concrete example: it returns an `EvmWalletSigner(BaseAccount)` for any
custody backend (CDP, smart wallets, or a custom signer such as an AWS
KMS-backed one).

When such a signer reaches `ExactEvmScheme._sign_authorization`, it's called
with x402's positional `ClientEvmSigner` convention —
`sign_typed_data(domain, types, primary_type, message)` — but
`EvmWalletSigner.sign_typed_data` (and `eth_account`'s own convention) expects
keyword arguments `(domain_data, message_types, message_data, full_message)`.
`primary_type` lands in the `message_data` slot and `message` lands in
`full_message`, so the "already has a message" branch is taken with an
incomplete dict, and `eth_account.messages.encode_typed_data` raises
`KeyError: 'types'`.

Discovered while wiring an AWS KMS-backed signer (subclassing `BaseAccount`,
not `LocalAccount`) into an x402 buyer paying a real x402 v2
`GatewayWalletBatched` payment requirement.

## Fix

Broaden the `isinstance` check from `LocalAccount` to `BaseAccount`
(`LocalAccount`'s own parent class) in all three occurrences. Every
`LocalAccount` is already a `BaseAccount`, so this is a strict superset — no
behavior changes for existing raw-private-key callers, and non-`LocalAccount`
`BaseAccount` signers are now correctly wrapped instead of crashing.

## Testing

Added `TestBaseAccountAutoWrap` to `tests/unit/mechanisms/evm/test_client.py`
with a minimal `BaseAccount`-only signer (not a `LocalAccount`), mirroring
third-party wallet-provider adapters. Verified the new tests fail without the
fix (`KeyError`/`ValueError` from `eth_account.messages`) and pass with it.
[...]

PR #2885, opened 2026-07-17
fix(evm): sign EIP-3009 domain against extra.verifyingContract, not asset

0xkurious:
## Problem

`_sign_authorization` in the exact scheme (both `mechanisms/evm/exact/client.py`
V2 and `mechanisms/evm/exact/v1/client.py` V1 legacy) reads `extra["name"]` and
`extra["version"]` from the seller's payment requirements to build the EIP-712
domain — but hardcodes `requirements.asset` (the token contract address) as
the domain's `verifyingContract`, ignoring `extra["verifyingContract"]`
entirely:

```python
domain, types, primary_type, message = build_typed_data_for_signing(
    authorization,
    chain_id,
    requirements.asset,   # always the token address, even when the seller
    name,                 # specified a different verifyingContract in `extra`
    version,
)
```

This is correct only when the token contract itself is the EIP-3009 verifier
(the common case). It silently signs against the **wrong domain** whenever a
seller settles through a different verifying contract than the token —
Circle Gateway's `GatewayWalletBatched` batch-settlement scheme is a concrete,
real-world example: its `extra` looks like

```json
{"name": "GatewayWalletBatched", "version": "1",
 "verifyingContract": "0x77777777Dcc4d5A8B6E418Fd04D8997ef11000eE"}
```

— a contract address distinct from the USDC token (`0x833589fC...`). Signing
against `requirements.asset` here produces a well-formed but cryptographically
different signature: the EIP-712 domain separator depends on
`verifyingContract`, so a different value changes the final hash entirely.
The real verifier (Circle's Gateway facilitator) recomputes the domain with
its own contract address and rejects the mismatch as an invalid signature.

## Fix

Prefer `extra["verifyingContract"]` when the seller provides one, falling back
to `requirements.asset` only when absent — unchanged behavior for plain token
EIP-3009 domains (the existing test suite already covers this fallback path
and continues to pass unmodified). Applied to both the V2 and V1 legacy
clients, since both had the identical bug.

## Testing

Added `TestSigningDomainVerifyingContract` to
`tests/unit/mechanisms/evm/test_client.py`:
- `test_signs_against_custom_verifying_contract_when_provided` — independently
  recomputes the EIP-712 digest against both the seller-provided
  `verifyingContract` and the asset address, and asserts the produced
  signature recovers against the former, not the latter.
- `test_falls_back_to_asset_when_no_verifying_contract_given` — regression
  guard for the existing plain-token-domain behavior.

Verified the new test fails without the fix (signature recovers against the
asset address instead) and passes with it. Full `tests/unit/mechanisms/evm`
suite (663 tests) passes.

**This fix is verified against a real settlement, not just unit tests**: I
reproduced the exact failure against a live x402 v2 `GatewayWalletBatched`
payment requirement (a production seller on Base mainnet), signed with an AWS
KMS-backed signer. Before this fix, the seller rejected the payment as
`PAYMENT_INVALID_SIGNATURE`. After it, the payment settled and the request
reached the seller's own business-logic validation (a `400` for missing query
parameters, unrelated to payment) — confirming the facilitator accepted the
signature.

## Disclosure

Most of this change (diagnosis, fix, and tests) was written with AI assistance
(Claude Code), reviewed and verified by me — including confirming the
failure/fix behavior by running the test suite and a real settlement before
and after the change — before opening this PR, per this repo's AI-assisted
contribution guidelines.


0xkurious:
Rebased onto `main` to resolve the conflict from #2884 merging (both touched `tests/unit/mechanisms/evm/test_client.py`). No changes to the fix itself — just picked up `main`'s new `TestBaseAccountAutoWrap` class alongside this PR's `TestSigningDomainVerifyingContract`. Full `tests/unit/mechanisms/evm` suite (665 tests) still passes.

0rkz:
A code-reading note on where this lands relative to the rest of the repo — three observations, no security claim:

1. **No reference facilitator reads `extra.verifyingContract`.** Every verify implementation rebuilds the EIP-712 domain from `requirements.asset` and takes only `name`/`version` from `extra`: TS v2 (`mechanisms/evm/src/exact/facilitator/eip3009.ts` — `getAddress(requirements.asset)` → `verifyingContract`), TS v1 (`exact/v1/facilitator/scheme.ts`), Python v2 (`mechanisms/evm/exact/facilitator.py` — `normalize_address(requirements.asset)`), Python v1, and Go (`exact/facilitator/eip3009.go` — `NormalizeAddress(requirements.Asset)`). So a client signing against a seller-supplied `extra.verifyingContract` produces a digest no reference facilitator reproduces — the payment is rejected as an invalid signature. After this merges, the Python client diverges not just from the TS and Go clients (both still pin `requirements.asset`) but from both Python facilitator versions in this same repo. The reverse direction is real too — per the docstring you hit a live `GatewayWalletBatched` requirement that rejected an `asset`-domain signature. So the tension isn't that the field is wrong, it's that today it only works against verifiers outside this repo, while the change makes the Python client the one SDK that can't transact with the facilitators inside it.

2. **The field isn't in the spec.** `specs/schemes/exact/scheme_exact_evm.md` enumerates the eip3009 `extra` keys as `assetTransferMethod`, `name`, `version`; `verifyingContract` appears nowhere in the file. Merging the client first would make behavior lead spec.

3. **The repo already has a precedent for "verifier is not the token," and it has a stricter shape.** The test docstring names Circle Gateway's `GatewayWalletBatched` — a verifier that isn't the token contract. The existing case of that shape here is batch-settlement, and it doesn't take the address off the wire: `getBatchSettlementEip712Domain` pins `verifyingContract: getAddress(BATCH_SETTLEMENT_ADDRESS)` in `mechanisms/evm/src/batch-settlement/utils.ts`, a vetted constant for x402's own contract. Not an argument that Gateway belongs in batch-settlement — it's a different contract — but the precedent for a non-token verifying contract in this repo is a pinned constant inside a dedicated mechanism, not a free-form `extra` passthrough on `exact`.

Suggested order: spec the field and its constraints first — including what a facilitator is expected to accept — then land client support in all three SDKs together, so the same challenge keeps producing the same signature everywhere.

*(Disclosure: I work on PayPerByte; we run a seller/facilitator stack on x402, so cross-SDK signature agreement is the part I care about.)*

0xkurious:
@0rkz thanks for the thorough read — this is exactly the kind of cross-SDK/facilitator consideration I didn't fully account for. Responding point by point:

**On the spec gap (2)**: agreed, straightforward miss on my part. I'll add `verifyingContract` as a documented optional `extra` key in `scheme_exact_evm.md` in this same PR.

**On facilitator/cross-SDK divergence (1)**: I'd push back gently on framing this as the fix creating divergence — the reference facilitators (all languages) share the same bug this PR fixes in the Python client: they only produce a *valid* signature when `verifyingContract == asset`, which happens to be true for plain-token settlement but is cryptographically wrong whenever a seller genuinely settles through a distinct contract (which is the whole reason this PR exists — a real, production `GatewayWalletBatched` rejection). I agree the fix should extend to the facilitators and to TS/Go for consistency, and I'm happy to open follow-up PRs for those — but I don't think that should block this one, since today *no* client in this repo can correctly pay a seller with a genuinely distinct verifying contract, and this PR fixes that for at least one path.

**On the trust model (3)** — this is the one I think is genuinely the strongest point. You're right that a bare `extra`-sourced value is a weaker trust model than the pinned-constant precedent in batch-settlement, even though I don't think it's directly exploitable on its own (a signature is only useful against a contract that already has custodial authority over the payer's funds via some separate action, e.g. a prior Gateway deposit). Rather than hardcoding an allowlist of specific external contracts (which doesn't fit a protocol-agnostic SDK), I'd propose making this opt-in at the call site: default behavior stays exactly as today (ignore `extra.verifyingContract`, always use `requirements.asset`), and callers who need to trust a seller-specified verifying contract (like us, for Gateway) have to explicitly pass something like `trust_verifying_contract=True` (or a validator callback) to opt in. That keeps the safe default for everyone else and makes the trust decision visible and intentional at the integration point, rather than implicit.

Happy to push that revision if this shape sounds right — want to align before I touch the diff again.

0rkz:
Yes — that shape sounds right, and the default-unchanged half is the part that matters: a payer who never opts in keeps signing against `requirements.asset`, which is what every verifier in this repo reproduces today.

One preference between your two options, argued from the client's own surface rather than taste: **the validator callback, not the boolean.**

- A scheme instance is registered once per network and then reused for every payment on it — `self._schemes: dict[Network, dict[str, SchemeNetworkClient]]` (`client_base.py:237`), populated by `register` (`:253`). So a constructor-level `trust_verifying_contract=True` is blanket trust of any seller-supplied address on that network for the life of the client, not a per-counterparty decision. A callback lets an integrator pin the one contract they actually mean, which is the pinned-constant discipline you already agreed is the stronger shape.
- Caller-supplied decisions are already callables here: `PaymentPolicy` and `PaymentRequirementsSelector` on `x402ClientConfig`, and `BeforePaymentCreationHook`, which receives `selected_requirements` and can return `AbortResult` (`schemas/hooks.py:343`, `:21`).
- A callback subsumes the flag — one that returns `True` unconditionally *is* `trust_verifying_contract=True` — so shipping only the callback avoids maintaining two opt-in surfaces.

On placement: `SchemeNetworkClient.create_payment_payload(self, requirements)` is protocol-pinned with no options parameter (`interfaces.py:116`), so the constructor looks like the natural seam; the SVM exact schemes already take optional config there (`rpc_url`, `mechanisms/svm/exact/client.py:48`).

On the facilitator framing — your read is fair, and I'd revise mine. The asset-as-verifying-contract assumption sits on both sides equally: TS v2 `exact/facilitator/eip3009.ts:93`, TS v1 `evm/src/exact/v1/facilitator/scheme.ts:296`, Python v2 `exact/facilitator.py:185`, Python v1 `exact/v1/facilitator.py:167`, Go `exact/facilitator/eip3009.go:47` all rebuild the domain from `requirements.asset`. Calling that divergence the PR creates described the transitional state, not its cause; "one assumption, fixed on one side first" is the more accurate way to put it.

That transitional state is worth a line in the changelog or the `extra` entry you're adding, because it still holds at current `main`: nothing in the repo reads `extra.verifyingContract`, so once a payer opts in against a seller that advertises one, in-repo facilitators will reject that payment as an invalid signature until the follow-ups land. Better read than discovered.

*(Disclosure: I work on PayPerByte; we run a seller/facilitator stack on x402, so cross-SDK signature agreement is the part I care about. Drafted with AI assistance.)*

0xkurious:
@0rkz pushed the revision — went with the validator callback, as discussed:

**`VerifyingContractValidator = Callable[[str, PaymentRequirements], bool]`**, passed as an optional `verifying_contract_validator` kwarg to `ExactEvmScheme.__init__`/`ExactEvmSchemeV1.__init__` (same shape on both, mirroring the SVM scheme's existing `rpc_url` optional-constructor-kwarg precedent). It receives the candidate `extra["verifyingContract"]` address and the full requirements, returns `True` to trust it.

Default is `None` → the field is never trusted, signing always falls back to `requirements.asset` — identical to pre-#2885 `main` behavior for anyone who doesn't opt in. `_resolve_verifying_contract()` centralizes the logic (candidate present + validator present + validator approves → use candidate; any of those false → asset), called from `_sign_authorization` in both V2 and V1 clients.

Also landed, in separate commits for reviewability:
- `specs/schemes/exact/scheme_exact_evm.md` now documents `extra.verifyingContract` under the `eip3009` section (was completely unspec'd before, as you flagged), including explicit MUST NOT-trust-by-default language.
- Reworded the changelog fragment to describe the final opt-in behavior instead of the intermediate unconditional-trust state.
- Extended `TestSigningDomainVerifyingContract` (fixed the one existing test that assumed unconditional trust, added 4 new cases: validator-approves, no-validator-default-deny, validator-rejects, and exact call-signature assertion via `Mock`), and added an equivalent test file for the V1 client, which had no test coverage at all before this.

Full `tests/unit/mechanisms/evm` suite: 675 passed (up from 665 pre-PR).

On the facilitator/cross-SDK divergence point from earlier: still open, still agree it's real, still think it shouldn't block this PR per the reasoning above — happy to open those as separate follow-ups once this lands. Let me know if the callback shape here looks right or if you'd tweak anything before this goes up for a real review pass.

0rkz:
Checked the revision at head `1873fa59` (`1873fa59bc437a6582b02b7c32b7c9191e343c0f`). The callback shape matches what we aligned on: `verifying_contract_validator` as an optional constructor kwarg on both `ExactEvmScheme` and `ExactEvmSchemeV1`, default `None`, falling back to `requirements.asset` whenever the candidate is absent, the validator is absent, or the validator returns `False`. `specs/schemes/exact/scheme_exact_evm.md` now documents `extra.verifyingContract` with explicit MUST-NOT-trust-by-default language.

One scope note: the new V1 test file closes the V1-client test-coverage gap, which is real and welcome, but that's distinct from the facilitator/cross-SDK divergence we raised — TS v1/v2, Python v1/v2, and Go still all rebuild the domain from `requirements.asset` only, and you've already flagged that one as still open rather than closed by this PR.

0xkurious:
Opened the facilitator-side follow-up flagged above (both instances — [thread 1](https://github.com/x402-foundation/x402/pull/2885#issuecomment-5112127349), [thread 2](https://github.com/x402-foundation/x402/pull/2885#issuecomment-5302313356)): #3174. Every facilitator in this repo (this Python one included) still rebuilds the EIP-712 domain from `requirements.asset` only, so a client that opts in here still gets rejected by this facilitator's own `verify`/`settle`. #3174 adds the same `verifying_contract_validator` opt-in on the facilitator side (V1 + V2), used consistently for both the signature check and the actual `transferWithAuthorization` call during settle. It's stacked on this branch (`fix/exact-scheme-wrong-verifying-contract`) and shouldn't merge before this one does. TS/Go facilitator and client follow-ups are still open, tracked separately per the discussion above.

0xkurious:
TS follow-up for the same gap, both directions (client didn't have the option at all; facilitator ignored the field): #3175. Mirrors this PR's opt-in `verifyingContractValidator` design exactly.

0xkurious:
Go follow-up for the same gap, both directions (client + facilitator): #3176. Same opt-in design as #3174/#3175, adapted to Go idiom -- functional option on the client side (its config is a shared type alias with `upto`), a direct struct field on the facilitator side (already scheme-dedicated).

0xkurious:
Thanks for the thorough review throughout this thread, @0rkz — the facilitator/cross-SDK divergence you flagged turned into three follow-ups (#3174 Python, #3175 TS, #3176 Go) that wouldn't have happened without you catching it. Appreciated.

0rkz:
Thanks for the credit, @0xkurious, and for turning the three languages around this fast.

Checked all three at their pinned heads:

- **#3174 (Python)** at `9ffcb8b7053f64b07baab8751d98573fce64bf12` — confirmed the branch is genuinely stacked on this PR: its commit history includes this PR's final commit (`1873fa59bc437a6582b02b7c32b7c9191e343c0f`) verbatim, not a rebase or reimplementation. `_resolve_verifying_contract()` is invoked in both `_verify()` and `settle()`, for both the V2 facilitator (`mechanisms/evm/exact/facilitator.py`) and the V1 facilitator (`mechanisms/evm/exact/v1/facilitator.py`) — the resolved address replaces `requirements.asset` at both call sites, so a trusted candidate is verified and settled against consistently rather than checked under one domain and transferred under another.
- **#3175 (TS)** at `24181d911226f5f357feffc4bcc93bf943862bcd` — same shape: `resolveVerifyingContract()` feeds both `verifyEIP3009`/`settleEIP3009` (V2) and the V1 facilitator's `_verify`/`Settle`, including the post-settlement `Transfer`-event check, which is easy to miss and wasn't missed here.
- **#3176 (Go)** at `02f3b8004f2b66ca41c42a62b9ba564f294d6fe7` — same shape: `resolveVerifyingContract()` feeds `verifyEIP3009`/`settleEIP3009` (V2) and `verify`/`Settle` (V1).

Default behavior is consistent and safe across all three: candidate absent, validator absent, or validator returns false all fall back to `requirements.asset`/`Asset`, matching pre-existing behavior — no facilitator that leaves the validator unconfigured newly trusts `extra.verifyingContract`.

One divergence is worth a look, but it's Go-specific and only actionable in that diff, so I've left the detail on #3176 rather than here: once a validator approves a candidate, the three languages don't validate its address format equally before using it. Python and TS both reject a malformed candidate outright; Go's `NormalizeAddress()` doesn't check length or hex, and nothing downstream errors either — the value gets silently coerced into a different, well-formed address. Not introduced by these PRs, but #3176 is the first place that normalizer meets a seller-influenced value.

0rkz:
One correction to my note above, since it restated the Go point: the claim that a malformed `verifyingContract` candidate gets silently coerced into a different valid address is wrong. It's rejected at EIP-712 domain hashing, before the conversion I described is reached — details and the withdrawal are on #3176. What's left there is a diagnostics difference, not a missing backstop.

The rest of that comment stands as written.

0xkurious:
Closing this — @phdargen's review on the facilitator follow-up (#3174) invalidates the premise this PR was built on: https://github.com/x402-foundation/x402/pull/3174#issuecomment-5304764975

Three points, in order of weight:
1. Circle Gateway doesn't actually settle via an on-chain `transferWithAuthorization` against a distinct verifying contract — the EIP-3009 authorization is used off-chain to update Circle's internal DB, and an independent facilitator can't do that; it has to go through Circle's API. The GatewayWalletBatched scenario this PR was built to fix doesn't occur in practice.
2. Without that motivating case, there's no legitimate use for `extra.verifyingContract`: a passive receiver is just a regular `payTo`, and a receiver that needs to act on the authorization should use `receiveWithAuthorization` (msg.sender == to), not `transferWithAuthorization`, which is front-runnable and can strand funds.
3. That leaves `extra.verifyingContract` as a server-provided field with no real purpose except added attack surface, even gated behind opt-in validation.

Closing rather than merging a fix for a bug that doesn't exist in the scenario that motivated it.

PR #2886, opened 2026-07-17
spec: eip3009-forwarder-split — atomic non-custodial fee splitting for exact on EVM (with a JPY stablecoin worked example)

cipherwebllc:
Per CONTRIBUTING's "PR 1 (specs only)" flow, this adds a spec for an atomic, non-custodial fee-splitting settlement under the `exact` scheme on EVM, via `extra.assetTransferMethod: "eip3009-forwarder-split"`.

**Why**: marketplaces and managed facilitators need per-payment fees. Today that means custody (receive-then-forward), a second transfer (two signatures, non-atomic), or unverifiable vendor extensions. This proposal makes the **EIP-3009 authorization nonce a commitment hash over the exact split** (merchant, amounts, fee receiver, validity window, chain, forwarder), so one payer signature atomically binds the split, an immutable forwarder contract enforces it on-chain, and the facilitator never holds funds.

**Production evidence**: running since June 2026 on a JPY-stablecoin marketplace (open-pay.jp) with settled Polygon mainnet sales, including agent-to-agent commerce — an AI agent (Claude via MCP) purchased a consultation from another AI agent (elizaOS on Internet Computer, fronted by a 402 gateway) for 2 JPYC: tx [`0xa9e6c6…da3249`](https://polygonscan.com/tx/0xa9e6c6a9ce10fd26ec2fab0d367de31d7fb0918c79d5e932b8566816ecda3249). Open-source buyer SDK/MCP ([openpay-x402-sdk](https://www.npmjs.com/package/openpay-x402-sdk), [openpay-x402-mcp](https://www.npmjs.com/package/openpay-x402-mcp)), an importable seller gate, and a deploy-in-5-minutes gateway template exist and would form the basis of the PR-2 reference implementation.

The worked example intentionally uses **JPYC v3 on Polygon** — the `exact` scheme already supports arbitrary EIP-3009 assets, and we would like the ecosystem's examples/test vectors to include a JPY-denominated case.

Open questions for reviewers are listed at the end of the spec (extension point vs. distinct scheme variant, `commitVersion` namespacing, receipt format scope). Happy to iterate on naming and structure.

### DIFF
--- a/specs/schemes/exact/scheme_exact_evm_forwarder_split.md
+++ b/specs/schemes/exact/scheme_exact_evm_forwarder_split.md
@@ -1,0 +1,177 @@
+# `eip3009-forwarder-split`: atomic fee-splitting settlement for the `exact` scheme on EVM
+
+Status: Proposal (spec-only PR per CONTRIBUTING "PR 1")
+Target: `specs/schemes/exact/` (extension of `scheme_exact_evm.md` via `extra.assetTransferMethod`)
+Authors: OpenPay (cipherweb LLC) — production operator of a JPY-stablecoin x402 facilitator
+
+## Motivation
+
+Marketplaces and managed facilitators need to take a service fee per payment. Today the
+`exact` scheme on EVM defines a single-recipient transfer (`transferWithAuthorization` to
+`payTo`), so fee collection requires either (a) custody (facilitator receives, then
+forwards), (b) a second transfer (two signatures, non-atomic), or (c) vendor-specific
+extensions that conformant clients cannot verify.
+
+This proposal specifies a **non-custodial, single-signature, atomic fee split**: the payer
+signs one EIP-3009 `receiveWithAuthorization` to an immutable forwarder contract, and the
+**authorization nonce itself is a commitment hash over the exact split** (merchant,
+merchant amount, fee receiver, fee amount, validity window, chain, forwarder). The
+forwarder recomputes the commitment on-chain and reverts unless the split matches, so:
+
+- the facilitator never holds funds (the token contract pays the forwarder, which splits
+  in the same transaction);
+- neither the facilitator nor the merchant can alter destination or amounts after
+  signing;
+- clients can verify the full economics of a payment offline before signing.
+
+It has been in production since June 2026 on a JPY-stablecoin marketplace (details in
+*Production evidence*), including agent-to-agent purchases where one AI agent bought a
+consultation from another AI agent.
+
+## PaymentRequired additions
+
+`accepts[]` entries use the existing `exact` scheme shape with:
+
+- `payTo`: the forwarder contract address (not the merchant).
+- `maxAmountRequired`: `merchantValue + feeValue` (the total the payer authorizes).
+- `extra.assetTransferMethod`: `"eip3009-forwarder-split"`.
+- `extra.forwarderSplit`:
+
+| Field | Type | Meaning |
+|---|---|---|
+| `forwarder` | address | Immutable split-forwarder contract (equals `payTo`) |
+| `merchant` | address | Final recipient of the price |
+| `merchantValue` | uint256 (string) | Amount forwarded to `merchant` |
+| `feeReceiver` | address | Final recipient of the fee |
+| `feeValue` | uint256 (string) | Amount forwarded to `feeReceiver` |
+| `commitVersion` | bytes32 | Version tag of the commitment encoding (below) |
+
+`extra.name` / `extra.version` carry the token's EIP-712 domain as in `exact` today, so
+**any EIP-3009 token works unchanged** — see the JPY stablecoin worked example.
+
+## Payment payload
+
+The payload is a standard EIP-3009 `ReceiveWithAuthorization` typed-data signature:
+
+- `from`: payer
+- `to`: `forwarder`
+- `value`: `merchantValue + feeValue`
+- `validAfter` / `validBefore`: validity window
+- `nonce`: **the split commitment hash** (this is the core of the proposal):
+
+```
+nonce = keccak256(abi.encode(
+  bytes32 commitVersion,
+  address from,
+  address merchant,
+  uint256 merchantValue,
+  address feeReceiver,
+  uint256 feeValue,
+  uint256 validAfter,
+  uint256 validBefore,
+  bytes32 intentSalt,
+  uint256 chainId,
+  address forwarder
+))
+```
+
+`intentSalt` is a client-generated random 32-byte value providing uniqueness across
+otherwise-identical intents. `commitVersion` pins the encoding (the production deployment
+uses `keccak256("openpay.eip3009.forwarder.v1")` = `0x7ff4e43ca5ec8a7745cdb456a45dc2f4787e1a8dc0ab9121c9941bfb1028ce89`; the value would be standardized or
+namespaced per deployment — open question below).
+
+Because EIP-3009 nonces are single-use per (`from`, `nonce`), replay is excluded by the
+token contract itself.
+
+## Verification logic (facilitator / server)
+
+1. Validate the typed-data signature for the token's EIP-712 domain.
+2. Recompute the commitment hash from `extra.forwarderSplit` + authorization fields and
+   require it to equal `nonce`.
+3. Require `value == merchantValue + feeValue`, `to == forwarder`, time window open,
+   and (via RPC) that the authorization is unused.
+
+## Settlement logic
+
+The facilitator (or anyone) submits one transaction to the forwarder:
+`settle(authorization, signature, split params)`. The forwarder:
+
+1. recomputes the commitment hash from the split params and requires it to equal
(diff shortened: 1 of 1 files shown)


cipherwebllc:
All commits are now signed and verified ✅ — the branch was rebuilt into a single verified commit (68d7f67) with the spec content unchanged from the original proposal. (The PR was briefly auto-closed mid-rebase and has been reopened.)

Ready for review whenever the team has bandwidth. Happy to iterate on the spec format or split anything out if that helps. 🙏

whawk46 (commented):
Reviewed as an EIP-3009 facilitator operator (we run 3009 + a permit2 witness proxy in production on Flare). The trust architecture here is **right**: welding the split into the signed payload via the nonce-commitment means the facilitator cannot alter destinations or amounts after signature, and the payer sees the exact fee before signing — better fee transparency than out-of-band accounting. The stateless forwarder (replay delegated to the token's own `(from, nonce)` state) is elegant. This mechanism deserves to land. Three things stand between it and a second implementation:

1. **Where does `intentSalt` travel?** The commitment includes a client-generated random 32-byte `intentSalt`, but it appears in no wire field: `extra.forwarderSplit` doesn't carry it and the payload section defines only the standard 3009 authorization fields. Verification step 2 ("recompute the commitment hash… require it to equal `nonce`") is therefore unimplementable by a facilitator, and the forwarder's on-chain recomputation has the same problem. Your production system clearly transports it somewhere — the spec just never says where. A full `PaymentPayload` JSON example (mirroring the base spec's per-method examples) plus a concrete `settle()` signature / minimal forwarder interface would make this buildable by someone who isn't you.

2. **Make the on-chain invariants normative.** (i) The forwarder MUST require `authorization.value == merchantValue + feeValue` — otherwise a larger `value` strands the residual in the forwarder. (ii) The method MUST use `ReceiveWithAuthorization` typed data and MUST NOT accept `TransferWithAuthorization` — and it's worth saying why: with `transferWithAuthorization`, any observer of the signature can submit it directly to the token, moving payer funds into the forwarder with **no** disbursement. That stranding/front-running vector is currently absent from Security Considerations (as is `cancelAuthorization` front-running as facilitator-gas griefing). (iii) Disbursement transfers MUST revert on failure (SafeERC20 semantics). (iv) "any EIP-3009 token works unchanged" is too strong — fee-on-transfer/rebasing tokens break the disbursement arithmetic; scope them out explicitly.

3. **Position against `auth-capture`.** `specs/schemes/auth-capture/scheme_auth_capture_evm.md` already does atomic 3009 collection with an on-chain-committed fee split using the same nonce-as-commitment technique (`charge()` with `autoCapture` is close to this proposal plus escrow). The legitimate differences — no operator role, no escrow dependency, pure `exact` semantics — are worth one paragraph, and aligning commitment-encoding conventions would help both. Related: consider adding the `asset` address to the commitment (one `abi.encode` word) so offline verification is self-contained rather than leaning on the EIP-712 domain check; and the worked example mixes v1 `maxAmountRequired` with v2 CAIP-2 networks — the v2 shape uses `amount`.

Happy to review the revision.

---
*Disclosure per CONTRIBUTING: review performed with significant AI assistance under significant human monitoring, direction, and correction; mechanism analysis checked against EIP-3009 semantics and the repo's existing schemes rather than generated from memory.*

cipherwebllc:
@whawk46 Thank you — this is exactly the review the spec needed, and coming from someone running 3009 + permit2 witness proxies in production it carries weight. Revision pushed; point-by-point:

**1. Where `intentSalt` travels.** You found a real spec/wire gap: the production wire carries it as `payload.authorization.intentSalt` (alongside the standard EIP-3009 authorization fields), and `nonce` is *derived, never trusted* — verifiers MUST recompute the commitment and require equality with `authorization.nonce`. The revision now defines the full `payload.authorization` field table, adds a **complete PaymentPayload JSON example** mirroring the base spec's per-method examples, and a **minimal `IEip3009ForwarderSplit` interface** with the concrete `settle()` signature (taken from the deployed contract, where `token` / `feeReceiver` / `commitVersion` are constructor immutables; the spec notes a variant MAY take them as parameters iff commitment-covered). Verification step 2 is now implementable by a third party.

**2. Normative on-chain invariants.** Added a dedicated MUST section covering all four, plus one you'll recognize from our deployment:
- `ReceiveWithAuthorization` only, `TransferWithAuthorization` forbidden — with your front-running rationale spelled out (the payee guard `to == msg.sender` is what makes the forwarder the only executor; a transfer-type signature could be submitted by any observer straight to the token, stranding funds at the forwarder with no settlement attached);
- `value` MUST be derived internally as `merchantValue + feeValue` (residue rationale included);
- SafeERC20 semantics with full-transaction revert;
- fee-on-transfer / rebasing tokens explicitly out of scope — "any EIP-3009 token works unchanged" is now qualified accordingly;
- (from production) `intentSalt == 0` MUST revert, and `merchant == feeReceiver` / `merchant == forwarder` SHOULD revert.
`cancelAuthorization` front-running is now in Security considerations as griefing-only, with pre-submission simulation as the suggested mitigation. For the record, the deployed forwarder already enforces all of the MUSTs — the gap was in the prose, not the contract.

**3. auth-capture positioning and the details.** Added a "Relationship to `auth-capture`" section up front: same nonce-as-commitment trick, but single-phase — no operator role, no escrow, no holding period, pure `exact` semantics; use auth-capture when you need delayed capture. On `asset` in the commitment: agreed it makes offline verification self-contained — the v1 encoding is frozen in production, so I've listed it as the natural spec-level `commitVersion` v2 change, ideally aligned with auth-capture's commitment encoding (open question 3). The v1/v2 mixing in the worked example is fixed — it now uses v2 shapes (`amount`, CAIP-2 network) throughout.

Happy to iterate further — and if aligning the two commitment encodings is something the auth-capture authors are open to, we'd participate.

cipherwebllc:
Branch hygiene fixed — **all commits are now verified** (`68d7f67` + `7da6e30`), and `check-verified-commits` passes. The revision-2 content is unchanged from what I described above; it was only re-committed with a proper signature (the earlier revision-2 push slipped through unsigned, which is why the bot was holding review).

The only remaining non-green check is the Vercel preview ("Authorization required to deploy"), which is on the org side for external contributor PRs — nothing actionable from my end.

Ready for re-review whenever the team or @whawk46 has bandwidth. 🙏

whawk46 (approved):
Re-reviewed at `7da6e30` against the three items from the first pass —
each one verified in the revised text, not from the changelog:

1. **`intentSalt` wire gap: closed.** The `payload.authorization`
   field table, the derived-never-trusted rule with a MUST-recompute,
   the complete PaymentPayload example, and the concrete `settle()`
   interface together make verification step 2 implementable by
   someone who isn't you — which was the whole ask.
2. **Invariants: closed, and stronger than requested.** Deriving
   `value` internally rather than validating an input removes the
   residue class entirely instead of guarding it. The two
   production-sourced additions (zero-salt revert, merchant/feeReceiver
   distinctness) are exactly the kind of deployment scar tissue a spec
   should capture.
3. **`auth-capture` relationship: closed.** The single-phase vs
   two-phase split is crisp, and parking `asset`-in-commitment as a
   `commitVersion` v2 aligned with auth-capture's encoding is the
   right sequencing given your frozen v1.

On the open questions, briefly: (1) `extra.assetTransferMethod` is the
right home — it composes with the `erc20ApprovalGasSponsoring`
precedent and keeps `exact` semantics intact; (2) deployment-scoped
`commitVersion` published in `extra` until a spec version exists, then
a standardized constant — the example already demonstrates the
namespaced form; (4) receipts feel like a separate, composable
proposal rather than scope here — the settlement event you emit is
already the anchor a receipt layer would bind to, and I'd be glad to
engage on that draft when you open it.

Approving. The mechanism was right on the first pass; now the spec is
buildable from the document alone, which is the bar.

---
*Disclosure per CONTRIBUTING: review performed with significant AI
assistance under significant human monitoring, direction, and
correction; revision verified against the spec text at head rather
than the revision notes.*

cipherwebllc:
Thank you for the thorough second pass — verifying against the spec text at head rather than the revision notes is exactly the standard this needed.

On the open questions, your answers match where the document already stands, so no further edits are needed from my side:

1. **`extra.assetTransferMethod`** stays as the extension point — agreed that composing with the `erc20ApprovalGasSponsoring` precedent keeps the exact-amount semantics intact.
2. **`commitVersion`** remains deployment-scoped in `extra` (namespaced form as shown in the example) until a spec version exists to anchor a standardized constant.
4. **Receipts** — agreed this is a separate, composable proposal rather than scope here. The settlement event is the natural anchor a receipt layer would bind to. I'll open that as its own draft and ping you on it.

The `Vercel` check failure is the fork preview-deploy authorization gate (`Authorization required to deploy`), not a content issue — `check-verified-commits` and `labeler` are green at 7da6e30.

Ready for maintainer review; happy to address anything further that comes up.

PR #190, opened 2025-05-19
[Chore] [Docs] Improve formatting of `json` blocks

stefanoamorelli:
Based on [this gist](https://gist.github.com/DamianEdwards/31d2457737304ca73556), I suggest improving the rendering of `json` code blocks in `README.md` file to better support comment highlighting.

PR #2890, opened 2026-07-17
feat(bsv): add exact scheme support for BSV

sirdeggen:
## Summary

Adds BSV (Bitcoin SV) as an x402 payment mechanism for the `exact` scheme: a new `@x402/bsv` package (client / server / facilitator), a per-chain spec, example wiring, and a USD price feed. It builds on the BRC-121 "Simple 402 Payments" flow (BRC-29 payments over BRC-42 key derivation) already used in production BSV tooling.

Per CONTRIBUTING's new-chain process this PR is the **spec + TypeScript reference implementation**; Go/Python ports would follow once this is accepted. The CAIP-2 `bsv` namespace is registered ([ChainAgnostic/namespaces#190](https://github.com/ChainAgnostic/namespaces/pull/190), merged 2026-08-18; spec: [bsv/caip2.md](https://github.com/ChainAgnostic/namespaces/blob/main/bsv/caip2.md)).

## How BRC-121 maps onto x402

- `payTo` = the recipient wallet's BRC-100 identity public key. It never appears on chain — every payment pays a fresh key derived from it (BRC-42), so on-chain observers cannot link payments to the payee.
- Payload = a BRC-29 remittance: `{ transaction (base64 BEEF), derivationPrefix, derivationSuffix, senderIdentityKey, outputIndex }`. The suffix decodes to a Unix-ms timestamp (BRC-121 freshness).
- The client's BRC-100 wallet funds, fees, and signs the transaction via `createAction` — the payer covers the miner fee, so there is no facilitator gas sponsorship.
- **The facilitator is the recipient's own wallet.** Because BRC-42 destinations are ECDH-derived, a third party cannot take custody and cannot verify the destination unilaterally. `verify` checks structure, wallet-chain agreement, freshness, exact amount, and the derived destination; `settle` internalizes the output via `internalizeAction`, backed by a txid dedup cache plus the wallet's `isMerge` signal for replay defense. (Either counterparty *can* voluntarily prove a payment's linkage via `revealSpecificKeyLinkage` — BRC-69, verifiable through BRC-94's Schnorr ZKP — see the spec's auditability appendix and note 2 below.)
- Networks: `bsv:mainnet` / `bsv:testnet` / `bsv:ttn` / `bsv:tstn`.

## What's included

**`@x402/bsv`** (`typescript/packages/mechanisms/bsv`), modeled on the Concordium mechanism:
- `exact/client` — derives the per-payment key and builds the funded BEEF payment
- `exact/facilitator` — verifies and settles into the recipient wallet
- `exact/server` — builds `PaymentRequirements` with satoshi price parsing (8 decimals)
- `createWhatsOnChainMoneyParser` — converts USD prices (`"$0.001"`) to satoshis via the WhatsOnChain exchange-rate API (60 s cache, single-flight refresh, bounded stale fallback)
- constants / types, README, package + tsup + tsconfig + vitest config, changeset, npm publish workflow

**Spec**: `specs/schemes/exact/scheme_exact_bsv.md` (full `exact`-on-BSV spec with a Key Linkage Revelation & Auditability appendix), plus BSV safety invariants added to the shared `scheme_exact.md` appendix.

**Examples**: all three `all_networks` examples wired —
- client: `BSV_WALLET=true` + a running BRC-100 wallet (e.g. BSV Desktop)
- facilitator: `BSV_SERVER_PRIVATE_KEY` + `BSV_WALLET_STORAGE_URL` via `@bsv/simple` `ServerWallet`
- server: `BSV_IDENTITY_KEY`, priced in USD through the money parser

## Security

Key facilitator invariants (also in the spec's Critical Validation Requirements):
- **Destination correctness** — `verify` derives the expected BRC-42 key (`getPublicKey` `forSelf`) and matches it against the P2PKH output, so an unfundable or mispaid BEEF is rejected before any resource work runs.
- **Subject-tx integrity** — the subject transaction is resolved via the Atomic BEEF `atomicTxid`, i.e. exactly the transaction the wallet will internalize (not "the last tx in the BEEF").
- **Exact amount** — strict equality with `requirements.amount` (stricter than plain BRC-121, which tolerates overpayment).
- **Freshness** — ±30 s at verify; at settle the window extends by `maxTimeoutSeconds` so slow handlers don't void a valid payment.
- **Replay** — facilitator-side txid dedup cache plus the wallet's `isMerge` (a wallet-toolbox extension, not core BRC-100); `verify` re-runs at settlement.
- **Chain agreement** — the facilitator checks `getNetwork()` against the requested network; `transaction` is empty on any failure per the core v2 schema.

## Testing

- **103 unit tests** (`pnpm test` in the package) — payload construction, verify/settle rules, Atomic BEEF subject resolution, replay paths, timestamp windows, and the rate-feed parser (cache / stale / error paths).
- **Live smoke** — money parser against the real WhatsOnChain API ($0.001 → 7,440 sats at $13.44/BSV).
- **Live wallet round trip** — `BSV_INTEGRATION=true pnpm test:integration` with a BRC-100 wallet running (create → verify → settle → replay-reject); requires a funded wallet.

## Notes for reviewers

1. **Network identifiers / namespace.** `bsv:mainnet` / `bsv:testnet` / `bsv:ttn` / `bsv:tstn` use human-readable network names (following the `casper` CAIP-2 precedent). The `bsv` namespace **has been registered** ([ChainAgnostic/namespaces#190](https://github.com/ChainAgnostic/namespaces/pull/190), merged 2026-08-18; spec: [bsv/caip2.md](https://github.com/ChainAgnostic/namespaces/blob/main/bsv/caip2.md)). `bip122` could alternatively identify BSV by its fork-block hash (as it does for Bitcoin Cash); happy to switch the network IDs if that's preferred.
2. **Facilitator trust model.** The recipient-wallet facilitator is faithful to BRC-29 but differs from the third-party-facilitator model on other chains. Extension paths, if a shared/third-party facilitator is wanted: a plain-P2PKH addressing variant (drops the privacy model), or a key-linkage-proof flow (BRC-69/BRC-94) that gives destination verification without custody — also the auditability route for regulated assets such as BSV-issued stablecoins.
3. **Strict amount equality** for `exact` diverges from BRC-121's overpay tolerance; documented in the spec.
4. **Scope.** No core packages are modified. Reference implementation is TypeScript only; Go/Python would be follow-up PRs.

---

*AI-usage disclosure (per CONTRIBUTING): the implementation, tests, spec, and this description were drafted with AI assistance (Claude Code) and reviewed by the author before submission; an AI-driven adversarial review pass produced 24 fixes now folded into the branch.*


BSVanon:
### On (1) — network identifiers

Agree with names over `bip122` — fork-block-hash is genuinely ambiguous given the shared genesis, and the BSV app layer (wallets, overlays, ARC, SPV clients) identifies networks by name anyway. It also lines up with the CAIP-2 namespace being registered in parallel (#190), which uses the same `bsv:mainnet` / `bsv:testnet` strings — worth keeping those two deliberately in lockstep.

One wrinkle worth pinning while it's cheap: `bsv:testnet` is getting ambiguous. Alongside classic testnet there's STN (the Scaling Test Network) and now Teratestnet (the Teranode test network) — and a per-request, high-throughput payments protocol is exactly what people will want to exercise against Teranode. Not suggesting you enumerate them all now, just: is `bsv:testnet` meant as "the one canonical test target," or should the syntax extend to something like `bsv:stn` / a teranode testnet later without overloading `testnet`? Probably worth deciding at the same time as the CAIP-2 registration, so x402 and CAIP-2 don't diverge on what "testnet" means. Happy to raise it on #190 too.

### On (3) — strict amount equality

Keep rule 6 exact. From designing a BSV merchant gateway on WooCommerce, I'd argue the case for equality in a machine protocol is stronger than just "defensible."

The core reason isn't tolerance bands — it's that a checkout has out-of-band recovery a 402 exchange doesn't. In my gateway an overpayment becomes "contact the customer to arrange a refund," and an underpayment becomes "waiting for the remaining amount," which only resolves because a reused address lets the buyer top up or the storeowner steps in. A machine-to-machine 402 has neither channel: the payer builds one funding tx to a unique derived output and that's the whole conversation. Tolerating a mismatch there doesn't degrade gracefully, it produces a state with no automated resolution. Equality means the only outcomes are "paid" and "retry" — which is what you want when there's no human in the loop.

Where my own code does tolerate a shortfall, it's context-specific rather than a global policy: a background poller watching a reused address slackens by `max(1000 sats, 0.1%)` because it's answering "did value show up here?" across dust and top-ups, while the structured PaymentACK receiver rejects any underpayment outright. Different jobs, different rules — which is exactly why baking one tolerance value into a shared spec is the thing to avoid.

The drift one would otherwise use a tolerance band to absorb — price/FX movement between quote and pay — is better handled the way this scheme already does it: pin the amount at challenge issuance, bound it with the freshness window, and fold any buffer into the quoted price. It's a quote-validity lever, not an accepted-amount lever. Strict equality + the 30s / `maxTimeoutSeconds` window is the cleaner composition.

### On conformance / test vectors

A note for when the Go/Python ports arrive. The pieces that already have test vectors are BRC-42 key derivation (the sample vectors in the BRC-42 spec) and the BRC-100 wire format (vectors live with that project). What has none yet is the part this PR newly adds: the exact-amount check at `outputIndex`, the millisecond timestamp in `derivationSuffix` plus the freshness-window edges, and the replay signals (txid dedup and the `isMerge` / no-new-satoshis case). Those are where two independent implementations are most likely to disagree, because there's nothing to test against yet.

I've spent a good amount of time working in a Rust port of the BRC-100 wallet-toolbox, so the replay/internalize path in particular is familiar. Happy to help put together a small known-answer fixture for just those new rules as a follow-up, once there's a second implementation to check against.

sirdeggen:
Wrinkle on network is fair. Perhaps we just register them all now. 

mainnet
testnet
ttn
tstn

I think no one is using stn since Teranode. We also have mockchain in some software. Which is modern regtest with spv, but neither are needed for this which would only make sense if there is _some_ network being used.

So perhaps just adding two additional networks would suffice.

andyrowe:
> Wrinkle on network is fair. Perhaps we just register them all now.
> 
> mainnet testnet ttn tstn
> 
> I think no one is using stn since Teranode. We also have mockchain in some software. Which is modern regtest with spv, but neither are needed for this which would only make sense if there is _some_ network being used.
> 
> So perhaps just adding two additional networks would suffice.

Registering them explicitly now seems right — naming a network later is much cheaper than migrating clients that already guessed.

One implementer data point, since we're running a plain-P2PKH exact mode on mainnet today: we currently emit
bip122:000000000019d6689c085ae165831e93 , and we treat that as a bug rather than a choice. It can't distinguish BSV from BTC, and the failure it enables is unrecoverable — a client that guesses wrong pays the right amount on the wrong chain. So whatever set gets registered, I'd argue the spec should require clients to refuse to spend on an ambiguous Bitcoin-family identifier rather than fall back to a default. We'll move to bsv:mainnet the moment #190 lands.

Agree on skipping mockchain/regtest — there's nothing to pay on.

sirdeggen:
### Network identifiers update

Following the discussion on registering the Teranode test networks explicitly (and @andyrowe’s note that ambiguous Bitcoin-family IDs are an unrecoverable footgun), this PR is now aligned with [ChainAgnostic/namespaces#190](https://github.com/ChainAgnostic/namespaces/pull/190):

**Registered CAIP-2 networks**

| Network | Identifier |
| --- | --- |
| Mainnet | `bsv:mainnet` |
| Testnet | `bsv:testnet` |
| Teranode Test Net (Teratestnet) | `bsv:ttn` |
| Teranode Scaling Test Net | `bsv:tstn` |

**Ambiguous `bip122` refusal**

The spec and TypeScript reference now require clients/facilitators to **refuse** ambiguous Bitcoin-family identifiers (including genesis-only `bip122:000000000019d6689c085ae165831e93`) rather than defaulting them to BSV. Paying the right amount on the wrong chain is unrecoverable.

**Implementation notes**

- `@x402/bsv` constants: `BSV_TTN_CAIP2`, `BSV_TSTN_CAIP2`, `BSV_NETWORKS`, `toBsvWalletNetwork()` mapping to BRC-100 `getNetwork()` names (`mainnet` / `testnet` / `ttn` / `tstn`)
- Facilitator wallet-chain agreement uses that map
- Examples accept `BSV_NETWORK=bsv:ttn|bsv:tstn`
- 121 unit tests passing

Namespaces PR updated in lockstep: https://github.com/ChainAgnostic/namespaces/pull/190

milai123:
nice

vegardwikeby:
x402 for sure will live up to its aim; include everyone, so I’m sure this will be added without any fuzz.

tonesnotes:
Extending the already class leading x402 support! Yeah!

davewantsmoore:
X402

PhillipJFryz:
good

gptix:
BSV!

mo-jaber:
+1 on getting this merged. We have users and use cases lined up ready to build on x402 payments over BSV as soon as it lands.

Colin-jpg:
gm

SteffenNS:
It is obviously right to have good BSV support because BSV has proof-of-concept tested over 1 million peer-to-peer transactions per second, and has very low transaction fees.

Zwilla:
Really thorough write-up — you clearly went deep on the BRC-121 → x402 mapping rather than bolting on a generic "chain #N" integration. A few things that stand out:

Explicit registration of bsv:ttn and bsv:tstn alongside bsv:mainnet / bsv:testnet, together with the hard refusal of ambiguous bip122 (genesis-hash) identifiers, is the correct hardening. Paying the right amount on the wrong chain is unrecoverable in an automated protocol.

The divergence from [BRC-121][brc121]'s overpayment tolerance to strict amount equality is well-justified for machine-to-machine flows. Equality produces only two terminal states ("paid" or "retry"), which is exactly the contract an automated agent needs. Pinning the amount at challenge time and bounding it with the freshness window is the cleaner composition.

Keeping the facilitator as the recipient's own [BRC-100][brc100] wallet preserves the core privacy property (the identity key never appears on-chain) and eliminates custody risk. The optional spvOnVerify callback is a useful early filter.

Citing the specific BRC standards ([BRC-121][brc121], [BRC-29][brc29], [BRC-42][brc42], [BRC-100][brc100], [BRC-69][brc69], [BRC-94][brc94]) and showing how they compose end-to-end is a strong signal that the underlying specs were actually read.

103 unit tests + live integration smoke is solid. Known-answer vectors for the exact-amount check at outputIndex, the ms timestamp in the derivation suffix (+ freshness edges), and the replay signals (txid dedup TTL + isMerge / no-new-satoshis) would help future Go/Python ports.

The Key Linkage Revelation & Auditability appendix ([BRC-69][brc69]/[BRC-94][brc94]) is a thoughtful addition, especially for regulated-asset use cases.

[brc121]: https://brc.dev/121
[brc29]: https://brc.dev/29
[brc42]: https://brc.dev/42
[brc100]: https://brc.dev/100
[brc69]: https://brc.dev/69
[brc94]: https://brc.dev/94

BMX-Casey94:
x402 on BSV is the only viable solution despite any surrounding politics, imo. 
Following this closely, nice one!

BSVanon:
Great job Deggen!  AI will find BSV the undeniably optimal choice for machine 2 machine economies.

jefflightweb:
I'd like to see this merged.

nedzof:
I think this is what we need for the x402. Identity-based atomic microtransactions for the m2m economy at the speed of light. Awesome work Deggen!

vincemedia:
Yes please

samooth:
The x402 protocol is explicitly designed to be chain-agnostic — any server can return 402, any client can satisfy it, and the underlying settlement happens on-chain without card-network intermediaries. BSV is not just another chain to tick off a list; it is an exceptionally strong fit for what x402 actually promises.
**1. Sub-cent micropayments natively**
x402's killer use case is machine-to-machine payments: API calls priced at $0.01, $0.001, or less. BSV transaction fees are natively sub-cent, routinely below $0.0001. A 1-satoshi payment is economically viable on the base layer — no approval transactions, no gas abstraction layer. This PR implements the exact scheme using native UTXO transfers.
**2. Stateless SPV verification**
The BSV exact scheme uses BEEF format, which carries full SPV ancestry in a single base64 payload. A server can verify a payment statelessly by checking the BEEF Merkle proof against a block header — no RPC node, no facilitator dependency. This is a genuine cost reduction for operators: a lightweight BSV overlay/SPV node can run on a $5/month VPS.
**3. Unbounded throughput on the base layer**
BSV's unbounded block sizes mean the base layer absorbs high-volume machine payments without routing, channel management, or batching hacks. For agent-to-agent commerce — where call volumes scale to millions per hour — this matters. You do not need deferred-payment schemes or consumption metering when the base layer itself handles the throughput.
**4. Protocol stability ("set in stone")**
BSV's protocol is locked to Satoshi's original design. Developers do not rebuild around annual hard forks, fee-market redesigns, or opcode deprecations. For infrastructure like x402 — where server operators, facilitators, and wallet vendors must coordinate — this predictability is a feature, not a bug. The PR already aligns BSV network identifiers with ChainAgnostic CAIP-2 namespaces (bsv:mainnet, bsv:testnet, etc.), so it plays nicely with x402's cross-chain registry.
**5. Security hardening sets a good precedent**
This PR includes concrete hardening that raises the bar for exact-scheme implementations:
BSV-only asset enforcement at verify time (prevents asset spoofing).
BRC-29 derivation prefix length check (minimum 8 bytes).
Optional spvOnVerify callback to close the unfunded-BEEF DoS surface.
Duplicate-settlement dedup TTL sized to the full settlement window with rollback on soft wallet rejection.
Explicit satoshi validation (positive integer within supply) at parse time.
**6. Already running in production**
This is not theoretical. BSV micropayments via x402 are live on Cloudflare Workers (Rust/WASM overlay stack) and in autonomous-agent deployments. The TypeScript reference implementation in this PR is compatible with those production systems.

As someone following x402 closely, this is the kind of chain integration that actually fits the protocol's mission. BSV makes x402 viable at the price points it was invented for. The PR is clean, tested, spec'd, and adds one (scheme, network) pair without breaking existing paths.

Strong support from me — would love to see the maintainers give this serious consideration.

sonerseo-stack:
Great work. This PR strengthens x402's goal of remaining a truly network-agnostic payment standard.

BSV brings a production-ready implementation based on existing BRC-121/BRC-29/BRC-42 standards rather than introducing an entirely new payment model.

I also like that the implementation preserves x402's security expectations while demonstrating an alternative facilitator model that avoids gas sponsorship and third-party custody.

Supporting multiple payment networks is exactly what makes an open standard valuable. I hope this proposal is considered on its technical merits and becomes part of the ecosystem.

blockdeev:
Send it! #BSV

BSVArchie:
We build Hodos, a browser with a native BRC-100 wallet (Rust + C++). We've had the BRC-121 402 flow in production for a while — users navigate to a 402-gated page, the browser mints the BRC-29 payment and retries transparently. We test against now.bsvblockchain.tech, which serves live 402-priced articles.

To be clear about what that does and doesn't mean: we implement BRC-121 today, not x402. But since this spec profiles BRC-121, our implementation is a useful independent check on whether the payload shape is right. It appears to be.

Field mapping is 1:1 between our BRC-121 headers and the payload object here:

| Ours         | This spec         |
|--------------|-------------------|
| x-bsv-beef   | transaction       |
| x-bsv-nonce  | derivationPrefix  |
| x-bsv-time   | derivationSuffix  |
| x-bsv-sender | senderIdentityKey |
| x-bsv-vout   | outputIndex       |

And the derivation is byte-compatible, arrived at independently: 8-byte random prefix (spec: "minimum 8 bytes"), suffix = base64(utf8(decimal ms)), invoice 2-3241645161d8-{prefix} {suffix}. Two implementations landing on identical encodings without coordinating is decent evidence the spec is unambiguous.

One piece of feedback on the freshness window, which I think is the one place our perspective differs usefully from a machine-to-machine client.

The spec sets a 30-second symmetric window on derivationSuffix. Our wallet caches and reuses a minted payment for up to 25 seconds for the same (URL, amount) — so we don't re-mint on rapid repeat requests.

That leaves ~5 seconds to absorb network RTT to the facilitator plus clock skew between the payer's clock and the verifier's. Internet clock skew alone can eat that. A reused payment at 24s age against a verifier running a few seconds fast gets rejected as stale, and the failure would look intermittent and hard to diagnose.

Not arguing the window is wrong — 30s is reasonable for automated flows. But it's worth deciding explicitly whether clients may reuse a payload at all within the window, or whether the spec should say mint-fresh-per-request. Right now it's silent, and a conforming client could reasonably do either. A sentence either way would prevent a real interop failure mode.

(Related: human-in-the-loop wallets have a second timing consideration — a user staring at an approval dialog. We mint after approval so the clock starts clean, but that's an implementation choice the spec might want to make explicit for wallets that don't.)

Happy to run cross-implementation interop testing — our client against the facilitator in this PR — if that's useful for validating the spec beyond a single-language reference. Independent client/server pairs tend to surface exactly this kind of ambiguity.

andyrowe:
Chiming in as another live exact-on-BSV implementer, with one caveat up front: bsv.cx runs the plain-P2PKH addressing variant (the extension path noted in
reviewer note 2), not the BRC-29 derivation flow — so we don't use `derivationSuffix`.  Offering this as an architectural data point, not direct experience with the mechanism under discussion.

On @BSVArchie's freshness-window question, which is the right thing to pin down: we sidestep the clock-skew squeeze entirely by anchoring freshness to a server-issued absolute `expiresAt` carried in the challenge (default 15 min), rather than a client-stamped timestamp the verifier re-checks against its own clock. The payer never stamps a time both sides must agree on, so there's no skew/RTT window to tune — the only clock that matters is the issuer's.

That's possible because our replay defense is structural, not temporal: the invoice is single-use, the amount is pinned at issuance, and payment lands on a
unique per-invoice output. The timestamp isn't load-bearing for replay.

Which is the underlying point for the BRC-29 flow too: `derivationSuffix` is being asked to do double duty — anti-replay and freshness — and the squeeze is the symptom. Since this PR already has txid-dedup + `isMerge` doing the replay work, a tight symmetric ±30s window mostly imports clock-skew fragility without
adding much security (a replay is caught by dedup regardless of its timestamp). 

Two things that would help, in rough priority:

1. Answer the reuse question explicitly (the silence @BSVArchie flagged) — state whether a client MAY reuse a payload within the window. If replay rests on dedup + `isMerge`, reuse is safe and the spec can say so.
2. Make the window asymmetric — generous on the past side (to absorb client-side caching + RTT + skew), near-zero on the future. Or better, define the verifier's tolerance relative to the challenge's issuance/expiry rather than the verifier's "now," so payer clock skew drops out entirely.

Either way, +1 that this belongs in the spec rather than left for implementations to guess — it's exactly what two conforming clients will diverge on. Happy to compare notes on the addressing-variant side if it's useful for the extension path.

jkramsky:
"The facilitator is the recipient's own wallet."

This is the x402 I want to use. 👍

MalkavianBilbao:
Great work and great proposal!

bitcoinsv-data:
Really solid work on this. The attention to interoperability, network identification, replay protection, and the BRC-121/x402 mapping gives this a strong foundation. Excited to see BSV support land in x402. +1 from me. 🚀

m-russ:
🤝

mrz1836:
Been building on BSV since the early days, so flagging interest here.

This follows the new-chain process pretty closely — spec plus TS reference, core packages untouched, CAIP-2 registration open in parallel, tests and a live wallet round trip. The two things that look like maintainer calls rather than more work for the author are the network identifier (`bsv:mainnet` vs. a `bip122` fork-hash) and the recipient-wallet facilitator model in note 2. Both have offered alternatives in the description, so a direction from someone on the team would unblock it.

If the TS lands, I'm happy to take the Go port — that's most of what I do in this ecosystem. Also glad to review or run the suite in the meantime if that helps.

yusufidimaina9989:
X402

cammyjee:
The original Bitcoin design (BSV) is a blueprint for scaling 'small casual payments' (on-chain micropayments), I'm convinced it is the ultimate solution for x402

Neggia:
BSV is the best fit for x402

kurtwuckertjr:
With most blockchains, there's inherent scaling problems, so "if you win, you lose." 

With BSV, I think the cool kids still say "WGMI."

andyrowe:
Quick update on the network-identifier question flagged above: the CAIP-2 registration this PR was waiting on just merged — [ChainAgnostic/namespaces#190](https://github.com/ChainAgnostic/namespaces/pull/190) (2026-08-18). `bsv` is now a registered CAIP-2 namespace with references `bsv:mainnet`, `bsv:testnet`, and the Teranode test nets `bsv:ttn` / `bsv:tstn`.

So `bsv:mainnet` is no longer a *proposed* identifier — it's the registered one, and `bip122:<genesis-hash>` is now the strictly-worse option, since BSV shares that genesis hash byte-for-byte with BTC and BCH and so can't actually name the chain. That resolves one of the two open maintainer calls without further work.

As an interop datapoint (not a competing proposal): we switched our live BSV x402 service over to emit `network: "bsv:mainnet"` today. Settlement keys off the broadcast tx, never the network string, so it was a clean swap and round-trips fine. Our plain-P2PKH variant is the companion you scoped out of this PR, so nothing here overlaps the BRC-29/42 exact scheme — just confirming the registered identifier works end to end.

andyrowe (approved):
LGTM, approving.  Confirmed this tracks the upstream `SchemeNetworkServer`/`MoneyParser` changes cleanly — `defaultAssetTransferMethod = "default"` + the single `authorization` flow lines up with the other non-EVM mechanisms (near/aptos/hedera/keeta), and widening the WhatsOnChain parser to `string | number` keeps the guards intact (empty string → 0 and non-numeric → NaN are both still rejected).  

One non-blocking note: the fork's CI is sitting at `action_required`, so unit/lint haven't actually run on this commit. — might be worth kicking off the workflow so it's green on record before merge.

BSVArchie:
**Independent implementer report: BEEF in a request header hits a 100 KB wall in practice**

We ship a browser with a native BRC-100 wallet and an independent BRC-121 implementation, so this profile is directly relevant to us. Passing on a failure we measured on mainnet this week — it affects the wire format proposed here rather than any one implementation.

BEEF must carry every direct parent in full (BRC-62), so the header's size is set by the payer's transaction history, not by the payment. Three consecutive payments from one wallet to one site, in one session:

| BEEF bytes | base64 in header | Result |
|---|---|---|
| 779 | ~1.0 KB | 200 |
| **92,837** | **123,784 B** | **431** — `Exceeded maximum HTTP header buffer size of 100KB` |
| 849 | ~1.1 KB | 200 |

A 110× spread between neighbouring payments, decided by which coin funded them. The failing one was 150 sats, between the other two in amount.

Three points that may be worth reflecting in the profile:

1. **The usable budget is not the quoted cap.** base64 is ×4/3, and the limit covers the whole header block — the page's own `Cookie` and `User-Agent`, plus this profile's `x-bsv-sender`, `x-bsv-nonce`, `x-bsv-time` and `x-bsv-vout`. On a 100 KB cap that lands near 68–73 KB of BEEF, and it varies per site and per user.

2. **A "prefer small parents" rule in the wallet is necessary and not sufficient.** We had already shipped one after hitting the same root cause on a 1 MiB message relay, and it was enabled on this path. It still let the 92,837-byte bundle through, because its threshold had been calibrated to that other channel's cap. A preference is an ordering rather than a guarantee, so a hard size check before minting is a separate layer.

3. **Whether a header is the right carrier.** A request body is bounded by a policy the server controls and can publish; a header block is bounded by whatever proxy sits in front of it, which the sender never chose and cannot query. The failure grows with the payer's history, so it worsens as a wallet is used. We recognise that is a wire-format change and are not asking for it lightly.

Full measurements and what we changed on our side are in the BRC-121 discussion: https://github.com/bsv-blockchain/BRCs/issues/261

No position on the merge — this is a data point from running the same wire format independently.

PR #2899, opened 2026-07-18
fix(python): preserve streaming bodies on payment retry

### LINKED ISSUE #2847: x402AsyncTransport._send_retry() fails against streaming-body clients (e.g. MCP's streamable-http transport)
## Bug

`x402AsyncTransport._send_retry()` in `x402/http/clients/httpx.py` accesses
`request.content` directly when building the retry request after a 402:

```python
retry_request = Request(
    method=request.method,
    url=request.url,
    headers=new_headers,
    content=request.content,   # <-- fails here
    extensions=new_extensions,
)
```

If the original `httpx.Request` was constructed with a streaming body rather
than pre-buffered bytes, `request.content` raises `httpx.RequestNotRead`:

```
httpx.RequestNotRead: Attempted to access streaming request content, without having called `read()`.
```

The fix is to `await request.aread()` before accessing `.content` (this
buffers the stream and makes `.content` safe to read), e.g.:

```python
content = await request.aread()
retry_request = Request(
    method=request.method,
    url=request.url,
    headers=new_headers,
    content=content,
    extensions=new_extensions,
)
```

## Impact

This breaks any client whose HTTP requests use a streaming body rather than
plain bytes - notably the official MCP Python SDK's
`mcp.client.streamable_http.streamablehttp_client`, which lets callers pass
a custom `httpx_client_factory`. Wrapping that factory's `httpx.AsyncClient`
with `x402AsyncTransport` (the documented pattern for adding x402 payment
handling to an MCP client) fails on every paid tool call that actually
needs a retry, since the MCP transport's POST requests aren't built with a
pre-buffered body.

## Reproduction context

Found while adding x402 support to an MCP server (FastAPI + the official
MCP Python SDK) and testing it with a client built exactly per the
documented pattern:

```python
def httpx_factory(headers=None, timeout=None, auth=None):
    return httpx.AsyncClient(
        transport=x402AsyncTransport(payment_client),
        follow_redirects=True,
        headers=headers, timeout=timeout, auth=auth,
    )

async with streamablehttp_client(MCP_URL, httpx_client_factory=httpx_factory) as (read, write, _):
    async with ClientSession(read, write) as session:
        await session.initialize()
        await session.call_tool("some_paid_tool", {...})  # raises PaymentError wrapping RequestNotRead
```

The first request correctly gets a 402 and `x402AsyncTransport` correctly
parses it and builds a signed payment payload - the failure is purely in
`_send_retry()`'s handling of the *original* request's body when
constructing the retry.

Confirmed end-to-end on a real funded Base Sepolia testnet wallet: a
transport that reads the body first (`await request.aread()`) before
retrying works correctly and settles on-chain; the SDK's own
`x402AsyncTransport`, unmodified, fails every time on this exact client
shape.

## Environment
- `x402` package version: 2.15.0
- Python 3.11
- `mcp` (official Python SDK) `streamablehttp_client`

realmehmetali:
## Summary

- buffer async HTTPX request bodies before the initial send
- replay identical bytes when a 402 response triggers a paid retry
- add regression coverage for one-shot streaming POST bodies
- add the required Python changelog fragment

## Root cause

HTTPX streaming bodies can be consumed by the initial request. Reusing that request after a payment challenge could therefore send an empty payload or fail instead of replaying the original body.

## Agent impact

AI agents frequently call MCP and paid tool endpoints with streamed POST bodies. A 402 challenge must preserve the task payload across payment recovery; otherwise an autonomous tool call can pay successfully but execute with missing input. Replaying the exact bytes strengthens x402 as infrastructure at the AI-agent frontier.

Fixes #2847.

## AI assistance

Codex assisted with implementation and regression-test preparation. I reviewed the public code path, HTTPX streaming semantics, and final diff before submission.


phdargen:
Thanks for the contribution @realmehmetali, LGTM 🚀

phdargen (approved):

PR #2900, opened 2026-07-18
fix(fetch): preserve Request body during recovery

realmehmetali:
## Summary

- retain the normalized cloned `Request` for one payment-hook recovery
- preserve POST bodies and headers when the caller supplies a `Request` object
- add a regression test covering the initial request, paid retry, and recovery retry
- add the required patch changeset

## Root cause

Constructing the first normalized `Request` can mark a caller-provided `Request` body as used. The recovery path rebuilt from that consumed input, causing body-bearing requests to throw before the retry.

## Agent impact

AI agents frequently purchase APIs through stateful POST requests. Losing the task payload during payment recovery breaks autonomous execution and forces human intervention. Preserving the body makes x402 more robust as payment infrastructure at the AI-agent frontier.

## Validation

- dependency-aware build for `@x402/core` and `@x402/fetch`
- `@x402/fetch`: 21 tests passed
- lint check
- format check

## AI assistance

Codex assisted with implementation and test preparation. I reviewed the diff, Fetch `Request` body semantics, and validation results before submission.

### DIFF
--- a/typescript/.changeset/fix-fetch-recovery-request.md
+++ b/typescript/.changeset/fix-fetch-recovery-request.md
@@ -1,0 +1,5 @@
+---
+"@x402/fetch": patch
+---
+
+Preserved Request bodies when a payment hook triggers a recovery retry.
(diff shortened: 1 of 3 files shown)


phdargen (approved):
Good catch @realmehmetali, thanks for the fix!

PR #2914, opened 2026-07-21
fix(go): preserve request bodies across payment retries

wnjoon:
## Description

Preserve one-shot and streaming Go request bodies across x402 authentication, payment, and corrective recovery retries.

The root cause was that `PaymentRoundTripper.RoundTrip` sent the initial request through the transport, while `sendHeaderRetry` and `sendPaymentRetry` could only recreate a body when `GetBody` was available (`go/http/client.go:204`, `414-470`). Requests with consumed, non-replayable bodies were therefore retried with an empty or closed body.

The request body is now buffered before the first transport call and attached to a working clone with a replayable `Body` and `GetBody`, while leaving the caller's request fields unchanged.

The implementation:

- skips buffering for nil bodies, `http.NoBody`, and requests that already provide `GetBody`
- replays identical payload bytes for authentication, payment, and corrective recovery retries
- closes the original body exactly once
- returns read, close, and cancellation errors before invoking the transport
- preserves request metadata and transport ownership of replay bodies
- uses `context.AfterFunc` to attempt to unblock body reads when the request is canceled

Finite one-shot bodies are fully buffered in memory. This change does not introduce an arbitrary body-size limit; large or non-terminating streams remain an explicit tradeoff of automatic replay.

## Tests

- Added one-shot body replay coverage across payment retries
- Added authentication followed by payment retry coverage
- Added corrective recovery retry coverage
- Added read, close, and cancellation error tests
- Added request metadata and caller immutability checks
- Added skip-path tests for nil, `http.NoBody`, and already replayable bodies
- Ran `go test ./http`
- Ran `go test -race ./http`
- Ran `go vet ./http`
- Cross-validation repeated cancellation and body-error race scenarios 100 times
- Full `make test` passed, including ``, `gin`, and `echo`

## Checklist

- [x] My commits are signed
- [x] I added a changelog fragment for the user-facing fix
- [x] Significant AI assistance was used; the implementation and tests were manually reviewed and independently cross-validated

### DIFF
--- a/go/.changes/unreleased/fixed-20260721-111633.yaml
+++ b/go/.changes/unreleased/fixed-20260721-111633.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Preserved one-shot request bodies across payment retries.
+time: 2026-07-21T11:16:33.744189+09:00
--- a/go/http/client.go
+++ b/go/http/client.go
@@ -1,9 +1,11 @@
 package http
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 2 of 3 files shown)


phdargen (approved):
Thanks for the fix @wnjoon!

Issue #2918, opened 2026-07-22
Proposal: upto scheme for Hedera (first non-EVM upto)


Madhav-Gupta-28:
### Problem

Metered services on Hedera — LLM token generation, bandwidth metering, dynamic compute — cannot price their work up front. The only `upto`-style scheme in x402 today is EVM-only (Permit2), so Hedera's only option is [`exact`](specs/schemes/exact/scheme_exact_hedera.md), which must fix the price *before* the work runs. A metered seller is forced to quote a worst case and keep the difference, or absorb the overage.

### High-level approach

Bring [`upto`](specs/schemes/upto/scheme_upto.md) to Hedera with the same protocol types and a network-specific enforcement contract, `X402UptoProxy` — the Hedera counterpart of the EVM scheme's `x402UptoPermit2Proxy`. The client grants the proxy a one-time HTS allowance ([HIP-336](https://hips.hedera.com/hip/hip-336)), then authorizes each metered payment with an off-chain EIP-712 signature; the facilitator settles the actual amount via the HTS ERC-20 facade's `transferFrom` ([HIP-376](https://hips.hedera.com/hip/hip-376)). The proxy enforces all five `scheme_upto.md` MUSTs on-chain.

### Why existing schemes don't suffice

- `exact` on Hedera cannot express "up to" — it requires the price at signature time.
- A bare HTS allowance (HIP-336 alone) cannot satisfy the `upto` MUSTs: allowances are not single-use, do not expire, and do not bind a recipient. This is the same reason Permit2 exists on EVM.

### Status

Spec written (mirrors `scheme_upto_evm.md` section-for-section); reference implementation deployed and verified on Hedera testnet (contract `0.0.9556979`), with packages published to npm. A PR referencing this proposal is ready to open.

PR #2919, opened 2026-07-22
Add upto scheme implementation for Hedera (first non-EVM upto)

Madhav-Gupta-28:
## Summary

Adds `specs/schemes/upto/scheme_upto_hedera.md` — the Hedera implementation of the [`upto`](specs/schemes/upto/scheme_upto.md) scheme — and lists it in `scheme_upto.md`'s network index.

`upto` currently exists only for EVM (Permit2). **This is its first non-EVM implementation.** It settles HTS fungible tokens through a small, admin-less enforcement contract, `X402UptoProxy` — the Hedera counterpart of the EVM scheme's `x402UptoPermit2Proxy`. A reference implementation is deployed and verified on Hedera testnet, with published npm packages for the client, resource-server, and facilitator roles.

Proposal issue: #2918

## Motivation

Hedera already contributes [`exact`](specs/schemes/exact/scheme_exact_hedera.md). `exact` fixes the price before the work runs, so a metered Hedera service — LLM token generation, bandwidth, dynamic compute — has to quote a worst case and keep the difference, or absorb the overage. `upto` is the scheme built for that shape of workload, and until now it was EVM-only. This brings it to Hedera the same way `exact` was: same protocol, same `PaymentRequirements` / `PaymentPayload` / `SettlementResponse` types, with network-specific enforcement.

## Why a new asset-transfer method is required

A bare HTS allowance ([HIP-336](https://hips.hedera.com/hip/hip-336)) **cannot** satisfy the `upto` MUSTs: allowances are not single-use, do not expire, and do not bind a recipient — a holder can move the owner's funds anywhere, up to the limit. This is exactly why Permit2 exists on EVM, and why an equivalent enforcement contract must exist on Hedera.

`X402UptoProxy` is that contract. It composes two native Hedera mechanisms:

- **HIP-336** — an account can grant an HTS token allowance to a *contract*.
- **HIP-376** — the HTS ERC-20 facade's `transferFrom` spends against *the caller's* allowance (the caller being the contract).

The client grants the proxy a one-time allowance, then authorizes each metered payment with an off-chain EIP-712 signature (no transaction, no fee). The proxy enforces all four MUSTs plus facilitator binding on-chain — so neither a malicious facilitator nor a malicious resource server can take more than the client signed, take it twice, take it late, or send it anywhere the client did not name.

## Conformance to `scheme_upto.md`

| `scheme_upto.md` property (MUST) | Enforcement on Hedera |
| --- | --- |
| 1 — Single-Use Authorization | `nonceUsed[from][nonce]`, consumed inside `capture()` even when the settled amount is `0` |
| 2 — Time-Bound | `validAfter <= block.timestamp <= deadline`, both inside the signed `Authorization` struct |
| 3 — Recipient Binding | `payTo` is a signed field; altering it invalidates the signature |
| 4 — Maximum Amount | `require(amount <= maxAmount)` on-chain; the settled `amount` MAY be `0` |
| 5 — Phase-dependent `amount` + settle-time re-verification | the facilitator re-verifies the signature against the signed **maximum**, never the settlement amount, then submits `capture` with the actual amount; the contract independently re-checks `amount <= maxAmount` |

The doc mirrors its sibling [`scheme_upto_evm.md`](specs/schemes/upto/scheme_upto_evm.md) section-for-section (Summary / AssetTransferMethod / PaymentRequirements / SettlementResponse / Error Codes / Annex / Security Considerations), including its **Settle-Time Verification** subsection, so the two chain docs read in parallel.

## Changes

1. **Add** `specs/schemes/upto/scheme_upto_hedera.md`.
2. **Edit** `specs/schemes/upto/scheme_upto.md` — add Hedera to the *Network-Specific Implementation* list:

   ```diff
     - EVM chains: See [`scheme_upto_evm.md`](./scheme_upto_evm.md)
   + - Hedera: See [`scheme_upto_hedera.md`](./scheme_upto_hedera.md)
   ```

## Reference implementation

- **Contract** — `X402UptoProxy.sol`: admin-less and immutable (no owner, pause, or upgrade path), custodies no funds. Deployed and verified on Hedera **testnet** at contract id `0.0.9556979`. Effects precede interactions (nonce burns before transfer) and the high-`s` signature twin is rejected, so each authorization has exactly one valid settlement. `capture()` costs 81,308 gas. Every invariant is asserted against the live deployment, and each negative case is checked twice: that the transaction reverted on-chain, and that the revert names the expected custom error.
- **Packages** — [`x402-hedera-upto`](https://www.npmjs.com/package/x402-hedera-upto) (client / resource-server / facilitator scheme implementations against `@x402/core`) and [`x402-hedera-receipts`](https://www.npmjs.com/package/x402-hedera-receipts) (verifiable offer / price / meter receipts + an audit CLI).
- **Source** — https://github.com/Madhav-Gupta-28/Tally

## An open question — `upto` receipts carry no meter reading (on every network)

This applies to `upto` on **every** network, so I'm raising it here rather than editing the extension in this PR. `scheme_upto_evm.md`'s own Security Considerations note that "malicious servers could charge up to `amount` regardless of actual usage" — the server alone decides the final charge, and nothing in the receipt artifacts lets anyone check it. The canonical [`offer-receipt`](specs/extensions/extension-offer-and-receipt.md) payloads commit the offer `amount` and the receipt's `payer` / `issuedAt` / `transaction`, but **neither commits the unit price or the units consumed** — so an `upto` receipt is not independently verifiable as written.

§5 of the Hedera doc describes a minimal fix that needs no protocol change: sign a price schedule alongside the offer and a meter reading alongside the receipt, and bind both into `resourceHash` — which `X402UptoProxy` already emits on-chain in its `Captured` event. `unitPrice × units == captured amount` then becomes arithmetic anyone can check against the ledger (the reference implementation anchors those artifacts to HCS). If maintainers agree this belongs in the `offer-receipt` extension, I'm glad to open a follow-up.

## Checklist (per `specs/CONTRIBUTING.md`)

- [x] Proposal issue opened first (#2918)
- [x] Uses the `scheme_impl_template.md` structure; mirrors `scheme_upto_evm.md`
- [x] File at `specs/schemes/upto/scheme_upto_hedera.md`; listed in `scheme_upto.md`
- [x] References the core spec (`x402-specification-v2.md`) for shared types
- [x] MUST / SHOULD / MAY throughout, with concrete JSON examples
- [x] Security Considerations section (replay, authorization scope, settlement atomicity)
- [x] Working, deployed, verified reference implementation

### DIFF
--- a/specs/schemes/upto/scheme_upto.md
+++ b/specs/schemes/upto/scheme_upto.md
@@ -72,3 +72,4 @@ The following patterns are NOT supported by `upto` and would require different s
 Network-specific rules and implementation details are defined in the per-network scheme documents:
 
 - EVM chains: See [`scheme_upto_evm.md`](./scheme_upto_evm.md)
+- Hedera: See [`scheme_upto_hedera.md`](./scheme_upto_hedera.md)
--- a/specs/schemes/upto/scheme_upto_hedera.md
+++ b/specs/schemes/upto/scheme_upto_hedera.md
@@ -1,0 +1,367 @@
+# Scheme: `upto` on `Hedera`
+
+## Summary
+
+The `upto` scheme on Hedera enables usage-based payments where the client authorizes a **maximum amount** and the facilitator settles the **actual amount used** at the end of the request. It is the scheme for variable-cost resources — LLM token generation, bandwidth metering, dynamic compute — where the final price is not knowable until after the work is done.
+
+This is the first non-EVM implementation of [`upto`](./scheme_upto.md). Hedera already contributes the [`exact`](../exact/scheme_exact_hedera.md) scheme; `exact` requires the price to be fixed before the work is done, so a metered seller must quote a worst case and keep the difference, or absorb the overage. `upto` closes that gap.
+
+| AssetTransferMethod | Use Case | Notes |
+| :--- | :--- | :--- |
+| **`X402UptoProxy`** | HTS fungible tokens. Client signs a maximum with EIP-712; facilitator settles the actual. | Admin-less, immutable contract. The Hedera counterpart of the EVM scheme's `x402UptoPermit2Proxy`. |
+
+> **Note**: Native **HBAR** (`asset: "0.0.0"`) is **not supported** for the `upto` scheme and MUST be rejected. An HBAR allowance is not reachable from a contract through the HTS ERC-20 facade; it requires the Hedera Account Service path (`hbarApprove` / `hbarAllowance`, [HIP-906](https://hips.hedera.com/hip/hip-906)) and a `cryptoTransfer` with the approval flag set. `upto` on Hedera is a stablecoin-metering scheme; a future revision MAY add native HBAR.
+
+## Use Cases
+
+- **LLM token generation**: client authorizes up to $0.50, actual charge based on tokens generated.
+- **Bandwidth / data transfer**: pay per byte transferred in a single request, up to a cap.
+- **Dynamic compute**: authorize a maximum cost, charge based on actual resources consumed.
+
+## Prerequisites
+
+1. **The client account MUST be ECDSA (secp256k1) with an EVM alias.** `ecrecover` recovers the alias-form address, and on Hedera the long-zero address of an alias-bearing account does not resolve in HTS calls (`INVALID_ALLOWANCE_OWNER_ID` as a sender, `INVALID_ALIAS_KEY` as a recipient). An ED25519 account cannot sign an `upto` authorization.
+2. The client MUST be associated with the `asset` and hold a balance.
+3. `payTo` MUST be associated with the `asset`, or hold an available auto-association slot.
+4. **One time, per (client, token):** the client grants `X402UptoProxy` an HTS allowance covering the ceilings it intends to sign (see Phase 1).
+
+## Protocol Flow
+
+1. **Client** requests a protected resource.
+2. **Resource Server** responds `402` with `PaymentRequirements` where `scheme: "upto"` and `amount` is the authorized **maximum**.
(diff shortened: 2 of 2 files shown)


Eras256:
Independently verified the concrete, checkable pieces of this: the reference contract at `0.0.9556979` is really deployed and live on Hedera testnet (confirmed against the mirror node, `evm_address` matches exactly), and the published packages (`x402-hedera-upto@0.1.0`, `x402-hedera-receipts@0.1.1`) match what's described here. Read through `X402UptoProxy.sol` in full — the approach (allowance via HIP-336 + `transferFrom` through the ERC-20 facade per HIP-376, authorized off-chain per-payment via EIP-712) is sound: nonce burns before transfer, malleable signatures are rejected, facilitator binding is enforced, no owner-gated function in the contract itself.

One open question, not a blocker I'm certain about: the mirror node reports a non-empty `admin_key` on the deployed contract's account. That's a Hedera platform-level field, separate from the Solidity having no owner logic — if it's still live, whoever holds that key could in principle update or delete this specific instance, which sits a bit oddly next to "no owner, no admin, no upgrade path" in the Security Considerations section. Might just be a leftover from the deploying account and worth clearing on a fresh deploy, or worth a line clarifying it either way -- didn't dig further into whose key it actually is.

This looks stalled on process, not substance -- `check-verified-commits` already passes, the only red check is Vercel's external-contributor deploy-authorization gate, and there's no review on either the PR or the issue yet. I've since built and run a full working demo against the real `x402-hedera-upto` package end to end on testnet (resource server, facilitator, client), so this design holds up in practice, not just on paper. Happy to help get this unstuck if useful.

Issue #835, opened 2025-12-21
Add Optimism and Optimism Sepolia Support


fretchen:
# Summary
Proposal  to add support for Optimism (OP Mainnet) and Optimism Sepolia testnet to the x402 protocol. Both chains are EVM-compatible L2s with native USDC support and seem to fit naturally into the existing `exact` scheme for EVM chains.

# Motivation
Optimism is a major Ethereum L2 with:
- High transaction volume and active ecosystem
- Native USDC Circle deployment (not bridged)
- Lower gas fees than Ethereum mainnet
- Growing adoption for payment use cases

Adding Optimism support would enable x402 to serve the large Optimism ecosystem and provide users with a more cost-effective payment option.

# Technical Details

**OP Mainnet (Chain ID: 10)**
- Network identifier: `optimism`
- Chain ID: `10`
- USDC Contract: `0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85`
- USDC Version: `2`
- RPC: Available via public endpoints

**OP Sepolia Testnet (Chain ID: 11155420)**
- Network identifier: `optimism-sepolia`
- Chain ID: `11155420`
- USDC Contract: `0x5fd84259d66Cd46123540766Be93DFE6D43130D7`
- USDC Version: `2`
- RPC: Available via public endpoints

# Implementation Scope

Since Optimism is EVM-compatible, the implementation could follow the same pattern as existing EVM chains (Base, Polygon, etc.):

1. **TypeScript SDK** (`@x402/shared`):
   - Add network mappings in `EvmNetworkToChainId`
   - Add USDC configurations for both chains
   - Extend network validation

2. **Go SDK**:
   - Add network constants
   - Add USDC configuration

3. **Python SDK**:
   - Add network mappings
   - Add USDC configuration

4. **Tests**:
   - Verification tests for both networks
   - Settlement tests for both networks
   - Integration tests with the `exact` scheme

# Verification
- ✅ USDC contracts deployed and verified on both chains
- ✅ EIP-3009 `transferWithAuthorization` function available
- ✅ Compatible with existing x402 `exact` scheme
- ✅ No new cryptographic mechanisms required

# Additional Context
I'm currently trying out a facilitator service that supports Optimism through custom configuration and would be happy to migrate to native x402 support once available. I'm willing to contribute the implementation if that helps move this forward. However, obviously I am still new to the whole topic and would most likely diverge from best practices on certain occasions.

### References
[...]

PR #2924, opened 2026-07-22
fix(evm): batch-settlement honors caller-supplied asset on networks outside DEFAULT_STABLECOINS

### LINKED ISSUE #2910: `batch-settlement`: is there a supported way to use a network that isn't in `DEFAULT_STABLECOINS`?
### What I'm trying to do

Run the `batch-settlement` scheme on Optimism mainnet, paying in Circle USDC (`0x0b2C…Ff85`). The batch-settlement contract seems to be deployed there, and the token implements EIP-3009.

Optimism isn't in `DEFAULT_STABLECOINS`, so I followed the guidance I was given in [#835](https://github.com/x402-foundation/x402/issues/835):

> v2 supports any EVM-compatible chain as long as the payment asset implements EIP-3009 […] You can either implement a moneyParser […] or specify 'amount' (in atomic units) and 'asset' instead of 'price'

That works for `exact`. With `batch-settlement` I can't get either approach to work, and I suspect I'm either holding it wrong or hitting something unintended — hence the question.

### Reproduction

```js
// npm i @x402/evm && node repro.mjs
import { BatchSettlementEvmScheme } from "@x402/evm/batch-settlement/server";

const scheme = new BatchSettlementEvmScheme("0x1111111111111111111111111111111111111111");

await scheme.enhancePaymentRequirements(
  {
    scheme: "batch-settlement",
    network: "eip155:10",
    amount: "1420",                                       // explicit atomic amount
    asset: "0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85",  // Circle USDC on OP
    payTo: "0x1111111111111111111111111111111111111111",
    extra: { name: "USD Coin", version: "2" },
  },
  {},
  [],
);
```

```
Error: No default asset configured for network eip155:10
    at getDefaultAsset (chunk-DQI2DTA4.mjs:151:11)
    at BatchSettlementEvmScheme.enhancePaymentRequirements (batch-settlement/server/index.mjs:1525:23)
```

### What I tried to narrow it down

Varying one thing at a time (full script at the bottom):

```
ok      batch-settlement  eip155:8453          asset kept: 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913
THROWS  batch-settlement  eip155:10            No default asset configured for network eip155:10
ok      exact             eip155:10            asset kept: 0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85
THROWS  batch-settlement  createChannelManager No default asset configured for network eip155:10
THROWS  batch-settlement  + moneyParser        No default asset configured for network eip155:10
```

My reading of this:

- Rows 1–2: the same call succeeds on Base and fails on OP, so the requirements object itself seems to be accepted.
- Rows 2–3: `exact` accepts the same network and asset that `batch-settlement` rejects.
- Row 4: `createChannelManager` seems to depend on the registry as well, so this may not be limited to one call path.
- Row 5: a `registerMoneyParser` doesn't change the outcome — from the source it looks like it's consulted by `parseMoney`, which isn't on this path.

### What I found in the source

In `batch-settlement/server`, `enhancePaymentRequirements` looks up the asset before reading the caller's requirements:

```js
const assetInfo = getDefaultAsset(paymentRequirements.network);   // index.mjs:1525
…
extra: { ...paymentRequirements.extra, name: assetInfo.name, version: assetInfo.version }
```

whereas `exact`'s implementation of the same method passes the requirements through:

```js
enhancePaymentRequirements(paymentRequirements, supportedKind, extensionKeys) {
  return Promise.resolve(paymentRequirements);
}
```

I couldn't find an option to supply the asset explicitly — as far as I can tell neither the client nor server `BatchSettlementEvmSchemeOptions` takes one — but I may have missed it.

One thing I wasn't sure whether to flag: because a resource server builds `accepts[]` by calling this per network, a single unlisted network appears to throw for the whole 402 rather than just that entry. That could be intended.

### Questions

1. Is `batch-settlement` intended to work on networks outside `DEFAULT_STABLECOINS`, and if so, what's the supported way? Happy to just use it if I've missed the mechanism.
2. If the caller's `asset`/`extra` should be honoured here the way `exact` does, I'd be glad to attempt a PR. The part I'd need guidance on is whether the EIP-712 `name`/`version` are deliberately taken from the registry rather than the caller — that looked intentional, and I didn't want to assume.

Separately, if it's simply that Optimism should be in the registry, these are the values I read from the contract on 2026-07-20:

```js
"eip155:10": {
  address: "0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85",
  name: "USD Coin",
  version: "2",
  decimals: 6,
},
```

Glad to send that as a PR too if it's useful — though I assume the broader question above matters more than my particular chain.

### Environment

- `@x402/evm` 2.18.0, `@x402/core` 2.18.0
- Node v24.16.0
- Same behaviour on 2.17.0

### Full script

```js
// npm i @x402/evm && node repro.mjs
// No facilitator, no RPC, no wallet, no funds.
import { BatchSettlementEvmScheme } from "@x402/evm/batch-settlement/server";
import { ExactEvmScheme } from "@x402/evm/exact/server";

const PAY_TO = "0x1111111111111111111111111111111111111111";
const AUTHORIZER = "0x2222222222222222222222222222222222222222";
const USDC_OP = "0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85"; // OP mainnet — not in DEFAULT_STABLECOINS
const USDC_BASE = "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913"; // Base mainnet — in DEFAULT_STABLECOINS

// Explicit atomic `amount` + `asset`: the documented way to use an unlisted chain (#835).
const reqs = (network, asset) => ({
  scheme: "batch-settlement",
  network,
  amount: "1420",
  asset,
  payTo: PAY_TO,
  extra: { name: "USD Coin", version: "2" },
});

const batch = new BatchSettlementEvmScheme(PAY_TO, { receiverAuthorizerSigner: { address: AUTHORIZER } });

// Calls enhancePaymentRequirements and reports whether the caller's `asset` survived.
const enhance = async (label, scheme, network, asset) => {
  try {
    const r = await scheme.enhancePaymentRequirements(reqs(network, asset), {}, []);
    console.log(`ok      ${label.padEnd(38)} asset kept: ${r.asset}`);
  } catch (e) {
    console.log(`THROWS  ${label.padEnd(38)} ${e.message}`);
  }
};

// Control: Base is in the registry. Shows the call is well-formed and the asset is honored.
await enhance("batch-settlement  eip155:8453", batch, "eip155:8453", USDC_BASE);

// The bug: identical call, unlisted network. Throws even though `asset` is supplied.
await enhance("batch-settlement  eip155:10", batch, "eip155:10", USDC_OP);

// The inconsistency: same network, same asset, same method — `exact` honors it.
await enhance("exact             eip155:10", new ExactEvmScheme(), "eip155:10", USDC_OP);

// A second, independent registry dependency: fixing enhancePaymentRequirements alone is not enough.
try {
  batch.createChannelManager({}, "eip155:10");
  console.log("ok      batch-settlement  createChannelManager eip155:10");
} catch (e) {
  console.log(`THROWS  ${"batch-settlement  createChannelManager".padEnd(38)} ${e.message}`);
}

// The other workaround from #835: registerMoneyParser only reaches parseMoney/defaultMoneyConversion.
const withParser = new BatchSettlementEvmScheme(PAY_TO, { receiverAuthorizerSigner: { address: AUTHORIZER } });
withParser.registerMoneyParser(() => ({ amount: "1420", asset: USDC_OP }));
await enhance("batch-settlement  + moneyParser", withParser, "eip155:10", USDC_OP);
```

fretchen:
## Description

Fixes #2910. The `batch-settlement` scheme threw `No default asset configured for network …` on any EVM network absent from `DEFAULT_STABLECOINS` (e.g. Optimism `eip155:10`), even when the caller supplied an explicit `amount` + `asset` — the documented path for unlisted chains (#835). The `exact` scheme already works this way; `batch-settlement` did not.

Root cause: `enhancePaymentRequirements` called `getDefaultAsset(network)` and re-derived `name`/`version`/`assetTransferMethod` from the registry, ignoring the asset metadata that `parsePrice` (or the caller) had already placed on the requirements. `createChannelManager` had the same hard dependency on the registry.

## Tests

Ran `pnpm test` for `@x402/evm` — all 818 unit tests pass; lint clean. Added coverage in `test/unit/batch-settlement/server.test.ts`:

- `enhancePaymentRequirements` succeeds on an unlisted network (`eip155:10`) with an explicit asset, preserving the caller's `asset`/`name`/`version`.
- `createChannelManager` builds with an explicit token on an unlisted network, and still resolves via the registry when none is given.
- `parsePrice`/`defaultMoneyConversion` sets `assetTransferMethod` for permit2 tokens (both the EIP-2612 and non-2612 branches).

Each new test was confirmed to fail against the pre-fix code and pass after the fix.

## Checklist

### DIFF
--- a/typescript/.changeset/batch-settlement-explicit-asset.md
+++ b/typescript/.changeset/batch-settlement-explicit-asset.md
@@ -1,0 +1,5 @@
+---
+"@x402/evm": patch
+---
+
+Fixed `batch-settlement` throwing `No default asset configured for network …` on EVM networks outside `DEFAULT_STABLECOINS` when the caller supplies an explicit `amount` + `asset`. Asset metadata now flows through `parsePrice`/the caller instead of being re-derived from the registry in `enhancePaymentRequirements`, `createChannelManager` accepts an optional explicit token, and `defaultMoneyConversion` sets `assetTransferMethod` for permit2 tokens (matching the `exact` scheme).
--- a/typescript/packages/mechanisms/evm/src/batch-settlement/server/scheme.ts
+++ b/typescript/packages/mechanisms/evm/src/batch-settlement/server/scheme.ts
@@ -429,12 +429,23 @@ export class BatchSettlementEvmScheme implements SchemeNetworkServer {
     const assetInfo = getDefaultAsset(network);
     const tokenAmount = convertToTokenAmount(numberToDecimalString(amount), assetInfo.decimals);
 
+    // EIP-3009 tokens always need name/version for their transferWithAuthorization domain.
+    // Permit2 tokens only need them if the token supports EIP-2612 (for gasless permit signing).
+    // Omitting name/version for permit2 tokens signals the client to skip EIP-2612 and use
+    // ERC-20 approval gas sponsoring instead.
+    const includeEip712Domain = !assetInfo.assetTransferMethod || assetInfo.supportsEip2612;
+
     return {
       amount: tokenAmount,
       asset: assetInfo.address,
       extra: {
-        name: assetInfo.name,
-        version: assetInfo.version,
+        ...(includeEip712Domain && {
+          name: assetInfo.name,
+          version: assetInfo.version,
+        }),
+        ...(assetInfo.assetTransferMethod && {
+          assetTransferMethod: assetInfo.assetTransferMethod,
+        }),
       },
     };
   }
--- a/typescript/packages/mechanisms/evm/src/batch-settlement/server/scheme.ts
+++ b/typescript/packages/mechanisms/evm/src/batch-settlement/server/scheme.ts
@@ -255,7 +255,10 @@ export class BatchSettlementEvmScheme implements SchemeNetworkServer {
 
   /**
    * Injects batched-specific fields into the payment requirements returned to
-   * the client (receiverAuthorizer, withdrawDelay, EIP-712 domain info).
+   * the client (receiverAuthorizer, withdrawDelay). Asset metadata (name,
+   * version, assetTransferMethod) is left untouched — it is already set by
+   * `parsePrice` or supplied explicitly by the caller, and is not re-derived
+   * from the default-asset registry here so unlisted networks keep working.
    *
    * @param paymentRequirements - Base payment requirements from the middleware.
    * @param supportedKind - Matched scheme/network kind (extra may contain overrides).
@@ -278,8 +281,6 @@ export class BatchSettlementEvmScheme implements SchemeNetworkServer {
   ): Promise<PaymentRequirements> {
     void _extensionKeys;
 
-    const assetInfo = getDefaultAsset(paymentRequirements.network as Network);
-
     const receiverAuthorizer =
       this.receiverAuthorizerSigner?.address ??
       (typeof supportedKind.extra?.receiverAuthorizer === "string"
@@ -299,10 +300,6 @@ export class BatchSettlementEvmScheme implements SchemeNetworkServer {
         ...paymentRequirements.extra,
         receiverAuthorizer: getAddress(receiverAuthorizer),
         withdrawDelay: this.withdrawDelay,
-        name: assetInfo.name,
-        version: assetInfo.version,
-        assetTransferMethod:
-          paymentRequirements.extra?.assetTransferMethod ?? assetInfo.assetTransferMethod,
       },
     };
   }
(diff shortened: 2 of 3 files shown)


phdargen (approved):
Thanks for the fix and catching the issue in the first place @fretchen, looks great 🚀

PR #2931, opened 2026-07-23
Fix: Algorand CAIP-2 network IDs

### LINKED ISSUE #2904: Algorand network identifiers are not valid CAIP-2 (facilitator + docs)
## Summary

The `x402.org` facilitator and the Networks & Token Support page advertise Algorand
networks using the standard base64 genesis hash. That form is not a valid CAIP-2
reference, and the CAIP-2 Algorand namespace profile rules it out explicitly. Clients
that validate CAIP-2 — as the docs say x402 uses — will reject every Algorand `kind`.

## Reproduce

```bash
curl -s https://x402.org/facilitator/supported | jq '.kinds[] | select(.network | startswith("algorand"))'
```

```json
{
  "x402Version": 2,
  "scheme": "exact",
  "network": "algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=",
  "extra": { "feePayer": "G7QWRIJODICBDG6JAVXNKHNTCKTBJZBXTSCGQLSMXSCIKEJ5SNFPEJSFQQ" }
}
```

The same form appears in the default-asset table on
[Networks & Token Support](https://docs.x402.org/core-concepts/network-and-token-support)
for Algorand mainnet: `algorand:wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=`.

## Why this is not CAIP-2

**1. It violates the CAIP-2 base specification** ([CAIP-2] (https://chainagnostic.org/CAIPs/caip-2),
Status: **Final**), which defines the syntax as:

```
chain_id:    namespace + ":" + reference
namespace:   [-a-z0-9]{3,8}
reference:   [-_a-zA-Z0-9]{1,32}
```

The advertised reference is 44 characters and contains `/` and `=`, none of which is
permitted. CAIP-2's own test cases note the maximum total length as `8+1+32 = 41`; the
advertised identifier is 53. The `/` also breaks two of CAIP-2's stated design goals
directly — an identifier "can be used unescaped in URL paths" and "can be used as
filename in a case-sensitive UNIX file system".

This holds regardless of any namespace profile.

**2. The Algorand namespace profile defines the correct form.** The
[CAIP-2 Algorand profile](https://namespaces.chainagnostic.org/algorand/caip2)
(Status: Draft) addresses precisely this case:

> Due to length and character limitations of CAIP-2 reference identifiers, the commonly
> referenced standard base64-encoded genesis hash of an Algorand chain cannot be used
> directly as a reference. The genesis hash must be encoded with a URL-safe base64
> alphabet (the characters `+` and `/` must be replaced with `-` and `_`, respectively).
> Only the first 32 characters of the URL-safe base64-encoding will be used.

Its **Test Cases** section lists the canonical identifiers for both networks, and they
match the values suggested below character for character.

## Suggested values

| Network | Currently advertised | Per the CAIP-2 Algorand profile |
|---|---|---|
| Testnet | `algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=` | `algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDe` |
| Mainnet | `algorand:wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=` | `algorand:wGHE2Pwdvd7S12BL5FaOP20EGYesN73k` |

Both right-hand values appear verbatim in the profile's Test Cases section, so this is
not an interpretation on my part.

To be clear about scope: the **network assignment is correct**. I verified the genesis
hashes directly against `GET /v2/transactions/params` on public Algorand nodes —
`mainnet-v1.0` returns `wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=` and
`testnet-v1.0` returns `SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=`, matching how
x402 labels them. Only the encoding of the identifier deviates.

Derivation: apply the URL-safe alphabet, then truncate to 32 characters. For these two
values no `+` or `/` falls inside the first 32 characters, so the visible change is the
truncation alone. The substitution does matter in general — the namespace profile's own
example is `mFgazF+2uRS1tMiL9dsj01hJGySEmPN28B/TjjvpVW0=` →
`algorand:mFgazF-2uRS1tMiL9dsj01hJGySEmPN2`.

## Impact

No funds are at risk. The effect is interoperability: any client that validates network
identifiers against CAIP-2 discards the Algorand kinds, which affects network selection,
facilitator routing, and validation of `accepts` entries. Since both the hosted
facilitator and the documentation table use this form, the SDK registries are likely
affected too.

## Question

Is the deviation intentional? If x402 deliberately uses the untruncated base64 form, the
documentation should define that format rather than referring to CAIP-2. If it is an
encoding oversight, the two values above are the fix.

Happy to open a PR against the docs and the TypeScript/Go/Python registries once you
confirm which direction you'd prefer — I did not want to guess at that in a PR.

---

<sub>Found with [x402-conformance](https://github.com/moped2110/x402-conformance) while
running its facilitator checks against the public testnet facilitator.</sub>

phdargen:
## Description

Algorand network values use the full base64 genesis hash, which is not valid [CAIP-2](https://chainagnostic.org/CAIPs/caip-2). The correct form is defined by the [Algorand CAIP-2 namespace profile](https://namespaces.chainagnostic.org/algorand/caip2): URL-safe base64 of the genesis hash, first 32 characters.

Server registration and constants emit canonical truncated IDs; client and facilitator normalize legacy full-hash IDs on input for backwards compatibility.

Closes https://github.com/x402-foundation/x402/issues/2904

## Checklist

### DIFF
--- a/docs/core-concepts/network-and-token-support.mdx
+++ b/docs/core-concepts/network-and-token-support.mdx
@@ -10,7 +10,7 @@ x402 uses [CAIP-2](https://chainagnostic.org/CAIPs/caip-2) standard network iden
 - **EVM networks**: `eip155:<chainId>` — any EVM chain is supported (e.g., `eip155:8453` for Base)
 - **Solana**: `solana:<genesisHash>` (first 32 bytes of the genesis block hash)
 - **TON (TVM)**: `tvm:<workchain>` (e.g., `tvm:-239` for mainnet, `tvm:-3` for testnet)
-- **Algorand**: `algorand:<genesisHash>` (base64-encoded genesis hash)
+- **Algorand**: `algorand:<reference>` — URL-safe base64 of the genesis hash, first 32 characters ([Algorand CAIP-2 profile](https://namespaces.chainagnostic.org/algorand/caip2))
 - **Stellar**: `stellar:<network>` (`pubnet` or `testnet`)
 - **Aptos**: `aptos:<chainId>` (e.g., `aptos:1` for mainnet)
 - **Hedera**: `hedera:<network>` (`mainnet` or `testnet`)
@@ -147,8 +147,8 @@ When you use `"$0.01"` syntax, x402 needs to know which stablecoin to use. The f
 
 | Chain | CAIP-2 | ASA ID | Decimals |
 |-------|--------|--------|----------|
-| Algorand Mainnet | `algorand:wGHE2Pwdvd7S12BL5FaOP20EGYesN73ktiC1qzkkit8=` | `31566704` (USDC) | 6 |
-| Algorand Testnet | `algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDexi9/cOUJOiI=` | `10458941` (USDC) | 6 |
+| Algorand Mainnet | `algorand:wGHE2Pwdvd7S12BL5FaOP20EGYesN73k` | `31566704` (USDC) | 6 |
+| Algorand Testnet | `algorand:SGO1GKSzyE7IEPItTxCByw9x8FmnrCDe` | `10458941` (USDC) | 6 |
 
 ### Stellar
 
--- a/docs/getting-started/quickstart-for-sellers.mdx
+++ b/docs/getting-started/quickstart-for-sellers.mdx
@@ -136,7 +136,7 @@ Integrate the payment middleware into your application. You will need to provide
               {
                 scheme: "exact",
                 price: "$0.001",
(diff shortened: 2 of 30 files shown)


phdargen:
cc @emg110

CarsonRoscoe (approved):

PR #2944, opened 2026-07-24
feat: add Solana support to cloudfront-lambda-edge example

notorious-d-e-v:
## What

Adds Solana support to the CloudFront + Lambda@Edge example alongside the existing Base support. Each route's `accepts` now offers both an `eip155:84532` (Base Sepolia) and a `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1` (Solana Devnet) payment option, and the client picks one — following the pattern of `examples/typescript/servers/express`.

- `lambda/src/lib/server.ts` registers schemes per namespace (`eip155:*` → `ExactEvmScheme`, `solana:*` → `ExactSvmScheme`), so the concrete networks come entirely from the routes config; the single-network `network` field is gone from the middleware config.
- `lambda/src/config.ts` gains `EVM_*`/`SVM_*` constants and dual-entry `accepts` arrays per route.
- `@x402/*` deps bumped `^2.11.0` → `^2.19.0` (+ `@x402/svm`). Bundle stays small (237 KB) since `ExactSvmScheme` needs no RPC on the server side — the facilitator does the chain work, which suits Lambda@Edge's no-env-var, latency-sensitive constraints.
- Docs: both getting-started guides + README updated for dual-network setup (wallets — with example EVM and Solana wallets — Circle faucet for both testnets, mainnet network IDs); the mainnet section keeps the example's existing neutral facilitator guidance (links the [x402 Ecosystem facilitators list](https://www.x402.org/ecosystem?filter=facilitators); no facilitator is singled out), and notes the Solana payTo needs an existing USDC token account.
- Doc fixes in passing, scoped to the sections this PR already touches: the sample 402 body now matches the real v2 wire shape (was `x402Version: 1` with `price` fields), and the Test 3 client snippet now sets both `EVM_PRIVATE_KEY` and `SVM_PRIVATE_KEY` (the fetch client exits if either is missing).

No changes to `cdk/` (it has no network config) or to the handlers beyond dropping the removed config field.

## Why

The default `x402.org` facilitator, the v2 SDK, and the ecosystem around this example (including AWS's native CloudFront/WAF x402 monetization) are all dual-chain now; this example was the remaining Base-only piece.

## How it was tested

`tsc --noEmit` and the esbuild bundle are green. Full payment flows were exercised end-to-end against the **live x402.org facilitator** with **real on-chain settlement**, driving the built handlers with CloudFront-shaped `origin-request`/`origin-response` events via a local Lambda@Edge event harness (a real CloudFront distribution was not part of this run; the CDK stack is unchanged):

1. **Unpaid request → 402** with both networks offered (Solana entry includes the facilitator's `feePayer` from the live `/supported` data).
2. **Paid via Solana Devnet USDC** → verify → forward → settle → HTTP 200 with `x-payment-response`; settlement on-chain: [`aQRrXdZRkgBeZAiiYvpvjx5BFMQxbP9wM7Hu1NtA137zARCtxAYPx37Snf3KjNjLj84y1ciN5WoGotu36QFyoZB`](https://explorer.solana.com/tx/aQRrXdZRkgBeZAiiYvpvjx5BFMQxbP9wM7Hu1NtA137zARCtxAYPx37Snf3KjNjLj84y1ciN5WoGotu36QFyoZB?cluster=devnet) — payTo received exactly 0.005 USDC (the `/content/**` price).
3. **Paid via Base Sepolia USDC** (regression check on the wildcard EVM registration) → HTTP 200; settlement on-chain: [`0x5db0a99862d31bce69d1beed44dbc3a8b45666ef5e8c2c7fb22c8f666da3fdb2`](https://sepolia.basescan.org/tx/0x5db0a99862d31bce69d1beed44dbc3a8b45666ef5e8c2c7fb22c8f666da3fdb2) — payTo received exactly 0.005 USDC.

### DIFF
--- a/examples/typescript/pnpm-lock.yaml
+++ b/examples/typescript/pnpm-lock.yaml
@@ -2930,11 +2930,14 @@ importers:
   servers/cloudfront-lambda-edge/lambda:
     dependencies:
       '@x402/core':
-        specifier: ^2.11.0
-        version: 2.11.0
+        specifier: ^2.19.0
+        version: 2.19.0
       '@x402/evm':
-        specifier: ^2.11.0
-        version: 2.11.0(typescript@5.9.3)
+        specifier: ^2.19.0
+        version: 2.19.0(bufferutil@4.0.9)(typescript@5.9.3)(utf-8-validate@5.0.10)
+      '@x402/svm':
+        specifier: ^2.19.0
+        version: 2.19.0(@solana/kit@6.1.0(bufferutil@4.0.9)(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3)(utf-8-validate@5.0.10))(@solana/sysvars@6.1.0(fastestsmallesttextencoderdecoder@1.0.22)(typescript@5.9.3))
     devDependencies:
       '@types/aws-lambda':
         specifier: ^8.10.159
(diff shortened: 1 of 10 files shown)


notorious-d-e-v:
Note: the `lint-python` failure is pre-existing on main (black wants to reformat 7 Python files untouched by this PR — the diff here is TypeScript/Markdown/lockfile only). All other checks are green.

phdargen (approved):
Thanks for the update @notorious-d-e-v, LGTM 🚀

PR #2948, opened 2026-07-24
spec(near): add the `intents-verifier` transfer method to exact/NEAR

mikedotexe:
## Summary

This adds a second asset transfer method to the NEAR `exact` scheme: direct settlement through the [NEAR Intents](https://docs.near-intents.org) Verifier contract (`intents.near`), selected by `extra.assetTransferMethod: "intents-verifier"`.

The existing NEP-366 `delegate` method remains the default when the discriminator is absent. This follows the same one-scheme/one-network/multiple-method pattern used by EVM; it does not introduce a new scheme or mechanism package.

## Protocol shape

`delegate` and `intents-verifier` are both authorization-based, facilitator-settled methods on the destination chain:

- `delegate`: a NEAR-native payer signs a NEP-366 delegate action; funds remain in the payer's token account until settlement.
- `intents-verifier`: a Verifier-balance payer signs a supported Verifier payload; the facilitator submits `execute_intents`.

This is complementary to #2102, whose 1Click flow is client-settled cross-chain payment proof and is moving toward the `upfront` scheme under #2520.

## Exactness and failure model

The amendment:

- defines a standard-neutral decoded intent document while preserving the original signature-bound string byte-for-byte;
- requires `nep413` and `erc191`, with their different envelope shapes handled explicitly;
- closes the permitted intent fields for both delivery modes;
- requires `ft_withdraw.msg` to be absent, preventing `ft_transfer_call` partial acceptance and refund from violating exact-payment semantics;
- treats balance reads and simulation as non-reserving preflight;
- validates delivery-specific DIP-4 simulation events rather than claiming simulation proves an external wallet transfer;
- requires authoritative transaction/event/receipt evidence for success; nonce consumption alone proves at-most-once execution, not merchant delivery.

A synchronous `execute_intents` rejection rolls back Verifier state, including nonce consumption. For wallet delivery, a later token-transfer failure refunds the payer's Verifier balance but leaves the nonce consumed; settlement fails and the client must obtain a fresh 402 response and sign a fresh payload.

## Grounding

- **Mainnet wallet settle:** [`2R2ANbXf…`](https://www.nearblocks.io/txns/2R2ANbXfhhTC1XaVfYLWqu3but8FKdsimAEEex1tK1dp) — independently constructed `nep413`, submitted by an unrelated account, with exact USDC delivery.
- **Live `erc191` vector:** [`ALqQDmeT…`](https://nearblocks.io/txns/ALqQDmeTUouVv6vqG8bqbwBo7RzxBuB7gufPwtcLyw6E) — a Circle EVM-wallet signature accepted by the deployed Verifier.

The current documentation and deployed implementation agree that `simulate_intents` validates signatures. Simulation does not observe external asynchronous delivery. The pinned defuse v0.4.2 source applies the reported protocol fee in `token_diff`; this method excludes `token_diff` and verifies exact `transfer`/`ft_withdraw` events.

## Constraints

- `near:mainnet` only; no testnet Verifier deployment exists.
- Payer funds must already be deposited in the Verifier ledger.
- Wallet delivery succeeds only with the exact spawned token-transfer receipt; outer execution success is insufficient.
- Implementations must retain a scoped in-flight `(Verifier, signer, nonce)` claim and fail closed when a consumed nonce cannot be bound to exact effect evidence.


mikedotexe:
Note on the second commit (2739b5b8): it's a temporary CI unbreak, separate from the spec change. `lint-python` runs unpinned `uvx ruff`, and the new ruff 0.16.0 release began formatting Python code blocks in Markdown — so seven pre-existing `python/x402` READMEs currently fail `ruff format --check` on every open PR. This commit pins ruff to the version already locked in `python/x402/uv.lock` (0.15.17) so this PR reflects its actual contents.

#2946 fixes the root cause by reformatting those READMEs for 0.16.0; once it lands I'll drop this commit and rebase back to the single spec commit.

mikedotexe:
As promised: #2946 has merged, so I've dropped the temporary CI commit and force-pushed back to the original spec-only commit (29622f58). `lint-python` passes against the unpinned workflow now that main's READMEs are formatted for ruff 0.16.0. This PR is back to a single +131-line spec diff.

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -274,6 +290,118 @@ Failure:
...
+- **`erc191`** — flat form: `{ "standard": "erc191", "payload": "<JSON string>", "signature": "secp256k1:<base58 of 65 bytes r||s||v>" }`. There is no envelope: the signed JSON string itself carries `signer_id`, `verifying_contract: "intents.near"`, `deadline`, `nonce`, and `intents`. The digest is EIP-191 `personal_sign` over the exact payload string bytes; the signer is recovered from the signature (`v` of 27/28 MUST be normalized to 0/1). Reference vector: the accepted `execute_intents` args of mainnet transaction [`ALqQDmeTUouVv6vqG8bqbwBo7RzxBuB7gufPwtcLyw6E`](https://nearblocks.io/txns/ALqQDmeTUouVv6vqG8bqbwBo7RzxBuB7gufPwtcLyw6E).
+
+The message string MUST be carried byte-identical from signing through submission — no re-serialization, no key reordering, no whitespace changes.
+
+### Intent Binding (Exactness)
+
+`message.intents` MUST contain exactly one intent, bound to the requirements by delivery mode:
```
IkerAlus: notation detail: note that `message` is a string not a JSON object, there is no `message.intents`.

IkerAlus: more importantly, for erc191, there  is no message field at all.

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -274,6 +290,118 @@ Failure:
...
+9. **Delivery preflight.** For wallet delivery, `storage_balance_of(payTo)` on the token contract MUST be non-null.
+10. **Simulation.** `simulate_intents([signedIntent])` (a read-only Verifier view taking the same input as `execute_intents`) MUST complete without panic. Empirical note (deployed contract, defuse v0.4.2): simulation **does validate signatures** — upstream documentation stating simulation skips signature validation does not match the deployed contract — but implementations MUST NOT rely on simulation alone for rule 3. Simulation excludes external async effects (a simulated `ft_withdraw`'s outer token transfer is not reflected) and reads state that can change before settlement.
+11. **Exactness under fees.** The Verifier reports a protocol fee (`state.fee`, pips) in simulation output. If simulation shows `payTo` receiving less than `amount`, verification MUST fail closed. (Empirically the fee applies to `token_diff` matching rather than plain `transfer`/`ft_withdraw`, but implementations MUST verify delivery equality rather than assume.)
+12. **Fail closed** on any RPC error, unparsable value, or undeterminable state.
+
+### Settlement
```
IkerAlus, line 395: Failure aftermath suggestion: On failed wallet delivery the debit is refunded to the payer's **Verifier balance** and the nonce is consumed. The same payload cannot be retried, the client must sign a fresh one against a fresh 402. Stating this avoids implementers retrying dead payloads.

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -8,6 +8,19 @@ The client signs a NEP-366 `SignedDelegateAction` that authorizes one exact `ft_
...
 NEAR account keys and signatures may use either `ed25519` or `secp256k1`; implementers should account for both when validating signed delegate actions.
 
+## Asset Transfer Methods
+
+NEAR `exact` payments are implemented via one of two asset transfer methods:
+
+| AssetTransferMethod | Mechanism | Payer universe | Networks |
+| :--- | :--- | :--- | :--- |
+| **1. `delegate`** (default) | NEP-366 `SignedDelegateAction` wrapping one exact `ft_transfer`; a facilitator relayer submits and sponsors gas | A NEAR account holding the token, signing with a **full-access** key | `near:mainnet`, `near:testnet` |
+| **2. `intents`** | A [NEAR Intents](https://docs.near-intents.org) multi-standard signed payload settled through the Verifier contract `intents.near` | Any signature standard the Verifier accepts — including EVM keys via `erc191` with **no NEAR account at all** | `near:mainnet` only |
```
IkerAlus: suggestion: having "intents" as a method name while "NEAR Intents" brands the whole system (including the 1Click method) may be confusing. This could become `"intents-verifier"` to be unambiguous

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -274,6 +290,118 @@ Failure:
...
+2. **Standard allowlist.** `signedIntent.standard` MUST be a supported standard; reject others deterministically.
+3. **Signature.** Recompute the signing bytes per standard (NEP-413 prehash for `nep413`; EIP-191 digest for `erc191`) and verify the signature. For recovery-based standards, the recovered key defines the signer.
+4. **Signer authorization.** For implicit signers, the id derived from the (recovered or declared) public key MUST equal `signer_id`. For named accounts, the Verifier MUST report the key as registered for `signer_id` (`has_public_key` view). NEAR account access keys are NOT the authority here — the Verifier's own key registry is.
+5. **Nonce.** The message/envelope nonce MUST be 32 bytes and unused: `is_nonce_used(signer_id, nonce)` MUST return `false`. The Verifier enforces single-use per signer on chain, so no facilitator nonce storage is required.
+6. **Deadline bounds** per the mapping above.
+7. **Intent binding.** Exactly one intent, matching the delivery-mode binding rules exactly (token/receiver/amount equality against `PaymentRequirements`).
+8. **Balance preflight.** `mt_balance_of(signer_id, "nep141:<asset>")` on the Verifier MUST be ≥ `amount`.
```
IkerAlus: Note that Verifier balances aren't locked by a signed intent, so settlement can still fail on insufficient balance after a passing preflight.

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -274,6 +290,118 @@ Failure:
...
+9. **Delivery preflight.** For wallet delivery, `storage_balance_of(payTo)` on the token contract MUST be non-null.
+10. **Simulation.** `simulate_intents([signedIntent])` (a read-only Verifier view taking the same input as `execute_intents`) MUST complete without panic. Empirical note (deployed contract, defuse v0.4.2): simulation **does validate signatures** — upstream documentation stating simulation skips signature validation does not match the deployed contract — but implementations MUST NOT rely on simulation alone for rule 3. Simulation excludes external async effects (a simulated `ft_withdraw`'s outer token transfer is not reflected) and reads state that can change before settlement.
+11. **Exactness under fees.** The Verifier reports a protocol fee (`state.fee`, pips) in simulation output. If simulation shows `payTo` receiving less than `amount`, verification MUST fail closed. (Empirically the fee applies to `token_diff` matching rather than plain `transfer`/`ft_withdraw`, but implementations MUST verify delivery equality rather than assume.)
```
IkerAlus: Rule 10 correctly notes simulation excludes external async effects, so for `ft_withdraw` there is no simulated "payTo receives" artifact for Rule 11 to compare. Suggest specifying, per delivery mode, exactly which simulation output is checked (e.g. the DIP-4 withdraw/transfer event amounts and the payer debit) for the fail-closed equality.

specs/schemes/exact/scheme_exact_near.md
```diff
@@ -274,6 +290,118 @@ Failure:
...
+- **`erc191`** — flat form: `{ "standard": "erc191", "payload": "<JSON string>", "signature": "secp256k1:<base58 of 65 bytes r||s||v>" }`. There is no envelope: the signed JSON string itself carries `signer_id`, `verifying_contract: "intents.near"`, `deadline`, `nonce`, and `intents`. The digest is EIP-191 `personal_sign` over the exact payload string bytes; the signer is recovered from the signature (`v` of 27/28 MUST be normalized to 0/1). Reference vector: the accepted `execute_intents` args of mainnet transaction [`ALqQDmeTUouVv6vqG8bqbwBo7RzxBuB7gufPwtcLyw6E`](https://nearblocks.io/txns/ALqQDmeTUouVv6vqG8bqbwBo7RzxBuB7gufPwtcLyw6E).
+
+The message string MUST be carried byte-identical from signing through submission — no re-serialization, no key reordering, no whitespace changes.
+
+### Intent Binding (Exactness)
```
IkerAlus, line 363: Intent binding should close the optional fields, concretely `msg` MUST be absent. The binding constrains `token`/`receiver_id`/`amount` but not `FtWithdraw`'s optional fields. If `msg` is present, delivery becomes `ft_transfer_call`, where the receiver's `ft_on_transfer` can partially accept: the resolve logic refunds the unused remainder, so the merchant can receive less than `amount` while settlement reports success.

specs/schemes/exact/scheme_exact_near.md
IkerAlus: On the underlying fee question: my understanding is the Verifier protocol fee applies to `token_diff` matching, consistent with your observation. A quick LLM research  to the source code also supports this.

IkerAlus (commented):
Clear support to this amendment  from my side (NEAR Foundation / NEAR Intents).

I believe that complementary framing is accurate: this covers the facilitator-settled, same-chain, authorization-based case (same trust shape as EIP-3009). While #2102 covers client-settled cross-chain prepayment and is moving to `upfront` scheme, pending #2520. 

Together with `delegate`, the NEAR routes now partition cleanly: 
1. NEAR-native payers → `delegate`
2. Verifier-balance payers, any signature standard, same asset → `intents`
3. any-chain/any-asset cold wallets → #2102 method under `upfront` scheme. 

Wrt the actual spec amendment, it looks strong and almost ready, with minor fixes or improvements as suggested in my comments.

mikedotexe:
Thank you—this is exactly the contract-level review this amendment needed.

I agree with the route partition and the points raised in all six threads. For maintainers reading across networks, the central distinction is settlement authority:

- `delegate` and the renamed `intents-verifier` are facilitator-settled authorizations on the destination chain.
- #2102 is a client-settled, cross-chain payment-proof flow under `upfront`.

I’ll revise the specification to:

1. define a standard-neutral decoded intent document while preserving the original signed bytes;
2. close optional fields that could introduce partial, additional, or asynchronous effects;
3. describe balance reads and simulation as non-reserving preflight;
4. bind exactness to delivery-specific simulated events plus authoritative final settlement evidence; and
5. document rollback, refund, and consumed-nonce behavior so dead authorizations are never retried.

I also rechecked the simulation documentation and source. The documentation does say signatures are validated, and the implementation confirms that. The actual limitation is that simulation cannot observe external asynchronous delivery. I’ll correct that wording as part of this revision.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Addressed in 472759bc. The spec now defines a standard-neutral decoded intent document: NEP-413 parses the exact payload.message string, while ERC-191 parses the exact payload string. The original signature-bound string remains byte-identical for submission.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Addressed in 472759bc. Settlement now distinguishes synchronous rollback from asynchronous wallet failure. The latter refunds the payer’s Verifier balance, consumes the nonce, returns failure, and requires a fresh 402 response and payload. Nonce consumption alone is not delivery proof.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Agreed and renamed throughout to `intents-verifier` in 472759bc, including the discriminator, headings, examples, and comparison table.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Addressed in 472759bc. Balance and simulation are now explicitly non-reserving, point-in-time preflight checks. A late check is required, but settlement must still handle the TOCTOU race as a normal failure path.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Addressed in 472759bc. The rule now validates per-mode intents_executed plus exact signer-attributed DIP-4 ft_withdraw/transfer events; wallet delivery remains final receipt evidence. The fee wording now cites the v0.4.2 token_diff source and no longer asks simulation to prove that payTo received funds.

specs/schemes/exact/scheme_exact_near.md
mikedotexe: Addressed in 472759bc. The wallet intent shape is closed: msg, memo, storage_deposit, min_gas, and unknown fields must be absent. Internal notification fields must also be absent.

IkerAlus (approved):
After the latest changes, I consider all my comments and raised issues addressed. Hence, from the perspective of NEAR blockchain and NEAR Intents, this method extension is good to go. Good work.

PR #2952, opened 2026-07-25
spec(batch-settlement-xrpl): add scheme description for xrpl batch-se…

desimmons:
## Description

This PR adds a draft XRPL network binding for the x402 `batch-settlement` scheme.

The binding uses a pre-funded XRPL Payment Channel and a reusable signed claim as a cumulative authorization cap. For each paid request, the facilitator reserves the maximum charge, records the actual charge, and returns a stable off-ledger commitment identifier. The channel destination later redeems the accumulated amount with an XRPL `PaymentChannelClaim`.

The specification defines:

* XRPL network, native XRP asset, account, and amount formats
* Payment requirements and reusable authorization payloads
* Claim signing and verification against validated ledger state
* Durable channel state, atomic reservations, and request idempotency
* Dynamic per-request charging and settlement identifiers
* Channel redemption, expiry, closure, and recovery
* Optional opaque payout routing through a collector
* Trust assumptions, security considerations, and minimum conformance tests

The same signed authorization can be reused across requests. Each logical request therefore carries a unique `payment-identifier`, which the facilitator uses to prevent duplicate charges and the resource server uses to prevent duplicate execution.

This is a specification-only change. It does not modify the TypeScript, Python, or Go implementations.

## Tests

Documentation-only change; no runtime code paths are modified.

Validation performed:

* Parsed the JSON examples
* Checked the document with a Markdown linter
* Reviewed claim signing and redemption against the XRPL Payment Channel format
* Reviewed the binding against the generic x402 `batch-settlement` requirements
* Reviewed replay, concurrency, recovery, dynamic-pricing, payout-routing, and redemption flows

## Checklist

* [x] I have formatted and linted my changes
* [x] This is a documentation-only change; runtime tests are not applicable
* [x] My commits are signed
* [x] A changelog fragment is not required for this specification-only change

### DIFF
--- a/specs/schemes/batch-settlement/scheme_batch_settlement_xrpl.md
+++ b/specs/schemes/batch-settlement/scheme_batch_settlement_xrpl.md
@@ -1,0 +1,888 @@
+# Scheme: `batch-settlement` on XRPL with opaque payout routing
+
+Status: Draft  
+Version: `xrpl-paychan-v1`
+
+## 1. Scope
+
+This document defines the XRPL network binding for the x402
+`batch-settlement` scheme.
+
+The binding uses an XRP Ledger Payment Channel with per-request cumulative,
+capital-backed authorizations:
+
+1. The payer funds a channel whose destination is the settlement recipient.
+2. For each request, the payer signs an XRPL claim whose cumulative amount is
+   the committed channel amount plus that request's maximum charge.
+3. The resource server verifies that exact cumulative amount before executing
+   the request.
+4. Each request has a unique x402 `payment-identifier`.
+5. The facilitator records the actual charge for each successful request.
+6. The channel destination later redeems the accumulated charge with an XRPL
+   `PaymentChannelClaim`.
+
+The client supplies a cumulative XRPL claim for each logical request. A retry
+of the same logical request reuses the same claim and payment identifier.
+
+This version supports native XRP only.
+
+## 2. Conformance
+
+The key words **MUST**, **MUST NOT**, **REQUIRED**, **SHOULD**, **SHOULD NOT**,
+and **MAY** are to be interpreted as described in BCP 14 when they appear in
+all capitals.
+
+An implementation conforms to this binding only if it implements the x402 v2
+facilitator interface and all requirements in this document.
+
+## 3. Roles and trust model
+
+- **Payer:** The XRPL account that creates and funds the payment channel.
+- **Channel destination:** The XRPL account in `PayChannel.Destination`. It is
+  the x402 `payTo` account and can redeem the signed authorization.
+- **Resource server:** The HTTP server providing the paid resource.
+- **Facilitator:** The service that verifies authorizations, reserves capacity,
+  commits charges, and exposes `/verify`, `/settle`, and `/supported`.
+- **Beneficiary:** An optional downstream recipient identified by the opaque
+  `PaymentRequirements.extra.payoutReference`.
+
+The payer-to-channel-destination leg is capital-backed by XRP allocated to the
+channel.
+
+When `extra.payoutReference` is present, the channel destination is a
+collector. The facilitator resolves the reference in its registry to the
+beneficiary's payout configuration. The collector owes a separate payout to
+that beneficiary. The reference does not prescribe a payout network or wallet
+format. That downstream obligation is credit-backed: the XRPL channel does not
+cryptographically earmark funds for the beneficiary.
+
+Each XRPL signature authorizes the channel destination to claim up to its
+signed cumulative amount. The resource server's authoritative state determines
+the expected amount for each request and records the lower actual charge.
+Clients **SHOULD** treat the highest amount they have signed as their exposure
+to the channel destination.
+
+## 4. Identifiers and amounts
+
+### 4.1 Network
+
+`PaymentRequirements.network` **MUST** be:
+
+```text
+xrpl:<network_id>
+```
+
+`<network_id>` is the unsigned `network_id` returned by XRPL `server_info`.
+Examples are `xrpl:0` for livenet and `xrpl:1` for testnet.
+
+The facilitator **MUST** confirm that its XRPL endpoint reports the requested
+`network_id`.
+
+### 4.2 Asset
+
+`PaymentRequirements.asset` **MUST** identify native XRP:
+
+```text
+xrpl:<network_id>/slip44:144
+```
+
+For example:
+
+```text
+xrpl:0/slip44:144
+```
+
+Issued currencies, MPTs, LP tokens, NFTs, and non-XRP native assets are outside
+this version.
+
+### 4.3 Amounts
+
+All monetary amounts **MUST** be non-negative base-10 integer strings
+representing drops. JSON numbers, signs, decimal XRP values, separators,
+whitespace, and scientific notation **MUST NOT** be used.
+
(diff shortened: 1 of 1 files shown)


whawk46 (commented):
Read the full binding as an operator building XRPL facilitator lanes. Overall: this is a strong spec — recommend landing it after minor clarifications.

**What stands out (worth naming, because most specs don't do this):**
- The invariant chain `ledgerBalance <= committedAmount <= authorizedAmount <= PayChannel.Amount` is stated once and re-checked at every transition, including recovery.
- §16 is honest about the trust boundaries instead of hiding them: 16.1 names the destination's over-claim power up to the signed max; 16.2 admits the claim signature binds only channel+amount, making the mandatory off-ledger checks load-bearing and explicit; 16.4 forbids collectors from calling the credit leg capital-backed. That candor is exactly what lets an implementer reason about what they're actually trusting.
- The corrective-402 recovery (§14.2) puts re-verification on the client against a validated ledger before adopting any server-supplied baseline — right trust direction.
- `minSettleDelay >= maxTimeoutSeconds + redemptionSafetySeconds` correctly defends the in-flight window against payer-initiated closure. And the §17 conformance list is more thorough than most implementations' test suites.
- Verified the claim format against XRPL `PaymentChannelClaim` semantics (`Amount` = signed max, `Balance` = claimed cumulative): correct, and the §11 field table gets the two the right way around.

**Three comments:**

1. **Make the payer's safety property explicit.** §7's rule that independent facilitators MUST NOT accept the same channel is a deployment-topology MUST that no protocol party can observe or verify. The payer's actual protection is real but currently implicit: their maximum exposure is the highest `authorizedAmount` they ever signed, regardless of any split-brain on the metering side (split-brain double-metering is destination-risk, not payer-risk, since redemption is capped by the signature). Suggest stating that as the payer-facing safety property in §16, and adding: a client that receives a corrective 402 it cannot reconcile per §14.2 SHOULD treat it as possible split-brain evidence and retire the channel (§14.2's last paragraph is close — this makes the *why* explicit).

2. **Question — `resource` in the payload example.** §6.3's `PaymentPayload` carries a top-level `"resource": {"url": ...}`. I don't find that member in the v2 envelope or the base batch-settlement scheme — is it inherited from somewhere, or new here? If new, it needs a field table row.

3. **Nit — commitment identifier vs tooling expectations.** `SettleResponse.transaction` carrying `xrpl-paychan:<...>` instead of a tx hash is well-prefixed, but downstream consumers that assume an XRPL hash will mis-parse. One sentence telling consumers to discriminate on the prefix (and that redemption hashes surface separately, if they do) would close it.

---
*Disclosure per CONTRIBUTING: review performed with significant AI assistance under significant human monitoring, direction, and correction; claims checked against XRPL protocol semantics rather than generated from memory.*

desimmons:
> Read the full binding as an operator building XRPL facilitator lanes. Overall: this is a strong spec — recommend landing it after minor clarifications.
> 
> **What stands out (worth naming, because most specs don't do this):**
> 
> * The invariant chain `ledgerBalance <= committedAmount <= authorizedAmount <= PayChannel.Amount` is stated once and re-checked at every transition, including recovery.
> * §16 is honest about the trust boundaries instead of hiding them: 16.1 names the destination's over-claim power up to the signed max; 16.2 admits the claim signature binds only channel+amount, making the mandatory off-ledger checks load-bearing and explicit; 16.4 forbids collectors from calling the credit leg capital-backed. That candor is exactly what lets an implementer reason about what they're actually trusting.
> * The corrective-402 recovery (§14.2) puts re-verification on the client against a validated ledger before adopting any server-supplied baseline — right trust direction.
> * `minSettleDelay >= maxTimeoutSeconds + redemptionSafetySeconds` correctly defends the in-flight window against payer-initiated closure. And the §17 conformance list is more thorough than most implementations' test suites.
> * Verified the claim format against XRPL `PaymentChannelClaim` semantics (`Amount` = signed max, `Balance` = claimed cumulative): correct, and the §11 field table gets the two the right way around.
> 
> **Three comments:**
> 
> 1. **Make the payer's safety property explicit.** §7's rule that independent facilitators MUST NOT accept the same channel is a deployment-topology MUST that no protocol party can observe or verify. The payer's actual protection is real but currently implicit: their maximum exposure is the highest `authorizedAmount` they ever signed, regardless of any split-brain on the metering side (split-brain double-metering is destination-risk, not payer-risk, since redemption is capped by the signature). Suggest stating that as the payer-facing safety property in §16, and adding: a client that receives a corrective 402 it cannot reconcile per §14.2 SHOULD treat it as possible split-brain evidence and retire the channel (§14.2's last paragraph is close — this makes the _why_ explicit).
> 2. **Question — `resource` in the payload example.** §6.3's `PaymentPayload` carries a top-level `"resource": {"url": ...}`. I don't find that member in the v2 envelope or the base batch-settlement scheme — is it inherited from somewhere, or new here? If new, it needs a field table row.
> 3. **Nit — commitment identifier vs tooling expectations.** `SettleResponse.transaction` carrying `xrpl-paychan:<...>` instead of a tx hash is well-prefixed, but downstream consumers that assume an XRPL hash will mis-parse. One sentence telling consumers to discriminate on the prefix (and that redemption hashes surface separately, if they do) would close it.
> 
> _Disclosure per CONTRIBUTING: review performed with significant AI assistance under significant human monitoring, direction, and correction; claims checked against XRPL protocol semantics rather than generated from memory._

@whawk46 Thanks for the review! I've pushed changes which should address your three comment:

1. **Payer safety and split-brain:** Agreed. Section 16 now explicitly states that, absent a separate payer-authorized on-ledger transaction, the destination cannot increase the channel’s cumulative balance beyond the highest valid `authorizedAmount` the payer has signed or `PayChannel.Amount`. Split-brain may cause duplicate service delivery or uncollectible destination-side charges, but it cannot make XRPL pay each branch independently. Section 14.2 also now identifies an unreconcilable corrective 402 as possible split-brain or state-corruption evidence and retains the stronger requirement that the client stop using and retire the channel.

2. **`resource` in the payload:** This is inherited from the optional core x402 v2 `PaymentPayload.resource` (`ResourceInfo`) field; it is not introduced by the XRPL binding. It is exposed by both SDKs and the core x402 v2 scheme:

   * [Python: `PaymentPayload.resource: ResourceInfo | None`](https://github.com/Dhali-org/x402/blob/xrpl-batch-settlement/python/x402/schemas/payments.py#L81-L96)
   * [JavaScript/TypeScript: `PaymentPayload.resource?: ResourceInfo`](https://github.com/Dhali-org/x402/blob/xrpl-batch-settlement/typescript/packages/core/src/types/payments.ts#L30-L35)
   * [Core x402 v2 scheme: `PaymentPayloadV2Schema.resource`](https://github.com/x402-foundation/x402/blob/main/specs/x402-specification-v2.md#52-paymentpayload-schema)

   The base batch-settlement document does not reproduce the complete v2 envelope. I added a sentence in Section 6.3 making this inheritance explicit, rather than adding a row to the XRPL-specific field table.

3. **Commitment identifier:** Agreed. Section 9.1 now requires consumers to discriminate the `xrpl-paychan:` prefix before applying XRPL transaction-hash parsing. It also clarifies that any later redemption transaction hashes are separate redemption records and do not replace the per-request commitment identifier in `SettleResponse`.

Thanks again! These clarifications make the operational boundaries easier to consume without changing the protocol behavior.

PR #2958, opened 2026-07-26
fix(http): JSON 402 response body should mirror the PAYMENT-REQUIRED header, not default to {}

StelarDigital:
`x402HTTPServerBase._create_http_response()` builds the `PAYMENT-REQUIRED`
header from the `PaymentRequired` model (`model_dump_json(by_alias=True,
exclude_none=True)`, via `encode_payment_required_header`), but the
non-browser JSON response body defaults to `{}` — an empty challenge.

Impact: any client that reads the payment challenge from the response body
instead of decoding the header (e.g. clients that do not special-case the
`PAYMENT-REQUIRED` header) receives an empty object and fails closed with
no challenge to act on, even though a spec-compliant challenge is present
in the header the whole time.

Fix: default the body to
`payment_required.model_dump(by_alias=True, exclude_none=True, mode="json")`
— the same field set, aliasing, and null-exclusion the header uses;
`mode="json"` preserves the same type coercion as `model_dump_json`. This
only changes the *default* — an explicit route-level `unpaid_response`
(`unpaid_response_body` hook) is unaffected and still overrides the body,
same as before.

Adds a unit test asserting the JSON 402 body deep-equals the decoded
header payload (and is non-empty), a serialization-flag-parity test, and
a test confirming the route-level override still wins.

(Found via a public endpoint checker exercising the JSON 402 path against
a live x402 resource server.)

### DIFF
--- a/python/x402/http/x402_http_server_base.py
+++ b/python/x402/http/x402_http_server_base.py
@@ -822,9 +822,13 @@ class x402HTTPServerBase:
                 is_html=True,
             )
 
-        # API response
+        # API response. Default the body to the same payload the
+        # PAYMENT-REQUIRED header carries so body-reading clients see the
+        # full challenge rather than an empty object.
         content_type = "application/json"
-        body: Any = {}
+        body: Any = payment_required.model_dump(
+            by_alias=True, exclude_none=True, mode="json"
+        )
 
         if unpaid_response:
             content_type = unpaid_response.content_type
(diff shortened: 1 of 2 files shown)


phdargen:
Thanks for the suggestion but this would only lead to more confusion. Clients should not expect the PaymentRequired response in the body. The PAYMENT-REQUIRED header is the only canonical, spec compliant path. We are currently considering non-empty body defaults with hints

PR #2962, opened 2026-07-27
fix(go): prevent settlement override percent overflow

wnjoon:
## Description

Prevent overflow when resolving large percentage settlement overrides in Go.

The root cause was in `ResolveSettlementOverrideAmount` (`go/server.go:30`): the percentage integer part was parsed into an `int64` while parse errors were ignored, and the value was multiplied by 100 using `int64` arithmetic. 
Large percentage values could therefore overflow and produce a negative or zero settlement amount.

A Go changelog fragment is included.

## Tests

- `go test ./...`
- `go vet ./...`
- `gofmt -d server.go server_test.go`

## Checklist

- [x] My commits are signed
- [x] I added a changelog fragment for the user-facing change

### DIFF
--- a/go/.changes/unreleased/fixed-20260727-140953.yaml
+++ b/go/.changes/unreleased/fixed-20260727-140953.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Prevented overflow when resolving large percentage settlement overrides.
+time: 2026-07-27T14:09:53.922115+09:00
--- a/go/server.go
+++ b/go/server.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"math/big"
 	"regexp"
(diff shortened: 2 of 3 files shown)


wnjoon:
@phdargen 

**Quick follow-up**: 

I rechecked current main after #3156. The percentage path still parses the integer part with `strconv.ParseInt(..., 64)`, ignores the error, and multiplies using `int64`, so the overflow remains reproducible.

This PR is currently mergeable and its signed CI checks passed, although it is now behind main. 
Would maintainers prefer a rebase as-is, or should very large percentage values be rejected instead of handled with `big.Int`?

phdargen (approved):
good catch, thanks for the patch @wnjoon

Issue #2963, opened 2026-07-27
[Bug]: dollar settlement overrides resolve inconsistently across SDKs


wnjoon:
## What happened?

Dollar-denominated settlement overrides resolve to different atomic-unit amounts across the Go, TypeScript, and Python SDKs.

For a token with 6 decimals, the current implementations produce:

| Override | Go | TypeScript | Python |
|---|---:|---:|---:|
| `$0.0000005` | `0` | `1` | `0` |
| `$0.0000009` | `0` | `1` | `1` |
| `$1.0000005` | `1000000` | `1000001` | `1000001` |

The implementations currently use different conversion semantics:

- Go uses `big.Float.Int()`, which truncates toward zero for these positive values.
- TypeScript uses `Math.round(parseFloat(...))`.
- Python uses `round(float(...))`, whose result can be affected by both Python's rounding semantics and binary floating-point representation.

As a result, the same x402 settlement override can represent different atomic amounts depending on the SDK processing it. Using binary floating point for monetary decimal conversion can also make edge-case behavior dependent on representation details.

I expected all SDKs to resolve the same decimal input deterministically.

Relevant implementations:

- Go: `go/server.go` (`ResolveSettlementOverrideAmount`)
- TypeScript: `typescript/packages/core/src/server/x402ResourceServer.ts` (`resolveSettlementOverrideAmount`)
- Python: `python/x402/http/x402_http_server_base.py` (`_resolve_settlement_override_amount`)

Before implementing a fix, could maintainers confirm the canonical behavior for inputs with more fractional digits than the asset supports?

1. Truncate/floor to atomic units.
2. Round using a defined rule such as round-half-up.
3. Reject the override as having excessive precision.

Whichever policy is selected, I propose parsing the decimal string directly rather than converting through binary floating point, and adding shared cross-SDK conformance vectors.

It would also be helpful to confirm whether the SDK fixes should be submitted as separate language-specific PRs linked to this issue or as one atomic cross-SDK PR. I can implement the Go portion once the expected behavior is agreed.

## Steps to reproduce

Using the current settlement override resolver in each SDK:

1. Set the asset decimals to `6`.
2. Resolve each of the following dollar overrides:
   - `$0.0000005`
   - `$0.0000009`
   - `$1.0000005`
3. Compare the resulting atomic-unit strings across Go, TypeScript, and Python.

The results differ as shown in the table above.

## Environment

- Repository: `x402-foundation/x402`
[...]

PR #2964, opened 2026-07-27
fix(go): make dollar settlement override conversion exact

wnjoon:
## Description

Fix the Go dollar settlement override conversion so it does not lose precision for large decimal values.

The root cause was in `ResolveSettlementOverrideAmount` in `go/server.go`: dollar amounts were parsed with a fixed 256-bit `big.Float`, multiplied by `10^decimals`, and then truncated. 

Decimal values exceeding that precision could be rounded before truncation and produce an incorrect atomic-unit amount.

This change parses the validated decimal string directly and constructs the atomic-unit value using decimal digits and `big.Int`. It supports arbitrarily long valid dollar inputs without binary floating-point conversion.

The current Go behavior for fractional digits beyond the asset precision is intentionally preserved: positive values are truncated toward zero. The cross-SDK rounding policy remains under discussion in #2963, so this PR should remain Draft until maintainers confirm the canonical behavior.

Related to #2963.

## Tests

Added coverage for:

- exact representable dollar values
- 6-decimal precision boundaries
- 0, 8, and 18 asset decimals
- a decimal value exceeding 256-bit floating-point precision
- preservation of existing truncation behavior

The long-precision regression test failed against the parent implementation and passes with this change.

Validation:

- `make fmt`
- `make lint`
- `make build`
- `make test`
- `git diff --check`

## Checklist

- [x] My commit is signed
- [x] I added a Go changelog fragment

### DIFF
--- a/go/.changes/unreleased/fixed-20260727-161020.yaml
+++ b/go/.changes/unreleased/fixed-20260727-161020.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Made dollar settlement override conversion exact for large decimal values.
+time: 2026-07-27T16:10:20+09:00
--- a/go/server.go
+++ b/go/server.go
@@ -45,16 +45,24 @@ func ResolveSettlementOverrideAmount(rawAmount string, requirements types.Paymen
 	}
 
 	if m := dollarRegex.FindStringSubmatch(rawAmount); m != nil {
-		dollarFloat, ok := new(big.Float).SetPrec(256).SetString(m[1])
+		parts := strings.SplitN(m[1], ".", 2)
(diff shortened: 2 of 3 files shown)


wnjoon:
@phdargen 

Closing this as superseded by #3156, together with #3154 for the TypeScript and Python paths.

The merged implementation adopts exact decimal-string conversion and truncation toward zero, which addresses the core issue raised here.

The production change in this PR is therefore no longer needed.

This PR does include an additional regression vector using a decimal value beyond the previous 256-bit `big.Float` precision. I would be happy to submit that as a small test-only follow-up if maintainers think it would still be useful.

Issue #2967, opened 2026-07-27
Proposal: trusted caller attribution for settlement analytics


mohamedkuch:
## Problem

A successful settlement identifies payment, but not why the payer arrived. Merchants can therefore count verifier calls, prearranged integration tests, and one-call ecosystem samplers as organic customer demand.

This has occurred in production. One merchant initially counted verifier settlements as customer demand, while another observed a newly created wallet paying 47 services once each. Payer-wallet heuristics cannot correct this reliably and create privacy and portability problems.

## Proposed extension

Add an optional `caller-attribution` extension to successful settlement responses. It carries one or more independently signed assertions:

- `verifier`: a verification program known to the asserting party
- `prearranged_integration`: a known partner or SDK validation call
- `catalog_referral`: an acquisition path attributable to a catalog
- `unknown`: no trustworthy attribution information

Each assertion contains `callerClass`, `assertedBy`, an optional or conditionally required namespaced `sourceId`, and a settlement-bound proof. Multiple assertions are required because a catalog can assert referral provenance while a facilitator separately identifies a verification program.

Facilitator and catalog assertions would reuse the JWS and EIP-712 formats plus DID-based signer authorization already defined by `offer-receipt`. The signed payload would bind the assertion to the settlement network, transaction, and payer.

## Trust and privacy rules

- Client assertions are debugging hints and MUST NOT count toward demand, customer, or revenue metrics.
- Absence, an empty assertion array, and `unknown` all mean no information. None implies organic demand.
- `sourceId` MUST be present for `catalog_referral` and globally namespaced.
- Merchants MUST NOT infer verified attribution from payer-wallet, IP, user-agent, or fingerprinting heuristics.
- A trusted verifier or prearranged assertion excludes that settlement from organic customer-demand metrics, even when a separate catalog referral is present.

## Why existing extensions do not cover this

`builder-code` records participating application, wallet, and service codes, including a client-supplied service code that is explicitly self-reported. It does not classify acquisition intent or provide trusted verifier and integration-test cohorts.

`offer-receipt` proves server terms and delivery. Acquisition provenance has different signers and trust semantics, so it should remain a separate composable extension.

## Reviewed draft

The design was reviewed by a merchant operating across three chains and two facilitators. The current draft and review history are available here:

https://gist.github.com/mohamedkuch/881d4d37c243efcb287ce4bc2fe6e19d

mohamedkuch:
A spec-only draft is available in #2968.

[...]

PR #2968, opened 2026-07-27
docs: propose trusted caller attribution

mohamedkuch:
## Description

Addresses #2967.

This draft adds a version 2 `caller-attribution` extension for signed, settlement-bound caller classification.

The proposal distinguishes:

- verifier activity,
- prearranged integration tests,
- catalog referral provenance, and
- unresolved attribution.

Each trusted assertion is independently signed and bound to the settlement network, transaction, and payer. The specification defines JWS and EIP-712 artifacts, signer authorization, issuer-role trust, multiple-assertion handling, conservative conflict resolution, privacy limits, and analytics semantics.

The draft keeps gross settled revenue separate from organic customer revenue. It also makes client self-reporting non-authoritative and forbids wallet, IP address, user-agent, and fingerprint heuristics from becoming trusted attribution.

This is intentionally spec-only. It does not add SDK or facilitator implementations.

The design was informed by production settlement reconciliation and review from a merchant operating across multiple chains and facilitators:

https://gist.github.com/mohamedkuch/881d4d37c243efcb287ce4bc2fe6e19d

AI assistance disclosure: Codex helped draft and validate this specification. The contribution was reviewed against the repository contribution guides, the x402 version 2 core specification, and the existing `builder-code` and `offer-receipt` extensions.

## Tests

- Parsed every JSON example.
- Validated both extension `info` examples against their Draft 2020-12 JSON Schemas with Ajv.
- Ran Prettier 3.6.2 in check mode.
- Ran `git diff --check`.
- Scanned the new specification for Unicode em dashes.

No runtime tests are applicable because this PR changes only a Markdown specification.

## Checklist

- [x] All applicable tests pass
- [x] My commit is SSH-signed
- [x] This is a docs-only change, so no changelog fragment is required

### DIFF
--- a/specs/extensions/caller-attribution.md
+++ b/specs/extensions/caller-attribution.md
@@ -1,0 +1,455 @@
+# Extension: `caller-attribution`
+
+## Summary
+
+The `caller-attribution` extension lets trusted settlement participants attach signed caller-classification assertions to a successful x402 `SettlementResponse`.
+
+The extension addresses a narrow analytics problem: a successful payment proves that value settled, but it does not prove why the caller paid. Verifier checks, prearranged integration tests, and catalog referrals can otherwise be misclassified as organic customer demand.
+
+This extension:
+
+- binds every assertion to the settlement network, transaction, and payer,
+- distinguishes verifier and prearranged integration activity from unresolved demand,
+- records catalog referral provenance without claiming buyer intent,
+- supports multiple independently signed assertions for one settlement,
+- forbids wallet, IP address, user-agent, and fingerprint-based attribution, and
+- keeps gross settled revenue separate from organic customer revenue.
+
+This specification applies to x402 version 2.
+
+---
+
+## 1. Relationship to Other Extensions
+
+The `builder-code` extension records applications, wallets, and services that participated in an x402 payment. Its service code is client supplied and is not a trusted statement of acquisition intent.
+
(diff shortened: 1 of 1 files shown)


epistemedeus:
Thanks for separating trusted issuer assertions from client-supplied metadata
and for keeping catalog referral distinct from organic demand. This maps to a
live merchant residual we see: discovery and challenge source can be known,
while a later successful construction or settlement becomes unattributed.

Two binding gaps appear to prevent the current draft from closing that join:

1. The signed payload does not identify the selected resource or offer. It
   binds `network`, `transaction`, and `payer`, while core
   `SettlementResponse` has no resource field and `sourceId` identifies a
   catalog surface or program rather than the paid operation. Once the artifact
   is evaluated outside the original HTTP exchange, a verifier cannot establish
   which resource the assertion classifies. The existing offer-receipt payloads
   already bind `resourceUrl`, and signed offers bind the terms. Could this
   payload add `resourceUrl` and require an exact comparison? If exact offer
   selection is in scope, composing or digesting the existing signed offer
   would avoid another terms identity.

2. The executable catalog path into `SettlementResponse` is not defined. The
   core settlement path has the client, resource server, and optional
   facilitator; a Bazaar is a discovery surface. Section 7 lets a catalog issue
   `catalog_referral`, but the draft does not say how that signed assertion
   reaches the facilitator or resource server, how the handoff is
   authenticated, or which participant merges it into the response. One option
   is a bounded detached-assertion handoff and merge rule. Another is to limit
   direct response production to an in-path participant while allowing it to
   carry an already signed catalog assertion.

These are spec-level questions only. I agree that an implementation should not
precede the authority and transport contract.

PR #2973, opened 2026-07-28
fix(go): bound HTTP response body reads

### LINKED ISSUE #2920: fix(go): bound HTTP response body reads in payment and facilitator clients
## Summary

Several non-legacy Go SDK HTTP paths read response bodies with `io.ReadAll` without applying a size limit.

Unexpectedly large facilitator or resource-server responses can therefore cause excessive memory use before the SDK parses or rejects them.

I could not find an existing open issue or PR covering response body limits in the Go SDK.

## Affected paths

The current implementation contains unbounded reads in:

- `go/http/facilitator_client.go`
  - `GetSupported`
  - `verifyHTTP`
  - `settleHTTP`
- `go/http/client.go`
  - initial 402 response handling in `PaymentRoundTripper.RoundTrip`
  - auth-style retry response handling
- `go/extensions/bazaar/facilitator_client.go`
  - facilitator discovery responses

Legacy/v1 code is intentionally out of scope.

## Current behavior

For example, the facilitator client performs:

```go
responseBody, err := io.ReadAll(resp.Body)
```

before validating the status code or decoding the expected JSON response.

The same pattern is used for 402 responses returned by protected resource servers. Although normal x402 responses are small JSON documents, the SDK currently has no upper bound on the amount of response data buffered in memory.

## Expected behavior

The SDK should stop reading and return a clearly identifiable error when an HTTP response body exceeds a documented maximum size.

The limit should apply consistently to:

- successful facilitator responses;
- facilitator error responses;
- initial and corrective 402 responses;
- Bazaar facilitator responses.

## Proposed approach

Introduce a shared bounded-read helper that:

1. reads at most `maxResponseBodyBytes + 1`;
2. reports a distinct error when the response exceeds the limit;
3. preserves existing parsing and error excerpts for responses within the limit;
4. closes the response body appropriately on all paths.

For example:

```go
func readResponseBodyLimited(body io.Reader, limit int64) ([]byte, error) {
    data, err := io.ReadAll(io.LimitReader(body, limit+1))
    if err != nil {
        return nil, err
    }
    if int64(len(data)) > limit {
        return nil, ErrResponseBodyTooLarge
    }
    return data, nil
}
```

I suggest keeping the initial change internal rather than adding public configuration unless there is a demonstrated need for caller-specific limits.

## Open question

What default limit would maintainers prefer?

- 1 MiB, which is already much larger than expected x402 JSON responses; or
- 4 MiB, for additional compatibility with facilitator-specific metadata.

## Suggested tests

- a response exactly at the limit is accepted;
- a response one byte over the limit returns the size-limit error;
- oversized 200, 402, and non-2xx facilitator responses are covered;
- the response body is closed on the oversized-response path;
- existing error excerpts remain unchanged for normal-sized responses.

## Scope

This proposal only covers non-legacy Go HTTP clients. It does not change the wire protocol or response schemas.

I would be happy to submit a focused Go PR with tests and a changelog fragment after confirming the preferred default limit.

wnjoon:
## Summary

- limit buffered HTTP response bodies to 4 MiB across Go payment, facilitator, and Bazaar clients
- return `ErrResponseBodyTooLarge` so callers can identify oversized responses with `errors.Is`
- preserve existing parsing, typed errors, error messages, and body-closing behavior
- add boundary and regression coverage for all affected response paths

## Root cause

The response boundaries in `go/http/client.go:227`, `go/http/facilitator_client.go:365`, and `go/extensions/bazaar/facilitator_client.go:217` used `io.ReadAll` directly. 
A facilitator or resource server could therefore cause the SDK to buffer an arbitrarily large response before validating its status or decoding the expected JSON.

## Implementation

A shared internal helper reads at most 4 MiB plus one byte. Responses exactly at the limit are accepted, while larger responses return an exported sentinel error.

The 4 MiB limit leaves compatibility room for Bazaar resources containing larger extension payloads or MCP JSON schemas while still placing a firm bound on memory usage. The helper and limit remain internal; only the sentinel error is public.

Oversized bodies are closed without draining the remaining untrusted data.

## Testing

- `make fmt`
- `make lint`
- `make build`
- `make test`

Tests cover exact-limit responses, one-byte-over responses, reader errors, all seven affected call paths, body closure, and oversized `429` responses without retrying.

Fixes #2920

### DIFF
--- a/go/.changes/unreleased/fixed-20260728-101028.yaml
+++ b/go/.changes/unreleased/fixed-20260728-101028.yaml
@@ -1,0 +1,4 @@
+kind: fixed
+body: Bounded buffered HTTP response bodies in Go payment, facilitator, and Bazaar
+  clients (1 MiB for control-plane responses, 4 MiB for Bazaar discovery).
+time: 2026-07-28T10:10:28+09:00
--- a/go/extensions/bazaar/facilitator_client.go
+++ b/go/extensions/bazaar/facilitator_client.go
@@ -214,7 +214,7 @@ func (c *BazaarFacilitatorClient) ListDiscoveryResources(
 	defer resp.Body.Close()
 
 	// Read response body
-	body, err := io.ReadAll(resp.Body)
+	body, err := readLimitedBody(resp.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response body: %w", err)
 	}
@@ -275,7 +275,7 @@ func (c *BazaarFacilitatorClient) SearchDiscoveryResources(
 	}
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(resp.Body)
+	body, err := readLimitedBody(resp.Body)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read response body: %w", err)
 	}
@@ -367,4 +367,22 @@ func (c *BazaarFacilitatorClient) buildSearchURL(params *SearchDiscoveryResource
 
 	u.RawQuery = q.Encode()
 	return u.String(), nil
+}
+
+// maxDiscoveryResponseBytes bounds discovery responses. Catalog pages carry
+// resource extensions and MCP JSON schemas, so they need more room than
+// control-plane JSON.
+const maxDiscoveryResponseBytes int64 = 4 << 20
+
+// readLimitedBody reads at most maxDiscoveryResponseBytes from r. A larger body
+// returns x402http.ErrResponseBodyTooLarge without buffering the rest.
+func readLimitedBody(r io.Reader) ([]byte, error) {
+	body, err := io.ReadAll(io.LimitReader(r, maxDiscoveryResponseBytes+1))
+	if err != nil {
+		return nil, err
+	}
+	if int64(len(body)) > maxDiscoveryResponseBytes {
+		return nil, fmt.Errorf("%w: limit %d bytes", x402http.ErrResponseBodyTooLarge, maxDiscoveryResponseBytes)
+	}
+	return body, nil
 }
--- a/go/extensions/bazaar/facilitator_client_test.go
+++ b/go/extensions/bazaar/facilitator_client_test.go
@@ -3,6 +3,8 @@ package bazaar
 import (
 	"context"
 	"encoding/json"
+	"errors"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"testing"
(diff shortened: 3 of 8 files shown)


wnjoon:
@phdargen 

**Quick follow-up**: 

#2920 remains open, and current main still uses unbounded `io.ReadAll` in the affected Go response paths. #2987 was closed without merge and did not resolve the issue.

This PR is currently mergeable and its signed CI checks passed, although it is now behind main. If the proposed 4 MiB limit and sentinel error are acceptable, I can rebase and refresh the validation.

PhilBot402:
Thanks for tackling #2920. The issue is real and the bounded-read approach is right. A few fixes before this is ready:

**Blocking**
- `TestPaymentRoundTripper_RejectsOversizedAuthRetry402` uses a 2-arg `OnPaymentRequired` hook; current API needs `(ctx, paymentRequired, requestURL string)`. CI fails on this.

**Structure**
- Drop `go/internal/httpbody/` and `go/http/response_body.go`. Heavy for ~30 lines and introduces the repo's first `internal/` package.
- Inline bounded reads in the three existing call sites instead:
  - `client.go`: payer 402 / auth retry (~1 MiB)
  - `facilitator_client.go`: verify / settle / supported (~1 MiB)
  - `bazaar/facilitator_client.go`: discovery (~4 MiB)
- Export one shared `ErrResponseBodyTooLarge` from `go/http/http.go` (neutral package surface; not facilitator- or payer-specific).

**Limits**
- Consider tiered limits now rather than one global 4 MiB: control-plane JSON stays tight; Bazaar catalog pages get the higher cap.

AI disclosure: automated by @phdargen. Use your own judgement.

phdargen:
> Thanks for tackling #2920. The issue is real and the bounded-read approach is right. A few fixes before this is ready:
> 
> **Blocking**
> 
> * `TestPaymentRoundTripper_RejectsOversizedAuthRetry402` uses a 2-arg `OnPaymentRequired` hook; current API needs `(ctx, paymentRequired, requestURL string)`. CI fails on this.
> 
> **Structure**
> 
> * Drop `go/internal/httpbody/` and `go/http/response_body.go`. Heavy for ~30 lines and introduces the repo's first `internal/` package.
> * Inline bounded reads in the three existing call sites instead:
>   
>   * `client.go`: payer 402 / auth retry (~1 MiB)
>   * `facilitator_client.go`: verify / settle / supported (~1 MiB)
>   * `bazaar/facilitator_client.go`: discovery (~4 MiB)
> * Export one shared `ErrResponseBodyTooLarge` from `go/http/http.go` (neutral package surface; not facilitator- or payer-specific).
> 
> **Limits**
> 
> * Consider tiered limits now rather than one global 4 MiB: control-plane JSON stays tight; Bazaar catalog pages get the higher cap.
> 
> AI disclosure: automated by @phdargen. Use your own judgement.

Please fix the above @wnjoon

wnjoon:
@phdargen 

Rebased onto current main (626df07e) and addressed all three points in dd97877.

**Blocking — hook signature**

Fixed. The branch was based on a July main, so `TestPaymentRoundTripper_RejectsOversizedAuthRetry402` still used the 2-arg hook. It now uses `(context.Context, types.PaymentRequired, string)`, matching the rest of `client_test.go`.

**Structure**

Dropped `go/internal/httpbody/` and `go/http/response_body.go`. `ErrResponseBodyTooLarge` is now exported once from `go/http/http.go`, alongside an unexported `readLimitedBody` used by the five call sites in that package.

One deviation worth calling out: `go/extensions/bazaar` is a separate package and can't reach an unexported helper, so it carries its own small `readLimitedBody` in `facilitator_client.go` and wraps the shared `x402http.ErrResponseBodyTooLarge`. Duplicating ~10 lines once seemed better than expanding the read inline at all seven call sites. Happy to fully inline in bazaar's two sites instead if you'd rather have zero duplication.

**Tiered limits**

Applied:

- `client.go` — payer 402 and auth retry: 1 MiB
- `facilitator_client.go` — verify / settle / supported: 1 MiB
- `bazaar/facilitator_client.go` — discovery: 4 MiB

Tightening the control plane to 1 MiB is safe for the payer paths: both reads in `client.go` are only reached for `402` responses, since non-402 responses return before the body is buffered (`RoundTrip` returns early, and `tryPaymentRequiredHooks` returns the auth response before reading it). Resource payloads are never buffered by these limits.

The sentinel message is now `http response body too large`, wrapped with the applicable limit (`: limit 1048576 bytes`) since the limit is no longer global. `errors.Is` behavior is unchanged.

**Validation**

`make fmt`, `make lint` (0 issues), `make build`, `make test` all pass on the rebased branch. Test coverage is unchanged in shape: exact-limit, one-byte-over, reader errors, all seven call paths, body closure, and oversized `429` without retry — the helper unit tests moved from the deleted `httpbody_test.go` into `client_test.go`.

phdargen (approved):
Great thank you @wnjoon

PR #2974, opened 2026-07-28
fix: add request timeouts to HTTPFacilitatorClient and guard eager init rejection

### LINKED ISSUE #2961: HTTPFacilitatorClient fetches have no timeout — one hung facilitator request wedges all protected routes until restart
## Problem

The three HTTP facilitator operations in `@x402/core@2.19.0` make `fetch()` calls without an application-level timeout or `AbortSignal`:

- `verify()` — `typescript/packages/core/src/http/httpFacilitatorClient.ts` L304
- `settle()` — L358
- `getSupported()` — L410

If a facilitator, proxy, or load balancer accepts a connection but never completes the response, the operation remains pending until the socket or runtime supplies its own deadline. Depending on the runtime, that can be several minutes or effectively unbounded.

### Global impact during initialization

With the default `syncFacilitatorOnStart: true`, the Hono, Express, Fastify, and Next adapters eagerly retain a shared `initPromise = httpServer.initialize()`.

`initialize()` awaits each facilitator's `getSupported()` sequentially. If one call remains pending:

- later/fallback facilitators are never tried;
- every protected route behind that middleware instance awaits the same pending promise;
- unprotected routes continue serving; and
- the retry path cannot run, because it clears `initPromise` only after rejection.

The protected surface remains blocked until the original fetch rejects—for example because the stalled socket closes, a runtime timeout fires, or the process restarts.

A stalled `verify()` or `settle()` does not share this global initialization latch, but it can still leave the individual paid request pending for an excessive period.

This was observed with a facilitator behind a load balancer that accepted a connection but stalled. Clients eventually hit their own deadlines (for example Undici's headers timeout), while the application continued waiting on the shared initialization promise.

## Reproduction

1. Run a listener that accepts the facilitator connection and keeps it open without sending an HTTP response, for example `nc -lk 4021`.
2. Configure any v2 resource server with `new HTTPFacilitatorClient({ url: "http://localhost:4021" })`, a protected route, and the default `syncFacilitatorOnStart` behavior.
3. Request the protected route. It remains pending in `getSupported()`.
4. Send additional requests to protected routes behind the same middleware instance. They all await the same `initPromise`; unprotected routes still respond.
5. Keep the original accepted connection open while making the facilitator healthy for new connections, as can happen when a load balancer shifts traffic away from a stalled backend. The application still does not recover because it never starts another initialization request.
6. Close the original stalled connection (or restart the process). Once the fetch rejects, the existing `initPromise = null` path can run and a later request can retry initialization.

Killing `nc` closes the accepted socket, so it demonstrates rejection and recovery rather than the persistent-latch condition in step 5.

## Suggested fix

- Add a configurable request timeout to `FacilitatorConfig`, with a default such as 30 seconds. This matches the existing Python and Go facilitator-client defaults.
- Apply the deadline to `verify()`, `settle()`, and every `getSupported()` fetch attempt, including response-body consumption.
- Convert timeout failures into a clear facilitator transport/timeout error instead of exposing a raw `DOMException`.
- Ensure the eagerly created initialization promise has a rejection handler immediately, so a timeout cannot become an unhandled rejection before the first protected request awaits it.
- Add regression tests covering stalled headers, a stalled response body, and successful initialization retry after timeout.

Once a stalled `getSupported()` rejects, the middleware's existing `initPromise = null` logic allows subsequent protected requests to retry.

For `settle()`, a client-side timeout is an indeterminate outcome: the facilitator may have received and completed the settlement after the caller stopped waiting. The timeout should therefore remain configurable and should not be reported as proof that settlement did not occur.

Related: #2157 concerns the `syncFacilitatorOnStart` default and cold-start latency. It is separate from the missing request deadline; the two fixes compose.

notorious-d-e-v:
## Description

Closes #2961.

`HTTPFacilitatorClient.verify()`, `settle()`, and `getSupported()` called `fetch()` with no deadline, so a facilitator (or an LB in front of it) that accepts a connection but never completes the response left the operation pending indefinitely. Through the eagerly created `initPromise` shared by the Hono, Express, Fastify, and Next middlewares, one hung `getSupported()` wedged every protected route behind that middleware instance until the socket happened to die or the process restarted.

This PR:

- **Adds `FacilitatorConfig.timeoutMs`** (default `30_000`, matching the Go client's 30s `Timeout` default and Python's `timeout: float = 30.0`). The deadline is applied via a fresh `AbortSignal.timeout()` per HTTP attempt — `verify()`, `settle()`, and each `getSupported()` retry attempt get their own budget — and the signal is passed to `fetch`, so it also covers response-body consumption. The 429 backoff sleep between `getSupported` attempts is deliberately outside the deadline, and a deadline abort raised while reading a non-2xx error body surfaces as the timeout error rather than a generic (and, for 429, retryable) HTTP failure. Invalid values throw a `RangeError` at construction: `timeoutMs` must be a positive integer of at most 2^31 − 1 ms, since `AbortSignal.timeout()` rejects non-integers and larger values overflow Node's 32-bit timers into an effective ~1ms deadline.
- **Introduces `FacilitatorTimeoutError`** (with `operation` and `timeoutMs` fields) instead of surfacing a raw `DOMException`. It extends `FacilitatorResponseError`, so the existing middleware handling surfaces timeouts as a clean 502 facilitator-boundary response with no adapter logic changes. Non-timeout failures are rethrown untouched (conversion requires both `signal.aborted` and an abort/timeout error in the cause chain, so a network error racing the timer isn't misclassified). Its docs record that a `settle()` timeout is an indeterminate outcome — the facilitator may still have completed the settlement.
- **Guards the eager `initPromise`** in all four middlewares with an immediately attached no-op rejection handler, so an initialization failure (now guaranteed to occur within the timeout) can no longer become an unhandled rejection before the first protected request arrives. The original promise stays stored, so the first protected request still observes the failure and the existing `initPromise = null` retry path recovers — including trying later/fallback facilitators, since `x402ResourceServer.initialize()` continues past a facilitator whose `getSupported()` rejects.

## Tests

- New core tests: stalled response headers (verify/getSupported), stalled response body (settle), `FacilitatorTimeoutError` shape and `FacilitatorResponseError` subclass/`getFacilitatorResponseError` contract, non-timeout errors passing through unchanged, a fresh deadline per `getSupported` retry attempt, config default and validation, and `AbortSignal` being passed to every fetch.
- New middleware tests (Hono, Express, Fastify, Next): an eager initialization rejection with no request in flight emits no `unhandledRejection`, the first protected request still observes the failure, and initialization then retries successfully. These use a hand-rolled `initialize` mock because `vi.fn()` instruments returned promises (for `mock.settledResults`), which attaches a rejection handler and would mask the very bug under test — verified the test fails when the guard is removed.
- Full suites pass locally: `@x402/core` 525, `@x402/hono` 38, `@x402/express` 53, `@x402/fastify` 44, `@x402/next` 56. Lint, prettier, and `tsup` builds pass for all five packages.

## Checklist

### DIFF
--- a/typescript/.changeset/facilitator-client-timeout.md
+++ b/typescript/.changeset/facilitator-client-timeout.md
@@ -6,6 +6,6 @@
 "@x402/next": patch
 ---
 
-Add a configurable per-request timeout to `HTTPFacilitatorClient` (`FacilitatorConfig.timeoutMs`, default 30s, matching the Go and Python facilitator clients). `verify()`, `settle()`, and each `getSupported()` attempt now reject with a typed `FacilitatorTimeoutError` — a `FacilitatorResponseError` subclass the HTTP middlewares already surface as a 502 — instead of hanging indefinitely when a facilitator accepts a connection but never completes the response. A `settle()` timeout is indeterminate: the facilitator may still have completed the settlement.
+Add a configurable per-request timeout to `HTTPFacilitatorClient` (`FacilitatorConfig.timeoutMs`, default 30s, matching the Go and Python facilitator clients; must be a positive integer of at most 2^31 - 1 milliseconds). `verify()`, `settle()`, and each `getSupported()` attempt now reject with a typed `FacilitatorTimeoutError` — a `FacilitatorResponseError` subclass the HTTP middlewares already surface as a 502 — instead of hanging indefinitely when a facilitator accepts a connection but never completes the response. A `settle()` timeout is indeterminate: the facilitator may still have completed the settlement.
 
 The Hono, Express, Fastify, and Next middlewares now attach a rejection handler to the eagerly created facilitator initialization promise, so an initialization failure before the first protected request no longer surfaces as an unhandled rejection; the first protected request still observes the failure and retries initialization.
--- a/typescript/.changeset/facilitator-client-timeout.md
+++ b/typescript/.changeset/facilitator-client-timeout.md
@@ -1,0 +1,11 @@
+---
+"@x402/core": minor
+"@x402/hono": patch
+"@x402/express": patch
+"@x402/fastify": patch
+"@x402/next": patch
+---
+
+Add a configurable per-request timeout to `HTTPFacilitatorClient` (`FacilitatorConfig.timeoutMs`, default 30s, matching the Go and Python facilitator clients). `verify()`, `settle()`, and each `getSupported()` attempt now reject with a typed `FacilitatorTimeoutError` — a `FacilitatorResponseError` subclass the HTTP middlewares already surface as a 502 — instead of hanging indefinitely when a facilitator accepts a connection but never completes the response. A `settle()` timeout is indeterminate: the facilitator may still have completed the settlement.
+
+The Hono, Express, Fastify, and Next middlewares now attach a rejection handler to the eagerly created facilitator initialization promise, so an initialization failure before the first protected request no longer surfaces as an unhandled rejection; the first protected request still observes the failure and retries initialization.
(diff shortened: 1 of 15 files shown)


matrix-agent116:
Preparing PR description...

matrix-agent116:
Preparing PR description...

phdargen (approved):
Thanks @notorious-d-e-v, LGTM 🚀

PR #2994, opened 2026-07-30
fix(ts/go/py): always add client buildercodes

phdargen:
## Description

Service codes are only attached when the resource server advertises the `builder-code` extension in its `PaymentRequired` response. Against servers that do not, the codes are dropped and the payment proceeds unattributed.

Problem: The core client only invokes enrichPaymentPayload when the extension key is present in PaymentRequired.extensions. This is the wrong gate for client-side enrichment. If a client registers an extension, the core client should run it. Server advertisement should govern what the server committed to, not whether the client may attach its own extension data.

Patch: Registered client extensions always get enrichPaymentPayload called. Extensions that require a server declaration must no-op internally when the server did not advertise them. Extensions that attach client-owned data (builder-code s) run unconditionally 

## Checklist


ethanoroshiba (approved):
🚀

PR #2979, opened 2026-07-29
spec(extensions): add discovery — DNS + .well-known facilitator records

whawk46:
> **Now also an IETF Internet-Draft** — [draft-hawkins-x402-dns-discovery](https://datatracker.ietf.org/doc/draft-hawkins-x402-dns-discovery/) (kept aligned with this PR), with IANA registrations filed for the `x402` well-known URI suffix and the `_x402` underscored node name.

## Summary

Adds `specs/extensions/discovery.md` — the `discovery` extension: out-of-band capability records for x402 hosts via a `/.well-known/x402` HTTPS manifest (RFC 8615) plus an optional `_x402.<domain>` DNS TXT pointer. Proposal issue with full motivation: #2978.

With the ecosystem page sunset, facilitator discovery currently relies on hand-curated lists. This extension lets any facilitator or resource server publish one JSON file (and optionally one TXT record) and become discoverable by any client or indexer with zero gatekeepers — curated directories become regenerable caches instead of load-bearing infrastructure. The DNS-pointer→HTTPS-authority split follows the proven MTA-STS pattern, so it stays safe without DNSSEC.

## What's in the spec

- Manifest schema + field semantics (`kind`, facilitator `baseUrl`/`endpoints`/`kinds`/`assets`, optional `resources` compatible with `bazaar`, optional `attestation` claims)
- TXT record grammar (`v=x402-1; wk=…; k=…; net=…; scheme=…`)
- Resolution algorithm (TXT → well-known fallback → live `/supported` cross-check, live data authoritative)
- Security considerations: discovery ≠ endorsement, same-domain constraint on the TXT pointer, spoofing analysis, divergence handling, attestation-claim verification
- Relationship to `bazaar`, curated lists, and future on-chain registries

## Reference implementation

Live manifest (public x402 facilitator on Flare Coston2):
`https://x402-facilitator-427961920698.us-central1.run.app/.well-known/x402`

A tested TypeScript resolver + TXT parser exists; happy to contribute it under `typescript/` in a follow-up if the spec direction is agreeable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

*Disclosure per CONTRIBUTING (added retroactively once we noticed the AI-assisted-contributions policy): this PR was built with significant AI assistance (Claude), under significant human monitoring, direction, and correction. Spec claims and reference-implementation behavior were verified against live networks and running code rather than generated from memory. Happy to answer any provenance questions.*

Issue #2978, opened 2026-07-29
Proposal: discovery extension — DNS + .well-known records for facilitator discovery


whawk46:
## Problem

x402 has three discovery directions, and one of them has no mechanism:

- *"What does this resource cost?"* — solved in-band: the 402 response's `accepts[]` is self-describing.
- *"What resources exist?"* — addressed by the `bazaar` extension and facilitator discovery APIs.
- *"What facilitators exist, and what do they support?"* — unsolved. Today this is hardcoded configuration and hand-curated lists (npm packages, directory sites). Lists rot, gatekeep, and centralize what should be edge-published data — the recent ecosystem-page sunset (in lieu of community aggregators) made the gap concrete.

## High-level approach

Let hosts publish their own capability records using two standard, DNS-era mechanisms — the same pattern email used (MX/SPF/DKIM, and most directly MTA-STS's "DNS TXT pointer → HTTPS policy file"):

1. **`/.well-known/x402`** (RFC 8615): an authoritative HTTPS JSON manifest — kind (facilitator / resource-server), facilitator base URL + endpoint paths, supported `kinds` (mirroring `GET /supported`), settleable assets, optional resources list (`bazaar`-compatible), optional attestation claims.
2. **`_x402.<domain>` DNS TXT** (optional): `v=x402-1; wk=<manifest url>; k=facilitator; net=…; scheme=…` — a pointer + coarse filter so crawlers can discover capability from DNS alone, before any HTTP request. The manifest stays the authority (TXT is unauthenticated without DNSSEC); the TXT `wk` URL must stay within the queried domain.

Resolution: one DNS query, fall back to well-known, then cross-check the live `supported` endpoint (live data wins). Discovery is explicitly **not** endorsement — trust still comes from on-chain settlement verification and consumer policy.

## Why existing mechanisms don't suffice

- `GET /supported` requires already knowing the facilitator's URL — it answers "what does this host support", never "what hosts exist".
- `bazaar` describes *resources* in-band, inside a 402, on a host you already found.
- Curated lists don't scale and centralize discovery; with this extension they become regenerable caches of crawlable data.

## Reference implementation (live)

- Manifest: https://x402-facilitator-427961920698.us-central1.run.app/.well-known/x402 (public x402 facilitator on Flare Coston2 — a platform-subdomain host, so it exercises the manifest-only path)
[...]

PR #3000, opened 2026-07-31
spec(extensions): add attestations — verifiable trust records

### LINKED ISSUE #2999: Extension proposal: attestations — verifiable trust records for x402 participants
## The gap

x402 makes *payment terms* self-describing (the 402 `accepts[]`), and the `discovery` extension proposal (#2978 / #2979) makes *capability* discoverable out-of-band. What remains unsolved at the protocol level is **trust**: an agent choosing between two facilitators, or deciding whether a resource server will deliver after an irreversible settlement, has no machine-verifiable signal — only curated lists and brand reputation.

Existing answers all centralize:

- Curated directories rot and conflate listing with endorsement.
- JWT-style identity networks are verifiable but single-issuer — the issuer becomes a trusted third party.
- Brand trust doesn't exist for the long tail, which is where open payment protocols matter most.

## Proposal sketch

An `attestations` extension standardizing two record types, both surfaced through the `/.well-known/x402` manifest:

1. **Execution attestation** — a facilitator's claim that its settlement key lives in a verifiable execution environment (e.g. a TEE), bound to a reproducible image digest and an on-chain key registration, with a published verification procedure anyone can run. Normative "honest manifest" rule: the block MUST only be served while the full chain currently verifies.
2. **Attestation records ("badges")** — on-chain, soulbound (ERC-721 + ERC-5192), issuer-explicit, typed, evidence-linked, revocable-with-history records about a subject address. `hasActiveBadge(subject, kind)` is the client query. A registry is a namespace, not an authority — clients allowlist issuers the way they pin facilitators today.

Trust flows both directions: services hold records agents check before paying; agents hold records services check before serving.

## Reference implementation (live)

- Facilitator serving a live TEE `attestation` block (Confidential Space TDX, reproducible image, 2-of-3 Safe key registration): https://x402-facilitator-427961920698.us-central1.run.app/.well-known/x402
- Soulbound badge registry deployed + source-verified on Flare Coston2: `0xb02f83e994830C4954c89C10482665A3963229c5`
- Open verifier re-deriving the full attestation chain (14 checks), client-side badge gate in an autonomous agent, and a first *paid* record issuance (audit purchased via x402 `exact`/EIP-3009, charged only on pass, payment tx in the evidence ref).

Spec text in the companion PR. Feedback welcome — particularly on whether `attestation` and `badges` should be one extension or two, and on registering attestation families beyond `tee`.

whawk46:
Adds `specs/extensions/attestations.md` — verifiable trust records for x402 participants.

Closes #2999. Companion to `discovery` (#2978 / #2979): **discovery finds a service; attestations tells you whether to trust it.** Both blocks ride the same `/.well-known/x402` manifest, so a crawler can go DNS TXT → manifest → on-chain trust records with no prior knowledge of the host.

Two record types:

1. **`attestation` block** — facilitator settlement key provably confined to a verifiable execution environment (TEE family defined; zk/MPC registrable later): reproducible image digest, on-chain key registration with attestation reference, published verification procedure. Normative **honest-manifest rule**: serve the block only while the full chain currently verifies — an advertised attestation a client can't reproduce is worse than none.
2. **`badges` block + registry contract requirements** — soulbound, issuer-explicit, typed, evidence-linked, revocable-with-history on-chain records; `hasActiveBadge(subject, kind)` client query; verification algorithm and trust model (record ≠ endorsement; clients allowlist issuers; no central issuer required — that's the point).

Everything specified is running today, not hypothetical:

- Live TEE-attested facilitator manifest: https://x402-facilitator-427961920698.us-central1.run.app/.well-known/x402
- Soulbound registry deployed + source-verified on Flare Coston2: `0xb02f83e994830C4954c89C10482665A3963229c5`
- Open 14-check verifier, agent-side badge gate (refuses revoked venues), and a first paid issuance where the audit itself was bought via x402 (`exact`/EIP-3009) with the payment tx embedded in the evidence ref.

Open questions flagged in #2999: one extension or two (`attestation` vs `badges`), and the registration process for attestation families beyond `tee`.

---

*Disclosure per CONTRIBUTING (added retroactively once we noticed the AI-assisted-contributions policy): this PR was built with significant AI assistance (Claude), under significant human monitoring, direction, and correction. Spec claims and reference-implementation behavior were verified against live networks and running code rather than generated from memory. Happy to answer any provenance questions.*

### DIFF
--- a/specs/extensions/attestations.md
+++ b/specs/extensions/attestations.md
@@ -1,0 +1,191 @@
+# Extension: `attestations` — verifiable trust records for x402 participants
+
+## Summary
+
+The `attestations` extension defines how x402 participants publish
+**machine-verifiable trust records** — claims about *how they operate* that a
+client can check independently, before or after paying. It standardizes two
+record types, both surfaced through the `/.well-known/x402` manifest defined
+by the `discovery` extension:
+
+1. An **execution attestation** — a facilitator's claim that its settlement
+   key lives inside a verifiable execution environment (e.g. a TEE), bound to
+   a reproducible image digest and an on-chain key registration, with a
+   published verification procedure any client can run.
+2. **Attestation records ("badges")** — on-chain, non-transferable,
+   revocable-with-history tokens issued by a named issuer about a subject
+   address (a facilitator, resource server, or agent), each linked to
+   evidence.
+
+Together these answer the question the base protocol leaves open: *x402 tells
+you what a service costs; nothing tells you whether the service can be
+trusted with the payment.* Today that gap is filled by curated lists and
+brand reputation — exactly the centralized pattern the `discovery` extension
+removed from capability lookup. This extension removes it from trust lookup.
+
+## Motivation
+
+x402 payments are final and machine-initiated. An agent choosing between two
+facilitators, or deciding whether a resource server will actually deliver
+after settlement, has no protocol-level signal. Existing approaches:
+
+- **Curated directories** — centralized, rot, and conflate listing with
+  endorsement.
+- **Bearer credentials (JWT-style identity networks)** — verifiable but
+  issued by a single company; the issuer is a trusted third party and a
+  single point of failure.
+- **Brand trust** — does not work for the long tail, which is precisely
+  where open payment protocols matter.
+
+The design goal: **trust claims should be verifiable the way payments are
+verifiable — on-chain or cryptographically, by anyone, with no privileged
+verifier.** A claim that cannot be independently checked MUST NOT be
+advertised (the "honest manifest" rule, normative below).
+
+## Execution attestation (`attestation` manifest block)
+
+A facilitator whose settlement key is provably confined to a verifiable
+execution environment MAY advertise it in its manifest:
+
+```json
+{
+  "x402Version": 1,
+  "kind": "facilitator",
+  "attestation": {
+    "type": "tee",
+    "platform": "confidential-space-tdx",
+    "imageDigest": "sha256:…",
+    "reproducible": true,
+    "keyRegistration": {
+      "network": "coston2",
+      "registry": "0x…",
+      "attestationRef": "x402-facilitator:sha256:…"
+    },
+    "verification": {
+      "procedure": "https://…/verify-facilitator",
+      "evidence": "https://…/attestation-token"
+    }
+  }
+}
+```
+
+- `type` — the attestation family. `"tee"` is defined here; other families
+  (zk, MPC) may be registered later.
+- `imageDigest` — digest of the exact image whose measurement appears in the
+  attestation evidence. `reproducible: true` claims the image can be rebuilt
+  bit-for-bit from public source.
+- `keyRegistration` — where the settlement address was registered on-chain
+  with a reference binding it to the attestation (so key custody claims are
+  auditable history, not just live state).
+- `verification` — a procedure (script, endpoint, or document) that lets a
+  third party re-derive the whole chain: evidence → measurement →
+  imageDigest → registered key → live settlement address.
+
+**Honest-manifest rule (normative):** a manifest MUST only carry an
+`attestation` block while the full verification chain currently passes.
+Servers SHOULD automate this — degrade the block (e.g. to `"type": "none"`
+or omission) the moment any link fails, and restore it only on re-verify.
+An advertised attestation that a client's check cannot reproduce SHOULD be
+treated as **worse than no attestation**.
+
+## Attestation records (`badges` manifest block + registry contract)
+
+### Manifest pointer
+
+Any participant MAY point to on-chain attestation records about itself:
+
+```json
+{
+  "badges": {
+    "registry": { "network": "coston2", "address": "0x…" },
+    "subject": "0x…"
+  }
+}
+```
+
+`subject` is the address the records are about — for a facilitator, its
+settlement address; for a resource server or agent, its receiving/paying
+address.
+
+### Registry contract requirements
+
+A conforming registry is a smart contract where records are:
+
+- **Non-transferable** — soulbound (ERC-721 with ERC-5192 lock semantics, or
+  equivalent). Reputation must not be tradeable.
+- **Issuer-explicit** — every record names its issuer; verifying a record
+  includes deciding whether you trust that issuer. A registry is a namespace,
+  not an authority.
+- **Typed** — each record carries a `kind` (e.g.
+  `x402-facilitator-attested`, `agent-audits-before-paying`) so clients can
+  query for the specific claim they care about:
+  `hasActiveBadge(subject, kind) → bool`.
+- **Evidence-linked** — each record carries an `evidenceRef` URI pointing to
+  the artifact that justified issuance (verification report, audit output,
+  payment receipt).
+- **Revocable with history** — revocation flips the record inactive but MUST
+  NOT erase it. "Held 2026-07 through 2026-09, then revoked" is signal;
+  deletion is not.
+
+### Client verification algorithm
+
+Given a service URL:
+
+1. Fetch `/.well-known/x402` (per the `discovery` extension). No manifest or
+   no `badges` block → status **unadvertised**.
+2. Read `badges.registry` and `badges.subject`; call
+   `hasActiveBadge(subject, kind)` on-chain for each `kind` the client
+   requires.
+3. Active → status **active** (proceed). Present-but-revoked or absent →
+   status **inactive** (clients SHOULD refuse or require explicit override).
+4. Optionally fetch `evidenceRef` and re-run the evidence check itself —
+   records are pointers to proof, not substitutes for it.
+
+Clients SHOULD treat **inactive** as a hard stop and **unadvertised** as a
+soft signal (many honest services simply have no records yet).
+
+### Trust model
+
+- A record is a **verifiable statement by a named issuer**, nothing more.
+  Advertising records is not endorsement by the protocol, the registry, or
+  any directory — the exact analog of the `discovery` rule that discovery ≠
+  endorsement.
+- Trust flows **both directions**: services hold records agents check before
+  paying ("this facilitator's key is enclave-held"), and agents hold records
+  services check before serving ("this agent verifies before it pays").
+- Issuers stake their own on-chain identity on every issuance. Sybil issuers
+  are possible and expected; clients maintain issuer allowlists the same way
+  they maintain facilitator configuration today. Nothing here requires a
+  central issuer — that is the point.
+
+## Interaction with `discovery`
+
+This extension defines two manifest blocks (`attestation`, `badges`) carried
+by the manifest the `discovery` extension standardizes. It is usable without
+DNS TXT records, but composes with them: a crawler can go TXT → manifest →
+on-chain records with no prior knowledge of the host and no HTTP request
+beyond the manifest fetch.
+
+## Reference implementation
+
+- Facilitator with live TEE `attestation` block (Confidential Space TDX,
+  reproducible image, Safe 2-sig key registration) and `badges` pointer.
+- `FCAttestationBadges` — soulbound ERC-721+5192 registry, Safe-issued,
+  deployed and source-verified on Coston2.
+- Open verifier re-deriving the full attestation chain (14 checks).
+- Client-side badge gate in an autonomous agent (refuses revoked venues).
+- First **paid** record issuance: audit purchased via x402 (`exact` /
+  EIP-3009), charged only on pass, payment tx embedded in `evidenceRef`.
+
+## Security considerations
+
+- **Stale attestation** — evidence tokens expire; verification MUST check
+  freshness (nonce/eat_nonce binding) not just signature validity.
+- **Registry substitution** — a malicious manifest can point to a lookalike
+  registry. Clients MUST pin or allowlist registries/issuers they accept,
+  exactly as they pin facilitators today.
+- **Revocation liveness** — clients MUST read active status from chain at
+  decision time; cached "active" results decay.
+- **Evidence rot** — `evidenceRef` URIs can die; issuers SHOULD use
+  content-addressed storage where practical. A record whose evidence is
+  unreachable SHOULD be down-weighted, not treated as revoked.
--- a/specs/extensions/attestations.md
+++ b/specs/extensions/attestations.md
@@ -140,6 +140,50 @@ A conforming registry is a smart contract where records are:
   NOT erase it. "Held 2026-07 through 2026-09, then revoked" is signal;
   deletion is not.
 
+### On-chain representation of the evidence triple
+
+A registry contract usually has **one** string field for evidence, not
+three — and adding typed fields means redeploying to a new address,
+which breaks every manifest already pointing at the registry. The
+triple must therefore survive inside a single string, in a form two
+independent implementations can compare **byte for byte**. Registries
+SHOULD use the following canonical grammar:
+
+```
+x402ev/1; digest=<alg>:<hex>[; anchor=<caip2>:<contract>:<record>][; ref=<uri>]
+```
+
+- `x402ev/1` — version tag. Parsers MUST reject unknown majors.
+- `digest=` — REQUIRED. Algorithm-prefixed so the format survives a
+  hash migration (`sha256:<64 lowercase hex>` today).
+- `anchor=` — OPTIONAL. Where the digest is independently recorded:
+  CAIP-2 chain, contract, record id. **Omit it when the record
+  carrying this string is itself the anchor** — a badge that already
+  contains the digest needs no second witness.
+- `ref=` — OPTIONAL. Where the artifact can be fetched. Link rot is
+  expected; the digest is the truth.
+
+Canonicalization, which is what makes comparison meaningful: fixed
+field order (`digest`, `anchor`, `ref`); separator exactly `"; "`;
+algorithm, digest hex and addresses lowercased; URIs verbatim
+(case-sensitive); no trailing separator. Parsers MUST accept fields in
+any order and MUST re-emit canonical order before comparing. Unknown
+keys MUST be ignored rather than rejected.
+
+Two references describe the same evidence when their **digests**
+match — never their `ref`. A relocated artifact is still the same
+evidence; different bytes at the same URL are not.
+
+Registries with typed fields MAY store the components separately, but
+MUST be able to produce this canonical string, so that cross-registry
+verification stays a string comparison rather than a schema
+negotiation.
+
+Serving the artifact at a path **named by its digest** (e.g.
+`/evidence/<sha256>.json`) is RECOMMENDED: a verifier then fetches and
+re-hashes without trusting the server about which file it received,
+and the URL is immutable by construction.
+
 ### Client verification algorithm
 
 Given a service URL:
--- a/specs/extensions/attestations.md
+++ b/specs/extensions/attestations.md
@@ -121,13 +121,65 @@ A conforming registry is a smart contract where records are:
 
 - **Non-transferable** — soulbound (ERC-721 with ERC-5192 lock semantics, or
   equivalent). Reputation must not be tradeable.
-- **Issuer-explicit** — every record names its issuer; verifying a record
-  includes deciding whether you trust that issuer. A registry is a namespace,
-  not an authority.
+- **Issuer-explicit** — verifying a record includes deciding whether you
+  trust its issuer, so the issuer MUST be determinable from chain state.
+  Two conforming shapes:
+  - **Multi-issuer registries MUST record the issuer per record** (an
+    `issuer` field set at issuance, emitted in the issuance event) and
+    SHOULD expose an issuer-scoped query, e.g.
+    `hasActiveBadgeFrom(subject, kind, issuer)`. Without it, records from
+    an issuer a client trusts are indistinguishable from records from one
+    it does not — and a permissionless registry is then unusable.
+  - **Single-issuer registries MAY** leave the issuer implicit in the
+    contract's own authority (e.g. an `onlyOwner` mint), in which case the
+    registry address IS the issuer identity and clients pin that address.
+
+  Implementers choosing the second shape should note two sharp edges we
+  hit ourselves. First, ownership is usually transferable, so a registry
+  that changes hands **retroactively re-attributes its entire history** —
+  clients pinning such a registry SHOULD also pin the expected owner and
+  re-check it. Second, the choice is one-way in practice: a deployed
+  single-issuer registry cannot later admit other issuers without a
+  migration, because its existing records carry no issuer to distinguish
+  them from the new ones.
+
+  A registry is a namespace, not an authority.
 - **Typed** — each record carries a `kind` (e.g.
   `x402-facilitator-attested`, `agent-audits-before-paying`) so clients can
   query for the specific claim they care about:
   `hasActiveBadge(subject, kind) → bool`.
+- **Durable in what they assert** — a record is a permanent, revocable
+  statement that a **verification happened and produced a given result**,
+  anchored to the evidence for it. A record MUST NOT be read as an
+  assertion about the subject's *present runtime state*. This distinction
+  is easy to lose and expensive to lose: a soulbound record is durable by
+  construction, so pointing one at a transient property produces a claim
+  that is true when written and silently false later, with nothing in the
+  record itself to reveal the drift.
+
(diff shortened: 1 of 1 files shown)


giskard09:
@whawk46 This is a serious design — the "honest manifest" rule (advertised claim MUST match currently-passing verification, degrade immediately on failure) is the right discipline; that's the same reason we don't let an operator's own log stand in for evidence.

One place this could compose with `action_ref`/AnchorRegistry rather than duplicate it: `evidenceRef` for badge issuance points to "the artifact that justified issuance." As written, that's a URI — same trust model as brand reputation if the pointer can be edited or taken down after the fact. If `evidenceRef` were (or could optionally be) an anchored digest instead of a bare link, the badge inherits tamper-evidence from the evidence itself, not from the registry's honesty about not editing the link. Concretely: `evidenceRef` = the `action_ref` of the audit/verification event, anchored on-chain (any permissionless registry works, ours is one instance) — a verifier resolves the digest independently instead of trusting that the URI still points to what it pointed to at issuance time.

Same shape applies to the execution-attestation `verification.evidence` field — the "attestation-token" endpoint is live state; anchoring the digest of a given verification event gives a permanent, independently-checkable record that it passed at a given time, separate from whether the endpoint keeps serving it.

Not proposing a merge of specs — badges/attestation is about who to trust before paying, ours is about what happened after. But the evidence layer underneath both could be the same primitive.

whawk46:
@giskard09 Thank you — this is exactly the right catch. Our Security considerations flagged "evidence rot" and hand-waved at content-addressed storage; your framing is the proper fix: the badge should inherit tamper-evidence from the evidence, not from the registry's honesty about not editing a link.

Amended in 1d72347:

- `evidenceRef` records SHOULD now carry a companion `evidenceDigest` (SHA-256 of the artifact at issuance) and MAY carry an `evidenceAnchor` — a chain reference to where that digest is anchored (any permissionless anchor registry; the spec privileges none). Verifiers SHOULD verify the digest rather than trust the URI, and MUST treat a mismatch as **evidence-invalid** — a stronger negative than unreachable.
- Same structure applied to the execution attestation's `verification.evidence`, for exactly the reason you give: anchoring a verification event's digest gives a permanent record that the chain passed at time T, separate from whether the live endpoint keeps serving the token.
- Security considerations now distinguish evidence *rot* (degrade gracefully) from evidence *tampering* (hard negative) — the digest is what keeps the second from hiding behind the first.

Agreed on the factoring: no spec merge — "who to trust before paying" and "what happened after" staying separate, sharing the evidence primitive underneath, is the right shape.

giskard09:
@whawk46 This lands exactly right — evidenceDigest as the trust anchor and evidenceAnchor as where it's recorded, with mismatch as a hard negative distinct from rot, is the correct shape. Applying the same structure to verification.evidence closes the loop cleanly: a verification event's outcome stays provable independent of whether the endpoint keeps serving.

Agreed on the factoring too. Good to see the primitive composing without either side needing to absorb the other.

whawk46:
@giskard09 — following up to close the loop: the amendment is live in `1d72347b` (evidence digests + optional anchors, digest mismatch treated as evidence-invalid, and the same shape applied to the execution attestation's `verification.evidence`).

Also wanted you to know we credited the catch publicly rather than quietly shipping it — an outside reviewer improving a spec inside a day is the part of this process worth advertising. Thank you for reading it carefully.

If it's useful to you: our reference implementation (attested facilitator manifest, soulbound registry on Flare Coston2, the 14-check verifier) is all open, so if you ever want a second implementation to test `action_ref` / anchor interop against, we're happy to be a guinea pig.

giskard09:
@whawk46 Took you up on the offer and checked it against the deployed registry, not just the spec text. Badge #1 on Flare Coston2 is still running the old evidenceRef string format (image sha256:...; ceremony...; enclave-settled...) — the evidenceDigest/evidenceAnchor shape from 1d72347b hasn't shipped in the contract yet, just the spec amendment. Architecturally it's the same primitive either way — SHA-256 digest plus a pointer into a permissionless registry, no privileged authority — same shape as action_ref/AnchorRegistry.

Can't run the byte-level comparison until there's a real badge minted with the new fields, though. If it's easier on your side: action-ref-conformance (giskard09/action-ref-conformance) is a public GitHub Action with the frozen vectors — verify-vendored mode lets you wire it into your own CI once the new fields ship, so it re-checks on every commit instead of a one-off comparison from our side.

whawk46:
@giskard09 Checking the deployed registry instead of the spec text was the right instinct, and you found a real gap — thank you.

You're right that badge #1 was still free text. But the root cause was bigger than our lag: **the spec never said how a registry with ONE evidence field carries the triple.** `FCAttestationBadges` has a single `string evidenceRef`, and so will most registries — adding typed fields means redeploying to a new address, which breaks every manifest already pointing at it. Every implementer hits this.

Fixed in both places.

**Spec** (`40e06db`) — new section *On-chain representation of the evidence triple*:

```
x402ev/1; digest=<alg>:<hex>[; anchor=<caip2>:<contract>:<record>][; ref=<uri>]
```

Fixed field order; separator exactly `"; "`; algorithm/digest-hex/addresses lowercased; URIs verbatim; parsers accept any order but MUST re-emit canonical order; unknown keys ignored. The digest is algorithm-prefixed so the format survives a hash migration. Two refs describe the same evidence when their **digests** match, never their `ref` — a relocated artifact is still the same evidence; different bytes at the same URL are not. Registries with typed fields MAY store components separately but MUST be able to emit the canonical string, so cross-registry checks stay a string comparison rather than a schema negotiation.

Note on `anchor=`: **omitted deliberately** when the record carrying the string is itself the anchor. Badge #6 contains the digest on-chain, so a second witness would be noise. Where our shapes differ from `action_ref`/AnchorRegistry, that's the seam — a badge is self-anchoring, an action record generally isn't.

**Implementation** — badge **#6** on Coston2, mint tx `0xe33d44aaa79fffcef5027f72dd150bdf82acc6d76938b24342d003a77af0d94d`:

```
x402ev/1; digest=sha256:ad77ae08d5d9b4efd54fa63f4426d90d92bd7063be7b2487bdd8ac439911acde; ref=https://x402-facilitator-427961920698.us-central1.run.app/evidence/ad77ae08d5d9b4efd54fa63f4426d90d92bd7063be7b2487bdd8ac439911acde.json
```

It supersedes #1 rather than replacing it — #1 stays exactly as issued, because a registry that quietly rewrites its own history is worth less than one that doesn't. The artifact is served **content-addressed**: the path is the digest, and the route re-hashes before serving, so it refuses to serve bytes that don't match their own name. Fetch it and re-hash — no trust in us required at any step.

On the conformance Action: genuinely appreciated, and I should be straight with you — our application code is local-only git by policy, so we can't wire a GitHub Action into it. What we can do is run your frozen vectors locally against `verify-vendored` and post the results here. If the vectors are consumable as a plain CLI/fixture set, point me at the entry point and I'll run them against badge #6 this week.

Reciprocally: our 14-check verifier and the registry are open if you want a second implementation to test `action_ref` interop against — happy to be the guinea pig, and equally happy to be told we got something else wrong.

whawk46:
@giskard09 Ran your frozen vectors against badge #6's canonicalization. **8/8 on `action-ref-v1-jcs-sha256`.**

```
POSITIVES (reproduce byte-identical)
  [PASS] 0001-giskard-baseline
  [PASS] 0002-dual-timestamps
  [PASS] 0003-scope-namespace-collision-proof-prefixed
  [PASS] 0003-scope-namespace-collision-proof-unprefixed

NEGATIVES (reject, or diverge from claimed_action_ref)
  [PASS] neg-a01-received-order          diverged -> == correct_action_ref
  [PASS] neg-b01-epoch-integer           rejected (grammar violation, fail-closed)
  [PASS] neg-b02-second-precision        diverged -> == correct_action_ref
  [PASS] neg-b03-microsecond-precision   diverged -> == correct_action_ref
```

Precise about what each negative exercised, since it matters: **b01 is the only one my grammar guard fired on** — an integer timestamp isn't RFC3339, so the adapter exits non-zero rather than coercing it. For a01/b02/b03 the `invocation_payload` you ship is already the well-formed tuple as received, so the correct behaviour is a single deterministic recompute, which lands on `correct_action_ref` and never touches `claimed_action_ref`. No retrying, no alternative preimages, no coercion — one canonicalization, one hash, compare.

On `verifier_must`'s "never reorder": we key-sort because JCS mandates it, not to hunt for a match. Deterministic canonicalization once ≠ trying orderings until one collides. Worth stating explicitly since the distinction is exactly where a buggy verifier hides.

**The function under test is the production one.** I first moved `canonicalJson` out of the badge-issuance script into `src/x402/evidence-ref.ts` and pointed the issuer at it, so the thing your vectors exercise is literally the thing that digests badge evidence — not a test-only reimplementation. Proof that the move was behaviour-preserving: badge #6's evidence digest is byte-identical before and after (`ad77ae08…acde`), which is also why the badge didn't need reissuing.

**Two honest limits**, because you'd find them anyway:
1. Our canonicalizer is a **JCS subset** — sorted keys, no insignificant whitespace, arrays order-preserved, primitives via `JSON.stringify`. That is exact for the artifacts we actually digest (strings, booleans, integers, nested objects). Full RFC 8785 also pins ECMAScript number formatting and UTF-16 code-unit key ordering; the day we digest floats or non-BMP keys, this needs the real implementation, and the code says so rather than pretending otherwise.
2. This was a **local run**, not a CI badge. Our application code is local-only git by policy, so I can't give you the pinned-Action green run your README (correctly) treats as the real claim. I've vendored your frozen vector set into our test tree so it re-checks on our side, and the runner is committed — but I'm not going to describe that as conformance in your sense, because it isn't. If you ever want an independent run, the adapter is 30 lines and reads a preimage on stdin, so it's trivially reproducible by anyone.

Thanks again for pushing on the deployed artifact rather than the prose. Two rounds of review, three concrete improvements to the spec — the `evidenceDigest`/`evidenceAnchor` shape, the rot-vs-tampering distinction, and now the on-chain encoding — none of which we'd have found by re-reading our own text.

giskard09:
@whawk46 Appreciate the precision — especially the note on b01 vs a01/b02/b03, that's exactly the distinction a buggy verifier hides behind.

Being straight back: I can verify the on-chain side myself (already did, last round), but the refactor and the vendored test tree aren't in any public branch I can find — which tracks with what you said about local-only git, no doubt there. Same bar we hold ourselves to, though: a row in PROVIDERS.md needs a verifiable artifact, not a report, however careful. Not calling the 8/8 into question — just not something I can independently confirm yet.

If it ever ends up on a public branch, gist, or even just a log output, happy to recheck and take it from there. No rush on our end.

whawk46:
@giskard09 Fair, and it's our own doctrine pointed back at us — we ask people not to trust claims, so a careful report shouldn't move you. Making it checkable instead.

**First, a correction I owe you.** I described our canonicalizer as a "JCS subset" and hedged about number formatting and UTF-16 key ordering. That was caution, not evidence. So I tested it against the RFC author's own reference vectors (`cyberphone/json-canonicalization`) — arrays, french, structures, unicode, values, weird — **6/6 byte-for-byte**. RFC 8785 is defined in ECMAScript terms, which is why the implementation is six lines: JS sorts by UTF-16 code unit as JCS mandates, and `JSON.stringify` already gives Number::toString formatting and minimal escaping. The honest boundary is narrower than I said: JCS canonicalizes *parsed* JSON, so parse-time losses (duplicate keys, `1.0` vs `1`) are out of scope by design.

That matters more than the 8/8 does — badge digests rest on this function, and third parties are now invited to build on the format.

**Second, the artifact.** Whole thing, Node stdlib only:

```ts
// canonical-json.ts — the function that digests our badge evidence
export function canonicalJson(value: unknown): string {
    if (value === null || typeof value !== 'object') return JSON.stringify(value);
    if (Array.isArray(value)) return `[${value.map(canonicalJson).join(',')}]`;
    const obj = value as Record<string, unknown>;
    const keys = Object.keys(obj).sort();
    return `{${keys.map(k => `${JSON.stringify(k)}:${canonicalJson(obj[k])}`).join(',')}}`;
}

// action-ref-cli.ts — preimage JSON on stdin -> hex sha256 on stdout
const preimage = JSON.parse(readFileSync(0, 'utf8'));                 // reject non-object
const ts = preimage.timestamp;                                         // fail closed, never coerce
if (ts !== undefined && !(typeof ts === 'string' &&
    /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/.test(ts))) process.exit(1);
process.stdout.write(createHash('sha256').update(canonicalJson(preimage), 'utf8').digest('hex'));
```

Copy those and run them against your vectors — that beats anything I can assert, and if it disagrees with my 8/8 I'd rather hear it from you.

**Third, a durable home.** Both suites plus the runners are staged for our public verifier repo (`gitlab.com/flareclaw-saving-demo-group/flareclaw-verifier`, MIT) — where our published trust surface already lives. It runs standalone from a clean checkout. I'll post the URL here when it's pushed; our GitLab credentials were rotated recently, so it needs a hand that isn't mine.

Understood on PROVIDERS.md, and I'm not asking for a row — a local run isn't the pinned-CI claim your README describes, and you're right to hold that line. Happy to take any further conformance detail to your repo rather than keep growing a sub-thread on a spec PR the maintainers have to read.

giskard09:
@whawk46 Ran it myself — no trust in your report anywhere in this. Your canonicalJson + the CLI against our 8 frozen vectors: 8/8, byte-identical on positives, correct fail-closed/recompute behavior on all four negatives. Then against cyberphone/json-canonicalization's own 6 reference pairs (arrays, french, structures, unicode, values, weird — including the ugly number cases like 1E30 and 4.50): 6/6. 14/14, independently confirmed, nothing taken on faith.

The honest boundary you drew earlier holds up: delegating to V8's native Number::toString instead of reimplementing number formatting is exactly why this doesn't break where most JCS implementations do.

Once flareclaw-verifier is public, this becomes citable the way your own bar requires — a pinned artifact someone else can re-run, not a comment thread. Until then: real, verified twice over on our end, just not durable yet. Good problem to have.

flarenetworkflr:
@giskard09 That's the outcome I wanted — you ran it, not my report. Thanks for spending the cycles.

Your note about delegating to `Number::toString` rather than reimplementing it is the useful generalization, so I've written it into the spec as implementation guidance (non-normative). Paraphrasing: RFC 8785 is defined in ECMAScript terms, so in a JS runtime a conformant canonicalizer is a few lines that *delegate* — UTF-16 key ordering is what `Object.keys().sort()` already does, and number formatting is already `Number::toString`. The failures come from hand-rolling that. Also folded in the two behaviours your negatives exercise: fail closed rather than coerce, and canonicalize **once** — deterministic ordering is required by the scheme, not licence to try orderings until one matches a claimed digest.

That's three rounds now where the spec got better because you tested the artifact instead of reading the prose. The evidence digest/anchor shape, rot-vs-tampering, the on-chain encoding, and now this — none of it would have come from us re-reading our own text.

On the durable home: still blocked on credentials that aren't mine to use, so I'm not going to put a date on it. When `flareclaw-verifier` is public I'll post the URL here and you can cite or ignore it as you see fit. Your framing is right — real and verified twice, just not durable yet.

Separately, and in the same spirit: we ran an adversarial audit of our own deployed surface this week and it found eight gaps between what we say and what we do, including two live security bugs. One is worth flagging to you specifically because it touches this spec — an `x402-facilitator-attested` record read `active` on-chain while our own manifest honestly reported `attestation: none`, because the TEE was down. The record wasn't wrong about what it attested; it was being *read* as a liveness signal a soulbound record cannot carry. The spec now separates the durable claim ("was verified") from the live one ("is operating that way right now") and requires clients to consult both. Flagging it because if you build against #3000, that distinction is load-bearing.

giskard09:
@flarenetworkflr appreciated — self-reporting live bugs against your own deployed surface is a stronger signal than any report we could run externally.

The durable/live split you landed maps directly onto something we went through with `outcome_handle`: a signed record can carry "this was true when observed," never "this is true right now" — the moment you read a soulbound attestation as a liveness check, you've smuggled in a claim the artifact never made. Good that it's explicit in the spec now instead of left implicit for clients to get wrong the same way you did.

No rush on the durable home for flareclaw-verifier — the offer to test interop stands whenever the credentials clear.

flarenetworkflr:
@giskard09 I tested the guidance I wrote from your note instead of leaving it as prose. It was **partly wrong**, so here is the correction and the two discriminating cases, in a form you can re-run.

**Key ordering — the case that actually discriminates.** UTF-16 code unit and code point disagree above the BMP, in *opposite* directions. `U+1F600` begins with code unit `0xD83D`, which sorts **below** `0xFFFF`; its code point `0x1F600` sorts **above** it.

```js
const emoji = '\u{1F600}', ffff = '\uFFFF';
[ffff, emoji].sort()                       // → [emoji, ffff]   ← JCS-correct
[ffff, emoji].sort(byCodePoint)            // → [ffff, emoji]   ← different document
```

Sorting by code point is a reasonable thing to write, and it yields a different canonical form — hence a different digest — for identical data. The reason it survives review: **every ASCII key orders identically under both rules.** That's now a test of its own, because the bug is invisible to any test suite written in ASCII.

**Number formatting — my guidance was wrong as published.** I wrote that failures come from "reimplementing number formatting by hand." Testing says otherwise. Implementations shaped like:

```js
String(parseFloat(n.toPrecision(17)))      // agrees with JSON.stringify on every case
```

look hand-rolled but **re-enter `Number::toString`** — re-parsing to a double and calling `String()` hands the job straight back to the runtime. They agree everywhere.

The real rule is narrower: *the digits must come from `Number::toString`*, whatever path you take to get there. What actually breaks is formatting the value **independently**. `Intl.NumberFormat` — the obvious tool to reach for — diverges on five of eight probes:

| input | `JSON.stringify` | `Intl.NumberFormat` |
|---|---|---|
| `1e30` | `1e+30` | `1000000000000000000000000000000` |
| `1e21` | `1e+21` | `1000000000000000000000` |
| `1e-7` | `1e-7` | `0.0000001` |
| `5e-324` | `5e-324` | `0` |
| `-0` | `0` | `-0` |

The last is the subtlest: `Intl` **keeps** the sign on negative zero where JSON drops it. Five inputs, five divergent canonical forms, five different digests.

Also pinned the boundary I claimed earlier by demonstration rather than assertion: `1.0` vs `1`, and duplicate keys, are parse-time losses — gone before a canonicalizer ever sees them, so they're out of scope by construction rather than by choice.

Spec text on our side now matches what the tests show, including the correction. 15 tests, all against the production `canonicalJson` — same function that digests badge evidence, not a test double.

Re-run whichever you like; if any of it disagrees with your implementation I'd rather hear it. And thanks — the useful half of this only exists because you pointed at `Number::toString` in the first place.

giskard09:
@flarenetworkflr following up on this — running your two discriminating cases against our own production canonicalizer (not just the guidance text) turned up the same bug in our code: `sort_keys=True` in Python sorts by code point, same divergence you found in JS. Found it in two places (the trail preimage and a downstream anchor-proof recomputation path that derives from the same preimage), fixed both, added your emoji/`￿` pair as a permanent regression test, deployed — commit e066abb, live since 17:01 UTC today.

Good reminder that "matches on every ASCII test" is exactly the failure mode that hides this class of bug until someone runs a non-BMP key through it. Appreciate you testing rather than trusting the prose — twice now it's found something real on our side, not just yours.

whawk46:
@giskard09 The durable home exists now:

https://gitlab.com/flareclaw-saving-demo-group/flareclaw-verifier/-/tree/main/conformance

What's there: the exact `canonicalJson` that digests `x402ev/1` badge evidence, plus both independently-authored suites:

- `run-jcs.mts` — RFC 8785 vectors from the RFC author's own repository (`cyberphone/json-canonicalization`), byte-for-byte: **6/6**
- `run-conformance.py` — your frozen `action-ref-v1-jcs-sha256` vectors: **8/8** (4 positives reproduce the digest byte-identically, 4 negatives reject or diverge)

Verified from a fresh `git clone` before posting this — only dependency is `tsx`, everything else is stdlib, so the two commands in the README are the whole ceremony.

Honest scope note, since your repo draws this line correctly: these are local runs, not the pinned-CI green run `action-ref-conformance` treats as the real conformance claim. Whether that meets the bar for a `PROVIDERS.md` row is your call — cite or ignore as you see fit.

Silentpartnercoding:
The discovery extension now constrains its manifest and resource fetches, but this attestation extension introduces new fetch-and-run surfaces without carrying those constraints forward. `verification.procedure`, `verification.evidence`, and the optional `evidenceRef` can all be remote URIs; the text invites clients to run the procedure or fetch the evidence, but does not bound their origin, redirects, resolved addresses, response size, or duration.

Could HTTP(S) references inherit a common safe-fetch profile: HTTPS, same-domain or an explicitly trusted origin, validation on every redirect hop, rejection of loopback/link-local/private destinations after resolution, and bounded time and bytes received? I would also say that clients MUST NOT automatically execute a fetched procedure as code. Content addressing and `evidenceDigest` protect artifact integrity after retrieval, but they do not make retrieval or execution safe.

Disclosure: this security review was prepared with AI assistance and manually checked against the current PR text and repository contribution rules.

giskard09:
Verified the clone, both suites pass as reported — appreciate the honest scope note, and it's the right call: local runs (however clean — fresh clone, stdlib-only, byte-identical on the 4 positives) aren't the pinned-CI green run `action-ref-conformance` treats as the real conformance claim for a `PROVIDERS.md` row. Same bar we've held everyone else to, including our own tooling.

This gives you a durable, reproducible home either way — happy to point people at it as evidence of interop, just not as a listed row yet. If/when it runs green under the pinned action (`@v1`) instead of local `tsx`/`python`, that's a one-line ping and it goes in.

whawk46:
@Silentpartnercoding Confirmed on every point — this extension introduced fetch surfaces (`verification.procedure`, `verification.evidence`, `evidenceRef`, `ref=` in the `x402ev/1` string) without carrying forward the constraints the `discovery` extension already established, and nothing forbade the worst reading of "a procedure any client can run." Fixed in both specs:

**This PR (`16d673c`)** — new normative section *Remote references: safe fetching*:
- All HTTP(S) references inherit the `discovery` manifest-fetch profile: HTTPS only; bounded deadline, response size, and redirect hops, re-validated on **every** hop.
- Clients MUST refuse a reference whose host is, or resolves to, a loopback/link-local/private-range address, re-checked per hop. Worth spelling out why the existing rules don't cover it: the publisher controls their own DNS, so an in-domain name can resolve anywhere, and DNS-01 issuance grants valid certificates to names that never point anywhere public.
- Origin: `verification.*` SHOULD be same-origin with the manifest; `evidenceRef`/`ref=` MAY be off-origin (content-addressed storage is legitimate) because the digest is the integrity control — with the explicit caveat you identified: a digest protects what you received, not what fetching it made you do.
- **Clients MUST NOT execute a fetched procedure as code.** Verification logic must be code the verifier already trusts (vendored, pinned, independently obtained); manifest fields supply only *parameters* — URLs, digests, addresses, chain coordinates. A manifest that can inject code into its own verifier verifies nothing.

**#2979 (`607a87a`)** — companion amendment: consumers MUST refuse private destinations before every fetch that spec invites (manifest, redirect hops, `resources[]` probes, live `supported` cross-check), with the rebinding caveat and an explicit escape hatch for intentionally-private deployments.

We also audited our reference resolver against your checklist rather than assuming it passed, and it didn't fully: the live `supported` cross-check was the one fetch still unbounded (no deadline, no size cap, silent redirect-following), and nothing rejected hosts resolving to private addresses. Both fixed, with regression tests covering the redirect-hop re-check and the resolves-to-metadata case specifically.

(Housekeeping in the same push, `cb66605`: three amendments that were discussed and agreed in the review thread above — the durable-vs-live record split, issuer determinability, canonicalization guidance — had been applied on our side but not yet to this branch. The branch now says what the thread says it says.)

whawk46:
@giskard09 Green under the pinned action: https://github.com/whawk46/flareclaw-conformance/actions/runs/30819828942 — `verify-impl` + `verify-vendored` on `action-ref-conformance@v1`, re-verifying on every commit at https://github.com/whawk46/flareclaw-conformance (same drop as the GitLab home, now with the CI claim your README correctly requires).

giskard09:
@whawk46 Confirmed — pinned action, green, re-verifying on every commit. That's the bar for the conformance-tooling side, not the Providers table (that one's about ARGENTUM production trails, which is a different claim). Added under "Verified conformant implementations" in PROVIDERS.md — second independent implementation to clear the frozen profile, which is exactly the kind of interop evidence that matters more than either of us saying so.

PR #3016, opened 2026-08-01
fix(site): validate facilitator request bodies against the real x402 schema

### LINKED ISSUE #3015: Hosted facilitator (x402.org/facilitator): malformed v2 request crashes with 500 instead of returning 400
## Description

`POST /verify` and `POST /settle` on the hosted facilitator (`typescript/site/app/facilitator/{verify,settle}/route.ts`) cast the request body directly to `PaymentPayload`/`PaymentRequirements` with no runtime schema validation:

```ts
const body = await req.json();
paymentPayload = body.paymentPayload as PaymentPayload;
paymentRequirements = body.paymentRequirements as PaymentRequirements;
```

When the body doesn't actually match the expected shape, a property access deep inside the scheme-specific handler (e.g. `typescript/packages/mechanisms/evm/src/exact/facilitator/eip3009.ts:75`'s `payload.accepted.scheme`) throws an uncaught `TypeError`, which the outer `catch` reports as an opaque HTTP 500 `unexpected_error` — with no indication of what was actually wrong with the request.

## Steps to reproduce

Sent to `https://x402.org/facilitator/verify` (live, reproducible right now):

**1. A v2 `paymentPayload` missing the required `accepted` field** (easy to construct if a client builds the v1 shape and just bumps `x402Version` to 2, since v1 and v2 `PaymentPayload` genuinely differ in required fields, not just version number):

```json
{"isValid":false,"invalidReason":"unexpected_error","invalidMessage":"Cannot read properties of undefined (reading 'scheme')","error":"Cannot read properties of undefined (reading 'scheme')"}
```

**2. A v2 `paymentRequirements`/`accepted` using v1's `maxAmountRequired` field name instead of v2's `amount`:**

```json
{"isValid":false,"invalidReason":"unexpected_error","invalidMessage":"Cannot convert undefined to a BigInt","error":"Cannot convert undefined to a BigInt"}
```

Both are HTTP 500. Neither tells the caller which field was wrong, or that the issue is even about the request shape rather than a server-side problem.

## Expected behavior

A malformed request body should fail schema validation and return **HTTP 400** with a specific, actionable reason — e.g. \"paymentPayload.accepted is required for x402Version 2\" — not a 500 with a generic JS error message.

## Suggested fix

`@x402/core/schemas` already exports `parsePaymentPayload`/`parsePaymentRequirements` (safeParse wrappers around `PaymentPayloadSchema`/`PaymentRequirementsSchema`) — validating with these before the routes ever touch a field turns every case above into a clear 400. I've implemented this and verified it fixes both repro cases above (confirmed the 500s disappear and a specific 400 with the Zod validation message is returned instead) — PR incoming.

## Context

Found while integrating a real WooCommerce x402 payment gateway against this facilitator for testnet verification. Once the request was reshaped to match the actual v2 schema, `/verify` and `/settle` both worked correctly and a real transaction settled on Base Sepolia: https://sepolia.basescan.org/tx/0x0c8d2ca373df569a6e5ae401b816355bf38065ce3c583ad57580accec7f0cc32

wakqasahmed:
## Summary

Fixes #3015 — `POST /verify` and `POST /settle` on the hosted facilitator (`typescript/site/app/facilitator/`) previously cast the incoming JSON body directly to `PaymentPayload`/`PaymentRequirements` with no runtime validation. A malformed or wrong-version-shaped body reached the scheme-specific handlers as an unvalidated value, and a missing/mis-shaped field access threw an uncaught `TypeError` surfaced as an opaque HTTP 500 `unexpected_error`.

Both routes now validate with `@x402/core/schemas`' `parsePaymentPayload`/`parsePaymentRequirements` (safeParse wrappers already exported from this monorepo, no new dependency) before ever touching a field, returning HTTP 400 with the Zod validation message on failure.

## Verified live, against the actual production endpoint

Both repro cases in #3015 confirmed fixed:

- v2 `paymentPayload` missing the required `accepted` field: was `500 {"invalidReason":"unexpected_error","invalidMessage":"Cannot read properties of undefined (reading 'scheme')"}`, now returns `400` with a specific Zod message naming the missing field.
- v2 `paymentRequirements`/`accepted` using v1's `maxAmountRequired` instead of v2's `amount`: was `500 {"invalidMessage":"Cannot convert undefined to a BigInt"}`, now returns `400` with a specific message.
- A correctly-shaped v2 request still verifies and settles exactly as before — confirmed with a real signed EIP-3009 authorization on Base Sepolia: [transaction 0x0c8d2ca373df569a6e5ae401b816355bf38065ce3c583ad57580accec7f0cc32](https://sepolia.basescan.org/tx/0x0c8d2ca373df569a6e5ae401b816355bf38065ce3c583ad57580accec7f0cc32) — `verify` returned `{"isValid":true,...}`, `settle` returned `{"success":true,"transaction":"0x0c8d...",...}`, and the transaction is confirmed on-chain (`status: 0x1`) moving the exact expected USDC amount from payer to payee.

## Scope note

`typescript/site` has no existing test infrastructure (no `*.test.ts`, no configured test runner in its `package.json`), so this PR doesn't add a new test harness for a route-level fix — happy to move the fix (or add a regression test) into `packages/mechanisms/evm/` instead if maintainers prefer, since `packages/mechanisms/evm/test/unit/exact/facilitator.test.ts` already exists there and would be a more natural home for a schema-shape regression test. Also happy to extend this same validation pattern to guard the unguarded `payload.accepted.scheme` dereferences in `permit2.ts`, `upto/permit2.ts`, and `batch-settlement/scheme.ts` if that's in scope for this PR rather than a follow-up.

## Test plan
- [x] `pnpm --filter=site exec tsc --noEmit` — no errors in either changed file (pre-existing, unrelated errors remain in other files in this package — `@x402/stellar`/`@x402/svm`/`@x402/paywall`/`@x402/next` module resolution and unrelated type mismatches in `facilitator/index.ts`/`proxy.ts`, none touched by this change)
- [x] `pnpm --filter=site run lint:check` — clean
- [x] Manually verified both bug repro cases now return 400 with a specific message, against the live production endpoint
- [x] Manually verified a correctly-shaped request still verifies and settles successfully, with a real on-chain transaction as proof

---

**AI disclosure:** drafted with an AI coding assistant. Reviewed by me against the real production endpoint and a live on-chain Base Sepolia transaction (not just unit assertions) before submitting, since this touches facilitator verify/settle.

### DIFF
--- a/typescript/site/app/facilitator/settle/route.ts
+++ b/typescript/site/app/facilitator/settle/route.ts
@@ -1,4 +1,5 @@
 import { PaymentPayload, PaymentRequirements, SettleResponse } from "@x402/core/types";
+import { parsePaymentPayload, parsePaymentRequirements } from "@x402/core/schemas";
 import { getFacilitator } from "../index";
 
 /**
@@ -9,13 +10,13 @@ import { getFacilitator } from "../index";
  */
 export async function POST(req: Request) {
   // Parse request body - only use "unknown:unknown" if parsing fails
-  let paymentPayload: PaymentPayload | undefined;
-  let paymentRequirements: PaymentRequirements | undefined;
+  let rawPaymentPayload: unknown;
+  let rawPaymentRequirements: unknown;
 
   try {
     const body = await req.json();
-    paymentPayload = body.paymentPayload as PaymentPayload;
-    paymentRequirements = body.paymentRequirements as PaymentRequirements;
+    rawPaymentPayload = body.paymentPayload;
+    rawPaymentRequirements = body.paymentRequirements;
   } catch (error) {
     const errorMessage = error instanceof Error ? error.message : String(error);
     console.error("Failed to parse request body:", errorMessage);
@@ -33,7 +34,7 @@ export async function POST(req: Request) {
   }
 
   // Check for missing parameters
-  if (!paymentPayload || !paymentRequirements) {
+  if (!rawPaymentPayload || !rawPaymentRequirements) {
     return Response.json(
       {
         success: false,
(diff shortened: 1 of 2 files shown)


wakqasahmed:
Commits are now signed and verified (head commit a22178d5, `check-verified-commits` passing in CI) - the unsigned-commit note above no longer applies. This PR is otherwise green (Vercel preview just needs a Coinbase-team member to authorize the deploy, which is outside contributor control). Ready for review whenever a maintainer has time.

PR #3017, opened 2026-08-01
feat(xrpl): add exact scheme Python implementation

dimitricole:
## What this adds

`python/x402/mechanisms/xrpl/`: client, resource server and facilitator for the `exact`
scheme on the XRP Ledger, following `specs/schemes/exact/scheme_exact_xrpl.md`. Python
has mechanisms for evm, svm and tvm and none for XRPL; the TypeScript SDK already has
one (#2801), and this brings Python to parity with it.

The facilitator holds no keys: the payer signs a `Payment` that carries its own fee, so
`get_signers` is empty and `areFeesSponsored` is always false. XRP and issued currencies
are supported on mainnet, testnet, devnet and custom networks, with both `sequence` and
`ticketSequence` transfer methods, invoice binding and destination tags.

## Also in this PR

- `pyproject.toml`: an `xrpl = ["xrpl-py>=4.2.0"]` extra, `xrpl` in the `mechanisms` and
  `all` bundles, and `xrpl-py` in the dev group. The floor is 4.2.0 because 4.0.0 has no
  `Simulate` request and 4.1.0's binary codec does not know the `Delegate` field.
- `tests/unit/mechanisms/__init__.py`: this marker file is missing today, and without it
  a test package named `xrpl` shadows the installed `xrpl-py`. Every other directory in
  the chain has one; XRPL is the first mechanism whose SDK module name matches its
  directory name.
- XRPL sections in the three `all_networks` examples, per the New Chain Family
  checklist. The server example prices XRPL as explicit drops, since XRPL has no
  on-ledger rate to convert a money price.

## Notes for reviewers

- Reason codes shared with the TypeScript mechanism use its exact spelling, so a client
  switching facilitators gets the same string for the same fault. A few deliberate
  differences on invalid input are commented at the site of each check.
- One finding may affect both implementations: the signature covers only the fields the
  binary format marks as signing, so appending another non-signing field (`Signature`,
  `MasterSignature`, and so on) to a validly signed blob changes its transaction hash
  without the payer's key, and the duplicate-settlement guard is keyed on that hash.
  This implementation rejects any non-signing field beyond `TxnSignature` and `Signers`.
  Since the first review it also holds a Payment to rippled's own field template, so a
  signing field a future amendment adds is refused until admitted deliberately.
  The TypeScript facilitator checks only `Signers` and may share the gap; happy to file
  that separately if useful.
- xrpl-py's sync client is an `asyncio.run` wrapper, so ledger reads move themselves to
  a worker thread when called on an event-loop thread (an async facilitator invoking
  this sync scheme). No sibling SDK behaves this way, hence no sibling needs this.
- Two open TypeScript PRs (#2873, #2874) would extend the reference; neither is in the
  merged spec yet, so neither is in this port. Parity follow-ups are expected work.

## Testing

392 unit tests, plus 6 opt-in live tests against the XRPL testnet
(`X402_XRPL_LIVE=1 pytest tests/integrations/test_xrpl_live.py`; faucet-funded, no
credentials needed). Coverage is 100%; ruff, black and mypy --strict are clean. Running
the whole `python/x402/tests` tree with and without this change leaves the failing and
erroring sets identical (modules whose optional SDKs are absent); the only delta is this
suite. Cross-implementation fixtures such as the transaction hash and invoice id are
computed by xrpl.js and asserted verbatim, so the two implementations cannot drift
silently.

## AI assistance

Per CONTRIBUTING.md: this was developed with significant AI assistance. I have reviewed
the code and tests, and the claims above were verified against a clean clone before
submission.


whawk46 (commented):
Ran the suite (381 passed / 1 skipped, py3.12 + xrpl-py 5.0.0) and read the mechanism end to end against the spec's facilitator MUST-list. Verdict from another XRPL implementer: **mergeable with two fixes** — spec conformance is essentially complete, the money path is correct (`Decimal` IOU arithmetic with a grammar that rejects `1e5`/`Infinity`, strict int drops, offline sig verification, hash computed from the blob and cross-checked against the node), and several defenses go beyond the spec. Naming those first because they deserve it:

- **Canonical re-encoding + non-signing-field rejection** (`utils.py:99-142`) closes hash-malleability against the dedup guard — an attack the spec never asks you to consider, with the docstring explaining it precisely.
- **Fully-canonical signature enforcement** (`utils.py:157-207`): minimal DER + low-S, and S < L for Ed25519, matching rippled's `RequireFullyCanonicalSig`. Without this, a malleated signature verifies locally but never settles.
- The concurrency suite is real (`sys.setswitchinterval(1e-6)` + barrier release to actually widen the race window), and tests assert wire strings rather than constants, so renames can't silently change the protocol.

**Should-fix 1 — unknown-field acceptance is delegated to xrpl-py's field classification** (`utils.py:139-141`). The check rejects unknown *non-signing* fields, but anything the codec classifies `is_signing=True` passes decode, and only `Delegate` gets an explicit facilitator check. Today that's safe — we verified empirically that XLS-68 `Sponsor`/`SponsorFlags`/`SponsorSignature` are unknown to xrpl-py 5.0.0's codec, so sponsored transactions hard-fail cleanly at decode (`invalid_exact_xrpl_payload_blob`), which is the correct default. But the day xrpl-py ships XLS-68 definitions (the amendment is live on Devnet and in rippled 3.3.0 RCs), a routine `pip install -U xrpl-py` silently converts "sponsored rejected" into "sponsored accepted" — with the sponsor paying and `Fee` checks passing trivially — with no code change on your side. `Delegate` being classified `is_signing=True` is the precedent. One-line hardening now: an explicit allowlist of accepted Payment fields, or a `Sponsor is None` structure check next to the existing `Delegate` one. (Context: there's a pending fee-sponsoring extension proposal, #3048 — if it progresses, this decode path plus `get_extra`'s hard-coded `areFeesSponsored: False` and the envelope check are the three coordinated touch points; happy to help with that when relevant.)

**Should-fix 2 — dedup guard injection point.** The spec requires dedup "across every process that serves /settle"; the in-process default is honestly documented and matches the svm/tvm mechanisms, but the constructor takes the concrete `SettlementCache` class rather than a Protocol, so backing it with Redis/Postgres means subclassing. A `Protocol` with `is_duplicate(key, ttl) -> bool` plus a short shared-store recipe in the README would make the spec-required deployment shape a first-class path.

**Two questions:** (1) Dedup TTL is wall-clock (`max_timeout + 120s`) while the spec words retention as "until `LastLedgerSequence` has passed" — at degraded ledger cadence those diverge; was keying to observed ledger index considered? Re-verification narrows the residual risk, so question not defect. (2) Envelope check compares `accepted.amount` to `requirements.amount` as exact strings while the tx-level check correctly uses `Decimal` — is a verbatim echo guaranteed by all clients cross-implementation, or should the envelope comparison be decimal-aware for IOU?

Nits, take or leave: `parse_money_to_decimal` returns `float` (name + money-through-binary-float; it never reaches verification, but `Decimal` costs nothing); `_loop_safe` builds a fresh `ThreadPoolExecutor` per ledger read; `enhance_payment_requirements` mutates its input while also returning it; `settle()`'s catch-all maps internal config errors to `transaction_failed`, which reads as a ledger rejection to the resource server.

---
*Disclosure per CONTRIBUTING: review performed with significant AI assistance under significant human monitoring, direction, and correction. The suite was actually run and the XLS-68 decode behavior verified empirically against xrpl-py 5.0.0, not assumed.*

dimitricole:
Thanks for the review. Both should-fixes are taken in `b53010f0`.

**Should-fix 1.** Taken as an allowlist rather than a `Sponsor is None` check, so it
covers any signing field a future amendment adds, not only the ones we can name
today. `decode_signed_transaction_blob` now holds a Payment to rippled's own field template (`ALLOWED_PAYMENT_FIELDS` in
`utils.py`), on top of the existing non-signing rejection. A sponsored blob fails
decode today as codec-unknown with `invalid_exact_xrpl_payload_blob`; once xrpl-py
ships the XLS-68 definitions it fails the allowlist with the same code, so
`pip install -U xrpl-py` no longer changes observable behaviour. (Checked against
xrpl-py main's `definitions.json` today: `Sponsor` and `SponsorFlags` are signing
fields, `SponsorSignature` non-signing. Since both extension modes present only
`Sponsor` and `SponsorFlags` at `/verify`, that is precisely the shape the old logic
would have admitted.) The test vectors are payer-signed blobs carrying
`OfferSequence`, `ClearFlag` and `TicketCount`: signing fields the codec already
encodes onto a Payment, exercising the same acceptance path a future signing field
would take. Fields with dedicated reason codes (`Memos`, `Paths`, `DeliverMin`,
`Delegate`, `Signers`) stay in the list so those codes remain reachable, and a test
walks the list against the codec so a misspelt entry fails the suite by name.

**Should-fix 2.** Taken. `SettlementCacheLike` is a `typing.Protocol` with the one
method the facilitator calls, `is_duplicate(key, ttl_seconds) -> bool`, with atomic
test-and-record as the stated obligation. The constructor accepts anything satisfying
it and defaults to the in-process cache; the README shows a Redis `SET NX EX`
implementation as the shared-store shape. Implementing it surfaced an adjacent bug:
the old `settlement_cache or SettlementCache()` would silently discard a shared store
whose `__len__` reports empty. The None test is now explicit, with a test.

**Question 1.** Considered; duration-based retention stays. The monotonic clock cannot step, while a
wall-clock jump would evict entries whose transactions can still land. The TypeScript
mechanism uses the same `maxTimeoutSeconds + 120s` retention. And the in-process
cache is cross-network, so ledger indexes from different networks are not comparable; keying on
them turns a lock-guarded dict into a networked check. On the residual: once a payment
validates, replays fail the sequencing checks with or without the cache (`Sequence`
consumed, ticket spent), so the guard's job is the in-flight window. For a 300s
timeout the expiry bound is 62 ledgers (`ceil(300/5) + 2`) against 420s of retention,
so divergence needs sustained average closes above ~6.8s for the whole window, with
the duplicate arriving after expiry. A deployment that wants exact retention can key
entries to `LastLedgerSequence` inside its shared `SettlementCacheLike` store.

**Question 2.** Deliberate. The spec words the envelope as field equality on
`accepted` against the requirements, and the TypeScript facilitator implements it as
`!==`; a decimal-aware comparison on one side would make the two implementations
disagree about the same payload. Decimal comparison lives at the transaction level
because that is where the ledger's canonicalisation ("1.50" to "1.5") bites; the
envelope compares the server's offered string against the client's echo of it.

**Nits.** We tried the executor change and backed it out. A shared module-level pool
inherited across `fork()` holds dead worker threads and a stale idle count, so the
first offloaded call in a pre-fork worker (gunicorn `--preload`, celery prefork)
queues forever with no exception. The per-call executor is fork-safe and costs
microseconds against a ledger RPC; its docstring now records why it stays. The other
three we held for parity: `parse_money_to_decimal` mirrors the EVM mechanism's parser
(its `MoneyParser` also takes `float`); `enhance_payment_requirements` mutates in
place as the EVM scheme does; and the settle catch-all keeps the `transaction_failed`
code the TypeScript catch-all also uses (TypeScript interpolates the detail into the
reason string, this implementation carries it in `error_message`). Agreed that a
shared `Decimal` parser and a distinct `unexpected_settle_error` would be
improvements, as coordinated changes across mechanisms rather than here.

Following up on #3048 about coordination.

---

*Disclosure per CONTRIBUTING: this response, like the PR, was developed with
significant AI assistance under human direction and review; the suite, type checks and
mutation tests were run against the updated branch before pushing.*

whawk46:
@dimitricole Checked `088ec644` — both are right, and the allowlist is a better answer
than the one I asked for.

Holding the Payment to rippled's own field template is the part that generalizes: it means
the rejection is derived from the network's definitions rather than from a list a
reviewer maintained by hand, so an amendment that adds a field can't silently widen what
you accept. I've said as much on #3048 and I'm rewriting my §9 sibling clause to point at
the allowlist as the normative mechanism, with the XLS-68 field names demoted to
informative examples. My denylist would have failed open on exactly the case XLS-68
already produced once — `FeeAmount` became `FeeAmountDelta` between the spec text and
Devnet rc5, which I only found by probing the template.

Same error code before and after xrpl-py ships the definitions is the right call too. A
payload that is unacceptable for two different reasons at two different times shouldn't
look like two different bugs to whoever is reading the logs.

Nothing further from me on this one.

---

*Disclosure: written with significant AI assistance, under human direction, review, and
correction.*

dimitricole:
Glad the commit checked out, and I appreciate the review; it made the mechanism
better. See you on #3048.

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human
direction and review.*

PR #3018, opened 2026-08-01
fix(stellar): let a contract account sign its own auth entries

tolgayayci:
## Description

A contract account (`C…`) cannot produce a payment with `exact` on Stellar. `createPaymentPayload` throws `invalid version byte. expected 48, got 16` (48 is the strkey byte for `G`, 16 for `C`). The same call from a classic account succeeds.

`signAuthEntries` delegates to the SDK's default `authorizeEntry`, which resolves every signer through `Keypair.fromPublicKey(publicKey)`. That rejects a contract address, so it throws before a payload is built. `signAuthEntries` already accepts an `authorizeEntry` override, but the Stellar client never passed one, so the keypair path was taken regardless of account type.

This PR adds an optional `authorizeEntry` to `ClientStellarSigner` and forwards it only when supplied. Keypair signers are unaffected. No wire format, payload, or facilitator changes.

`ClientStellarSigner` already documents support for both `G` and `C` accounts. This makes that true.

Contract accounts cannot sign transaction envelopes at all, so auth entries are their only route to paying.

## Tests

Two unit tests added to `test/unit/client.test.ts`: one asserts the override is forwarded, one asserts the key is absent when no override is supplied so the SDK default still applies. Reverting the source change fails the first; the other 158 pass.

Verified end to end on testnet. A `__check_auth` account with an ed25519 session key and an on-chain spending policy paid through the client and settled:

- Transaction: [`f1330843315ee93e…`](https://stellar.expert/explorer/testnet/tx/f1330843315ee93e8eb760511d046d573e8f3c6fad0aed3102c4c03f40f3c4c6) (ledger 3915905) · [raw on Horizon](https://horizon-testnet.stellar.org/transactions/f1330843315ee93e8eb760511d046d573e8f3c6fad0aed3102c4c03f40f3c4c6)

- Payer: [`CD752A22JHVS7ZWD…`](https://stellar.expert/explorer/testnet/contract/CD752A22JHVS7ZWDRBFPLNMNTLVCWZAO4EZUZN7FHYA76V362MON7MHZ), a contract account. It is the transfer's `from`.

Without the change, the same code against the same account never reaches the override.

`pnpm test` from `/typescript`: 159 passing. Lint and format clean.

## Checklist

### DIFF
--- a/typescript/.changeset/stellar-contract-account-signing.md
+++ b/typescript/.changeset/stellar-contract-account-signing.md
@@ -1,0 +1,5 @@
+---
+"@x402/stellar": patch
+---
+
+`ClientStellarSigner` accepts an optional `authorizeEntry`, forwarded to `signAuthEntries`. Contract (C) accounts can now sign their own auth entries; the SDK's default resolves signers through `Keypair.fromPublicKey`, which rejects a contract address. Keypair signers are unaffected.
--- a/typescript/packages/mechanisms/stellar/src/exact/client/scheme.ts
+++ b/typescript/packages/mechanisms/stellar/src/exact/client/scheme.ts
@@ -89,6 +89,8 @@ export class ExactStellarScheme implements SchemeNetworkClient {
       address: sourcePublicKey,
       signAuthEntry: this.signer.signAuthEntry,
       expiration: maxLedger,
+      // Only when supplied, so keypair signers keep the SDK default.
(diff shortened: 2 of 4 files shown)


tolgayayci:
@phdargen would appreciate a review when you get a chance.

@marcelosalloum thanks for writing `scheme_exact_stellar.md`, it helped a lot while I was figuring out the auth entry side. Flagging you in case the signing path here is worth a second pair of eyes.

PR #3021, opened 2026-08-02
feat(starknet): add exact scheme TypeScript reference implementation

adipundir:
## Description

This PR adds `@x402/starknet`, the TypeScript reference implementation of the Starknet `exact` scheme, which was specified in [`scheme_exact_starknet.md`](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_starknet.md) and merged in #2849. It is PR 2 of the New Chains process.

**Please merge #3126 before this one.** The implementation follows the spec as amended there, so merging this PR first would put code on main that disagrees with the spec.

The client signs a SNIP-9 Outside Execution that authorizes exactly one SNIP-2 `transfer` and binds it to the facilitator through `extra.feePayer`. The facilitator verifies it against the spec rules and then settles it with `execute_from_outside_v2`, paying the whole fee, so the payer never needs a gas balance.

The PR contains the package itself, with the client, resource server and facilitator schemes, a direct executor signer, a SNIP-29 paymaster signer, unit and integration tests and a README. It also contains the e2e wiring, the three `all_networks` examples, the publish workflow, the release script entries and a changeset.

When a broadcast cannot be confirmed in time, settlement returns `settlement_pending` with the transaction hash and remembers that hash in core's `PendingSettlementStore`, so that the resource server's single retry waits on the same transaction instead of broadcasting a second one. A consumed nonce on any other request is terminal. The store is keyed by the SNIP-12 message hash rather than by the signature, because a Starknet signature is an account-defined felt array. The receipt wait is bounded by `confirmationTimeoutMs`, which defaults to 75 s as in `@x402/cardano`, so that the response still arrives inside core's 90 s facilitator timeout.

`starknet` v8 is a peer dependency, because it is the only release line that supports both RPC 0.8.1 and 0.9.0.

A second implementation (Vauban Pay) has run against this code in both directions and agrees on all eleven cases of its refusal matrix, as described in the comments below.

## Tests

- Unit: 334 tests pass with the coverage thresholds (`pnpm --filter @x402/starknet test`).
- Integration: 7 pass and 1 is skipped, because that one needs funded Sepolia accounts.
- E2E: `pnpm test --testnet --min --families=starknet --versions=2 --transport=http` passes 5/5 on Sepolia at this head, on fastify ([`0x1eff…0877`](https://sepolia.voyager.online/tx/0x1eff627d5813c61498c02c2605ca9929e23716f4bcc69b563a6256aff0e0877) and [`0x4e5e…1bd9`](https://sepolia.voyager.online/tx/0x4e5e75692b19fff1b44630b3abbff2bf51d59434d62d7f4df82c4a0273a1bd9)), next ([`0x792d…a2ae`](https://sepolia.voyager.online/tx/0x792d388c382a86a307ffe7c1cd864adcd5567c48d5a4a37221d84dccbd9a2ae)), hono ([`0x7243…6813`](https://sepolia.voyager.online/tx/0x724323274129b6c499eacefa44ab1279c6fa6a21d899ef5b0944c4a6e8f6813)) and express ([`0x51a4…587e`](https://sepolia.voyager.online/tx/0x51a46226ff5328996e1cb4dd4b21325bf8746765a97889de94b34e3a7e9587e)).
- The SNIP-29 paymaster signer was exercised live against RPC 0.8.1 and 0.9.0 at the previous head, and it has not changed since.

## Checklist

## AI assistance

Following the AI-Assisted Contributions policy: this was developed with AI assistance against the spec and reviewed by me, and the flow was confirmed live on Sepolia with the transactions listed above.


phdargen:
Hi @adipundir, thanks for the contribution! We made a couple of high-priority sdk changes in the last 2 weeks that unblock several work streams. Could you please rebase to upstream/main and make sure to account for the following changes:

- payment flows: https://github.com/x402-foundation/x402/pull/3053
- spend controls and unified default asset declaration + money parsing: https://github.com/x402-foundation/x402/pull/3124
- e2e refactor: https://github.com/x402-foundation/x402/pull/2976

Please let me know if you have questions about any of the changes

adipundir:
Hi @phdargen, I've rebased the branch onto main and accounted for those changes (plus #3083, which landed after your comment). The spec follow-up #3126 should go in first, since the implementation follows it. Let me know if anything else needs changing.

seritalien:
Hi, a note from a second implementer, in case it is useful while #3021 waits for review.

We have both sides of `scheme_exact_starknet.md` running against each other and against Sepolia: a facilitator (Rust; canonical SNIP-9 v2 reconstruction per rule 2, the eight verification rules, settlement with the receipt effect check) and a client mechanism for `@x402/core`'s `x402Client` (`@vauban-pay/x402-starknet` on npm; any `starknet.js` Account signs). Driven by `@x402/fetch`, it settled 0.01 STRK on Starknet Sepolia today: tx `0x1b88b79ffaa84344d36fdab48e322a86d687a3eb4e9e2f45edae4c01b5b548d`, block 14708171. The test reads the transaction back from the chain before it believes the receipt.

Two things we hit that may be worth knowing:

1. We applied the payer-scoped `Transfer` criterion from #3126 before it merged. Implementing the merged rule 8 verbatim declared an honest STRK payment failed, exactly as #3126 describes, since our default asset is STRK.

2. `starknet-core` (Rust) hashes a `selector`-typed field through `get_selector_from_name` unconditionally, so a `Selector` written in hex, as the spec's own example does, hashes to the wrong value in that library while `starknet.js` passes it through. We reconstruct and hash from constants instead, and use the library only as an independent oracle with the selector spelled as a name. Mentioning it because a facilitator written in Rust against the example would refuse every wallet-signed payment with no useful error.

If a cross-implementation test would help #3021, our facilitator can be started in mock-chain mode (hermetic) or live against Sepolia and pointed at from your e2e client; happy to share the exact commands, and to run your client against it from our side.

Vauban Pay (pay.vauban.tech)

seritalien:
Follow-up, since the merchant side is the half that lets #3021's client be tested against something that is not ours.

The facilitator now serves the standard facilitator API at one URL, `https://demo.pay.vauban.tech/starknet` (`/supported`, `/verify`, `/settle`; `starknet:SN_SEPOLIA`; it announces its `feePayer` and pays the gas). A merchant built with `@x402/express` and `HTTPFacilitatorClient` pointed at it, with a server-side scheme that copies `feePayer` from `/supported` into the requirements (`@vauban-pay/x402-starknet/server`), settled this morning on Sepolia with its own `payTo` and price, paid by `@x402/fetch` with the client mechanism: tx `0x78f6773ecdff976e85de26742c7fe1f80f7ea50a1c8b492a245a56cac874912`, block 14732675 ; the merchant's balance moved by exactly the price, since the merchant pays no gas under this scheme.

So a client implementing this PR can be pointed at any route of such a merchant, or at `https://demo.pay.vauban.tech/starknet/v1/quote` (0.01 STRK, testnet), and settled by a facilitator written independently from the same text. If the eight rules disagree between the two implementations, the disagreement is the finding. The exact 402 and the exact `/supported` body are on https://demo.pay.vauban.tech/standard.

Vauban Pay (pay.vauban.tech)

seritalien:
Cross-implementation results against this PR's code, both directions, sources at `d495522` used verbatim (nothing patched on either side):

**1. This PR's client → our facilitator** (`typescript/packages/mechanisms/starknet` client scheme on `x402Client` + `@x402/fetch`, against `zkpay-facilitator` in mock-chain mode). With default spend controls the client refuses before any request: `All payment requirements were rejected by spendControls: only default assets or entries in spendControls.allowedAssets are allowed`, because `defaultAssets.ts` knows USDC only and our offer is STRK. With `client.setSpendControls({ allowedAssets: true })`: **200**, receipt `{"success":true,…,"amount":"10000000000000000","extra":{"mock":true}}`. The payload as this PR builds it (all felts hex, `Execute After` `0x1`, selector without leading zeros) passes our eight rules unchanged.

**2. Our client → this PR's facilitator, live on Sepolia** (facilitator scheme with `toFacilitatorStarknetSigner(<our fee-payer Account>)` and an `RpcProvider`; requirements carry its `getExtra()`; our `@vauban-pay/x402-starknet` client signs). `verify`: `{"isValid":true,"payer":"0x631ba1b3…c494c"}`. `settle`: `{"success":true,"transaction":"0x23da5044ac67446404542115508298cf95250b90752367b2abf4dcd70d8c352","network":"starknet:SN_SEPOLIA",…,"amount":"5000000000000000"}`. On chain: block 14734185, `SUCCEEDED`, `Transfer(payer → merchant, 5000000000000000)`, gas paid by the fee payer through this PR's `account.execute` path.

So the two implementations read, verify and settle each other's payments in both directions. The one thing a user hits is the default-asset table: a client of this PR does not pay a STRK offer without touching spend controls, and ours defaults to STRK because that is what Sepolia faucets hand out. We are adding USDC (same addresses as this PR's table) to our defaults so either client pays either facilitator's offer as shipped; if the PR wanted to add STRK to its Sepolia defaults, a fresh account could then pay a testnet offer with nothing but faucet funds.

Not covered yet: agreement on refusals (the eight rules exercised on invalid payloads, both directions). If the e2e fixtures of this PR can be pointed at an external facilitator URL, that is the run we would do next.

Vauban Pay (pay.vauban.tech)

seritalien:
The refusal side, since the previous comment left it open. Eleven documents built by our client (real Sepolia accounts, live reads, nothing signed by either facilitator), each broken in exactly one way, submitted to this PR's facilitator `verify()` and to ours :

| case | this PR | ours |
|---|---|---|
| valid | VALID | VALID |
| requirements amount above the signed calldata | `…_amount_mismatch` | same |
| requirements payTo differs from signed To | `…_recipient_mismatch` | same |
| requirements asset differs from signed contract | `…_asset_mismatch` | same |
| signed Caller is not the announced feePayer | `…_payload_caller` | same |
| Execute Before already passed | `outside_execution_expired` | same |
| window beyond now + maxTimeoutSeconds | `…_window_exceeds_max_timeout` | same |
| one signature felt altered | `…_payload_signature` | same |
| payer address not deployed | `account_not_deployed` | same |
| unknown key in `message` | `invalid_payload` | same |
| second call appended after signing | `invalid_payload` | was `…_payload_signature`, now `invalid_payload` |

Ten of eleven agreed token for token on the first run. The eleventh was ours to fix : a two-call document whose signature covers the one-call document is wrong twice, and we checked the signature before the shape ; this PR names the shape, which is the fault a client can act on. Our verifier now checks "exactly one call" in the structural pass, before the signature, so it answers the same token.

So, as of today : both directions settle, and eleven refusals agree. The script is `refusal-matrix.ts` in our repository's `docs/testing/interop-pr3021/` if the e2e suite here wants the same cases as fixtures.

Vauban Pay (pay.vauban.tech)

adipundir:
Thank you, it is genuinely useful to see this cross-checked against a second implementation.

I have added a note about the selector hashing to #3126. I am not going to add STRK to the default assets, because that table has to be USD-pegged, so setting `allowedAssets` as you did is the right way to pay a STRK offer.

One thing to be aware of is that settlement has changed since your run: a consumed nonce is now terminal, rather than being resolved to the transaction that consumed it. The reasoning is in #3126. Thank you for picking up the documentation side in #3415 as well.

seritalien:
@adipundir Glad the cross-check was useful. Since Vauban Pay is the second implementation your PR now cites, the matching docs entry (network id, facilitator URL, SDK) is #3415, open since the 8th with no review yet. A look from you or a maintainer would close the loop on both sides.

PhilBot402:
### Spec review — #3021

Reviewed `@x402/starknet` exact facilitator paths against `scheme_exact_starknet.md` on upstream `main` (post-#3126).

**Blocking**
- Spec SHOULDs are not reflected in the reference SDK: sponsored fee caps, per-payer rate limits, and restricting which resource servers may settle. Document them as mandatory operator configuration (with concrete knobs) or enforce them in facilitator config/verify paths so defaults are not silently permissive.

**Optional**
- Horizontal scale relies on in-process `SettlementCache`; spec allows replica races (gas-only). Shared duplicate guard and pending store remain deployment concerns.
- RPC/simulation infrastructure faults surface as `unexpected_verify_error` / `unexpected_settle_error` rather than `simulation_failed`. Fail-closed is correct; operators should treat these as retryable infra, not payer faults.
- Default `maxSignatureFelts` (32) may reject long account signatures (passkey/webauthn) unless raised.

**Nits**
- PR body still says to merge #3126 first; that amendment is already on `main`.
- `getExtra` advertises only the first configured `feePayer` when multiple signers exist.

PhilBot402:
### Cross-SDK review — #3021

Compared `@x402/starknet` exact to `@x402/evm` and `@x402/svm` exact (Cardano for pending-timeout patterns) at head `663f0990`.

**Optional**
- **Pending reconcile ordering:** `reconcilePendingSettlement` removes the `PendingSettlementStore` entry before `confirmExecution` finishes. SVM keeps the entry until confirmation succeeds or re-pends. After delete, another replica can miss the store and re-walk settle while the first tx is still confirming (on-chain nonce should prevent double payout, but extra gas and confusing terminal errors are likely). Consider confirm first, delete on success, re-`set` on another indeterminate timeout.
- **`getExtra` vs multiple `feePayer` signers:** Only `signer.getAddresses()[0]` is advertised while `getSigners` can list many (SVM randomizes). If multiple callers are intentional, document one-kind-per-feePayer or add explicit selection; otherwise operators may assume SVM-style rotation.
- **Paymaster submit timeout:** `TimeoutError` from `executeFromOutside` cannot emit `settlement_pending` (no hash per v2 §9) and holds the duplicate guard until execute-before. Worth a short operator note beside EVM/SVM ambiguous-broadcast guidance.

**Nits**
- Facilitator `scheme.ts` is a single large module vs EVM exact’s thin router; harder to review despite matching Starknet surface area.
- Verify always runs mandatory post-signature simulation (good for security vs EVM’s optional `simulateInSettle`); expect higher RPC cost on open `/verify`.

PhilBot402:
### Rebase review — #3021

Alignment check against current `x402-foundation/x402` `main` after a clean rebase (head ~`663f0990`).

**Blocking**
- Branch is merge-dirty against current `main` (overlapping regions in a merge-tree check). Rebase and re-run CI before relying on green checks.
- **CJS declaration mirroring:** Mechanism packages on `main` use `tsup && node ../../../scripts/mirror-cjs-dts.mjs` with CJS `dts: false` (#3522). `@x402/starknet` still uses `build: "tsup"` only with CJS `dts` enabled, so it will fail workspace `pnpm verify:exports` (#3550) after rebase.
- **Publish workflow:** `.github/workflows/publish_npm_scoped_x402_starknet.yml` builds and publishes but omits the `pnpm verify:exports` step other scoped mechanism workflows gained in #3550. Add it after the filtered build (match e.g. `publish_npm_scoped_x402_casper.yml`).
- **Upfront `paymentFlow`:** `ExactStarknetScheme` still advertises authorization-only `paymentFlows`, with no server unit test for upfront and no `/exact/starknet/upfront` in `e2e/config/mechanisms_starknet.json`. Post-#3240 peers (Stellar, NEAR, Aptos, Hedera) declare `supported: ["authorization", "upfront"]` with `default: "authorization"` and e2e upfront routes. Match that pattern unless Starknet documents an explicit authorization-only exception like Casper/Cardano on current `main`.

**Optional**
- If authorization-only is intentional for SNIP-9 outside execution, state that explicitly in the changeset/README so it is not read as accidental drift from #3240.

PhilBot402:
### Code quality review — #3021

Reviewed facilitator verify/settle, signer, simulation, typed-data parsing, tests, and release wiring at head `663f0990`.

**Optional**
- `src/exact/facilitator/scheme.ts` is ~1,680 lines with `max-lines` capped at 2,000. Security ordering reads sound, but verify/settle orchestration is still monolithic; siblings split earlier (simulate is already extracted).
- `settlement-cache.ts` repeats the same “optimization not authority / SNIP-9 nonce is authoritative” rationale in the module header and on `SettlementCache`.
- Facilitator `verify`/`settle` log unexpected failures with `console.error`; several sibling facilitators stay silent on the wire path unless documented as intentional.
- `signer.ts` is ~85% line coverage (timeout fetch and `createStarknetProvider` defaults); direct provider-timeout regression tests are thin.
- Default `pnpm test` skips `test/integrations/`; live Sepolia coverage depends on org secrets and author-reported e2e runs.
- `ExactStarknetPayload.outsideExecution.typedData: unknown` matches untrusted input; a dedicated unparsed typed-data alias would document intent without relaxing checks.

**Nits**
- `lint:check` emits `MODULE_TYPELESS_PACKAGE_JSON` for `eslint.config.js`; add `"type": "module"` to `package.json` or align with sibling mechanism packages.
- Changeset body is one very long paragraph; trim to user-visible behavior.

PR #3022, opened 2026-08-02
fix(go): propagate payment response hook errors

### LINKED ISSUE #2913: go/http: corrective retry silently discards payment response hook errors
## Summary

The Go HTTP payment client silently discards errors returned by payment response hooks after the final corrective retry.

In `go/http/client.go`, the final hook dispatch currently does this:

```go
if _, err := t.dispatchPaymentResponseHooks(
	ctx,
	&correctiveBuild,
	correctiveResp,
); err != nil {
	return correctiveResp, nil
}
```

This returns the HTTP response with a `nil` error even when reconciliation or a user-registered payment response hook failed.

The behavior also appears inconsistent with the nearby `dispatchPaymentResponseHooks` documentation, which says:

> Hook errors propagate so the caller can surface them.

Source: [`go/http/client.go`](https://github.com/x402-foundation/x402/blob/main/go/http/client.go)

## Current behavior

The V2 recovery flow is:

1. Send the initial request.
2. Receive a 402.
3. Build and send a payment payload.
4. A payment response hook reports `Recovered=true`.
5. Build a fresh payload and perform one bounded corrective retry.
6. Dispatch payment response hooks for the corrective response.

If the final hook dispatch in step 6 fails, the error is discarded:

```go
return correctiveResp, nil
```

The caller therefore cannot distinguish between:

- a response whose settlement state was reconciled successfully; and
- a response that was received successfully but whose local reconciliation failed.

This matters in particular for batch-settlement, where `OnPaymentResponse` updates the local channel session from the settlement response. A storage or reconciliation failure may leave the local session stale even though the caller observes no error.

## Related cleanup concern

The earlier hook dispatch path behaves differently:

```go
recovered, err := t.dispatchPaymentResponseHooks(ctx, build, newResp)
if err != nil {
	return nil, err
}
```

Here the error is propagated, but `newResp.Body` does not appear to be closed before returning it as `nil`. This may prevent connection reuse or leak response resources when a hook fails after an HTTP response has already been received.

It may be worth covering both paths when defining the intended error contract.

## Open question

What is the intended Go API behavior when an HTTP response was received successfully but a payment response hook or local reconciliation step failed?

Possible options include:

1. Return the response together with the hook error.
2. Close the response body and return only the hook error.
3. Return the response and expose reconciliation failure through a typed payment result.
4. Preserve the response and treat the failure as a non-fatal diagnostic.

Simply returning both `correctiveResp` and `err` may not be sufficient because `net/http.Client` does not generally expect a `RoundTripper` to return both a usable response and an error.

## Relationship to #2908

#2908 proposes typed corrective-402 recovery outcomes across SDKs. This issue is narrower: the Go transport currently drops an actual error returned by the existing `onPaymentResponse` path.

Should this be fixed independently as an error-propagation/resource-cleanup bug, or should it be handled as part of the typed recovery outcome work in #2908?

## Suggested acceptance criteria

- Errors from the final corrective-response hook are not silently discarded.
- The behavior for an already-received HTTP response is explicitly defined.
- Response bodies are closed or returned to the caller consistently on hook failures.
- A regression test covers a hook that fails:
  - after the first paid response; and
  - after the bounded corrective retry.
- Existing one-retry recovery behavior remains unchanged.

256dino:
## Description

Closes #2913.

Propagates payment-response hook errors from both the initial paid response and the corrective retry. On either failure, the response body is closed and `RoundTrip` returns `(nil, err)` instead of exposing a response whose reconciliation failed.

The existing recovery API and one-retry behavior remain unchanged.

## Tests

- Added regression coverage for hook failures on the first paid response and corrective response.
- Verified the affected response body is closed exactly once and the original hook error is returned.
- `cd go && make verify`

## Checklist

### DIFF
--- a/go/.changes/unreleased/fixed-20260731-220657.yaml
+++ b/go/.changes/unreleased/fixed-20260731-220657.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Propagate payment response hook errors after corrective retries and close response bodies on hook failure.
+time: 2026-07-31T22:06:57.340821-04:00
--- a/go/http/client.go
+++ b/go/http/client.go
@@ -275,6 +275,7 @@ func (t *PaymentRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 	// retry once more with a freshly built payload (mirrors @x402/fetch recovery).
 	recovered, err := t.dispatchPaymentResponseHooks(ctx, build, newResp)
 	if err != nil {
+		newResp.Body.Close()
 		return nil, err
 	}
 	if !recovered || newResp.StatusCode != http.StatusPaymentRequired {
@@ -311,7 +312,8 @@ func (t *PaymentRoundTripper) RoundTrip(req *http.Request) (*http.Response, erro
 	correctiveBuild.paymentPayload = freshPayload
 	correctiveBuild.payloadBytes = freshBytes
 	if _, err := t.dispatchPaymentResponseHooks(ctx, &correctiveBuild, correctiveResp); err != nil {
-		return correctiveResp, nil
+		correctiveResp.Body.Close()
+		return nil, err
 	}
 	return correctiveResp, nil
 }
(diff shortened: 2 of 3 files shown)


phdargen (approved):
Thanks for the contribution @256dino, LGTM 🚀

PR #3023, opened 2026-08-02
docs(xrpl): add upto XRPL scheme specification

dimitricole:
## Description

Adds `specs/schemes/upto/scheme_upto_xrpl.md`, the XRPL profile of the `upto` scheme,
alongside [`scheme_upto_evm.md`](specs/schemes/upto/scheme_upto_evm.md) and
[`scheme_upto_svm.md`](specs/schemes/upto/scheme_upto_svm.md).

XRPL realizes `upto` with Payment Channels, a native ledger primitive: the payer escrows
the ceiling in a channel and signs one off-ledger claim over `(channelId, maxAmount)`, and
after metering the destination claims the actual amount and closes the channel, refunding
the remainder.

## Notes for reviewers

- Closing a channel atomically requires the transaction to originate from the channel
  `Destination`, so the facilitator cannot settle with its own key. The spec has the
  resource server signing the `PaymentChannelClaim` and the facilitator verifying and
  relaying it, keeping the facilitator keyless as on this ledger's `exact` scheme. The
  claim rides in settle-time `extra.settlementTransaction`; no existing scheme puts
  signed material there, and `scheme_upto_svm.md` has the same server-signed voucher but
  does not say where it travels. If there is a preferred home for it, I'll move it.
- Fees are not sponsored, and unlike `exact` the `payTo` account pays the settlement fee
  too, so it must hold XRP. This diverges from the "clients and servers should not need
  to pay gas" preference in the authoring guidance. XRPL has no co-signing mechanism; the
  `Sponsor` amendment would close the gap but is not on Mainnet.
- One rule is new relative to the EVM and SVM profiles: settlement must not re-run
  verification's admission checks, since the payer can invalidate each of them itself
  after the work has run (see the Settle-Time Verification section).
- A facilitator implementation exists and has been run end to end against Testnet. It is
  not in this PR: spec first, as with `upto` on SVM, and it needs client and server
  implementations before it is a complete mechanism.

cc @aristotle-satoshi for visibility, in case this overlaps anything already planned.

## Tests

Documentation-only change; no runtime code paths are modified. The specified flow was
exercised against XRPL Testnet: channel preflight, off-ledger authorization, verification
rejections, metered settlement of 2.47 of 10 authorized XRP with the remainder refunded,
refusal of a claim whose `Amount` was not the authorized maximum, over-settlement refusal,
and replay refusal (`tecNO_TARGET`).

## AI assistance

Per CONTRIBUTING.md: this was developed with significant AI assistance. I have reviewed
the specification, and its behavioural claims were verified on Testnet.

## Checklist

- [x] I have formatted and linted my changes
- [x] This is a documentation-only change; runtime tests are not applicable
- [x] My commits are signed
- [x] Changelog fragment not required for a specification-only change

### DIFF
--- a/specs/schemes/upto/scheme_upto_xrpl.md
+++ b/specs/schemes/upto/scheme_upto_xrpl.md
@@ -67,13 +67,15 @@ drops. A facilitator MAY reject an amount it cannot verify without loss of preci
 
 ## Protocol Flow
 
-The protocol flow for `upto` on XRPL is client-driven.
+The protocol flow for `upto` on XRPL is client-driven, and follows the `authorization`
+payment flow of the v2 specification: verification, then the metered work, then
+settlement.
 
 1. **Client** makes a request to a **Resource Server**.
 2. **Resource Server** responds with `PaymentRequired` in the `PAYMENT-REQUIRED` header,
    with `amount` set to the maximum it may charge.
 3. **Client** submits a `PaymentChannelCreate` to `payTo` with `Amount` set to that
-   maximum, a `CancelAfter` expiry and a `SettleDelay`.
+   maximum, a `CancelAfter` expiry and a `SettleDelay` sized per verification rule 7.
 4. **Client** signs an off-ledger claim over `(channelId, maxAmount)`. This is not a
    ledger transaction.
 5. **Client** sends a new request with the `PAYMENT-SIGNATURE` header containing the
@@ -84,7 +86,7 @@ The protocol flow for `upto` on XRPL is client-driven.
    the actual consumption.
 8. **Resource Server** signs a `PaymentChannelClaim` for the actual amount and calls the
    facilitator's `/settle` endpoint with `amount` set to that actual charge and the signed
-   claim in `extra.settlementTransaction`.
+   claim added to the payment payload as `settlementTransaction`.
 9. **Facilitator** re-verifies, submits the claim, and confirms it succeeded in a
    validated ledger.
 10. **Facilitator** responds with a `SettlementResponse` carrying the settled amount.
@@ -134,8 +136,7 @@ client must authorize at verification, and the actual amount to charge at settle
   "extra": {
     "areFeesSponsored": false,
     "minSettleDelay": 60,
-    "validAfter": 1754136000,
-    "settlementTransaction": "12000F2280000000240000000120..."
+    "validAfter": 1754136000
   }
 }
 ```
@@ -147,7 +148,6 @@ client must authorize at verification, and the actual amount to charge at settle
 | `extra.areFeesSponsored`      | boolean | yes         | Always `false`; XRPL cannot sponsor either transaction                                          |
 | `extra.minSettleDelay`        | number  | no          | Minimum `SettleDelay` the channel must carry, in seconds                                        |
 | `extra.validAfter`            | number  | no          | Earliest activation time, Unix seconds                                                          |
-| `extra.settlementTransaction` | string  | settle only | Hex blob of the `PaymentChannelClaim` signed by `payTo`. MUST be absent from `PAYMENT-REQUIRED` |
 
 ## `PaymentPayload` for `upto`
 
@@ -187,6 +187,17 @@ client must authorize at verification, and the actual amount to charge at settle
 | `publicKey` | Channel `PublicKey`, against which the claim verifies                       |
 | `payer`     | Channel `Account`                                                           |
 
+At settle time the resource server adds one field to the payload before calling
+`/settle`:
+
+| Field                   | Description                                                                                    |
+| ----------------------- | ---------------------------------------------------------------------------------------------- |
+| `settlementTransaction` | Hex blob of the `PaymentChannelClaim` signed by `payTo`; see [Settlement](#settlement) |
+
+The field is server-authored and settle-only. It MUST be absent from the client's
+`PAYMENT-SIGNATURE`, and a facilitator MUST reject a payment presenting it at `/verify`
+with `invalid_upto_xrpl_payload_unexpected_settlement_transaction`.
+
 ## `SettlementResponse`
 
 ```json
@@ -199,7 +210,8 @@ client must authorize at verification, and the actual amount to charge at settle
 }
 ```
 
-`amount` carries the drops actually settled and MAY be `"0"`. `transaction` is the hash
+`amount` carries the settlement-time charge, delivered in full through the channel once
+the settlement validates, and MAY be `"0"`. `transaction` is the hash
 of the validated `PaymentChannelClaim`, including a zero settlement, which is still an
 on-ledger close.
 
@@ -245,6 +257,12 @@ simulation by establishing that a claim against this channel would succeed.
    produces canonical signatures, and rippled rejects a non-canonical claim signature, so
    a facilitator verifying with a generic library MUST enforce this rather than assume it.
 6. **Unclaimed** - `PayChannel.Balance` is `0`. The scheme is single-use.
+
+   A validated-ledger read cannot see a concurrent use: two in-flight payments
+   presenting the same channel both pass this rule, both metered works run, and only
+   one claim can settle, so the payer obtains the other work free. The
+   duplicate-settlement guard below protects settlement only, not this window. A
+   resource server SHOULD NOT run metered work concurrently against one channel id.
 7. **Time bound** - `PayChannel.CancelAfter` is present and at least
    `maxTimeoutSeconds` plus a landing margin beyond the validated ledger's close time,
    `PayChannel.Expiration` is unset, and `PayChannel.SettleDelay` is at least that margin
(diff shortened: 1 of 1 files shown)


dimitricole:
Pushed a tightening round based on self-review. What changed:

- Named a verify-time channel-reuse window under rule 6 (two in-flight payments on one channel both pass verify; only one settles) and pinned its mitigation: resource servers SHOULD NOT run metered work concurrently against one channel id.
- Fixed the landing margin at 20 seconds, with a facilitator interop ceiling and client-side construction guidance.
- Stated settle-time re-verification mechanically: substitute `payload.maxAmount` for `requirements.amount` before re-running rules 1 and 4.
- Moved the settlement claim into the settle-time payment payload (`payload.settlementTransaction`, server-added), aligning with the #3053 payment-flow machinery and the same transport #3094 uses for the SVM voucher; a client-supplied value is rejected at `/verify`.
- Corrected the settle-time Unclaimed rationale: the source can deliver drops unilaterally, so a nonzero `Balance` pre-pays rather than strands, and a fully pre-delivered charge settles as a bare close refunding the undrawn remainder.

Heads-up that the branch history was also rewritten to fix my commit email, so prior SHAs changed.

whawk46:
Reading this as the author of the `exact` XRPL scheme and an operator of a facilitator against these failure classes — and it's strong work. Before the findings, three things this spec gets right that the other profiles don't teach:

- **The verify-vs-settle rule-split table** is the best treatment of phase-dependent re-verification in any upto profile: naming which rules are admission control (payer-invalidatable once the work runs) versus properties of the authorization (re-check in full) is exactly the distinction the EVM profile leaves implicit, and the "a channel admitted at `/verify` but refused at `/settle` means the work ran unpaid" framing should probably migrate upstream into `scheme_upto.md`.
- **The expired-channel catch** — a claim against an expired channel closes it without delivering and still returns `tesSUCCESS` — is the kind of thing only found by reading rippled, and the delivered-amount-from-metadata confirmation is the right mitigation.
- **The dedup release asymmetry** versus `exact` (release on definitive failure, because a fresh server-signed claim has no double-land risk, versus retain, because a payer-presigned blob does) is correctly reasoned, and the release-only-your-own-entry rule closes the re-take race. Verified the supporting claims against rippled: the channel keylet derivation, `tecNO_TARGET` on the second claim, `tecNO_PERMISSION` for unrelated submitters, the NetworkID ≤1024 rule, and destination-close atomicity all check out.

**One design question — why is `CancelAfter` REQUIRED (rule 7)?** The payer's lockup is already bounded without it: a source-initiated `tfClose` schedules `Expiration = now + SettleDelay`, so the payer can always recover funds unilaterally after the delay. Requiring `CancelAfter` therefore doesn't add payer protection it doesn't already have — but it does add a hard deadline the settlement can miss, and the failure mode for missing it is precisely the expired-channel trap this spec flags: the claim "succeeds" and delivers nothing. If the rationale is abandoned-channel hygiene, that's legitimate — but worth stating in the text, and worth considering whether an absent `CancelAfter` (treated as unbounded) should verify, since it's strictly safer for the settlement leg. As written, the rule forces every channel to carry the one field that can turn a valid claim into a free close.

**One interop should-fix — pin the claim message bytes.** Rule 5 says the signature "verifies over `(channelId, maxAmount)`," and everything rides on that check (it's what stands between a facilitator and running metered work on an unverifiable authorization). But the byte format is never stated: the signed message is the `CLM\0` hash prefix (`0x434C4D00`) followed by the 32-byte channel id and the amount as a big-endian unsigned 64-bit drops value. A facilitator verifying with a generic crypto library — which rule 5's own low-`s` note anticipates — needs those bytes; without them, "verifies over" invites implementations that hash the wrong serialization and fail open or closed unpredictably. Suggest stating the format explicitly and citing the reference verifiers: rippled's `channel_verify` RPC and xrpl.js `verifyPaymentChannelClaim`.

**One suggestion from operating experience — give the rule-6 window a facilitator-side guard too.** The spec is honest that two in-flight verifications of one channel both pass and one metered work runs free, and puts the mitigation on the resource server (SHOULD NOT run concurrent work per channel). Keep that as primary — it's the only control that works across facilitators — but a facilitator can cheaply add defense-in-depth: an advisory in-flight record on `(network, channelId)` taken at `/verify` and released at settlement or timeout, the verify-side twin of the settlement dedup entry. We run the equivalent guard on the EVM side (in-flight nonce claims, distinct from consumed nonces, added after auditing against the class of concurrent-duplicate attacks in arXiv:2607.19545) and it has caught real concurrent duplicates in production. A SHOULD here would cost implementers a map and close most of the single-facilitator window.

Smaller notes, take or leave: the fixed 20-second landing margin is a good interop move (a channel built against the spec verifies anywhere), but it may be worth saying it's a floor for admission rather than a landing guarantee, since degraded consensus can exceed it and the settle-time `LastLedgerSequence` window is what actually bounds the wait. And on invoice binding being out of scope: `PayChannel.DestinationTag` exists and is fixed at creation — too small (uint32) to carry an invoice hash, but worth a sentence acknowledging it as the partial mechanism, so readers don't rediscover it.

On the composition question from #3048: once XLS-68 activates, the fee story here changes in exactly the way your prerequisite paragraph anticipates — a sponsored `PaymentChannelCreate` covers the payer's creation fee (`spfSponsorFee`), and reserve sponsorship could cover the owner reserve under an explicit policy, with the escrowed principal remaining the payer's own. That composes with this spec rather than changing it: `extra.areFeesSponsored` flips per-requirements when the extension is active, and the envelope equality rule already handles it. Happy to co-design that follow-on once this lands, as offered there.

None of the findings block the design — the channel-as-authorization model is the right XRPL realization of upto, and the Checks analysis explains why convincingly. With the claim-bytes fix and the `CancelAfter` question resolved, I'd consider this the strongest of the upto profiles.

*Disclosure per CONTRIBUTING: review performed with significant AI assistance under significant human monitoring, direction, and correction.*

dimitricole:
Thanks for the review; the verification down to rippled result codes is exactly what this spec needed. Everything below lands in `d9dbf2b3`.

**Design question, `CancelAfter`.** Kept, with the rationale now in rule 7's text. You are right that it is not payer protection. The rule exists because `scheme_upto.md` requirement 2 makes explicit time bounds a MUST on every authorization and the out-of-scope list rules out open-ended allowances, and `CancelAfter` is the only place XRPL can carry the end bound in consensus: without it the signed claim stays redeemable indefinitely, an open-ended allowance in the parent's own terms, and the SVM profile makes the same call by rejecting `expiresAt == 0` although its program supports it. An absent start time, by contrast, defaults to valid-now and stays bounded; that is the reading under which `validAfter` stays optional here, as in the SVM profile and as EVM expresses with `validAfter = 0`. The trade you named is real, and the text now says so, pointing at the settle-time headroom re-check as the mitigation.

Your question also flushed out a hole on the same mechanism's other edge. The `tfClose` exit is benign at rest but not mid-work: scheduled while the metered work runs, it expires the channel `SettleDelay` later, and rule 7's floor for `SettleDelay` was only the landing margin, so a payer could expire the channel under any claim whose work outran it, and the settle-time re-check would then refuse a claim the work had earned. Rule 7 now requires `SettleDelay` of at least `maxTimeoutSeconds` plus the margin, mirroring the `CancelAfter` clause. Verified against rippled's `PaymentChannelClaim` (a source `tfClose` on a non-dry channel sets `Expiration` to `parentCloseTime` plus `SettleDelay`) and `PaymentChannelFund` (which cannot set it sooner).

**Should-fix, claim message bytes.** Taken. Rule 5 now states the signing message and names rippled's `channel_verify` and xrpl.js `verifyPaymentChannelClaim` as reference verifiers, with a note that the two disagree on units (`channel_verify` takes drops, `verifyPaymentChannelClaim` takes XRP and converts). Confirmed against ripple-binary-codec 2.8.0's `signingClaimData`; it matches your description byte for byte.

**Suggestion, verify-side in-flight guard.** Taken as a SHOULD under rule 6: an advisory entry on `(network, channelId)`, checked and taken atomically at verification, refusing a verification that finds a live entry. Once a settlement attempt arrives the entry follows the duplicate-settlement entry's release rules, since a fixed verify-anchored timeout could expire while a late claim is still landable and reopen the window; with no settlement attempt it expires after `maxTimeoutSeconds` plus the landing margin. The text keeps the resource-server rule primary for the reason you gave.

**Smaller notes.** Both taken. Rule 7 now calls the 20-second margin an admission floor rather than a landing guarantee, with the claim's `LastLedgerSequence` window as what actually bounds the wait, and the invoice-binding limitation acknowledges `DestinationTag`: fixed at creation, but at 32 bits it labels a payment rather than carrying an invoice hash.

On moving the verify-vs-settle rule split upstream into `scheme_upto.md`: agreed it is generic, since each network profile re-derives it today, but I would rather that be a maintainer call and a follow-on PR than grow this one. Happy to draft it if they want it.

And glad the XLS-68 reading matches: `extra.areFeesSponsored` flipping per-requirements with envelope equality already covering it is how I hoped that would compose. See you on #3048 once this lands.

whawk46:
Read through d9dbf2b3 against the diff — it all checks out. A few things worth saying beyond that.

On CancelAfter, I'm convinced. I'd been reading it as payer protection, which it isn't. Reading it instead as the parent spec's mandatory end bound — with `CancelAfter` the only place XRPL consensus can actually hold that bound — settles the question, and the SVM profile refusing `expiresAt == 0` is good company to be in. The trade is in the text now, with the settle-time re-check named as the mitigation. That's all I wanted.

The tfClose catch is the part of this exchange I'll remember. I asked about one edge and you went and walked the other one, and there was a real hole waiting there: a payer schedules `tfClose` while the metered work runs, and the channel expires under a claim the work had already earned. Your fix reads right to me, and I checked the rippled behavior you cite — source `tfClose` on a non-dry channel sets `Expiration` to `parentCloseTime + SettleDelay`, and `PaymentChannelFund` can't pull it earlier. Raising the floor to `maxTimeoutSeconds` plus the margin covers the work, not just the landing.

The release rules on the in-flight entry are the subtle bit, and you got them right. A fixed verify-anchored timeout could lapse while a late claim is still landable — at which point the guard has quietly reopened the very window it exists to close. Tying release to the settlement entry avoids that.

Verdict updated accordingly: with both fixes in and the mid-work edge closed, this is the strongest of the upto profiles. And agreed that lifting the verify-vs-settle split into scheme_upto.md is the maintainers' call — if they want it, I'll second your draft.

See you on #3048.

PR #3027, opened 2026-08-03
fix(ts/go/py): merge server and client builder-code s arrays

ethanoroshiba:
## Description

Resource servers that declare their own builder-code `s` (service codes) previously had the client's `s` silently dropped: `mergeExtensions` always preferred the server's leaf value, including arrays, so a server-declared `s` clobbered whatever the client attached. This scoped the fix to builder-code's `s` field specifically (other extensions' array fields, and payment-requirements `extra` matching, keep exact-match semantics) and then went a step further per [this thread](https://coinbase.slack.com/archives/C0AKTLL5ADP/p1785784884516169): instead of one shared cap that either side can still crowd out, `s` now has dedicated, non-overlapping reservations per party.

**Core merge/echo (`@x402/core`, `go`, `python`):**
- `mergeExtensions` concatenates client + server `s` (client first, deduped) instead of the server clobbering the client.
- Extension echo validation treats builder-code's `s` as additive (an echo may be a superset of what was advertised, with scalar/array coercion), while every other extension's array fields still require an exact match.
- As a defensive backstop, echo validation now rejects a client echo whose `s` exceeds the combined client+server budget outright, rather than accepting an oversized echo as a valid "superset" and relying on the facilitator to silently truncate it later.

**`builder-code` extension:**
- `s` reservations are now per-party instead of one shared cap: `MAX_CLIENT_SERVICE_CODES` (5) for `BuilderCodeClientExtension`, `MAX_SERVER_SERVICE_CODES` (5) for `declareBuilderCodeExtension`, and a new `MAX_FACILITATOR_SERVICE_CODES` (1) for the facilitator. `MAX_SERVICE_CODES` is now the sum (11).
- `declareBuilderCodeExtension`/`DeclareBuilderCodeExtension`/`declare_builder_code_extension` accepts an optional service code(s) argument, so a resource server can attribute its own dependencies (e.g. a server-side SDK) via `s`.
- `BuilderCodeFacilitatorExtension` gained a new `serviceCode`/`ServiceCode`/`service_code` option so the facilitator can append its own service code at settlement, within its own reservation, deduplicated against whatever the client/server already echoed.

## Tests

- TS/Go/Python: core unit tests for the additive `s` merge and echo validation (including that non-builder-code extensions still require exact array matches, and that an oversized echo is rejected rather than truncated); builder-code integration tests for server+client `s` and the facilitator's new service code.
- Regression tests confirming client and server can each use their full reservation simultaneously without either side's entries being dropped.
- Changelog fragments across all three SDKs.

## Checklist


phdargen (approved):
LGTM 🚀

PR #3030, opened 2026-08-03
Add settlement_pending error code for EVM exact settlement

ethanoroshiba:
## Description

Adds `settlement_pending` for EVM settlement receipt-wait failures across exact, upto, and batch-settlement flows. A transaction can be broadcast even when its receipt cannot be retrieved, so callers receive its hash and can reconcile before retrying.

Also preserves confirmed batch-deposit success when post-receipt RPC state reads fail, validates broadcast hashes, and ensures terminal Go settlement errors do not expose malformed transaction values.

## Receipt-wait decision

Treat every receipt-wait exception after a valid broadcast hash as `settlement_pending`. The exception type cannot prove that the transaction will not confirm; preserving the hash is safer than reporting a terminal failure that could lead callers to retry a completed payment.

## Tests

- TypeScript, Go, and Python unit coverage includes receipt-wait failures and invalid broadcast hashes.

## Checklist


phdargen:
Continued in https://github.com/x402-foundation/x402/pull/3083

PR #3031, opened 2026-08-03
feat(evm): add Flare mainnet (14) default stablecoin

whawk46:
Adds Flare mainnet (`eip155:14`) to the EVM default-asset registry across TypeScript, Go and Python, plus the v1 legacy network maps and the support-table docs. Follows `DEFAULT_ASSETS.md`; no paywall regeneration needed (6 decimals).

**Asset:** USD₮0 `0xe7cd86e13AC4309349F30B3435a9d337750fC82D` — EIP-712 name `USD₮0`, version `1`, 6 decimals, EIP-3009.

### Why this asset

Flare has not endorsed a stablecoin for x402, so per the selection policy I used ecosystem alignment as the tiebreaker:

| token | address | supply on Flare | EIP-3009 |
|---|---|---|---|
| **USD₮0** | `0xe7cd86e13AC4309349F30B3435a9d337750fC82D` | **~23.1M** | ✅ |
| USDC.e | `0xFbDa5F676cB37624f28265A144A48B0d6e87d3b6` | ~3.25M | ✅ |

USDC.e is a fully verified alternative (EIP-712 name `Bridged USDC (Stargate)`, version `2`, 6 decimals, EIP-3009 confirmed) — happy to switch on request, it's one line per SDK. If the Flare team would rather make the call themselves, please treat this as a starting point rather than a claim.

### Verification

The policy asks for correct EIP-712 domain parameters, so both facts were checked against the deployed contract rather than read from a docs page:

- **EIP-3009 is present.** `transferWithAuthorization` called with a junk signature reverts `TetherToken: invalid signature` — the function exists and validates. (USDC.e answers `FiatTokenV2: invalid signature`.) A missing function reverts differently, which is what distinguishes "supported" from "absent". `authorizationState(address,bytes32)` also returns cleanly.
- **The domain is correct.** I recomputed `hashDomain({name, version: "1", chainId: 14, verifyingContract})` from the values **as written in each of the three SDK files** and compared against the token's on-chain `DOMAIN_SEPARATOR`:

```
on-chain: 0x4c53d754dcb2f406ecf83695856d8c10b13296a318d9a481043412bf253d43ae
TypeScript  name="USD₮0" → MATCH
Go          name="USD₮0" → MATCH
Python      name="USD₮0" → MATCH
```

One note worth flagging for implementers: this token's EIP-712 name contains **₮ (U+20AE)**. That is a legitimate name and a genuine interop probe — an implementation that mishandles non-ASCII in the domain name produces a different digest and every signature fails. The TypeScript entry encodes it as a `₮` escape, verified above to decode to the same 5-character string in all three SDKs.

### Disclosure

We operate an x402 facilitator on Flare and issue our own EIP-3009 token on Coston2 testnet. This PR deliberately proposes **mainnet only**, with a third-party asset we do not control — a testnet default naming our own token would be self-dealing dressed as a registry entry. Happy to add a Coston2 entry separately if maintainers want one, with the conflict stated up front.

---

*Disclosure per CONTRIBUTING (added retroactively once we noticed the AI-assisted-contributions policy): this PR was built with significant AI assistance (Claude), under significant human monitoring, direction, and correction. Spec claims and reference-implementation behavior were verified against live networks and running code rather than generated from memory. Happy to answer any provenance questions.*

### DIFF
--- a/docs/core-concepts/network-and-token-support.mdx
+++ b/docs/core-concepts/network-and-token-support.mdx
@@ -126,6 +126,7 @@ When you use `"$0.01"` syntax, x402 needs to know which stablecoin to use. The f
 | XDC Network | `eip155:50` | `0xfA2958CB79b0491CC627c1557F441eF849Ca8eb1` | USDC | 6 | EIP-3009 |
 | XDC Apothem Testnet | `eip155:51` | `0xb5AB69F7bBada22B28e79C8FFAECe55eF1c771D4` | USDC | 6 | EIP-3009 |
 | Igra | `eip155:38833` | `0xA5b8BF902b2844dA17d4506cc827F7F1681735E7` | USDC | 6 | Permit2 |
+| Flare | `eip155:14` | `0xe7cd86e13AC4309349F30B3435a9d337750fC82D` | USD₮0 | 6 | EIP-3009 |
 | Celo | `eip155:42220` | `0xcebA9300f2b948710d2653dD7B07f33A8B32118C` | USDC | 6 | EIP-3009 |
 | Celo Sepolia | `eip155:11142220` | `0x01C5C0122039549AD1493B8220cABEdD739BC44E` | USDC | 6 | EIP-3009 |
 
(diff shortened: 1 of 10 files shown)


phdargen (approved):
Thanks for the addition @whawk46

whawk46:
Thanks @phdargen for the review and merge — much appreciated. Happy to keep the Flare entries maintained as the SDKs evolve; ping me anytime something breaks on our side of the table.

PR #3039, opened 2026-08-04
fix: reject external / in bazaar discovery schema (SSRF)

CarsonRoscoe:
## Description

  Fixes an SSRF / local-file-read vulnerability in the Bazaar discovery extension's schema validation,
  present in all three SDKs.

  `paymentPayload.extensions.bazaar.schema` is a client-controlled JSON Schema document that gets passed into
  each language's JSON Schema validator during `validateDiscoveryExtension`/`validate_discovery_extension`. In Go
  and Python, the default schema loader/resolver dereferences any `$ref`/`$id` value that isn't a same-document
  fragment — meaning an attacker could embed a `$ref` pointing at `http://169.254.169.254/...` (cloud metadata),
  an internal service, or a `file://` path, and the facilitator would fetch/read it before ever validating the
  actual `info` payload.

  - **Go** (`go/extensions/bazaar/facilitator.go`): `gojsonschema`'s default loader resolves external
  `$ref`/`$id` via `http.Get`/`os.Open`.
  - **Python** (`python/x402/extensions/bazaar/facilitator.py`): `jsonschema`'s default registry resolves
  external `$ref`/`$id` via `urllib.request.urlopen` (both `http(s)://` and `file://`).
  - **TypeScript** (`typescript/packages/extensions/src/bazaar/facilitator.ts`): Ajv's synchronous `compile()`
  already throws on unresolved external `$ref` by default, but silently accepted a standalone external `$id` with
  no fetch and no rejection.

  **Fix:** each language now recursively scans the schema for `$ref`/`$id` values before compiling/validating it,
  and rejects the schema outright unless every such value is a same-document JSON Pointer fragment (starts with
  `#`). Same-document refs (`#/definitions/foo`) continue to work normally.

  The rule is now documented in `specs/extensions/bazaar.md` so all current and future SDKs implement the same
  behavior consistently.

  ## Tests

  Added targeted regression tests per language, each proving (via a local HTTP server / temp file with a hit
  counter) that no outbound request or file read occurs and the schema is rejected:

  - **Go** (`go/extensions/bazaar/bazaar_test.go`): `TestValidateDiscoveryExtension_SSRF` (5 subtests: `$ref`
  over HTTP, `$ref` via `file://`, `$ref` nested in `properties`, external `$id`, local-fragment-only schema
  still validates) + `TestExtractDiscoveredResourceFromPaymentPayload_SSRF` (end-to-end via a full
  `PaymentPayload`).
  - **Python** (`python/x402/tests/unit/extensions/bazaar/test_facilitator.py`):
  `TestValidateDiscoveryExtensionSSRF` (6 tests, same shape as Go, using a real `HTTPServer`) +
  `TestHasExternalSchemaReference` (8 direct unit tests of the helper).
  - **TypeScript** (`typescript/packages/extensions/test/bazaar.test.ts`): new `describe("SSRF via external
  $ref/$id in schema", ...)` block, 6 tests using a local `node:http` server.

  For each language, I confirmed the tests are meaningful by temporarily reverting just the guard invocation
  (keeping the helper defined) and re-running:
  - Go: caught a real bug this way — the guard call compiled but was a complete no-op due to a type-switch
  identity mismatch (`extension.Schema` is the named type `types.JSONSchema`, which never matches a `case
  map[string]any:`). Fixed with an explicit conversion at the call site.
  - Python: confirmed a real outbound HTTP hit occurs without the fix (and `jsonschema` itself emits a
  deprecation warning calling out this exact behavior as a security concern).
  - TypeScript: confirmed 5 of 6 tests already pass without the fix (Ajv fails closed on `$ref` by default) —
  isolating the guard's real value to the external-`$id` case.

  ## Checklist

### DIFF
--- a/go/.changes/unreleased/fix-bazaar-schema-ref-ssrf.yaml
+++ b/go/.changes/unreleased/fix-bazaar-schema-ref-ssrf.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: Reject bazaar discovery extension schemas containing external "$ref"/"$id" values (anything other than a same-document "#" fragment) before validation, preventing an attacker-controlled schema from triggering an outbound HTTP request or local file read (SSRF/LFI, CWE-918)
+time: 2026-08-04T00:00:00Z
--- a/go/extensions/bazaar/bazaar_test.go
+++ b/go/extensions/bazaar/bazaar_test.go
@@ -2,7 +2,12 @@ package bazaar_test
 
 import (
 	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
 	"strings"
+	"sync/atomic"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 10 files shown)


phdargen (approved):
LGTM 🚀

PR #3044, opened 2026-08-04
fix: match payment-gated routes on the escaped request path

CarsonRoscoe:
PR description

 ## Description

 The Go payment middleware matched routes against the percent-decoded `URL.Path`,
 while HTTP routers dispatch on the escaped path. A request that encodes a
 separator inside a path parameter — `GET /api/users/x%2Fy` against a
 `"GET /api/users/:id"` route — splits into an extra segment for the middleware
 but not for the router. The compiled pattern `^/api/users/[^/]+$` misses,
 `RequiresPayment` returns false, and the router invokes the paid handler anyway:
 the protected resource is served with no `VerifyPayment` and no
 `ProcessSettlement`.

 This is a path-equivalence bug. Four vectors were confirmed, and three
 of them live in the shared `normalizePath`/`parseRoutePattern` helpers in
 `http/server.go`, so **all three adapters (Echo, Gin, net/http) were affected**:

 | Vector | Root cause | Adapter-specific |
 | --- | --- | --- |
 | `%2F` in a `:param` | Middleware used decoded `URL.Path` | Yes |
 | `%252F` | `normalizePath` ran `url.PathUnescape` on an already-decoded path | No |
 | `%5C` | `normalizePath` rewrote every `\` to `/` | No |
 | Trailing `/` on a `/*` route | Trailing slash trimmed, but the pattern required it | No |

 Worth calling out: `%252F` and `%5C` arrive with an **empty** `URL.RawPath`, so
 they bypass purely through `normalizePath` regardless of how the router resolves
 the path. An adapter-only fix would have left those open.

 ### Changes

 - `http/echo`, `http/gin`, `http/nethttp` — pass `URL.EscapedPath()` instead of
   `URL.Path`. `reqCtx.Path` only feeds `getRouteConfig`, so this is contained to
   route matching. `Adapter.GetPath()` is intentionally left decoded: it feeds the
   bazaar extension's path-param *values* after the gate, where decoded is right.
 - `http/server.go` — `normalizePath` decodes one segment at a time and re-escapes
   any decoded `/` and `\`, so a decoded byte can never create a segment boundary
   the router didn't see. Only one decode pass now happens.
 - `http/server.go` — `parseRoutePattern` emits `(?:/.*?)?` for a trailing `/*` so
   the bare prefix matches after the trailing slash is trimmed.

 ### Behavior change

 With a `GET /api/premium/*` route config, a bare `/api/premium` now returns 402
 instead of falling through. Echo and Gin 404 that path anyway, so no
 previously-working request changes outcome — but the gate is now fail-closed.

 ## Tests

 `go test ./...` from `/go` — all 30 packages pass.

 New regression coverage:

 - `http/server_test.go` — `TestRouteMatching_PathNormalizationBypass` (18 cases)
   at the matching layer, plus new `TestNormalizePath` cases. Includes negative
   cases so the fix doesn't over-match: a genuine extra segment
   (`/api/users/x/y`) still misses `:id`, and `/api/premiumx` still misses
   `/api/premium/*`.
 - `http/{echo,gin,nethttp}/middleware_test.go` —
   `TestPaymentMiddleware_EncodedPathDoesNotBypassPaymentGate`, end-to-end through
   each adapter, asserting both that the paid handler never runs *and* that the
   response is 402.

 These were verified to **fail without the fix**: with the four source files
 reverted, all four packages fail and 15 `payment bypassed` assertions fire across
 the three adapter suites.

 Adapter-specific notes discovered while testing:

 - Gin only routes on `RawPath` when `UseRawPath = true`, so the Gin test sets it
   explicitly — otherwise the `%2F` vector 404s and the subtest would silently
   assert nothing.
 - The net/http test uses `http.ServeMux` with Go 1.22+ patterns; every listed
   path was confirmed to dispatch to the paid handler (no 301, no 404) before the
   fix.

 ## Checklist

 - [x] I have formatted and linted my code (`go fmt ./...`, `go vet ./...` clean)

### DIFF
--- a/go/.changes/unreleased/fixed-20260804-142718.yaml
+++ b/go/.changes/unreleased/fixed-20260804-142718.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: 'Match payment-gated routes on the escaped request path so percent-encoded separators cannot bypass the payment gate: the Echo, Gin, and net/http middlewares now pass URL.EscapedPath() instead of the decoded URL.Path, normalizePath decodes one segment at a time (re-escaping any decoded / or \) instead of decoding the whole path twice, and a trailing /* route pattern now also matches its bare prefix'
+time: 2026-08-04T14:27:18.169799-04:00
--- a/go/http/echo/middleware.go
+++ b/go/http/echo/middleware.go
@@ -296,8 +296,11 @@ func createMiddlewareHandler(server *x402http.HTTPServer, config *MiddlewareConf
 			adapter := NewEchoAdapter(c)
 			reqCtx := x402http.HTTPRequestContext{
 				Adapter: adapter,
-				Path:    c.Request().URL.Path,
-				Method:  c.Request().Method,
+				// EscapedPath, not Path: routers dispatch on the escaped path, so
+				// matching on the decoded one lets "%2F" split a segment here but
+				// not in the router, bypassing the payment gate.
+				Path:   c.Request().URL.EscapedPath(),
+				Method: c.Request().Method,
 			}
 
 			// Check if route requires payment before waiting for initialization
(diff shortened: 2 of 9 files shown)


phdargen (approved):
LGTM 🚀

Issue #3047, opened 2026-08-04
Proposal: xrplFeeSponsoring extension — facilitator-sponsored fees for the XRPL exact scheme via XLS-68


whawk46:
## Gap

`scheme_exact_xrpl.md` states it "does not support facilitator-sponsored network fees… Supporting fee sponsorship would require a different payment model, not only a facilitator implementation change", and mandates `extra.areFeesSponsored` MUST be `false`.

That payment model now exists: **XLS-68 (`Sponsor` amendment)** is enabled on XRPL Devnet (verified against the `Amendments` ledger object, ID `BE1F90581635DBCEBFC4678C4B54FEDDC1A17B50FD02CFE765A4132A342126AC`; not on Mainnet — `Majorities` empty as of 2026-08-04).

## Proposal

An extension, `xrplFeeSponsoring`, following the pattern set by `eip2612_gas_sponsoring` / `erc20_gas_sponsoring`:

- Payer signs a `Payment` carrying `Sponsor` + `SponsorFlags` (`spfSponsorFee`); the facilitator pays the XRPL fee.
- Two modes: `cosigned` (facilitator adds `SponsorSignature` per tx) and `prefunded` (on-ledger `Sponsorship` object; no per-tx facilitator signature; bounded by `FeeAmount` pool + per-tx `MaxFee`).
- Sponsor and payer sign the identical signing payload, so neither party can alter terms after the other signs (verified against rippled `STTx::checkSingleSign`).
- Paired minimal relaxation of the base scheme's `areFeesSponsored`/safety-check MUSTs, applying only when the extension is active.

## Working reference implementation

https://github.com/whawk46/xls68-x402-fee-sponsoring — end-to-end on Devnet: zero-XRP payer onboarded (`tfSponsorCreatedAccount`), sponsored TrustSet, IOU payment with facilitator paying exactly the fee, both modes, and negative controls (tamper-after-sign, sponsor-signature replay, fee-drain above `MaxFee` — all rejected). Validated-run evidence with tx hashes included.

Implementer note: Devnet (rippled `3.3.0-rc5`) diverges from the XLS-0068 draft text (`SponsorshipSet` is tx type 91; pre-funding is additive `FeeAmountDelta`). Details in the repo README.

## Scope and timing

[...]

PR #3048, opened 2026-08-04
feat(specs): xrplFeeSponsoring extension — facilitator-sponsored fees via XLS-68

whawk46:
Implements the proposal in #3047: facilitator-sponsored network fees for the XRPL `exact` scheme via the XLS-68 `Sponsor` amendment (Devnet-enabled; amendment status must be verified per network).

## What's in this PR

- **`specs/extensions/xrpl_fee_sponsoring.md`** (new) — follows the `eip2612_gas_sponsoring` pattern: advertisement via the `extensions` object, `cosigned` and `prefunded` modes, verification deltas to the base scheme's MUST list, fee-drain bounds (`maxFee` per tx, `Sponsorship.FeeAmount` pool), and a settlement rule for XRPL's `ter`-retry semantics (a sponsored tx rejected for missing sponsorship is *pending, not dead* — it applies when the sponsorship is funded).
- **`specs/schemes/exact/scheme_exact_xrpl.md`** (7 surgical edits) — `areFeesSponsored` and the safety-check rejections of sponsor fields now read "unless a fee-sponsoring extension is active". No behavior change for implementations not using the extension.

## Reference implementation

https://github.com/whawk46/xls68-x402-fee-sponsoring — end-to-end Devnet proof with validated-run evidence: zero-XRP payer onboarded, sponsored TrustSet, IOU payment with the facilitator paying exactly the fee, both modes, and negative controls (tamper-after-sign, sponsor-signature replay, `MaxFee` fee-drain — all rejected). Includes an implementer note where Devnet (rippled 3.3.0-rc5) diverges from the XLS-0068 draft text (`FeeAmountDelta`, tx type 91).

## Review

@aristotle-satoshi — this touches your scheme; the base-scheme edits are deliberately minimal and I'll happily restructure (or split the PR) per your preference.

No changeset: spec-only.

---

*Disclosure per CONTRIBUTING: built with significant AI assistance (Claude), under significant human monitoring, direction, and correction. Field names and behavior claims verified against live Devnet (`server_definitions`, `Amendments` ledger object) and rippled source rather than generated from memory.*

### DIFF
--- a/specs/extensions/xrpl_fee_sponsoring.md
+++ b/specs/extensions/xrpl_fee_sponsoring.md
@@ -1,0 +1,153 @@
+# Extension: `xrplFeeSponsoring`
+
+## Summary
+
+The `xrplFeeSponsoring` extension enables facilitator-sponsored network fees
+for the [`scheme_exact_xrpl.md`](../schemes/exact/scheme_exact_xrpl.md) scheme
+using XLS-68 (`Sponsor` amendment): the payer signs a `Payment` carrying
+`Sponsor` and `SponsorFlags` fields, and the facilitator — not the payer —
+pays the XRPL transaction fee.
+
+When this extension is active, `extra.areFeesSponsored` is `true` and the
+base scheme's statement that fee sponsorship "would require a different
+payment model" is satisfied: XLS-68 is that payment model.
+
+The extension only applies on networks where the `Sponsor` amendment is
+enabled (currently Devnet; not Mainnet).
+
+## Prerequisites
+
+- `Sponsor` amendment (`BE1F90581635DBCEBFC4678C4B54FEDDC1A17B50FD02CFE765A4132A342126AC`)
+  enabled on the target network. The facilitator MUST verify this against the
+  `Amendments` ledger object, not static configuration.
+- Two operating modes:
+  - `"cosigned"` — the facilitator co-signs each transaction
+    (`SponsorSignature`). No ledger setup required.
+  - `"prefunded"` — a `Sponsorship` ledger object (sponsor = facilitator's
+    account, sponsee = payer) authorizes fee payment with no per-transaction
+    facilitator signature. Bounded by the object's `FeeAmount` pool and
+    per-transaction `MaxFee`.
+
+## `PaymentRequired`
+
+A facilitator advertises support by including the `xrplFeeSponsoring` key in
+the `extensions` object of the `402 Payment Required` response.
+
+```json
+{
+  "x402Version": 2,
+  "accepts": [
+    {
+      "scheme": "exact",
+      "network": "xrpl:2",
+      "asset": "524C555344000000000000000000000000000000",
+      "payTo": "rN7n3473SaZBCG4dFL83w7a1RXtXtbk2D9",
+      "amount": "2.5",
+      "maxTimeoutSeconds": 600,
+      "extra": {
+        "areFeesSponsored": true,
+        "issuer": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q"
+      }
+    }
+  ],
+  "extensions": {
+    "xrplFeeSponsoring": {
+      "info": {
+        "description": "The facilitator sponsors XRPL network fees via XLS-68.",
+        "version": "1",
+        "sponsor": "rSponsor1VktvzBz8JF2oJC6qaww6RZ7Lw",
+        "mode": "cosigned",
+        "maxFee": "1000"
+      }
+    }
+  }
+}
+```
+
+| Field           | Type   | Required | Description                                        |
+| --------------- | ------ | -------- | -------------------------------------------------- |
+| `info.sponsor`  | string | Yes      | Classic address that pays the fee                  |
+| `info.mode`     | string | Yes      | `"cosigned"` or `"prefunded"`                      |
+| `info.maxFee`   | string | Yes      | Maximum sponsored `Fee` in drops, per transaction  |
+
+## `PaymentPayload`
+
+The client builds the full transaction — including `Sponsor = info.sponsor`,
+`SponsorFlags = 1` (`spfSponsorFee`), its own `SigningPubKey`, `Fee <=
+info.maxFee`, and sequencing per the base scheme — signs it, and sends the
+blob as in the base scheme.
+
+- `"prefunded"`: the signed blob is complete and submittable as-is.
+- `"cosigned"`: the facilitator computes `SponsorSignature` over the same
+  signing payload during settlement and inserts it before submission.
+  `SponsorSignature` is a non-signing field, so adding it does not invalidate
+  the payer's signature — and the facilitator cannot alter any other field
+  without invalidating it.
+
+## Verification Logic
+
+All base-scheme verification rules apply, with these deltas:
+
+- `extra.areFeesSponsored` MUST be `true` (envelope check §1 inverted).
+- `tx_json.Sponsor` MUST equal `info.sponsor`.
+- `tx_json.SponsorFlags` MUST equal `1` (`spfSponsorFee`). A facilitator MUST
+  reject `spfSponsorReserve` unless it has an explicit reserve-sponsorship
+  policy: reserve sponsorship locks the sponsor's XRP per created object.
+- `tx_json.Fee` MUST be `<= info.maxFee` (tightens safety check §9's "Fee
+  above facilitator policy").
+- `"prefunded"` mode: the `Sponsorship` ledger object for
+  `(info.sponsor, tx_json.Account)` MUST exist with `FeeAmount >= tx_json.Fee`;
+  `tx_json.SponsorSignature` MUST be absent.
+- `"cosigned"` mode: `tx_json.SponsorSignature` MUST be absent at `/verify`
+  (the facilitator adds it); the facilitator MUST co-sign only transactions
+  that passed every verification rule — the sponsor signature is an
+  endorsement of exactly these payment terms.
+- The `Sponsor` amendment MUST be enabled on the target network.
+- Simulation (§11) MUST account for the fee being charged to the sponsor:
+  the payer needs no XRP balance for the fee.
+
+## Settlement Logic
+
+Base-scheme settlement applies unchanged, plus:
+
+- A rejection of `terNO_PERMISSION` for a `"prefunded"` payment whose
+  `Sponsorship` object does not yet exist is **pending, not dead**: `ter`
+  results are retriable, and the transaction applies automatically once the
+  sponsorship is funded. A facilitator MUST NOT create or fund a
+  `Sponsorship` object while such a transaction may still be live (its
+  `LastLedgerSequence` has not passed) unless it intends that payment to
+  settle.
+- The settlement response SHOULD include `feeSponsored: true` and the sponsor
+  address, so clients can attribute network costs.
+
+## Security Considerations
+
+- **Mutual binding.** Sponsor and payer sign the identical
+  `STX\0`-prefixed signing payload (XLS-68; rippled `STTx::checkSingleSign`).
+  Neither party can alter amount, destination, fee, or sequence after the
+  other signs; sponsor signatures cannot be replayed on other transactions
+  because `Account` and `Sequence` are signed.
+- **Fee drain is the abuse model.** Per-transaction exposure is bounded by
+  `info.maxFee` (and on-ledger by `Sponsorship.MaxFee` in prefunded mode);
+  total prefunded exposure is bounded by `FeeAmount`. Cosigned mode is
+  bounded by facilitator policy. Facilitators SHOULD rate-limit sponsorship
+  per payer account.
+- **Account-creation sponsorship is out of scope** for this extension.
+  `tfSponsorCreatedAccount` onboarding (1 XRP reserve per account) is
+  unbounded sponsor cost and MUST be gated separately if offered.
+
+## Reference Implementation
+
+End-to-end Devnet proof (all modes; negative controls including
+tamper-after-sign, sponsor-signature replay, and MaxFee fee-drain rejection),
+with a validated-run evidence file:
+https://github.com/whawk46/xls68-x402-fee-sponsoring
+
+## Implementation-vs-draft note (for the XLS-68 author)
+
+Observed on Devnet (rippled `3.3.0-rc5`), diverging from the XLS-68 draft
+text: `SponsorshipSet` is transaction type 91 and accepts additive
+`FeeAmountDelta` rather than absolute `FeeAmount`/`ReserveCount` (which are
+rejected by the transaction template); creating a budget-less `Sponsorship`
+fails with `tecNO_PERMISSION`. Implementers should code against
+`server_definitions` of the target network, not the draft field tables.
--- a/specs/extensions/xrpl_fee_sponsoring.md
+++ b/specs/extensions/xrpl_fee_sponsoring.md
@@ -27,6 +27,13 @@ enabled (currently Devnet; not Mainnet).
     account, sponsee = payer) authorizes fee payment with no per-transaction
     facilitator signature. Bounded by the object's `FeeAmount` pool and
     per-transaction `MaxFee`.
+- Codec support for the XLS-68 fields in the client's signing stack — the
+  practical blocker at the time of writing: `xrpl.js` has not shipped the
+  `Sponsor`/`SponsorFlags`/`SponsorSignature` field definitions, and
+  `xrpl-py` carries them on `main` only (absent from the latest release,
+  checked against v5.0.0). Implementations SHOULD load definitions from the
+  target network's `server_definitions` rather than a library snapshot (see
+  the implementation-vs-draft note below).
 
 ## `PaymentRequired`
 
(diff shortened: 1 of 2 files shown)


dimitricole:
From the base-scheme side of the coordination you offered on #3017: the Python
mechanism now pins Payment field acceptance to an explicit allowlist, so a facilitator
without the extension rejects sponsor fields by default and an extension-aware one
would admit `Sponsor` and `SponsorFlags` deliberately. That composes with your §9 edit
("unless a fee-sponsoring extension is active"). The other two touch points on our
side are `get_extra`'s advertisement and the envelope check. Happy to coordinate when
this progresses.

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human
direction and review.*

whawk46:
@dimitricole An explicit allowlist is the better shape, and it's strictly stronger than
what this PR proposes. My §9 sibling clause is a denylist ("reject `Sponsor`,
`SponsorFlags`, `SponsorSignature` unless the extension is active"), which fails open the
moment XRPL adds a field I didn't enumerate — and XLS-68 already renamed one on me
between spec and implementation (`FeeAmount` → `FeeAmountDelta`, measured on Devnet
rc5). An allowlist fails closed against exactly that. If your Python change lands, I'll
rewrite §9 to reference the allowlist as the normative mechanism and keep the field names
as informative examples rather than as the check itself.

On the two touch points:

- **`get_extra` advertisement** — the extension needs to say *which* sponsorship mode it
  supports, not just that it supports sponsorship. The two are not interchangeable: in
  cosigned mode the sponsor signs alongside the payer, so the facilitator must verify two
  signatures and can be drained by a payer who abandons half-built transactions; in
  delegated mode it can't. Our implementation gates on that distinction
  (`src/x402/xrpl-lane.ts`), and a facilitator that advertises the extension without
  naming the mode is advertising an ambiguity.
- **Envelope check** — agreed this is where the composition actually has to hold. Worth
  a joint test vector: one sponsored payload that a non-extension facilitator must
  reject, and one that an extension-aware facilitator must accept, so the two
  implementations are checked against the same bytes rather than against each other's
  prose.

Happy to write those vectors against the Devnet transactions this PR is derived from —
they're real signed blobs, not constructed examples, so they'll catch encoding drift too.

---

*Disclosure: written with significant AI assistance, under human direction, review, and
correction.*

whawk46:
@dimitricole §9 is rewritten — `4164d46`. The allowlist is now the normative
mechanism and the sponsorship field names are demoted to informative examples, with
the reason stated in the spec rather than left in this thread:

> **Field acceptance is an allowlist, not a denylist.** The facilitator MUST hold the
> decoded `Payment` to an explicit set of permitted fields and reject any transaction
> carrying a field outside it, unless an active extension admits that field deliberately.
>
> This is normative because the alternative fails open. A denylist of named fields
> silently widens what a facilitator accepts every time the XRPL adds a field, and the
> field names in any such list are not stable: XLS-68 renamed `FeeAmount` to
> `FeeAmountDelta` between the published draft and the Devnet `3.3.0-rc5`
> implementation, so a list written from the draft would already be wrong.

I cited your derivation-from-rippled's-own-template approach as the way to keep it
correct without maintenance, because that is the part that generalizes — a reviewer
maintaining a hand-written list is the failure mode, not the list itself.

On the joint test vectors I offered: still happy to write them, and I'd rather do it
against the real Devnet blobs than constructed examples so encoding drift gets caught
too. Two vectors, one sponsored payload a non-extension facilitator MUST reject and one
an extension-aware facilitator MUST accept, checked by both implementations against the
same bytes rather than against each other's prose. Say the word and I'll put them in a
form your test suite can consume.

---

*Disclosure: written with significant AI assistance, under human direction, review, and
correction.*

dimitricole:
Thanks for turning the §9 rewrite around so quickly, and for the citation. I read
`4164d46` against the Python behavior, and the normative sentence describes exactly
what the implementation does. One small note from that read: in the base scheme a
`SponsorshipSet` is refused at §3 (`TransactionType` must be `"Payment"`), so the
budget fields the informative list mentions never decide an outcome there. The example
is harmless; it documents the extension's territory rather than the base scheme's.

As for the vectors, yes, please write them. The form the Python
suite consumes is a hex blob asserted verbatim, the same way it already pins fixtures
produced by the TypeScript implementation byte for byte (transaction hash, invoice
id), so real Devnet blobs will slot straight in. The must-reject vector can land in
the base suite as soon as it exists. The must-accept half runs naturally against your
reference implementation today, and the Python suite will pick it up once an
extension-aware mode exists here; in the meantime the same blob can earn its keep in
the base suite as a second must-reject fixture. Rewrapped under base-scheme terms it
must still be refused, and that refusal lands on the bytes themselves.

And I agree on naming the mode in the advertisement.

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human
direction and review.*

whawk46:
@dimitricole Vectors delivered — real Devnet blobs, validated on-ledger, asserted verbatim:
[`test-vectors/sponsoring-test-vectors.json`](https://github.com/whawk46/xls68-x402-fee-sponsoring/blob/main/test-vectors/sponsoring-test-vectors.json) (commit `720b017`).

**`sponsored-payment`** — the base suite's must-reject, usable today exactly as you framed it: rewrapped under base-scheme terms it must still be refused, and the refusal lands on the bytes (`Sponsor` + `SponsorSignature` are right there in the blob, outside the Payment allowlist). Validated ledger 4211027, tx `F306E305…CDCB32A`: a 1-drop payer pays 2.5 FCD while the facilitator carries the 12-drop fee.

<details><summary>hex (350 bytes)</summary>

```
120000240040414E201B0040417A204A0000000161D488E1BC9BF0400000000000000000000000000046434400000000000F8AE3322434A9F800DE6F3F8A3CE4571651C33D68400000000000000C7321EDA1EC51E363E86F50C845611412C6CB8F5736BA25EF9406DA516D5F8A543A394C7440EE1C0CED21D3794FBFBA5E61F4BC44B880BD3A591118138DDCF20075731D355F8EEDF56EAD11F4A662E3B974BF7D3F4665CAEF4EAF61ACB4015B2E7C36CF11018114C226FF9458B8B0D6EEF1C504BC3178F68FD9E1BE83146E7807830BEA4C9D495B091E8CC2FFDB39BD0278801B146C92943ED80CB008FBC02B38EC7D35F74B8A3FE5E0267321EDE870C1A491E87A212D41DBF634476D4A1E50AF3D7CE94CA9BFB8BD476F4CF9317440BCAA29225213CE6A1B85D85756A55B7C5B6F5FB9213D294190F9851BFA8CAD9BEF43BF1E2E47B0F472E042D986E0CB3B22446F1F8DB6B1E7B06F410CD8B86F07E1
```
</details>

**`plain-payment`** — the base must-accept: ordinary issued-currency Payment, every field in the allowlist. Ledger 4211024, tx `A13433C5…870703`.

<details><summary>hex (217 bytes)</summary>

```
120000240040414A201B0040417761D4C38D7EA4C6800000000000000000000000000046434400000000000F8AE3322434A9F800DE6F3F8A3CE4571651C33D68400000000000000C7321ED43DBB35E73B8FC96EC70B956E76CF4C3B4C9611BA9FDD290A90BD8831FD80D32744009C627784FE8F420E459904E98E3AB7BAA6A216F7352F498848185C8B945861CB772254695A2F1D98C47EB893C1705AC19575CA1D7F1294886F23CE68248730681140F8AE3322434A9F800DE6F3F8A3CE4571651C33D8314C226FF9458B8B0D6EEF1C504BC3178F68FD9E1BE
```
</details>

Both are re-derivable without trusting the repo: `tx` with `binary: true` against public Devnet history returns the same bytes, and the JSON carries hashes, ledger indexes, and the account map. When an extension-aware mode exists on the Python side, `sponsored-payment` flips to its must-accept role unchanged.

On §3 vs §9: confirmed and fixed — `942a3b7`. You're right that a `SponsorshipSet` never reaches the field allowlist in the base scheme (§3 refuses any `TransactionType` other than `Payment`), so the informative bullet was attributing an outcome to the wrong check. It now names only the Payment-level sponsorship fields and states where `SponsorshipSet` actually dies, with the budget fields left to the extension whose territory they are.

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human direction and review.*

dimitricole:
Both vectors are in the base suite as of `927dc5a8`. `sponsored-payment` is the
must-reject fixture: the suite reproduces its hash from the bytes and asserts the
refusal on the blob under base-scheme terms, pinning the exception type and the
stable reason code rather than the message, since the failing layer migrates from
codec-unknown to the blob field guards when xrpl-py ships the XLS-68 definitions.
`plain-payment` rides along as the acceptance control at the same layer: every field
sits in the allowlist and the Devnet signature verifies offline. One scope note on
`plain-payment`: it carries no SendMax, so full base-scheme verification would stop at the
IOU SendMax rule; the byte-level claim is what the fixture pins. I re-derived both
blobs from public Devnet history as well, and they came back byte-identical to your
JSON.

Before an extension suite gets wired from the JSON labels: the extension's `/verify`
path expects `SponsorSignature` absent in both modes, and this blob carries it,
so its must-accept role under the extension lives at the settlement layer rather than
at `/verify`. A layer qualifier on `extensionAware` would save a future implementer
the surprise.

The §9 fix reads right. I appreciate the quick work on all of it.

---

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human
direction and review.*

whawk46:
@dimitricole `927dc5a8` noted with real satisfaction — and your independent re-derivation coming back byte-identical is the provenance model doing exactly what it was built for: the repo is a convenience, the ledger is the authority, and nobody has to trust the JSON.

The layer qualifier is in — `ad031d6`. `extensionAware` on `sponsored-payment` now reads `{layer: "settlement", outcome: "MUST_ACCEPT"}` with an explicit warning that the blob is fully countersigned and the extension's `/verify` path expects `SponsorSignature` absent in both modes — feeding it to `/verify` and expecting acceptance is precisely the surprise you flagged, and now the fixture says so itself. `plain-payment` gains the SendMax scope annotation in the same commit: byte-level claim, full base verification stops at the IOU SendMax rule. Pinning the reason code rather than the message was the right call too, given the failing layer migrates when xrpl-py ships the XLS-68 definitions.

When an extension-aware mode grows on the Python side, the `/verify`-layer must-accept fixture (payer-signed, `Sponsor` present, `SponsorSignature` absent) is a fresh Devnet ceremony away — say the word and it lands next to these two.

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human direction and review.*

dimitricole (commented):
I'm reading this as an implementer: I maintain the Python exact mechanism in #3017 and the TS allowlist port in #3114, both built against the joint Devnet vectors from your xls68 repo. The extension design is sound: I verified the signing classifications against rippled source (Sponsor/SponsorFlags signing, SponsorSignature kNotSigning, which is what the whole security model rests on), recomputed the amendment ID, and checked amendment status on Devnet and Mainnet. Two things worth calling out: the spfSponsorReserve rejection is load-bearing, since Payment is in rippled's v1 reserve-sponsorship allowlist and only your rule stops reserve-lock abuse; and the terNO_PERMISSION pending-not-dead rule is the paragraph I most needed as an implementer.

Two should-fixes on the §9 rewrite, plus one open question:

1. The rewrite silently drops the MUSTs for Memos, Fee-above-policy, and Delegate. The old "The facilitator MUST reject transactions with:" line is gone and the list now sits under "informative examples of what an allowlist excludes". §6 still normatively catches Paths/DeliverMin, tfPartialPayment for IOU (the ledger refuses XRP partials outright), and the Amount/DeliverMax pairing, but those three have no other normative home, and they are not really allowlist material anyway: both implementations keep Memos and Delegate inside ALLOWED_PAYMENT_FIELDS so each rejection keeps its own reason code, and Fee-above-policy is a value bound on a required field. Suggest keeping the allowlist paragraph as the lead but retaining "MUST additionally reject" over the list.

2. "Deriving the permitted set from rippled's own field template for the transaction type keeps the check correct without maintenance" fails open if taken literally: the Payment template is the per-type fields plus getCommonFields(), which includes Sponsor, SponsorFlags, SponsorSignature, Delegate, and Memos, so a derived set admits nearly everything the list below it excludes. Both implementations pin a static snapshot instead, for exactly this reason. The MUST itself is fine (a pinned set is "an explicit set"); the rationale sentence is what steers people toward the fail-open version.

3. (question) Cosigned mode changes the transaction hash after verification: SponsorSignature is non-signing, but the txid is computed over the full serialization. The base scheme dedups, polls, and reports on the hash of the received blob, and the client can no longer predict its own hash. Which hash should settle() key on and return? This was the first thing I hit thinking through the implementation.

Smaller things, take or leave: prerequisites don't mention codec support, which is the actual blocker today (xrpl.js hasn't shipped the fields at all, xrpl-py only on main); the prefunded rule could use the Sponsorship keylet or a ledger_entry example since no shipped client library knows the entry type yet (and it doesn't check lsfSponsorshipRequireSignForFee or the object's MaxFee, which can still sink the submit); nothing operationally binds "active extension" to the PaymentPayload extensions echo, and there's no payload example; and one sponsored tx_json example per mode would make the spec and your vectors mutually checkable.

Also curious whether you see this generalizing to other payer-signed transaction types down the line (upto's channel create is the case I care about), but that's future work, not this PR.

Sources checked: rippled develop (TxFormats.cpp, sfields.macro, SponsorHelpers.cpp, Transactor.cpp, STTx.cpp), the Amendments ledger object on Devnet and Mainnet via public JSON-RPC, xrpl.js ripple-binary-codec definitions.json on main, and xrpl-py main against the v5.0.0 release.

Happy to turn any of these into suggested text if that's useful.

whawk46:
This is the review this PR needed — verified against rippled source rather than the draft tables, which is exactly the discipline the implementation-vs-draft note asks for. Everything is adopted and pushed in 23ff907; point-by-point:

**Both should-fixes confirmed and taken.** You're right that the §9 rewrite silently demoted the MUSTs — my rewrite, my regression. The allowlist paragraph stays as the lead mechanism, and the list is normative again ("MUST reject … each with its own reason code"), with your implementer rationale written in: implementations legitimately keep `Memos` and `Delegate` inside the permitted decode set precisely so each rejection carries its own reason instead of a generic field-not-permitted. And the derivation sentence was steering readers into the fail-open version, exactly as you said — the template's common fields admit most of what §9 exists to exclude. The rewritten rationale now requires a pinned explicit snapshot reviewed on amendment activation, and permits derivation only as an alarm: **derive to detect, pin to admit.**

**The txid question — good catch, and here's the answer now in the spec.** Three rules, one per party:

1. The duplicate-settlement cache key is computed over the blob **as received** from the client (stable across retries of the same submission; under cosigning it is not the on-ledger txid). The base scheme's mitigation section now says this explicitly.
2. `settle()` polls for and returns the **final** txid — the hash of the cosigned blob as submitted — in `SettlementResponse.transaction`.
3. A cosigned-mode client MUST NOT predict its own hash; it takes the txid from the response. For hash-independent reconciliation, `(Account, Sequence or TicketSequence)` identifies the payment in both modes (e.g. `account_tx`) — the same doctrine as EVM-side dedup on `(from, nonce)`: identity lives in the intent's uniqueness primitive; the hash is an output.

In prefunded mode the blob is submitted unchanged and the hashes coincide.

**Smaller items, all in:** codec support is now a named prerequisite (xrpl.js unshipped, xrpl-py main-only against v5.0.0 — with the `server_definitions` steer); the prefunded verify rule now checks the object-level `MaxFee` and requires `lsfSponsorshipRequireSignForFee` unset, with the `ledger_entry`-by-keylet route called out since no shipped library knows the entry type; the extension is now activated by the `PaymentPayload.extensions` echo (with a payload example), so sponsorship fields without the echo fail §9 deterministically instead of being inferred into the extension path; and there's one `tx_json` example serving both modes, so the spec and the Devnet vectors are mutually checkable.

**On generalizing to other payer-signed transaction types: yes, and your case is the one I'd do first.** XLS-68's binding is transaction-type-agnostic — sponsor and payer sign the same payload, so nothing in the mechanism is Payment-specific. This extension deliberately stays scoped to the base scheme's `Payment`-only rule, but a sponsored `PaymentChannelCreate` is the natural companion for `upto`: it would let a zero-XRP agent open the channel your scheme meters. Happy to co-design that as a follow-on once #3023 settles — it composes with your spec rather than living in this one.

Thanks also for the b53010f0 rebase heads-up — records updated on our side.




*Disclosure per CONTRIBUTING: prepared with significant AI assistance under significant human monitoring, direction, and correction.*

dimitricole:
Everything checks out. Verified `23ff907` against the review point by point: both should-fixes, the txid rules, and all four smaller items are in, and the diff matches your description throughout. Two touches beyond the summary also read right: the prefunded budget wording moving to the remaining fee budget, and the sponsorship-fields list losing its reader's-benefit hedge now that the list is normative again.

The txid answer settles the question cleanly. Identity living in the intent's uniqueness primitive with the hash as an output is the right doctrine, and keying the dedup cache on the blob as received keeps retries stable in both modes.

Derive to detect, pin to admit is a better sentence than my review had. That closes out my review.

And yes to the sponsored `PaymentChannelCreate` follow-on. Once #3023 settles I'll open that thread.

Issue #2993, opened 2026-07-30
Bazaar: a resource can be silently unpayable because its description is too long — document the limit and give it a distinct error


sukrutkrdg:
The Bazaar troubleshooting section lists what a declaration must contain to be catalogued, and says nothing about how *large* any part of it may be. On our deployment, one field's length is the difference between an endpoint that can be paid and one that silently cannot — and I think that belongs in the spec or the troubleshooting guide, whichever is the right home.

I know the facilitator is a third-party product and that verify behaviour is theirs; this is also filed at coinbase/cdp-sdk#759. I'm raising it here because the troubleshooting doc is where a seller looks first, and because two of the questions below are spec questions, not facilitator questions.

## The measurement

Seller: `402.com.tr`, x402 v2, `exact` scheme, USDC on Base mainnet, CDP facilitator, `@x402/*` 2.16.0, declarations built with `declareDiscoveryExtension`.

Five endpoints had never settled a single x402 payment in over a month. Every payment-carrying request came back `402` in ~1–2.5 s — before our handler ran — with an empty body. The same handlers, same inputs, served fine through our own prepaid-credit path, which does not involve the facilitator. Shortening `resource.description`, and changing nothing else, made four of them settle immediately:

| endpoint | description (chars) | challenge (b64 bytes) | before | after |
|---|---|---|---|---|
| `safe-check` | 647 → 469 | 3348 → 3112 | never settled | **settles** |
| `morpho-vault` | 608 → 488 | 3712 → 3548 | never settled | **settles** |
| `gas-payer` | 672 → 483 | 3108 → 2852 | never settled | **settles** |
| `morpho-liquidatable` | 517 → 438 | 2964 → 2856 | never settled | **settles** |
| `paymaster-check` | 582 → 469 | 2968 → 2816 | never settled | untested (test wallet ran dry) |

Settlements on Base mainnet, payer `0xFe21a68f21d556A3C4274a44c2Fb5410c50cDa1C` → `payTo 0x973a31858f4d2125f48c880542da11a2796f12d6`:
[`0x3ef3f3a6…f950`](https://basescan.org/tx/0x3ef3f3a6c76963eae355adad60f11477e5526300fc0d590f500e976b1429f950) ·
[`0xcf24a2b5…b9ee`](https://basescan.org/tx/0xcf24a2b558de3159953d7f661bf99dad5896db1dc72f5d90128b2887d62fb9ee) ·
[...]

Issue #3019, opened 2026-08-01
[Bug]: `withX402()` hardcodes a `"*"` route pattern → `routeTemplate: ":var1"` fails Bazaar’s required `matches_resource` check, so Next.js resources are never indexed


kaditang:
### What happened?

`@x402/next`'s `withX402()` hardcodes the route pattern as `"*"`:

```js
// @x402/next/dist/esm/index.js
function withX402(routeHandler, routeConfig, server, paywallConfig, paywall, syncFacilitatorOnStart = true) {
  const routes = { "*": routeConfig };          // ← hardcoded wildcard
  const httpServer = new x402HTTPResourceServer2(server, routes);
  return withX402FromHTTPServer(routeHandler, httpServer, paywallConfig, paywall, syncFacilitatorOnStart);
}
```

When the bazaar discovery extension is registered, `@x402/core` expands a `"*"` pattern into the
auto-generated placeholder `routeTemplate: ":var1"` — it even warns about this at boot:

```
[x402] Route "*": Wildcard (*) patterns with bazaar discovery extensions will auto-generate
parameter names (var1, var2, ...). Consider using named parameters instead (e.g. /weather/:city)
for better discovery metadata.
```

The warning undersells the consequence. CDP's Bazaar treats the mismatch as a **required**
failure, so the resource is **never indexed** — no matter how many payments settle:

```
POST https://api.cdp.coinbase.com/platform/v2/x402/validate
{"resource":"https://stockwaves.net/api/rotation","method":"GET"}

  check:    bazaar.routeTemplate.matches_resource
  severity: required
  passed:   false
  detail:   resource.url "https://stockwaves.net/api/rotation" does not match routeTemplate ":var1"
```

Because `withX402()` is the documented one-call entry point for Next.js, this appears to affect
every Next.js resource server that uses it together with bazaar discovery. In our case two
production services had hundreds of successful mainnet settlements through the CDP facilitator
and were entirely absent from the Bazaar catalog (0 hits for either domain across ~1,000
paginated entries) until we stopped using `withX402()`.

Related indexing reports that may share this cause: #2942, #2814, #2821, #3011.

### Steps to reproduce

1. Build a Next.js route handler with `withX402(handler, routeConfig, server, …)` where `server`
   has `bazaarResourceServerExtension` registered and `routeConfig` declares a discovery
   extension via `declareDiscoveryExtension(...)`.
2. Deploy on Base mainnet behind the CDP facilitator and settle at least one real payment.
[...]

PR #3049, opened 2026-08-05
specs/bazaar: metadata handling must not block payment

sukrutkrdg:
Closes the gap that #2993 has been circling: the spec says what to do with bad discovery metadata, but never says what happens to the **payment** when metadata is unacceptable.

Today those two things are entangled in at least one facilitator. A resource of mine is indexed, healthy, and refused at verify — a `402` in ~1.2s, before the handler runs, no settlement. That is byte-identical to a bad signature and to an unfunded payer, so from the seller's side it is an outage with no cause. Five of my endpoints sat unpayable for over a month on this before I worked out what changed.

The measurement behind it, on one resource inside two hours today:

| `resource.description` | paid attempts | result |
|---|---|---|
| 469 chars | 1 | settled |
| 494 chars | 1 | settled |
| 582 chars — original text | 2 | `402`, no settlement |
| 582 chars — **different text, same 582 chars / 584 bytes** | 3 | `402`, no settlement |

Same endpoint, handler, price, `payTo` and deploy; only the description string moved. The control rules out the obvious confound: the same buyer wallet settled a different endpoint of mine in the same minute, holding 1.73 USDC against a $0.05 price.

Note the last row — a completely different string of exactly the same length fails identically, so on this resource the trigger tracks length rather than content.

## What this PR changes

Two additions, no behaviour redefined that wasn't already implied:

1. **Facilitator Behavior** — cataloging and payment are independent. A field that fails a validation rule, a field with no rule (`description` has none), and a `bazaar` block rejected in full are all cataloging outcomes; the payment is verified and settled on its own terms. Facilitators MUST NOT return a payment error for a metadata problem.

2. **Verify and Settlement Response Header** — a facilitator that rejects discovery info SHOULD emit `EXTENSION-RESPONSES` with `status: "rejected"` and a `rejectedReason`. Once a metadata failure may not be reported as a payment error, this header is the only place left to report it.

The first is already the spec's own logic — the soft-drop rules exist precisely so that hostile metadata degrades the catalog entry rather than the transaction — it just is not stated for the payment. The second only moves an existing `MAY` to `SHOULD` for the rejection case, using the mechanism the spec already defines.

## What this PR deliberately does not do

It does not add a length rule for `description`, and it should not: @Nikolife2016 measured the full public index and found, in code points, 74 resources carrying 508 or more, 73 of which took paid calls in the last 30 days — with a high-water mark of 723 code points and 102 distinct payers. (An earlier revision of this body quoted 90 resources at 517+ characters; that count was taken in UTF-16 units and was corrected in #2993. The conclusion and the direction are unchanged.) Whatever produces the behaviour above, a global cap is not it, and writing one into the spec would codify a bug.

I have since pinned my own threshold and posted it in #2993: the count is **Unicode code points**, with the cutoff between 499 and 508. A probe carrying 20 non-BMP emoji settled at 498 code points while measuring 518 UTF-16 units and 558 bytes — values already refused in earlier rows — so a rule on UTF-16 units would have to refuse 519 and accept 518, and a rule on bytes would have to refuse 511 and accept 558. It is also not the specific text: two unrelated 582-code-point strings fail identically.

That number stays out of this diff, but the unit is worth a line here for anyone writing a checker: `length` in TypeScript, `len()` in Python and `len()` in Go give three different answers for a description containing an emoji, and only Python's matches. The spec already names that divergence as the reason `serviceName` is bounded to ASCII; `description` has no rule at all.

### DIFF
--- a/specs/extensions/bazaar.md
+++ b/specs/extensions/bazaar.md
@@ -428,6 +428,18 @@ When a facilitator receives a `PaymentPayload` containing the `bazaar` extension
 
 How a facilitator stores, indexes, and exposes discovered resources is an implementation detail. Facilitators may choose to catalog resources in a database, expose them via a discovery API, or process them in any manner they see fit.
 
+Discovery metadata **MUST NOT** affect payment verification or settlement.
+Cataloging and payment are independent: a field that fails a validation rule, a
+field that has no rule (such as `description`), and a `bazaar` block rejected in
+full are all cataloging outcomes, and the payment is verified and settled on its
+own terms regardless. Facilitators **MUST NOT** return a payment error for a
+metadata problem.
+
+The reason is that the seller cannot tell the difference. A verify-side
+rejection, an invalid signature, and an unfunded payer all surface as the same
+fast `402`, so a resource refused over its metadata looks to its operator like an
+outage with no cause — and to a buyer like a broken endpoint.
+
 ### Optional Discovery Endpoints
 
 Facilitators that implement Bazaar discovery may expose discovery APIs to let clients browse and search cataloged resources.
@@ -506,6 +518,12 @@ EXTENSION-RESPONSES: eyJiYXphYXIiOnsic3RhdHVzIjoicmVqZWN0ZWQiLCJyZWplY3RlZFJlYXN
 *(base64 of `{"bazaar":{"status":"rejected","rejectedReason":"info failed schema validation"}}`)*
 
 Clients that understand the `bazaar` extension SHOULD read the `bazaar` key of this header to confirm cataloging succeeded and surface any rejection reason for debugging.
+
+A facilitator that rejects discovery info **SHOULD** emit this header with
+`status: "rejected"` and a `rejectedReason`. Without it, rejection is silent:
+the resource never appears in the catalog and its operator has no signal to act
+on. This is the only place in the flow where a cataloging failure can be
+reported, because per the rule above it must not be reported as a payment error.
 
 ---
 


Nikolife2016:
Your fourth row settles it, and it settles it against the reading I was leaning toward. A
different 582-character string failing identically kills the per-resource-verdict-keyed-on-text
explanation on your deployment — if changing the text minted a fresh key, that row would have
passed. On your resource the trigger tracks length. I was wrong to weight candidate 1 as heavily
as I did, and the thing that decided it was you running the control rather than either of us
arguing about it.

## Data for the range you are still narrowing

You are looking for where your threshold falls between 494 and 582. I cannot see your side, but I
can say what that band does everywhere else. From the public index today:

| description length | resources | with paid calls (30d) |
|---|---|---|
| 490–499 | 73 | 72 |
| 500–509 | 34 | 34 |
| 510–519 | 5 | 5 |
| 520–579 | 21 | 20 |
| **494–582 total** | **133** | **131** |

The whole band is payable elsewhere. The high-water mark inside it is `2s.io/api/convert/unit` at
**515 characters with 105 distinct payers in 30 days** — comfortably past where your own endpoint
stops settling. So whatever number you find, it is a property of your path, not of the range, and
your PR is right that it belongs in an implementation rather than the spec.

## A correction to something I told you earlier

In #2993 I said the stuck state "is not something a lot of sellers are silently sitting in",
because only 3 resources in my index have been listed over two weeks with zero settlements. That
observation is still true, but I drew too strong a conclusion from it, and your result shows why.

If the trigger is length, and the fix is any edit to the text, then the cohort empties itself:
sellers stumble into a working description the way you did and never report it, because from
outside there is nothing to report. A small standing population does not bound how many endpoints
pass through the state — it is what a self-clearing, unreportable failure looks like. That is an
argument for your second change, not against it. From my side I index roughly 19k resources and I
cannot distinguish "seller misconfigured", "facilitator refused the metadata" and "buyer had no
funds" for any of them; `EXTENSION-RESPONSES` with a `rejectedReason` is the difference between a
diagnosable ecosystem and one where a month-long outage has no cause anyone can name.

## Practical, unrelated to the substance

`check-verified-commits` is failing on `cb9d4f9` and the bot gives you a week. We had a PR die at
exactly that gate, so in case it saves you the same hour: the check needs the commit itself signed
and the signing key registered on your GitHub account, and the API will not do it for you —
commits created through the REST API are unsigned unless they come from the web UI or a GitHub
App, which I found out by trying. Settings → SSH and GPG keys → New SSH key, key type **Signing
Key**, then `git config gpg.format ssh`, `git config user.signingkey <path to the .pub>`, rebase
with `-S`, force-push. The Vercel failure is not yours to fix — it wants authorization from the
Coinbase team.

Worth the trouble here. The substance of this PR is right, and it is right for a reason that
outlives the specific bug: the spec already decided that hostile metadata should degrade the
catalog entry instead of the transaction, and simply never said so where the payment path could
read it.

sukrutkrdg:
Two updates, neither of which changes the diff.

**Corrected figures.** The body cited "90 resources at 517+ characters, 89 with paid calls". That count was taken in UTF-16 units, and @Nikolife2016 re-measured it in code points in #2993: **74 resources carry 508 or more code points and 73 of them settled payments in the last 30 days**, with a high-water mark of **723 code points and 102 distinct payers**. The conclusion and the direction are unchanged — the band this PR declines to cap is demonstrably payable everywhere else — but a PR about counting in the right unit should not quote its own evidence in the wrong one. The body is updated.

**Overlap with #2998.** That PR also edits `specs/extensions/bazaar.md` and also says "Fixes #2993". The hunks do not conflict — it inserts near line 321, this one near 428 and 507 — so both can land. They disagree on one substantive point: #2998 writes a 1,000-character upper bound for `description` into the spec. No measurement supports that number. The refusals I recorded begin between 499 and 508 code points, and resources at 723 code points settle normally on the same facilitator, so a 1,000-character rule would neither describe the observed behaviour nor prevent it. This PR deliberately names no number for that reason. The `EXTENSION-RESPONSES` requirement is compatible between the two and could be merged into whichever lands first.

One correction I owe this thread, since it bears on the indexing half: every one of my own resources was also declaring `routeTemplate: ":var1"`, the generated placeholder from #3019. That is a second, independent reason a resource can be absent from the index, and I had been conflating it with the payment refusals. It is fixed on my side. The length-keyed refusals reproduce with the declaration otherwise unchanged, so the substance of this PR is unaffected.

@phdargen — this is the spec half of the behaviour your bazaar indexing troubleshooting guide (#2947) documents, if you have a moment for it. Commits are signed and `check-verified-commits` is green; the failing Vercel check needs authorization from the Coinbase team and is not something I can resolve.

PR #3051, opened 2026-08-05
fix(core): throw when no scheme server is registered

### LINKED ISSUE #2760: x402ResourceServer.buildPaymentRequirements() fails silently (empty array + console.warn) instead of throwing when no scheme is registered
## Problem

Source: `typescript/packages/core/src/server/x402ResourceServer.ts` lines 687-693 (also reproduced in the compiled `@x402/core@2.17.0` npm package, `dist/cjs/server/index.js` around line 1047-1060):

```ts
// Fallback to placeholder implementation if no server registered
// TODO: Remove this fallback once implementations are registered
console.warn(
  `No server implementation registered for scheme: ${scheme}, network: ${resourceConfig.network}`,
);
return requirements; // requirements = [] — empty array, not a throw
```

When no scheme is registered for the requested network, this resolves successfully with an **empty array** and only a `console.warn` side effect — it does not throw. The source's own `TODO` comment confirms this is known-provisional code, not an intentional API contract.

## Why this matters

We hit this while adding a second network scheme to an existing `x402ResourceServer`-based integration that only had one scheme registered. Its error handling was written expecting `buildPaymentRequirements()` to throw on misconfiguration:

```ts
try {
  requirements = await resourceServer.buildPaymentRequirements(resourceConfig);
} catch {
  return PRICE_COMPUTE_FAILED;
}
```

Since nothing throws, that catch never fires, and the caller proceeds to build a real 402-shaped response with `accepts: []`. A client sees "payment required" with zero payment options and no indication anything is actually broken server-side — the `console.warn` is server-side-only and invisible to the caller.

## Repro

Construct an `x402ResourceServer` with only one scheme registered, call `buildPaymentRequirements({scheme: "exact", network: "<a different network>", ...})`. Logs the warn, resolves with `[]`, no exception.

## Acceptance criteria

- [ ] `buildPaymentRequirements()` throws (or returns a typed error result) when no matching scheme is registered, instead of silently resolving to an empty array
- [ ] Consumers that currently rely on a thrown exception for their error-handling path get a real signal to act on instead of building a client-facing `accepts: []` response

VedantAnand17:
## Summary

This changes `@x402/core` behaviour, and per maintainer review the changeset requests a patch bump. `x402ResourceServer.buildPaymentRequirements()` now throws when no scheme server is registered for the requested scheme and network, where it previously logged a `console.warn` and resolved with an empty array. It also updates 21 tests across `@x402/express`, `@x402/fastify`, `@x402/hono` and `@x402/next` that were depending on the old lenient behaviour.

The old path produced a 402 response carrying an empty `accepts` list, so a client was told payment was required but given zero options, and a caller that wrapped the call in `try`/`catch` never saw the misconfiguration because the warning is only visible server side. An unregistered scheme is a server configuration error rather than a valid state, and the same function already throws fifteen lines further down when the facilitator does not advertise the scheme and network. The two guards now behave consistently.

Closes #2760.

### Changes

- `typescript/packages/core/src/server/x402ResourceServer.ts`: replaced the `console.warn` and empty array return with a thrown `Error`, and removed the obsolete `TODO`. The message mirrors the adjacent facilitator guard: `No server implementation registered for <scheme> on <network>. Make sure to call register() with a scheme server for this network.`
- `typescript/packages/core/test/unit/server/x402ResourceServer.test.ts`: the test pinning the old behaviour now asserts the throw via `rejects.toThrow`, renamed to `should throw if no scheme registered for network`.
- `typescript/packages/http/{express,fastify,hono,next}/src/*.test.ts`: six routing tests built an `x402ResourceServer` with no scheme registered and relied on the empty `accepts` array to still reach a 402. Each now registers the `exact` scheme with a stub facilitator advertising it, so the assertion is a genuine 402.
- `typescript/.changeset/resource-server-throw-on-unregistered-scheme.md`: patch changeset for `@x402/core` (downgraded from major per review).

### Why

The fallback was already flagged as temporary by the source itself:

> ```ts
> // Fallback to placeholder implementation if no server registered
> // TODO: Remove this fallback once implementations are registered
> ```

Registering scheme implementations is now the normal path. Leaving the fallback in place converts a configuration mistake into a malformed but superficially valid 402, which is harder to diagnose than an exception at the point of the mistake.

The maintainer asked for a patch bump, so the changeset now requests patch. A call that previously resolved with `[]` now rejects, and four in-repo packages relied on the lenient path without anyone noticing.

That is also a scope question for maintainers, since this began as a fifteen line fix and now touches tests in four more packages. Two alternatives are legitimate: remove the fallback on a planned major instead, or have the middleware layer translate a requirements-building failure into a clearer response. I am happy to rework, or to split the middleware test updates into a follow-up PR.

#2878 previously proposed the same core fix and was closed for unverified commits rather than on substance. Both commits here are signed and verified.

Every other caller was checked: only `buildPaymentRequirementsFromOptions()` consumes it, and it spreads the result and now propagates the error.

Disclosure per `CONTRIBUTING.md`: the majority of this PR was generated with an AI coding agent. I reviewed the diff and the test results before opening it. This description was revised after an earlier measurement error.

## Test plan

Blast radius, measured with `@x402/core` rebuilt in each arm:

| Core source | Middleware tests | express | hono | fastify | next |
| --- | --- | --- | --- | --- | --- |
| `upstream/main` | unchanged | 59/59 | 49/49 | 50/50 | 63/63 |
| this PR | unchanged | 10 failed | 1 failed | 5 failed | 5 failed |
| this PR | updated here | 59/59 | 49/49 | 50/50 | 63/63 |

- [x] `pnpm test` from `typescript/`: 55 of 55 tasks, 3740 tests passed, exit code 0.
- [x] `pnpm --filter @x402/core test`: 550 passed (550).
- [x] `pnpm format:check` and `pnpm lint:check` from `typescript/`: 27 of 27 tasks each.

## Checklist


typescript/.changeset/resource-server-throw-on-unregistered-scheme.md
```diff
@@ -0,0 +1,7 @@
+---
+"@x402/core": major
```
phdargen: Please change to patch
```suggestion
"@x402/core": patch
```

phdargen:
Thanks @VedantAnand17 , looks good but please downgrade to `patch`

VedantAnand17:
Thank you @phdargen for looking onto this :)

Done - the changeset now requests a patch bump (a1e3d509).

I also merged the latest main into the branch: main added construct-time route validation that requires registered scheme servers to declare a paymentFlows table, which broke 21 middleware tests across express, fastify, hono and next. Those stubs are updated and the full suite, format and lint are green.

phdargen (approved):
Thanks @VedantAnand17

PR #3053, opened 2026-08-05
feat(ts): payment flow handlers

phdargen:
## Description

x402 currently hardcodes a single verify → work → settle path, which works well for authorization-style payments but blocks escrow models like svm upto and prepaid models like upfront. This PR lets each mechanism declare which payment flow it needs, so core can order verify and settle around the resource handler without baking one trust model into the middleware.

Payment flow handlers are the named orderings of verify / settle relative to resource execution. Mechanisms opt in via a declared flow (signaled on the wire with extra.paymentFlow when not the default); clients and facilitators stay the same, while servers gain a first-class way to settle before work, settle twice around work, or skip settle entirely when payment was already committed.

verify should remain **read-only**: it may validate payment state but must not commit funds or write onchain state. Escrowing a ceiling (or opening a channel) is a durable commitment, so it belongs on the settle path; the settle itself is the check, and a later settle records the final charge.

**Payment flows:**
- `authorize (default)`: verify → work → settle → respond. Used by today’s exact flows and EVM upto
- `escrow`: settle → work → settle → respond. Used by SVM upto (deposit, then charge) and EVM auth-capture (authorize, then capture)
- `upfront`:  settle → work → respond. Used by upcoming exact with assetTransferMethod=transaction-proof, and BTC Lightning-style prepaid flows

Multiple payment flows can live under one scheme, eg exact can cover simultaneously:

Facilitator-settled flows:
```
{
 "extra": {
  "paymentFlow": "authorize"
 }
}
```
and 
```
{
 "extra": {
  "paymentFlow": "upfront"
 }
}
```
as well as client settled flows:
```
{
 "extra": {
  "assetTransferMethod": "tranasction-proof",
  "paymentFlow": "upfront"
 }
}
```

References:
- upfront PR: https://github.com/x402-foundation/x402/pull/2520 

## Checklist

### DIFF
--- a/specs/x402-specification-v2.md
+++ b/specs/x402-specification-v2.md
@@ -42,7 +42,7 @@ The x402 protocol follows a standard request-response cycle with payment integra
 3. **Payment Authorization Request**: Client submits a signed payment authorization in the subsequent request
 4. **Settlement Response**: Server verifies the payment authorization and initiates blockchain settlement
 
-This cycle describes the default `authorize` payment flow, in which the payment is verified before the resource executes and settled afterward. Schemes may declare other flows that settle before execution; see section 6.1 Payment Flow Models.
+This cycle describes the default `authorization` payment flow, in which the payment is verified before the resource executes and settled afterward. Schemes may declare other flows that settle before execution; see section 6.1 Payment Flow Models.
 
 **3. Protocol Components**
 
@@ -281,13 +281,13 @@ Individual schemes and their per-network bindings — including `exact`, `upto`,
 
 Schemes differ not only in how a payment is formed and validated, but in **when** settlement occurs relative to resource execution. A mechanism (a scheme on a specific network) declares its payment flow by name. The flow determines the ordering of the facilitator's read-only `/verify` (section 7.1) and state-committing `/settle` (section 7.2) around the resource server's execution of the protected request.
 
-When a mechanism does not declare a flow, the default is `authorize`. A mechanism selecting a non-default flow MUST signal it to the client on the wire via `extra.paymentFlow`, so the client can reason about the trust model before paying — in particular, whether payment becomes final before the resource executes: if the resource handler fails, an `authorize` payment flow is never settled, whereas an `upfront` payment flow has already been committed.
+When a mechanism does not declare a flow, the default is `authorization`. A mechanism selecting a non-default flow MUST signal it to the client on the wire via `extra.paymentFlow`, so the client can reason about the trust model before paying — in particular, whether payment becomes final before the resource executes: if the resource handler fails, an `authorization` payment flow is never settled, whereas an `upfront` payment flow has already been committed.
 
 The following flows are defined:
 
 | Flow                  | Ordering                                        | Description                                                                                                                              |
 | --------------------- | ----------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------- |
-| `authorize` (default) | verify → resource → settle → respond            | Read-only verify before the resource executes; funds move only after it completes successfully. |
+| `authorization` (default) | verify → resource → settle → respond        | Read-only verify before the resource executes; funds move only after it completes successfully. |
 | `upfront`             | settle → resource → respond                     | Payment is durably committed before the resource executes, giving the server finality first. Required by networks with no pull-settlement primitive. |
 | `escrow`              | settle → resource → settle → respond            | A first settle commits a deposit or ceiling, the resource executes, and a second settle records the final charge. |
 
--- a/specs/x402-specification-v2.md
+++ b/specs/x402-specification-v2.md
@@ -127,7 +127,7 @@ Each `PaymentRequirements` object in the `accepts` array contains:
 | `asset`             | `string` | Required | Token contract address or ISO 4217 currency code for fiat     |
 | `payTo`             | `string` | Required | Recipient wallet address or role constant (e.g., "merchant")                                                              |
 | `maxTimeoutSeconds` | `number` | Required | Maximum time allowed for payment completion                                                                               |
-| `extra`             | `object` | Optional | Scheme-specific additional information                                                                                    |
+| `extra`             | `object` | Optional | Additional information. Reserved protocol keys: `assetTransferMethod`, `paymentFlow` (section 6.1); other keys are scheme-specific |
 
 The `ResourceInfo` object contains:
 
@@ -273,15 +273,17 @@ Each scheme defines:
 
 - How to construct the `payload` field within `PaymentPayload`
 - Settlement and validation procedures
-- Scheme-specific requirements in the `extra` field of `PaymentRequirements`
+- Requirements in the `extra` field of `PaymentRequirements` (reserved protocol keys in section 6.1; remaining keys are scheme-specific)
 
 Individual schemes and their per-network bindings — including `exact`, `upto`, `batch-settlement`, and `auth-capture` — are specified under [`specs/schemes/`](./schemes/).
 
-**6.1 Payment Flow Models**
+**6.1 Asset Transfer Methods and Payment Flow Models**
 
-Schemes differ not only in how a payment is formed and validated, but in **when** settlement occurs relative to resource execution. A mechanism (a scheme on a specific network) declares its payment flow by name. The flow determines the ordering of the facilitator's read-only `/verify` (section 7.1) and state-committing `/settle` (section 7.2) around the resource server's execution of the protected request.
+An `assetTransferMethod` identifies **how** value is authorized or moved for a mechanism (a scheme on a specific network) — for example `eip3009` vs `permit2` on EVM `exact`, or `sequence` vs `ticketSequence` on XRPL `exact`. Allowed `assetTransferMethod` string values are mechanism-defined. `extra.assetTransferMethod` and `extra.paymentFlow` are protocol-reserved keys in `PaymentRequirements.extra`: clients and servers MUST interpret them as defined here rather than as opaque scheme-private fields.
 
-When a mechanism does not declare a flow, the default is `authorization`. A mechanism selecting a non-default flow MUST signal it to the client on the wire via `extra.paymentFlow`, so the client can reason about the trust model before paying — in particular, whether payment becomes final before the resource executes: if the resource handler fails, an `authorization` payment flow is never settled, whereas an `upfront` payment flow has already been committed.
+Schemes differ not only in how a payment is formed and validated, but in **when** settlement occurs relative to resource execution. A mechanism declares supported payment flows **per `assetTransferMethod`**, each with a default flow, plus a scheme-level default `assetTransferMethod` used when `extra.assetTransferMethod` is omitted. The resolved flow determines the ordering of the facilitator's read-only `/verify` (section 7.1) and state-committing `/settle` (section 7.2) around the resource server's execution of the protected request.
+
+Omitting `extra.assetTransferMethod` or `extra.paymentFlow` means the mechanism default when resolving. When the resolved payment flow is not `authorization`, `PaymentRequired` `accepts[].extra.paymentFlow` MUST be present so clients can reason about pre-handler fund commitment without scheme-specific knowledge (for example, distinguishing an SVM upto `escrow` default from an EVM upto `authorization` default). `authorization` MAY be omitted or explicit. Resource servers MUST reject unsupported `assetTransferMethod` / payment flow combinations.
 
 The following flows are defined:
 
--- a/specs/x402-specification-v2.md
+++ b/specs/x402-specification-v2.md
@@ -42,6 +42,8 @@ The x402 protocol follows a standard request-response cycle with payment integra
 3. **Payment Authorization Request**: Client submits a signed payment authorization in the subsequent request
 4. **Settlement Response**: Server verifies the payment authorization and initiates blockchain settlement
 
+This cycle describes the default `authorize` payment flow, in which the payment is verified before the resource executes and settled afterward. Schemes may declare other flows that settle before execution; see section 6.1 Payment Flow Models.
+
 **3. Protocol Components**
 
 The x402 protocol involves three primary components:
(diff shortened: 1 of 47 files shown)


specs/x402-specification-v2.md
```diff
@@ -273,61 +275,31 @@ Each scheme defines:
...
-For Solana (SVM), the `exact` scheme is implemented using `TransferChecked` for SPL tokens. Critical verification requirements include:
-
-- Enforcing a strict instruction layout (Compute Unit Limit, Compute Unit Price, TransferChecked)
-- Ensuring the facilitator fee payer does not appear in any instruction accounts and is not the transfer `authority` or `source`
-- Bounding compute unit price to mitigate gas abuse
-- Verifying the destination ATA matches the `payTo`/`asset` PDA and account existence rules
-- Requiring the transfer `amount` to exactly equal the `amount` specified in PaymentRequirements
-
-Full SVM details are specified in `specs/schemes/exact/scheme_exact_svm.md`.
+Invariant: at least one check — a verify before the resource or a settle before the resource — MUST run before the resource executes. The resource never executes with nothing checked.
```
CarsonRoscoe: nit: `a verify or settle before the resource` to avoid stuttering "before the resource"

specs/x402-specification-v2.md
```diff
@@ -415,12 +387,14 @@ Example with actual data:
 
 **7.2 POST /settle**
 
-Executes a verified payment by broadcasting the transaction to the blockchain.
+Durably commits payment state, typically by broadcasting a transaction to the blockchain. A settle need not be the final charge: it MAY establish an escrow, record a charge, or transfer funds, depending on the scheme and payment flow (see section 6.1 Payment Flow Models).
```
CarsonRoscoe: nit: Lets replace `blockchain` with `network ledger`

1. Moves language to be more neutral for fiat networks such as Cloudflare
2. Makes it clear that it does not HAVE to be committed to a blockchain, just that state is updated. e.g. the EVM `batch-settlement` voucher flow still handles `settle` locally at the application layer, but is not posting to the facilitator as its only managing a local ledger for keeping the in-memory state of voucher's cumulative amount up to date

specs/x402-specification-v2.md
```diff
@@ -415,12 +387,14 @@ Example with actual data:
...
 
 **Request:** Same structure as `/verify` endpoint (contains `paymentPayload` and `paymentRequirements`).
 
 > **Note**: While the request structure is identical, some payment schemes may assign different semantics to fields at settlement time versus verification time. For example, in the `upto` scheme, the `amount` field in `paymentRequirements` represents the maximum authorized amount at verification time, but the actual amount to settle at settlement time. See individual scheme specifications for details.
 
+> **Note**: `/settle` MAY be invoked more than once for a single payment (for example, the `escrow` flow settles a deposit before the resource executes and the final charge after). A scheme defining multiple settles MUST specify how the facilitator distinguishes them from payload content.
```
CarsonRoscoe: Should we call out that this distinguishing feature is likely to be server-lead?

e.g. in `escrow`, the client only signed one payload. If that same payload is passed through to both settles, there is no way for the facilitator to know whether it's operating the first or second settle operation based on the client provided payload alone.

Either:
1. Server is going to need to manipulate the payload, e.g. a scheme may need some sort of `step: deposit` vs `step: withdraw` property for the server to signal its intent
2. The facilitator is going to need to discern this distinction on its own via reading it's networks' ledger to understand if its on the first or second state

I imagine #1 is what all schemes we build are going to use for efficiency reasons, however technically #2 is valid

If you feel this is extraneous info and its fine to leave this ambiguous, thats fine

specs/x402-specification-v2.md
```diff
@@ -273,61 +275,31 @@ Each scheme defines:
...
-1. **Signature Validation**: Verify the EIP-712 signature is valid and properly signed by the payer
-2. **Balance Verification**: Confirm the payer has sufficient token balance for the transfer
-3. **Amount Validation**: Ensure the payment amount exactly matches the required amount
-4. **Time Window Check**: Verify the authorization is within its valid time range
-5. **Parameter Matching**: Confirm authorization parameters match the original payment requirements
-6. **Transaction Simulation**: Simulate the `transferWithAuthorization` transaction to ensure it would succeed
+The following flows are defined:
 
-**6.1.3 Settlement**
+| Flow                  | Ordering                                        | Description                                                                                                                              |
```
CarsonRoscoe, line 290: Validation is missing from this table, but exists in the code

Should we add it? Or is this a signal that it should be removed from code for now as not even batch-settlement's voucher verification intends to use it at this time?

typescript/packages/core/src/http/x402HTTPResourceServer.ts
```diff
@@ -726,6 +763,39 @@ export class x402HTTPResourceServer {
         },
       };
     }
+
+    const flow = this.ResourceServer.getPaymentFlow(paymentPayload, requirements);
```
CarsonRoscoe: The flow is re-resolving after the handler instead of carrying it through

The risk here is if the payload gets modified or the scheme juggles modes (not something we're prescribing, but technically possible), this can lead to bugs.

For example, if a scheme resolves `upfront` pre-handler and `authorization` post-handler, `phases.settleAfterHandler` is now true, the echo branch is skipped, and it settles a second time

No scheme trips this, but just calling it out. Might be worth resolving it once, putting the name in `CompletedSettlement` or a handle, and read it back?

(This is a nit, can be ignored if you disagree or see a flaw in my reasoning)

CarsonRoscoe (commented):
One risk is charged-but-failed requests carry no `PAYMENT-RESPONSE`. In every adapter (e.g. express), status >= 400 leads to `processSettlement` never being called, so the before-handler settle tx hash never reaches the client. Under `upfront` the payer paid and gets no receipt in this case.

The spec says the server bears no reversal obligation, which is fine, but the payer should still get the tx hash back imo

CarsonRoscoe (commented):
I think `beforeVerify` / `afterVerify` extension hooks are silently skipped for `upfront`/`escrow`. verifyPayment in x402ResourceServer.ts (line 976) is where  extension `beforeVerify` hooks can abort. Skip verify and any extension gating a request that way stops gating, with no warning and nothing in the changeset. Worth at least documenting / including in the changeset.

Could possibly worth firing a  gate hook independent of the verify call, but may not be necessary. Just documenting this behavior might be enough

CarsonRoscoe (commented):
If an unrecognized flow name is used, should we fail loud, or default to authorization?

Currently `resolvePaymentFlowPhases` is called in `x402HTTPResourceServer` outside of any try/catch. An unrecognized flow name from an untyped scheme returns undefined and throws a typeError out of `processSettlement` rather than defaulting

We should document what the default behavior is, and make the code match the decision

CarsonRoscoe (commented):
Test coverage is good on happy-paths, but lacks testing for pre-handler settle failure, the escrow cancel path, and MCP flows.

I also wonder if we should look into `@x402/core`'s integration tests, the ones that use the `Mock` network/scheme, and if we should enhance this suite to have `MockAuthorize`, `MockEscrow` and `MockUpfront`, and do testing around them, to prove out the happy paths are wired up for these payment flows that don't yet have implementing schemes

specs/x402-specification-v2.md
phdargen: Added

specs/x402-specification-v2.md
phdargen: Yeah couldnt think of a good description of the `validation` usecase here, so deleted it from the table. Lets remove also from sdk, can add it shall it be needed at some point

typescript/packages/core/src/http/x402HTTPResourceServer.ts
phdargen: Fixed: resolving it once, putting the name in CompletedSettlement

specs/x402-specification-v2.md
phdargen: removed validation from the sdk

phdargen:
> One risk is charged-but-failed requests carry no `PAYMENT-RESPONSE`. In every adapter (e.g. express), status >= 400 leads to `processSettlement` never being called, so the before-handler settle tx hash never reaches the client. Under `upfront` the payer paid and gets no receipt in this case.
> 
> The spec says the server bears no reversal obligation, which is fine, but the payer should still get the tx hash back imo

Agreed, good catch! Fixed

phdargen:
> If an unrecognized flow name is used, should we fail loud, or default to authorization?
> 
> Currently `resolvePaymentFlowPhases` is called in `x402HTTPResourceServer` outside of any try/catch. An unrecognized flow name from an untyped scheme returns undefined and throws a typeError out of `processSettlement` rather than defaulting
> 
> We should document what the default behavior is, and make the code match the decision

Better to fail loud imo

phdargen:
> I think `beforeVerify` / `afterVerify` extension hooks are silently skipped for `upfront`/`escrow`. verifyPayment in x402ResourceServer.ts (line 976) is where extension `beforeVerify` hooks can abort. Skip verify and any extension gating a request that way stops gating, with no warning and nothing in the changeset. Worth at least documenting / including in the changeset.
> 
> Could possibly worth firing a gate hook independent of the verify call, but may not be necessary. Just documenting this behavior might be enough

Good point, some extensions might rely on these hooks. So we should always try to run them and only gate the facilitator /verify http request. 

Changed to the following flow:
```
  // 1. beforeVerify hooks — always (abort / skip-with-result unchanged)
  // 2. if (verifyBeforeHandler) { facilitatorClient.verify(...) }
  // 3. afterVerify / onVerifyFailure — only when a VerifyResponse exists
  //    (facilitator result, or beforeVerify skip-with-result)
```

IkerAlus (commented):
_NOTE: Edited after checking `1e8cdcd`_

Hi @phdargen  I am very much in favour of this general idea: separating scheme / `assetTransferMethod` / `paymentFlow` pulls apart axes that were previously entangled, and the read-only-`verify` invariant is a good thing to have written down. I have not done a proper review (SDK files etc), but I have some general comments / questions on the spec:

**1. Does the facilitator `/verify` call run under `upfront`?**

From the hooks discussion above it reads like the facilitator `/verify` request is gated off for `upfront`/`escrow` (extension hooks still fire). For client-prepaid methods that leaves no obvious home for the checks that must happen before anything is committed, it is worth stating explicitly which is intended:

- `verify (read-only) → settle → work`, or
- proof verification folded into `settle`, with `/verify` unused for this flow.

Either works, but it determines where the checks live for every proof-based method. Note that verification for client-prepaid payments is read-only *by construction* — it only reads chain or backend state — so it is arguably the case that fits this PR's invariant best, and keeping it lets a facilitator reject a bad proof cheaply before engaging any settlement machinery.

**2. Intended scope of `assetTransferMethod: "transaction-proof"`.**

6.1 now states that allowed ATM values are mechanism-defined, which suggests per-mechanism naming. For the client-prepaid PRs there is a mayor shared surface (challenge binding, proof lifecycle/idempotency, finality declaration) while the proofs themselves differ (transaction id, invoice preimage, …). Also some of them settle synchronously and some with a backend phase between the client's payment and finality. Hence, is `transaction-proof` meant to be one name reused across mechanisms? or will it share verification core that each mechanism declares its own ATM against? Clarifying this will prevent that the incoming per-network schemes diverge on the common parts.

**3. Client-side behaviour on an unrecognized `paymentFlow`.**

I think here there is still unstated the client-side mirror of the fail-loud decision: what a client does with a flow name it does not recognize (a future flow, say). Server-side is fail-loud and now validated at construction. I suggest the mirror, "clients MUST NOT construct payment for a `paymentFlow` they do not recognize". Related: #2520 had "clients SHOULD prefer the post-handler-settlement option when both are offered for the same resource".

phdargen:
Thanks for the feedback @IkerAlus 

> **1. Does the facilitator `/verify` call run under `upfront`?**
> 
> From the hooks discussion above it reads like the facilitator `/verify` request is gated off for `upfront`/`escrow` (extension hooks still fire). For client-prepaid methods that leaves no obvious home for the checks that must happen before anything is committed, it is worth stating explicitly which is intended:
> 
> * `verify (read-only) → settle → work`, or
> * proof verification folded into `settle`, with `/verify` unused for this flow.
> 
> Either works, but it determines where the checks live for every proof-based method. Note that verification for client-prepaid payments is read-only _by construction_ — it only reads chain or backend state — so it is arguably the case that fits this PR's invariant best, and keeping it lets a facilitator reject a bad proof cheaply before engaging any settlement machinery.

The upfront flow would be just settle -> work, not verify -> settle -> work. 
Settle typically calls verify internally, so adding verify in front would be an unnecessary http roundtrip. 

For transaction proofs, I suppose settle would mark sth like a challengeID as consumed for replay protection which should be considered a durable write operation. Hence this would not be read-only

> 
> **2. Intended scope of `assetTransferMethod: "transaction-proof"`.**
> 
> 6.1 now states that allowed ATM values are mechanism-defined, which suggests per-mechanism naming. For the client-prepaid PRs there is a mayor shared surface (challenge binding, proof lifecycle/idempotency, finality declaration) while the proofs themselves differ (transaction id, invoice preimage, …). Also some of them settle synchronously and some with a backend phase between the client's payment and finality. Hence, is `transaction-proof` meant to be one name reused across mechanisms? or will it share verification core that each mechanism declares its own ATM against? Clarifying this will prevent that the incoming per-network schemes diverge on the common parts.
> 

This PR just declares assetTransferMethod as a reserved key in extra. 

assetTransferMethod names are defined per mechanism, eg assetTransferMethod=3009 only makes sense for EVM. 

For consistency I think it would be nice if we could define a set of properties an "transaction-proof" assetTransferMethod should fulfill at the network-agnostic exact scheme level: https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact.md
but this is outside the scope of this PR

> **3. Client-side behaviour on an unrecognized `paymentFlow`.**
> 
> I think here there is still unstated the client-side mirror of the fail-loud decision: what a client does with a flow name it does not recognize (a future flow, say). Server-side is fail-loud and now validated at construction. I suggest the mirror, "clients MUST NOT construct payment for a `paymentFlow` they do not recognize". Related: #2520 had "clients SHOULD prefer the post-handler-settlement option when both are offered for the same resource".

Added in [7817397](https://github.com/x402-foundation/x402/pull/3053/commits/7817397ffdc8195a43c9b2b69afcc337f661cc57)

IkerAlus:
hey @phdargen thanks for the prompt response and review, all three answers make sense.

- On the flow: agreed with `settle → work` with internal verification, and the proof consumption being a durable write resolves it for us. We will structure the NEAR Intents mechanism spec that way, with the checks under settle rather than as a separate `/verify` step.

- On network-agnostic `transaction-proof` properties assetTransferMethod should fulfill: Is it ok to propose them as a PR extending the current `scheme_exact.md`? Some thoughts on what this should specify:
  - what constitutes a proof and how it is presented in the payload
  - binding of the proof to specific `PaymentRequirements` (amount, recipient, freshness) and how the accepted amount relates to `amount`
  - single-use scope: the same proof rejected across resources and requirement sets, and concurrency behaviour
  - declaration of the finality boundary: which observable event counts as final
  - client recovery after a non-success terminal state 

  One design question for that PR: should conforming mechanisms reuse the literal `transaction-proof` ATM string, or use a mechanism-specific name that declares conformance to the property set? 

Last thing, not for this PR: `/supported` `kinds` still carries only `{x402Version, scheme, network}`, so a resource server cannot tell whether its facilitator implements the authorization methods it needs (an authorization and a client-prepaid proof can both be `exact` on the same network while requiring different settle paths). Same question applies to `eip3009` vs `permit2` today. I may open a separate issue rather than widen this one.

CarsonRoscoe (commented):
@IkerAlus 

Regarding the inability for facilitators to appropriately explain capabilities, I agree that there's a general gap in support declaration. I believe we should revisit [this PR](https://github.com/x402-foundation/x402/pull/2498) of mine to expand what can be expressed.

@phdargen 

PR is looking good to me, my main final question is `transaction-proof` as an `assetTransferMethod`.

There are two types of transaction submissions that can be handled for upfront:
1. (Preferred) signed transactions, similar to how Solana and other chains implement `exact` today. This is preferred as facilitators actually settle this on-chain, solving for double-spends in a stateless manner via leveraging the blockhains inherant security properties.

I don't think `transaction-proof` is the best name for this, since `transaction-proof` just sounds like a ledger-agnostic way of saying `transaction hash`

2. Transaction receipts, where `settle` is just verifying the transaction landed and the correct events were emitted showing funds were appropriately moved from the buyer to the seller.

This is the sub-ideal flow, as it requires servers and/or facilitators managing state to avoid double spends.

For this, `transaction-proof` is a fine name

My debate is:
1. Should these be two different `assetTransferMethod`'s, or one? (e.g. `signed-transaction` vs `transaction-proof`)
2. If not, what's a better universal name? e.g. `transaction`?

phdargen:
> My debate is:
> 
> 1. Should these be two different `assetTransferMethod`'s, or one? (e.g. `signed-transaction` vs `transaction-proof`)
> 2. If not, what's a better universal name? e.g. `transaction`?

Agreed on the distinction and I think they should be named differently like `signed-transaction` vs `transaction-proof`

Note that `transaction-proof` does not appear in any of the spec or sdk changes made in this PR. 
It is only mentioned in the PR description as a future use case motivating the different payment flows

CarsonRoscoe (approved):

CarsonRoscoe (commented):
Let's solve the capability expression separately. Merging this unblocks some workstreams already in progress

alftom:
Almost there!  A few issues I'm surfacing that were addressed in https://github.com/x402-foundation/x402/pull/1004 but not in this PR:

1. a single-use payment proof MUST be atomically claimed before resource execution; only one request may successfully claim it.  This prevents two simultaneous requests from claiming the same transaction.
2. do all `assetTransferMethod` values require a nonce?  There is no session and it's difficult to put the nonce in the transaction for EVM.  It's a server-side responsibility to make sure the transaction matches the service delivery and consume the ID to prevent replay.
3. for EVM at least, we need a signature or other proof that proves the client is authorized to use the transaction. This might be needed for other implementations as well.  Lots of ways to do this.  1004 lists one but there are others.
4. each implementation might want to define its own replay consumption identifier.  For EVM it would be `(network, txHash)` or something.
5. make it clear what criteria the facilitator uses for finality.  This could be part of the `/supported` discussion above.

CarsonRoscoe:
> 1. a single-use payment proof MUST be atomically claimed before resource execution; only one request may successfully claim it.  This prevents two simultaneous requests from claiming the same transaction.
> 2. do all `assetTransferMethod` values require a nonce?  There is no session and it's difficult to put the nonce in the transaction for EVM.  It's a server-side responsibility to make sure the transaction matches the service delivery and consume the ID to prevent replay.

Question on timing of "before resource execution". 

Per my comment about `signed-transaction` vs `transaction-proof`, an `upfront` payment of `assetTransferMethod: "signed-transaction"` would be broadcasted from the facilitator, submitted via a `POST /settle` before the the servers' route handler executes. So it would be broadcasted from the facilitator within the servers' execution, ensuring that settle response inherits the chains natural replay safety. This can be done statelessly

If `assetTransferMethod: "transaction-proof"`, it would have been executed before the middleware sees the request come in. This is where replay safety would be a concern. The transactions must have landed (cannot still be in flight) by the time the server processes it, and the server would be asking the facilitator to validate that payment did what was required to fulfill the payment requirement. This will require state to prevent double-spends, and is a harder problem to solve. I'm curious what chains would require this approach today, as the `signed-transaction` approach is inherently more secure.

As for requiring a nonce, I don't think we can prescribe that every scheme and every network require a nonce. Whether a nonce is included or not would have to be an implementation detail of a given mechanism (network/scheme tuple). 

> 3. for EVM at least, we need a signature or other proof that proves the client is authorized to use the transaction. This might be needed for other implementations as well.  Lots of ways to do this.  1004 lists one but there are others.

You make a good point that the `transaction-proof` style of submitting just a txn hash lacks proving that the intent of that payment was for this endpoint. The 1004 approach is one way, and good for making it a mechanism-driven way (e.g. every mechanism decides its own way to implement this). I could see the argument for pairing this with SIWX to create a pseudo-session, with the requirement being that the key used for SIWX must be the one who made the `upfront` payment, however that risks feeling heavy-handed by coupling the solution to an extension at the scheme level (something we want to avoid moving forward, as extension adoption is lower than scheme adoption, risking mismatch of support)

> 4. each implementation might want to define its own replay consumption identifier.  For EVM it would be `(network, txHash)` or something.

Good call. This is either a scheme spec detail, or we should come up with a universal one. The de-factor community standard way to do this is in web3 is using CAIP-10 paired with the txn hash, so for Ethereum it would be `eip155:1:0xtransactionHash`, so if we wanted to standardize it, that would be my pitch

PR #3055, opened 2026-08-05
fix(python): match wildcard routes containing a line feed

saneGuy:
## Description

The Python HTTP server compiles each route pattern to a regex where `*` becomes `.*?`. That regex is compiled with `re.IGNORECASE` but **not** `re.DOTALL`, so `.` does not match a line feed (`\n`).

A wildcard route such as `GET /api/premium/*` therefore fails to match a request path whose wildcard tail contains a decoded line feed. When the route misses, `_get_route_config()` returns `None`, `requires_payment()` returns `False`, and `_process_request_core()` returns `RESULT_NO_PAYMENT_REQUIRED` — the protected resource is served with no payment verification and no settlement.

This is the Python counterpart of the TypeScript fix in #3036 (which added the `s`/dotAll flag). The Python surface is narrower than the TS one: Python's `$` matches before a trailing `\n`, and Python's `.` only excludes `\n` (it already matches CR, U+2028, U+2029), so only a **line feed inside the wildcard tail** triggers the miss. The fix and its rationale are identical.

Fix: compile the route regex with `re.DOTALL` so a wildcard segment matches any character, including a line feed.

## Tests

`uv run pytest` from `python/x402/` — full suite **1855 passed, 65 skipped**. New file `tests/unit/http/test_route_matching.py` adds regression coverage that fails before the change and passes after:
- baseline: an ordinary sub-path of a wildcard route requires payment;
- guard: an unrelated path does **not** require payment (so the fix doesn't make everything match);
- decoded LF, percent-encoded LF, and LF-before-a-later-segment in the wildcard tail all still require payment.

`uvx ruff check` and `uvx ruff format --check` both clean.

## Checklist

- [ ] My commits are signed (will re-sign before merge per the template)

### DIFF
--- a/python/x402/changelog.d/+wildcard-line-feed-bypass.bugfix.md
+++ b/python/x402/changelog.d/+wildcard-line-feed-bypass.bugfix.md
@@ -1,0 +1,1 @@
+Fixed wildcard (`*`) route matching in the HTTP server when the wildcard tail contains a decoded line feed. Wildcard route regexes now compile with `re.DOTALL`, preventing protected routes from being missed before payment verification and settlement.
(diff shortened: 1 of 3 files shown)


phdargen (approved):
Thanks for the fix @saneGuy

PR #3056, opened 2026-08-05
docs(examples): add a recipient-screening client example

hypeprinter007-stack:
Adds a small, provider-agnostic example to `examples/typescript/clients/advanced` showing the canonical real-world use of `onBeforePaymentCreation`'s abort: **screen the payment recipient (`payTo`) before paying, and abort a bad address.**

The existing `hooks.ts` demonstrates the lifecycle hooks and notes you *can* abort; this shows the most common reason to — an autonomous payer shouldn't send funds to a sanctioned / known-malicious address without a human in the loop.

- New `screen-recipient.ts` mirroring the `hooks.ts` style; `onBeforePaymentCreation` reads `context.selectedRequirements.payTo`, calls a `screenRecipient(payTo)` check, and returns `{ abort, reason }` on a block.
- `screenRecipient` is a **stub** (inline deny-set) — the comment points at swapping in any sanctions/address-reputation API; the hook shape stays provider-agnostic.
- Wired into `index.ts` dispatch (`pnpm start screen-recipient`) + a README line.

Docs/example-only; no SDK behavior change, so no changelog fragment. Commit is signed.

### DIFF
--- a/examples/typescript/clients/advanced/README.md
+++ b/examples/typescript/clients/advanced/README.md
@@ -1,6 +1,6 @@
 # Advanced x402 Client Examples
 
-Advanced patterns for x402 TypeScript clients demonstrating builder pattern registration, payment lifecycle hooks, network preferences, and spend controls.
+Advanced patterns for x402 TypeScript clients demonstrating builder pattern registration, payment lifecycle hooks, network preferences, spend controls, and recipient screening (`onBeforePaymentCreation` → abort on a bad `payTo`).
 
 ```typescript
 import { x402Client, wrapFetchWithPayment } from "@x402/fetch";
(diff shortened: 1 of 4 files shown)


hypeprinter007-stack:
Rebased onto latest `main` (the branch was ~95 commits behind, which is what the stale Lint run was reacting to — not the example itself). Resolved the overlap with the newer `spend-controls` example: both patterns now coexist across the imports, JSDoc, usage list, `switch`, and README, and the `default:`-case "Available patterns" line lists both.

The example is lint-clean — `eslint . --ext .ts` and `prettier --check` both pass locally against this package's config and pinned tool versions. Could a maintainer approve/run the workflows on the new head when you get a chance? Thanks.

PR #3060, opened 2026-08-05
fix(mcp): withhold tool result when settlement resolves success:false

SashaMIT:
## Summary

`createPaymentWrapper`'s `settlePaymentResult` only treated a *thrown* error as settlement failure. `settlePayment` resolves (does not throw) on facilitator-side failures, so a `SettleResponse` with `success: false` fell through: `onAfterSettlement` fired and the paid tool result was returned with settlement metadata attached — **served for free**.

Concrete instance: the double-spend race. Two parallel calls present the same eip3009 authorization; both verify, both execute the tool, the first settles and consumes the nonce; the second's settle resolves `{ success: false, errorReason: nonce_already_used }` — and previously still received the paid content.

The HTTP resource server already checks `!settleResponse.success` and builds a failure response (`x402HTTPResourceServer`). This mirrors that check in the MCP wrapper: on `success: false` it returns the settlement-failure 402 result instead of the tool result (content withheld).

## Test plan

- [x] New unit test: handler executed, result `isError`, no paid content in body, no settlement metadata attached, `onAfterSettlement` not fired
- [x] 100/100 mcp tests pass (99 existing + 1 new); lint + format + build clean

Same repo-hygiene family as #3057/#3058/#3059 (settle-failure semantics parity across surfaces).

Made with Cursor

### DIFF
--- a/typescript/packages/mcp/src/server/paymentWrapper.ts
+++ b/typescript/packages/mcp/src/server/paymentWrapper.ts
@@ -425,6 +425,20 @@ async function settlePaymentResult(
       transportContext,
     );
 
+    // settlePayment resolves (does not throw) on facilitator-side failures:
+    // e.g. a double-spend race where the nonce was consumed between verify
+    // and settle. Returning the tool result here would serve paid content
+    // for free, so mirror the HTTP resource server and withhold it.
+    if (!settleResult.success) {
+      return createSettlementFailedResult(
+        resourceServer,
+        toolName,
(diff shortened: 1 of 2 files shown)


phdargen:
Thanks for the contribution, was meanwhile patched in https://github.com/x402-foundation/x402/pull/3246

SashaMIT:
Thanks Philippe. Glad this landed in #3246.

PR #3061, opened 2026-08-05
feat(mech/hedera): verify payee net credit from consensus record at settle

SashaMIT:
## Summary

The Hedera facilitator verified the **declared** transfer legs in the transaction body and treated receipt `SUCCESS` as full settlement, but never checked the **effective** balance changes. HTS tokens with custom fees settle `SUCCESS` while the payee receives less than the required amount:

- **Fractional custom fees** are deducted from the token receiver.
- **Fixed custom fees with net-of-transfers** are likewise deducted from what the receiver is credited.

So a resource server charging 1000 units of a custom-fee token gets underpaid (e.g. 950 + 50 to the fee collector) while `/settle` reports `success: true`. This is the same under-delivery class the eip3009 `Transfer` event check covers on EVM (and #3057–3059 extend to Permit2); Hedera has no events, but its consensus **record** already contains the effective transfers including assessed custom-fee legs.

## Fix

- `createHederaSignAndSubmitTransaction` now fetches the consensus record and returns the effective `hbarTransfers` / `tokenTransfers` alongside the transaction id (optional field — custom signers that omit it skip the check, documented on the interface).
- `settle()` confirms the payee's **net credit for the required asset** equals `requirements.amount` when transfers are present, failing with `settlement_transfer_amount_mismatch` (and the settled tx id, for traceability) otherwise.
- Fee legs denominated in a **different** token are ignored: fixed fees charged to the payer in another token don't reduce the payee's asset credit, so they must not false-fail.

## Test plan

- [x] 5 new settle tests: full credit → success; custom-fee underpayment → `settlement_transfer_amount_mismatch` with tx id; credit to a different account → mismatch; other-token fee legs → still success; signer without record support → success (guard preserved)
- [x] Default signer test updated to assert the record is fetched and flattened
- [x] `pnpm vitest run` 107/107, `pnpm run build`, `pnpm run format`, `pnpm run lint` all green

Made with Cursor

### DIFF
--- a/typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/hedera/src/exact/facilitator/scheme.ts
@@ -255,6 +255,30 @@ export class ExactHederaScheme implements SchemeNetworkFacilitator {
         requirements.network,
       );
 
+      // Consensus success only proves the transfer executed. When the signer
+      // returns the record's effective balance changes, confirm the payee's
+      // net credit equals the required amount — HTS tokens with custom fees
+      // (fractional fees are deducted from the receiver; fixed fees with
+      // net-of-transfers likewise) can otherwise settle SUCCESS while the
+      // payee received less than required.
+      if (settled.transfers != null) {
+        const assetLegs = isHbarAsset(requirements.asset)
+          ? settled.transfers.hbarTransfers
+          : (settled.transfers.tokenTransfers[requirements.asset] ?? []);
+        const netToPayTo = assetLegs
+          .filter(entry => hederaAccountIdsEqual(entry.accountId, requirements.payTo))
+          .reduce((sum, entry) => sum + BigInt(entry.amount), 0n);
+        if (netToPayTo !== BigInt(requirements.amount)) {
+          return {
+            success: false,
+            network: requirements.network,
+            payer: valid.payer || feePayer,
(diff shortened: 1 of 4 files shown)


Eras256:
Independently reproduced this against the currently published `@x402/hedera@2.25.0` (not just this branch) — built a real `TransferTransaction` and ran it through `ExactHederaScheme.verify()`/`settle()` with a signer mirroring the current `createHederaSignAndSubmitTransaction` (which only awaits `getReceipt()`, never `getRecord()`, confirmed by checking the installed dist). For a hypothetical 10% receiver-pays fractional-fee HTS token, `verify()` accepts the nominal transfer (fees aren't declared client-side) and `settle()` returns `success: true` unconditionally — no check of what `payTo` actually received. Matches the gap this PR describes exactly.

Read through the fix — the net-credit check against the consensus record's effective transfers, plus the `transfers` addition to `FacilitatorHederaSigner`, looks correct, and the added test cases (underpaid via custom fee, wrong recipient, different-token legs ignored, backward-compat when a signer doesn't return `transfers`) cover the real scenarios well. No concerns from my side — this looks ready.

SashaMIT:
Thanks for running it against published 2.25.0. That matches what we saw: getReceipt only, settle reports success without checking what payTo received. Glad the consensus-record net-credit check and the tests lined up for you.

PR #3064, opened 2026-08-06
fix(next): let withX402 take an explicit route pattern

### LINKED ISSUE #3019: [Bug]: `withX402()` hardcodes a `"*"` route pattern → `routeTemplate: ":var1"` fails Bazaar’s required `matches_resource` check, so Next.js resources are never indexed

AKnassa:
## Description

`withX402()` always registered its route as the wildcard `"*"`. With bazaar discovery extensions, that wildcard becomes an auto-generated `routeTemplate` (`":var1"`), which fails discovery validation's required `matches_resource` check — so Next.js resources using `withX402()` were never indexed, no matter how many payments settled.

This widens `withX402`'s second parameter from `RouteConfig` to `RoutesConfig`, so the route's real path can be given explicitly (e.g. `{ "/api/users/[id]": config }`) — the same convention `paymentProxy`, express, hono, and fastify already use. A bare `RouteConfig` still works unchanged: `@x402/core` applies its own `"*"` default for that shape, so existing callers are unaffected.

Closes #3019

## Tests

- Two new tests in `typescript/packages/http/next/src/index.test.ts` pin the routes shape reaching `x402HTTPResourceServer`: a pattern-keyed map passes through unchanged (failed before this fix), and a bare config passes through untouched.
- `pnpm test` from `/typescript`: full suite green.
- `pnpm lint:check` and `pnpm format:check` pass.

## Checklist

### DIFF
--- a/typescript/.changeset/next-withx402-route-pattern.md
+++ b/typescript/.changeset/next-withx402-route-pattern.md
@@ -2,4 +2,4 @@
 "@x402/next": minor
 ---
 
-`withX402` now accepts a `RoutesConfig`, so the route's path pattern can be given explicitly (e.g. `{ "/api/users/[id]": config }`) instead of always registering a wildcard route. This fixes bazaar discovery for Next.js resources: the hardcoded `*` pattern produced an auto-generated `routeTemplate` (`:var1`) that fails discovery validation, so resources were never indexed. Passing a bare route config remains supported and unchanged.
+`withX402` now accepts a `RoutesConfig`, so the route's path pattern can be given explicitly (e.g. `{ "/api/users/[id]": config }`) instead of always registering a wildcard route. This fixes bazaar discovery for Next.js resources: the hardcoded `*` pattern produced an auto-generated `routeTemplate` (`:var1`) that fails discovery validation, so resources were never indexed. Passing a bare route config remains supported and unchanged. When a pattern-keyed config matches no request path, the handler runs without payment and a warning is logged once per wrapper.
--- a/typescript/.changeset/next-withx402-route-pattern.md
+++ b/typescript/.changeset/next-withx402-route-pattern.md
@@ -1,0 +1,5 @@
+---
+"@x402/next": minor
+---
+
+`withX402` now accepts a `RoutesConfig`, so the route's path pattern can be given explicitly (e.g. `{ "/api/users/[id]": config }`) instead of always registering a wildcard route. This fixes bazaar discovery for Next.js resources: the hardcoded `*` pattern produced an auto-generated `routeTemplate` (`:var1`) that fails discovery validation, so resources were never indexed. Passing a bare route config remains supported and unchanged.
--- a/typescript/packages/http/next/README.md
+++ b/typescript/packages/http/next/README.md
@@ -60,17 +60,26 @@ const handler = async (_: NextRequest) => {
 export const GET = withX402(
   handler,
   {
-    accepts: {
-      scheme: "exact",
-      price: "$0.01",
-      network: "eip155:84532",
-      payTo: "0xYourAddress",
+    "/api/your-endpoint": {
+      accepts: {
+        scheme: "exact",
+        price: "$0.01",
+        network: "eip155:84532",
+        payTo: "0xYourAddress",
+      },
+      description: "Access to API endpoint",
     },
-    description: "Access to API endpoint",
   },
   server, // your configured x402ResourceServer
 );
 ```
+
+Keying the configuration by the route's path pattern (dynamic segments like
+`/api/users/[id]` are supported) is preferred over passing a bare route config.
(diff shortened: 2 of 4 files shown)


phdargen:
Thanks for the contribution @AKnassa , this has meanwhile been patched in https://github.com/x402-foundation/x402/pull/3309

PR #3066, opened 2026-08-06
docs(spec): add verification-gated release policy profile for auth-capture (#3065)

### LINKED ISSUE #3065: [Feature]: verification gated release policy for auth-capture (a pact-escrow profile)
### Problem or use case

auth-capture deliberately defines the hold and stays silent on the release: capture or void is a discretionary act of the captureAuthorizer. That's the right layering, but it leaves the two parties with no protocol-level way to (1) agree upfront what satisfactory delivery means for this payment, (2) verify delivery against that agreement, and (3) dispute a wrong release decision. Today the client trusts the captureAuthorizer's judgment, off-protocol.

That blocks the use case several open issues are circling: agent-to-agent task delegation where payment releases on verified completion (#1247), dispute handling when verification is contested (#2001), a correctness and dispute layer for what settlement can't see (#2887), and canonical delivery proof at settlement (#1195, #2833).

Concrete example: an agent pays $195 for a data cleaning job under auth-capture with autoCapture:false. What decides capture vs void? Today, nothing the counterparties agreed to in a machine checkable form. The acceptance tests passing or failing should be that decision, and both sides should have committed to those tests before work started.

### Proposed solution

A release-policy profile over auth-capture. No changes to the scheme itself, the profile specifies what a conformant captureAuthorizer executes:

  1. Before authorization, buyer and seller co-sign a Verifiable Task Contract: a JSON object fixing task scope by hash, price, verification method, challenge window, and the seller's bond. Canonicalized with JCS (RFC 8785), signed with JWS (RFC 7515).

  2. Release becomes mechanical. Verification passes: capture. Verification fails: void. On the optimistic path, capture fires at challenge-window expiry unless a fraud proof lands; the bond funds slashing, challenger cost reimbursement, and a capped bounty, because slashing alone doesn't fix the verifier's dilemma. Verification is graded per contract: deterministic re-execution, TEE attestation (EAT/RATS), zkML, or staked arbitration.

  3. At settlement, both parties co-sign a Work Attestation carrying the contract hash and outcome: a delivery receipt in the sense #1195 and #2833 are after, except only emittable through a bonded, escrowed contract, which is what makes accumulated receipts expensive to fake.
 
The full lifecycle is specified as an IETF Internet-Draft, draft-laxsharma-pact-00 (individual submission, Experimental). One honesty note: Section 8 of the -00 currently expresses the x402 binding as a standalone pact-escrow scheme, written against the escrow and deferred lineage that was closed in May (#873, #426). Reshaping that binding into a profile over the merged auth-capture machinery is exactly the change I'd make in -01, and this issue is me proposing it here first. Schemas and worked examples whose hashes verify in CI: https://github.com/pact-spec/spec . Draft: https://datatracker.ietf.org/doc/draft-laxsharma-pact/

Concrete ask: (a) is this better shaped as a profile doc under specs/schemes/auth-capture/ or an extension under specs/extensions/? (b) If maintainers see a path, I'll draft that spec doc as a PR. It's a -00, published to be demolished; sharp edges welcome.

jihadMo:
Resolves #3065.

Adds the verification-gated release policy profile specification under specs/schemes/auth-capture/pact-release-policy.md defining contract co-signing, mechanical verification settlement, and work attestation.

### DIFF
--- a/specs/schemes/auth-capture/pact-release-policy.md
+++ b/specs/schemes/auth-capture/pact-release-policy.md
@@ -1,0 +1,30 @@
+# Pact Release Policy Profile for x402 Auth-Capture
+
+This specification defines the Verification-Gated Release Policy profile over the x402 `auth-capture` payment scheme (#3065).
+
+## Overview
+
+The `auth-capture` scheme defines hold and capture mechanics. This profile specifies the mechanical execution rules for `captureAuthorizer` based on verifiable task completion.
+
+```
++----------------+      Co-sign Contract (RFC 8785 / JWS)      +----------------+
+|  Buyer Agent   | <=========================================> |  Seller Agent  |
++----------------+                                             +----------------+
+        |                                                              |
+        | 1. Auth Hold ($195)                                          | 2. Submit Work
+        v                                                              v
++-------------------------------------------------------------------------------+
+|                        x402 CaptureAuthorizer Engine                         |
+|  - Runs Verification Suite (Deterministic / TEE / zkML)                      |
+|  - Verification PASS -> Capture Payment                                      |
+|  - Verification FAIL -> Void Hold & Slash Bond                                |
++-------------------------------------------------------------------------------+
+```
+
+## Lifecycle Steps
+
+1. **Verifiable Task Contract:** Buyer and seller co-sign a contract fixing task scope hash, payout, verification method, and challenge window.
+2. **Mechanical Settlement:** On task submission, verification suite runs:
+   - **PASS:** Auto-capture released to seller.
+   - **FAIL:** Hold voided and seller bond slashed.
+3. **Work Attestation:** Both parties co-sign a Work Attestation upon settlement recording contract hash and completion status.


laxsharma:
Thank you for this, genuinely. You turned the issue into a document faster than I would have, and you got a framing right that my own Internet-Draft gets wrong: this is a release policy over auth-capture, not a new scheme. Section 8 of draft-laxsharma-pact-00 calls pact-escrow "an x402 payment scheme", and your title is the more accurate description. I'm changing the draft to match.

Before this goes further I owe you and any reviewer an update about the thing it builds on. After the review comment in #3065 I ran a full adversarial audit of the draft, and the settlement economics do not close. The findings are logged publicly at [pact-spec/spec#1](https://github.com/pact-spec/spec/issues/1). The short version is that the bond in the -00 is a fine rather than collateral, so a defrauded buyer recovers nothing; the amount released before verification exceeds the bond, so defection dominates; and challenger reimbursement is capped by the bond while re-execution costs roughly what execution costs, so the reimbursement requirement is unsatisfiable. I would rather say that here than have it found after the profile is merged.

Three things in this document specifically, independent of that redesign:

The bond has no instrument on this rail. Step 2 FAIL says "Hold voided and seller bond slashed", but auth-capture has no payee-side deposit. Void returns funds to the payer, refund returns captured funds to the payer, reclaim is payer-initiated. Nothing in the scheme can move the seller's own collateral, so the slash half of that step cannot execute. Any workable profile needs the bond named as a separate instrument, and we should say so explicitly rather than imply that auth-capture covers it.

The engine running the verification suite is the problem the issue raised. Your diagram has the CaptureAuthorizer Engine run verification and then capture or void. That collapses two properties a reviewer separated in #3065: how strongly the work is checked, and who checks it. If the captureAuthorizer both verifies and settles, it holds exactly the discretion this profile exists to remove, and a deployment could satisfy the profile with the seller's own designee verifying the seller's work. I think the profile needs the verifier named in the contract and required to be neither party, with the captureAuthorizer acting only on a verification outcome it did not produce.

An open question for the maintainers that blocks the checkpoint model: does auth-capture permit repeated partial capture against a single authorization? The words partial, incremental, multiple, and captureAmount do not appear in either scheme file. If it does not, then a contract with progress checkpoints needs one authorization per checkpoint and this document should say so.

One thing your document does better than the -00 and worth keeping: void returning funds to the payer is the correct restitution behaviour, and the -00 explicitly forbade paying "any party to the dispute", which meant the buyer got nothing. Your version is right and mine was wrong.

I'd rather build this with you than in parallel. My suggestion is to keep the PR open and narrow it to the parts that survive (the profile framing, the lifecycle shape, the co-signed contract fixing scope and verification method before work starts), and add the verifier and bond-instrument pieces once the -01 settles them. I'm targeting mid-September for that revision and will post the relevant sections here first. If you'd rather drive it, say so and I'll review rather than lead.

jihadMo:
Thanks @laxsharma! I really appreciate the detailed review and feedback. Keeping the PR open to refine the profile framing, lifecycle shape, and co-signed contract scope while aligning with the upcoming -01 revision sounds like a great path forward. Happy to collaborate on driving this together!

laxsharma:
@jihadMo  Sorry for the long silence on this one. I said I'd come back once the -01 was out, and then the -02 came along and changed the shape of the thing.

The -02 went up on 17 September: https://www.ietf.org/archive/id/draft-laxsharma-pact-02.html. The main change is that the draft no longer talks about money at all. It doesn't reserve a pact-escrow name, doesn't name an x402 scheme, and leaves who holds or releases funds to a separate profile that lives outside the document. So a release policy over auth-capture, the thing you wrote here, is now meant to live exactly where you put it: outside the draft. You had that framing right in August and I didn't.

Two small things in the text don't match any more. The outcome record is signed by the settlement side only, not by both parties. If the party that lost money had to co-sign, it could simply refuse, and the record could never show a loss. And "slash bond" is now a rule inside a profile, not part of the protocol.

Rather than keep patching this against a moving target, I think the cleanest thing is to close it. If you'd like to write an auth-capture release policy in your own name at some point, I'd be glad to read it. Thanks again for turning the issue into a document that fast.

PR #3069, opened 2026-08-06
fix(java): buffer response and settle before delivering content

### LINKED ISSUE #3068: [Bug]: [java] PaymentFilter delivers content before settlement premium resources served without guaranteed payment
### What happened?

PaymentFilter.doFilter() calls chain.doFilter() (which writes the full HTTP response to the client) before calling facilitator.settle(). If settlement fails for any reason network error, RPC timeout, facilitator outage the premium content has already been delivered and the seller receives no payment.

Once chain.doFilter() completes, response.isCommitted() is typically true. The respond402() fallback on settlement failure is effectively unreachable.

The TypeScript/Express middleware handles this correctly by monkey-patching res.write/res.end to buffer the response, settling first, and only flushing the buffer to the client after a successful settle. The Java SDK has no such buffering.

### Steps to reproduce

1. Set up a Spring Boot app with PaymentFilter protecting an endpoint
2. Configure a facilitator that intermittently fails settle() (e.g. simulate a network timeout)
3. Send a valid payment header (verify passes)
4. Observe: chain.doFilter() runs and the protected response is sent to the client
5. Observe: settle() then fails but the response is already committed
6. Result: client received premium content, seller received no payment

### Environment

java/x402-java SDK
PaymentFilter.java doFilter() method
Affects all servlet containers (Tomcat, Jetty, Undertow)
Reference (correct behavior): typescript/packages/http/express/src/index.ts "payment-verified" case

Sertug17:
## Summary

`PaymentFilter` called `chain.doFilter()` before `settle()`, meaning premium content was delivered to the client before payment was confirmed on-chain. If settlement failed, the content was irrecoverably served for free.

## Root cause

```java
// BEFORE (broken order):
chain.doFilter(req, res);   // ← content sent to client NOW
settle(header, ...);        // ← payment confirmed LATER (too late)
```

Unlike the TypeScript/Express middleware which buffers the response by monkey-patching `res.write`/`res.end`, the Java filter wrote directly to the real `HttpServletResponse` with no buffering.

## Fix

Introduce `BufferingResponseWrapper` (`HttpServletResponseWrapper`) that captures all handler output in memory. Settlement is attempted first; the buffer is flushed to the client only on success. On failure the buffer is discarded and a fresh 402 is returned — the client never sees the premium content.

```java
// AFTER (correct order):
BufferingResponseWrapper buffered = new BufferingResponseWrapper(response);
chain.doFilter(req, buffered);   // ← captured in memory
settle(header, ...);             // ← payment confirmed first
buffered.copyTo(response);       // ← only now sent to client
```

Also adds `Cache-Control: no-store` to `respond402()` to prevent CDN/proxy caching of 402 responses containing wallet addresses and payment nonces.

## Related

Fixes #3068

### DIFF
--- a/java/.changes/unreleased/fix-java-settle-before-deliver.yaml
+++ b/java/.changes/unreleased/fix-java-settle-before-deliver.yaml
@@ -1,0 +1,5 @@
+kind: fix
+body: >
+  Buffer HTTP response in PaymentFilter and settle payment before flushing to
+  client, preventing premium content from being delivered when settlement fails.
+  Also adds Cache-Control: no-store to 402 responses to prevent CDN caching.
--- a/java/src/main/java/org/x402/server/BufferingResponseWrapper.java
+++ b/java/src/main/java/org/x402/server/BufferingResponseWrapper.java
@@ -1,0 +1,52 @@
+package org.x402.server;
+
+import jakarta.servlet.ServletOutputStream;
+import jakarta.servlet.WriteListener;
+import jakarta.servlet.http.HttpServletResponse;
+import jakarta.servlet.http.HttpServletResponseWrapper;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.PrintWriter;
+
+class BufferingResponseWrapper extends HttpServletResponseWrapper {
+
+    private final ByteArrayOutputStream buffer = new ByteArrayOutputStream(4096);
+    private final ServletOutputStream   stream  = new BufferingOutputStream(buffer);
+    private       PrintWriter           writer;
+    private       int                   status  = SC_OK;
+
+    BufferingResponseWrapper(HttpServletResponse response) {
+        super(response);
+    }
+
+    @Override public void setStatus(int sc) { this.status = sc; super.setStatus(sc); }
+    @Override @SuppressWarnings("deprecation")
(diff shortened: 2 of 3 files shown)


Sertug17:
Happy to address any review feedback. This fixes a critical deliver before settle ordering bug. @phdargen

Sertug17:
Did you have a chance to review it? @phdargen

phdargen:
Thanks for the contribution @Sertug17 and identifying this bug in [(#3068).](https://github.com/x402-foundation/x402/issues/3068)

We merged the more complete patch in https://github.com/x402-foundation/x402/pull/3074 here because:
• It compiles on current main (jakarta.servlet-api 6.1.0) -> the `setStatus(int, String)` override in your PR fails to build
• It includes an embedded-Jetty regression test that reproduces the bug on pre-fix code (200 + leaked body) and passes with buffering
• It fixes a masked integration-test false positive and has a more complete wrapper (isCommitted(), flushBuffer(), writer/stream mutual exclusion)

Note: the Java SDK is legacy v1 only and effectively frozen. We no longer actively maintain this package or accept PRs in general, with the exception of security patches. If you are interested building on a Java stack, please consider community maintained Java sdks instead, see eg https://docs.x402.org/dev-tools/third-party-sdks

PR #3071, opened 2026-08-07
fix(python): cap solana below 0.40 in the svm extra

notorious-d-e-v:
`pip install x402[svm]` currently produces an installation in which nothing under `x402.mechanisms.svm` can be imported.

solana 0.40 removed the sync client (`solana.rpc.api`) and `solana.rpc.types.TxOpts`. `mechanisms/svm/signers.py` and `mechanisms/svm/utils.py` import both at module scope, and the `svm` extra declared `solana>=0.36.0` with no ceiling — so a fresh install on Python 3.11+ resolves 0.40.1 and every svm import raises. The `except ImportError` handler then advises `pip install x402[svm]`, which is the command the user just ran.

## Reproduction

On `main`, Python 3.12:

```
$ pip install 'x402[svm]'                        # resolves solana 0.40.1
$ python -c "import x402.mechanisms.svm.exact"
ImportError: SVM mechanism requires solana packages. Install with: pip install x402[svm]
```

Same tree with this patch applied: resolves solana 0.39.0, imports cleanly.

## Why CI does not catch it

`uv sync --all-extras --dev` resolves from `uv.lock`, which pins solana **0.36.12** — five minor releases behind what `pip install` hands a user. The lock has never seen 0.37+, so no amount of test coverage on the current setup would surface this. A lock-free `pip install x402[svm] && python -c "import x402.mechanisms.svm"` smoke job would; happy to open that separately if you want it.

## Scope

- Verified every version the new range admits — 0.36.12, 0.37.1, 0.38.0, 0.39.0 — imports all of `mechanisms/svm`.
- `uv.lock` resolution is unchanged: 0.36.12 already satisfied the new bound, so only the two recorded specifiers move.
- `1850 passed, 65 skipped` locally.
- `solders` is unaffected and left alone.

The cap is a stopgap. The real fix is moving those two modules off the removed sync client, which is an async-vs-solders-direct call I did not want to make on your behalf — glad to follow up if you name a direction.

---

Written with AI assistance. The command output, resolved versions and range results above are real, not generated.


phdargen:
Hi @notorious-d-e-v, there is some work in progress to migrate to >0.40 in https://github.com/x402-foundation/x402/pull/2862, would that work for you?

notorious-d-e-v:
Yep that works

PR #3072, opened 2026-08-07
spec: propose MCP URL elicitation transport v3

ethanniser:
As the author of the current x402 over MCP spec I realized that since I wrote that, [MCP Elicitation](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation) has progressed a lot. I think the current spec is quite jank in comparison to URL based eliciation, which I think actually fits x402 very well.

This is a AI-written version of what a v3 spec could look like. Can dig into it deeper if people agree this is the right direction to go

---

## Summary

Adds a proposed v3 MCP transport that composes MCP URL elicitation with the existing x402 HTTP transport.

MCP handles user consent, continuation state, and resuming the protected operation. All payment protocol messages remain on the HTTP leg.

## Design

- Signals payment requirements using MCP `InputRequiredResult` and URL mode elicitation.
- Requires MCP protocol version `2026-07-28` or later.
- Uses the existing x402 HTTP v2 transport without defining alternative payment encodings or headers.
- Converts successful HTTP settlement into a single-use entitlement bound to the authenticated MCP principal and exact operation.
- Resumes the original MCP request using opaque `requestState`.
- Returns an ordinary terminal MCP result without repeating settlement data in MCP metadata.
- Defines identity binding, replay protection, idempotency, cancellation, and post-settlement failure behavior.

This keeps `PaymentRequired`, `PaymentPayload`, and `SettlementResponse` objects out of MCP messages and allows existing x402 HTTP clients, wallets, facilitators, and middleware to be reused.

### DIFF
--- a/specs/transports-v3/mcp.md
+++ b/specs/transports-v3/mcp.md
@@ -1,0 +1,365 @@
+# Transport: MCP (URL Elicitation)
+
+## Summary
+
+This transport composes MCP URL elicitation with the existing x402 HTTP
+transport. MCP elicits consent to open a payment URL and resumes the protected
+operation after payment; all payment protocol messages use the HTTP transport.
+
+The MCP client never receives or submits `PaymentRequired`, `PaymentPayload`,
+or `SettlementResponse` objects. The elicited HTTP resource MUST implement the
+[x402 HTTP transport](../transports-v2/http.md) without modification.
+
+This document defines version 3 of the x402 MCP transport. The elicited HTTP
+resource continues to use x402 protocol version 2. Transport version and core
+x402 protocol version are independent.
+
+## Requirements
+
+This transport requires MCP protocol version `2026-07-28` or later and the MCP
+multi round-trip request (MRTR) pattern. It applies to MCP operations that may
+return an `InputRequiredResult`:
+
+- `tools/call`
+- `resources/read`
+- `prompts/get`
+
+Examples use `tools/call`; the same rules apply to the other supported
+operations.
+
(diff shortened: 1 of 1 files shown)


specs/transports-v3/mcp.md
```diff
@@ -0,0 +1,365 @@
...
+    Client->>Server: tools/call (id: 1)
+    Server-->>Client: InputRequiredResult(URL, requestState)
+    Client->>User: Show URL and request consent
+    User-->>Client: Accept
+    Client->>Browser: Open elicitation URL
+    Browser->>Server: x402 HTTP request
+    Server-->>Browser: x402 HTTP payment required
+    Browser->>Server: x402 HTTP paid request
+    Server->>Server: Settle and record entitlement
+    Server-->>Browser: x402 HTTP settlement response
```
erikreppel, line 90: The current design has the nice property of allowing for servers to be stateless, introducing state to the server is generally something I think we should try to avoid unless there are clear benefits.

The first benefit seems to be not requiring the x402 types to be included in the MCP payloads, I'm curious what else there is.

erikreppel (commented):
Good doc!

This design makes a ton of sense to me in the elicitation paradigm, I think the question in my mind is does opting into elicitation and the required server side state that accompanies it provide enough benefit to justify the incremental server requirement.

I do think we should revisit the MCP transport design with stateless MCP coming out

ethanniser:
@erikreppel agreed- made a fresh pr for a stateless formal "mcp extension": #3106

PR #3074, opened 2026-08-07
fix(java): buffer response body until settlement succeeds in PaymentFilter

### LINKED ISSUE #3068: [Bug]: [java] PaymentFilter delivers content before settlement premium resources served without guaranteed payment

rileybuilds:
Fixes #3068.

## Problem

`PaymentFilter.doFilter()` calls `chain.doFilter()` directly against the real `HttpServletResponse`, so the protected handler's output is written — and often committed — to the client **before** `facilitator.settle()` runs. If settlement then fails (network error, RPC timeout, facilitator outage), the buyer has already received the paywalled content while the seller never gets paid. The existing `respond402()` fallback on settlement failure is only reachable when `response.isCommitted()` happens to still be `false`, which in practice it usually isn't once the handler has written a response body.

The TypeScript/Express middleware in this same repo (`typescript/packages/http/express/src/index.ts`) already handles this correctly by buffering the response and only flushing it after a successful settlement — the Java SDK never got the equivalent fix.

## Fix

- `BufferedServletOutputStream` / `BufferingHttpServletResponseWrapper` (new, `org.x402.server`): a minimal in-memory `HttpServletResponseWrapper` that buffers everything written via `getOutputStream()`/`getWriter()` instead of sending it to the client. Status codes and headers set through the wrapper still forward to the real response immediately (safe, since nothing is ever written to the real output stream through this wrapper), matching how the TS version treats headers vs. body.
- `PaymentFilter.doFilter()`: runs `chain.doFilter()` against the buffered wrapper instead of the real response. The buffered body is only flushed to the real response after settlement succeeds (or the handler itself returned a 4xx/5xx, in which case it's flushed as-is since there's nothing to settle). On settlement failure or a settlement exception, the buffered body is discarded and `respond402()` is sent instead — deterministically, not dependent on whether the container had already committed the response.

## Testing

- Added `FilterSettlementFailureIntegrationTest` (new, embedded Jetty): a facilitator stub that always verifies but always fails to settle, hitting a servlet that would otherwise deliver paywalled JSON. **Verified this test actually reproduces the bug**: run against the pre-fix code it fails with `expected 402 but was 200` and a leaked response body; with the fix it passes (`402`, body never delivered).
- Fixed `FilterIntegrationTest`'s stub facilitator, which returned a default (`success = false`) `SettlementResponse` from `settle()` — its `validHeaderGets200` test was only passing *because of* the pre-fix bug (content already delivered before the always-failing settle() result could matter). Now returns a proper success response.
- Updated `PaymentFilterTest`'s `chain.doFilter()` verifications to accept an `HttpServletResponse` argument rather than asserting reference-equality against the raw mock, since the filter now passes a wrapper.
- Full suite: 42/42 passing (41 pre-existing + 1 new), verified locally on Java 17 / Maven.

No behavior change for the free-path (unprotected routes) or the pre-payment-verification error paths (malformed header, verification failure, etc.) — only the payment-verified → business-logic → settle path is buffered.

### DIFF
--- a/java/src/main/java/org/x402/server/BufferedServletOutputStream.java
+++ b/java/src/main/java/org/x402/server/BufferedServletOutputStream.java
@@ -1,0 +1,36 @@
+package org.x402.server;
+
+import jakarta.servlet.ServletOutputStream;
+import jakarta.servlet.WriteListener;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+/** In-memory {@link ServletOutputStream} backing {@link BufferingHttpServletResponseWrapper}. */
+final class BufferedServletOutputStream extends ServletOutputStream {
+
+    private final ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+
+    @Override
+    public void write(int b) throws IOException {
+        buffer.write(b);
+    }
+
+    @Override
+    public void write(byte[] b, int off, int len) throws IOException {
+        buffer.write(b, off, len);
+    }
+
+    @Override
+    public boolean isReady() {
+        return true;
+    }
+
+    @Override
+    public void setWriteListener(WriteListener writeListener) {
+        // Buffering is always synchronous; async write-readiness notifications don't apply.
+    }
+
+    byte[] toByteArray() {
+        return buffer.toByteArray();
+    }
+}
--- a/java/src/main/java/org/x402/server/BufferingHttpServletResponseWrapper.java
+++ b/java/src/main/java/org/x402/server/BufferingHttpServletResponseWrapper.java
@@ -1,0 +1,67 @@
+package org.x402.server;
+
+import jakarta.servlet.ServletOutputStream;
+import jakarta.servlet.http.HttpServletResponse;
+import jakarta.servlet.http.HttpServletResponseWrapper;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
+
+/**
+ * Buffers the response body written by downstream filters/servlets instead of sending it to the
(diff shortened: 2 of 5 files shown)


phdargen (approved):
Thanks for the patch @rileybuilds , LGTM. 

Please note that the Java SDK is legacy v1 only and effectively frozen. We no longer actively maintain this package nor accept PRs in general, with the exception of security patches. If you are interested building on a Java stack, please consider community maintained Java sdks instead, see eg https://docs.x402.org/dev-tools/third-party-sdks

PR #3078, opened 2026-08-07
fix: [Feature]: earn extension for discovering endpoints that pay agents (#3076)

### LINKED ISSUE #3076: [Feature]: earn extension for discovering endpoints that pay agents
### Problem or use case

Every discovery mechanism in production indexes the sell side. `bazaar` catalogs a
service when payment settles from the agent to the endpoint, and its `info` block
describes what a paid call costs and expects. On-chain indexers derive the same
direction from settlement flow.

There is no schema for the other direction. A resource server that **pays an agent
for completed work** cannot declare that in a form any catalog can index, so an
agent looking for paid work has no protocol-level way to find it. Today the only
route is a human wiring one venue into one agent by hand.

We hit this running a support-ticket bounty board that settles USDC on Base. Three
independent agents have found it and been paid, but every one of them arrived
through a human, because there is no field that says "this endpoint pays you."

### Proposed solution

Add `specs/extensions/earn.md`, following the v2 `info` + `schema` pattern used by
the other extensions. A resource server declares what work is available, what it
pays, what an attempt costs, what triggers payment, and its settled history.

Three fields are required specifically to make a declaration falsifiable:

- **`history`**, because any server can claim to pay agents. A claim with no
  settled-payment evidence is unfalsifiable, and agents acting on unfalsifiable
  claims lose money and leave the market. Servers with no history omit the
  extension rather than publishing zeros.
- **`latestPaymentTx`** when `paidCount > 0`, so the claim is checkable on chain
  without trusting the publisher.
- **`attemptCostUsd`**, because most earn-side venues charge something to attempt,
  and an undeclared attempt cost is how an agent discovers a negative expected
  value after paying it.

Anything a server could simply assert about its own quality is deliberately absent.
The spec also recommends catalogs rank on settled activity rather than declared
availability, since `open` and `openValueUsd` are self-asserted and `paidCount` is
not.

Reference implementation is live at https://deskcrew.io/.well-known/x402 under
`extensions.earn`. Human approval is the only payment trigger and the worker keeps
85%. Every `history` field is computed from settled payments and recorded verdicts:
currently 6 payouts totalling $3.00 across 3 distinct wallets, latest settlement
0x08fb51b2fc5123c76f37ac0f8f72edb337b90afe24a52dcd35e0958195b65a8c on Base.

That is a small sample, and I would rather say so than oversell it. What makes it
worth standardising is that the round trip works: an agent found the board from
this data alone, evaluated it with no human involved, paid to compete, and was paid.

Open questions I would want feedback on:

1. Should `history` require a bounded window so a dormant server cannot publish
   flattering lifetime figures?
2. Should `attemptCostUsd` support a range for venues that price per task?
3. Is `earn` the right key, or should this sit under a broader `work` namespace
   that could also express unpaid or reputation-only tasks?

Happy to revise the shape, narrow the scope, or split it. The gap matters more to
us than this particular spelling of it.

FrancescoGSR77:
Closes #3076

Resuelve el issue https://github.com/x402-foundation/x402/issues/3076 (bounty $3.00) — generado por el worker autónomo.

### DIFF
--- a/specs/extensions/earn.md
+++ b/specs/extensions/earn.md
@@ -1,0 +1,354 @@
+# Extension: `earn`
+
+## Summary
+
+The `earn` extension enables **demand-side discovery** for x402: a resource
+server that **pays an agent for completed work** declares what work is
+available, what it pays, what an attempt costs, what triggers payment, and its
+settled-payment history. Catalogs and discovery services can then index the
+demand side the same way the `bazaar` extension indexes the supply side.
+
+Where `bazaar` catalogs a service when payment settles **from the agent to the
+endpoint** (selling access to a resource), `earn` is the schema for the other
+direction: an endpoint that settles payment **to the agent** for completed
+work. An agent looking for paid work can find it at the protocol level, without
+a human wiring one venue into one agent by hand.
+
+A reference implementation is live at
+<https://deskcrew.io/.well-known/x402> under `extensions.earn` (a
+support-ticket bounty board that settles USDC on Base).
+
+---
+
+## Design Goals: Falsifiability
+
+Earn-side declarations are inherently self-asserted. Any server can *claim* to
+pay agents; a claim with no settled-payment evidence is unfalsifiable, and
+agents acting on unfalsifiable claims lose money and leave the market. Three
+fields are required specifically to make a declaration falsifiable:
+
+- **`history`** — required, because a claim that a server pays agents is
+  unfalsifiable without settled-payment evidence. Servers with no settled
+  history omit the extension rather than publishing zeros.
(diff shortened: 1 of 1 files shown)


FrancescoGSR77:
Hi! All commits in this PR have been updated with valid GPG signing and verified email identity matching DCO. The PR is ready for review. Thank you!

FrancescoGSR77:
Hi! Just checking in on this PR. All commits are now GPG-signed with verified email identity matching DCO requirements. Please let me know if any additional changes or tests are needed to proceed with the review. Thank you for your time!

PR #3080, opened 2026-08-07
fix(permit2): verify Transfer event in settle receipts (exact + upto)

SashaMIT:
## Description

A payer using a fee-charging (fee-on-transfer) or otherwise non-conforming token can underpay while the facilitator reports full success: every Permit2 settle path (exact and upto) trusted `receipt.status` alone, and the proxy's `settle()` does not revert when the token delivers less than the required amount.

**What changes:** `waitAndReturnSettleResponse` now takes the expected Transfer fields `{ token, from, to, value }` and requires a matching ERC-20 `Transfer` event in the receipt before reporting success, reusing the existing `verifyEip3009TransferEvent` helper and the same `invalid_exact_evm_transfer_event_mismatch` error reason. All six Permit2 settle branches are covered (exact: direct, EIP-2612 gas-sponsored, ERC-20-approval gas-sponsored; upto: the same three). Exact paths expect the full permitted amount; upto paths expect the actual settlement amount, which may be lower.

**Why:** receipt status proves only that the proxy call did not revert. With a fee-on-transfer token, the payer signs a Permit2 authorization for the full amount, the token delivers less to the recipient, the transaction still succeeds, and the merchant releases the resource underpaid. The proxy performs no balance-delta check, so the receipt is the only place to catch this.

**How:** this mirrors the structure of #2385, which added the identical check to the EIP-3009 settle path: same verifier, same error reason, same semantics (when logs are present, a matching Transfer is required). The expected Transfer is derived from the signed authorization (`from` = payer, `to` = witness recipient, `value` = settled amount, emitted by the permitted token).

**Scope and follow-ups:** the same status-only pattern exists in the Go SDK, the Python SDK, and the legacy v1 facilitator path. This PR is intentionally limited to the TypeScript exact and upto Permit2 paths (highest usage, direct analog of #2385) to stay reviewable; happy to follow up with the remaining SDKs if maintainers want.

Spotted during a security review of the facilitator settle paths. This PR was prepared with AI assistance (Cursor agent); every test assertion was validated by running the suite below, not generated blind.

## Tests

`vitest run` in `typescript/packages/mechanisms/evm`: 33 test files, 831 tests, all passing.

New regression test `test/unit/exact/permit2-settle-transfer-event.test.ts` (5 tests) drives `settlePermit2` with a real EIP-712 signature against mocked receipts:

- underpaying Transfer (900 of 1000 delivered): rejected with `invalid_exact_evm_transfer_event_mismatch`
- empty logs (token moved nothing): rejected with the same reason
- full amount transferred to a different recipient: rejected
- valid full-payment Transfer: succeeds
- reverted receipt: still reported as `invalid_transaction_state`

Pre-fix, the first three cases returned `success: true` (3 of 5 tests fail on 8bef6f7); post-fix all 5 pass.

`prettier --check` and `eslint` pass on all touched files.

## Checklist

### DIFF
--- a/typescript/.changeset/permit2-settle-transfer-event.md
+++ b/typescript/.changeset/permit2-settle-transfer-event.md
@@ -1,0 +1,5 @@
+---
+"@x402/evm": patch
+---
+
+Verify the ERC-20 Transfer event in Permit2 settle receipts (exact and upto) instead of trusting receipt status alone, so underpaying or non-conforming tokens are rejected at settlement
--- a/typescript/packages/mechanisms/evm/src/exact/facilitator/permit2.ts
+++ b/typescript/packages/mechanisms/evm/src/exact/facilitator/permit2.ts
@@ -36,6 +36,7 @@ import {
   checkPermit2Prerequisites,
   validateEip2612PermitForPayment,
   buildExactPermit2SettleArgs,
+  buildExpectedPermit2Transfer,
   splitEip2612Signature,
   waitAndReturnSettleResponse,
   mapSettleError,
@@ -438,7 +439,13 @@ async function settlePermit2WithEIP2612(
       dataSuffix,
     });
 
-    return waitAndReturnSettleResponse(signer, tx, payload, payer);
+    return waitAndReturnSettleResponse(
+      signer,
+      tx,
+      payload,
+      payer,
+      buildExpectedPermit2Transfer(permit2Payload),
+    );
   } catch (error) {
     return mapSettleError(error, payload, payer);
   }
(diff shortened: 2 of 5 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/3057

Please be mindful of maintainers time and refrain from opening duplicate PRs @SashaMIT, we already have enough incoming PRs to triage

SashaMIT:
Understood, apologies for the duplicate. #3057 is the surviving evm PR; #3058 and #3059 are the Go and Python SDK counterparts, not repeats. I will dedupe against open PRs before filing going forward.

PR #3081, opened 2026-08-07
spec: exact scheme for Sui (effects-only verification)

hayes-mysten:
## Description

Adds `specs/schemes/exact/scheme_exact_sui.md`, a Sui chain implementation of the `exact` scheme.

This is an **alternative design** to #2615 by @Sceat, whose work established gasless Address-Balance transfers on Sui and surfaced the core questions this spec builds on — full credit to that PR for the groundwork and discussion. The two differ in verification model:

- **Effects-only verification.** The facilitator confirms one thing — the declared recipient (`payTo`) was credited exactly `amount` in `asset`, read from the transaction's balance changes — and does not constrain how the transaction is built (which commands it uses, where funds are sourced, or how gas is arranged). Sui transactions are composable by design, so a payment may swap to the requested asset, withdraw from a vault, or write an on-chain receipt in the same transaction without the facilitator knowing or caring how. Clients may also self-sponsor gas.
- **Stateless replay + idempotent settlement.** Verify rejects an already-committed digest; Sui commits a transaction at most once, so a retried settle re-executes to the same effects. Checks-enabled simulation validates expiration, so there is no separate freshness field.
- **Optional server nonce.** A server MAY declare a Base64 `extra.nonce` the client carries as a `Pure` input for challenge binding or on-chain reconciliation (SHOULD be ≤32 bytes to stay gasless-eligible; MAY be referenced by a command). When absent, there is no nonce check.
- **Sponsorship as one field.** Gas sponsorship is a single `extra.feePayer` coordination field; what is sponsorable is entirely the sponsor's concern and out of x402 scope.

This is the base of a two-part stack; the `@x402/sui` implementation is in #3082, which depends on this spec.

## Tests

Docs-only. No code changes.

## Checklist

- [x] All new and existing tests pass (n/a — docs only)

---

*Disclosure: this spec was drafted with substantial AI assistance and reviewed by me before submission, per the AI-assisted contributions guidance.*


DrVelvetFog:
Really glad to see this — effects-only is how the facilitator I run validates payments, and the nonce-as-Pure-input design matches where the #2615 thread landed. Credit to @Sceat too; this builds on a lot of ground that PR broke.

One thing I'd push on before this merges: the freshness story has a gap I can back with data.

The spec's already-executed check is "recompute the digest and look it up." That lookup has a retention horizon. On Aug 7 I verified that digests from my July gasless runs already return NOT_FOUND on the public testnet fullnode — roughly a 2–3 week window ([data in the #2615 thread](https://github.com/x402-foundation/x402/pull/2615#issuecomment-5217061357)). So the verify-time replay guard silently stops working for any payment older than the node's retention, and the spec currently presents it as unconditional.

That would be fine if payment lifetime were bounded, but it isn't: the spec says checks-enabled simulation validates expiration so no separate freshness field is needed — except an Address-Balance-sourced transaction can legally carry `Expiration::None`, and the reference client in #3082 doesn't set an expiration at all. Default-built payments live forever; the lookup that's supposed to catch their replay doesn't.

To be clear about scope: this isn't a double-charge concern — execution-level dedup is the chain's job, as you said upthread. It's that the spec's stated verify guarantee doesn't hold past the retention window.

Two possible fixes, not mutually exclusive:

1. Have the spec RECOMMEND (or require) clients set an expiration, bounding validity well inside typical fullnode retention. The reference client setting one by default would make this real in practice.
2. Scope the already-executed check honestly ("best-effort within the node's retention window") and point servers at the declared nonce for anything that must stay verifiable longer — the nonce already solves this cleanly since the server retires the challenge, not the payment.

Happy to run this against my facilitator's testnet vectors once the shape settles — same offer as in the #2615 thread.

hayes-mysten:
Thanks for the review @DrVelvetFog.  I don't think this is actually a concern because execution will fail if you try to re-execute a tx that was already pruned.   I am updating the spec to mention this is a best effort check, but it is not critical, and there is no risk of an old tx somehow succeeding settlement. The simulation should also fail in this case, so you also should not end up in a situation where verify passes but settlement fails

Danny-Devs:
Read through at `9b1b1412`, with the `@mysten/sui` client and a testnet node open alongside.

One finding I would call load-bearing: a facilitator that implements §Effects check exactly as written can accept a payment that never happened. Then three smaller documentation notes.

I ran the call rather than reasoning from the type declarations, so the first one comes with output instead of a hypothesis.

## `balanceChanges` is opt-in on the transport this scheme now targets

§Verification step 6 and §Effects check both read the agreed asset's `balanceChanges`. On the gRPC surface that field is masked — `@mysten/sui@2.22.0`, `dist/client/types.d.mts`:

```ts
balanceChanges: Include extends { balanceChanges: true } ? BalanceChange[] : undefined;
```

`Include` defaults to `{}`. Simulating the same transaction twice against `fullnode.testnet.sui.io:443` (`sui-node/1.77.2`), one argument apart:

```
A. simulateTransaction({ transaction })
     status         : { success: true, error: null }
     balanceChanges : undefined

B. simulateTransaction({ transaction, include: { balanceChanges: true } })
     status         : { success: true, error: null }
     balanceChanges : [ { coinType: "0x2::sui::SUI", address: "0x1f3b…", amount: "-63520" }, … ]
```

Case A is a successful simulation. `status` arrives unconditionally, so there is no error for a facilitator to notice — the asset data is simply not there. Worth adding: `balanceChanges` is still **present as a key** on the response in case A, holding `undefined`, so a defensive `'balanceChanges' in result` check passes.

This is not a corner case for much longer. JSON-RPC on public fullnodes now answers *"Method not found. JSON-RPC on public fullnodes has been deprecated. Please migrate to gRPC or GraphQL endpoints"* (`sui_getChainIdentifier` against `fullnode.mainnet.sui.io`, 2026-08-16), so implementers are being moved onto the masked surface within this spec's own lifetime.

That matters because §Effects check states the assertion twice, and the two phrasings diverge on exactly this input. The positive one — *"asserts that the declared recipient, `payTo`, was credited (net) exactly `amount`"* — fails closed, since an empty sum is not `amount`. The closing line — *"A mismatch is rejected"* — presupposes a balance change exists to compare against, so an implementer looking for a contradicting entry finds none and passes.

Absent, empty, and present-but-wrong are three different states here, and the text currently distinguishes only the third.

**Suggested addition, one sentence in §Effects check:** the facilitator MUST request balance changes explicitly, and MUST treat their absence as a verification failure rather than as an absence of mismatch.

## Two smaller ones

**1. `extra.gasStation` disappears without a note.** Measured across the two revisions: `gasStation` 2 references on `main` → 0 here, `feePayer` 0 → 4. Since `extra` is matched by value against the client's `accepted.extra`, a stale client does not degrade — it stops matching. One line in §Gas sponsorship stating the field is removed would save a reader diffing revisions to find out.

**2. The reason-code table names a requirement the body disclaims.** The table describes `invalid_exact_sui_payload_transfer_mismatch` as *"Effects do not match the single-recipient requirement"*, but §Effects check and §Summary both say the scheme does not forbid other recipients. §Security Considerations already has the correct phrasing under *Authorization scope*: *"the recipient credited exactly `amount`"*. Worth fixing because the table is where a second implementer looks first — it misled this read, and a reader who stops there builds something that rejects valid composed payments.

---

*Drafted with AI assistance, verified by a human in the loop.*

hayes-mysten:
Thanks for the review. I updated the wording in a few places to clarify a few of the things you flagged

I am not sure the balance change missing thing your agent flagged is really a major concern, that kinda just falls into a category of a broken implementation, but added some small changes to cover this anyway

feePayer was based on comments on the other spec pr.  I'm not sure if migrations should be covered here or not (would love some input from a maintainer here, but since there was never an in-repo implementation of the spec, I think we can probably omit it?

DrVelvetFog:
Re-read at `5d5b8f21`. The fail-closed sentence on balance changes is the right shape — absent and zero being different answers is the kind of thing a second implementer gets wrong silently.

Coming back to the best-effort wording, though: I think step 4 now rules out the one case it was written for.

The paragraph opens by saying why the lookup exists ([L193](https://github.com/x402-foundation/x402/blob/5d5b8f2188ca0d1dde2bcdc4723d3a2b88f4cae9/specs/schemes/exact/scheme_exact_sui.md#L193)):

> A transaction sourced entirely from Address Balances has no object inputs, so simulation alone cannot detect a replay.

and then closes by saying why losing the lookup is survivable ([L196](https://github.com/x402-foundation/x402/blob/5d5b8f2188ca0d1dde2bcdc4723d3a2b88f4cae9/specs/schemes/exact/scheme_exact_sui.md#L196)):

> a replay whose digest has been pruned instead fails simulation or execution (its inputs are consumed or its expiration has passed)

Take an Address-Balance payment with no expiration set. There are no object inputs to consume — that's the first sentence — and there's no expiration to pass. Both fallbacks are empty for exactly the transaction shape the step exists to catch, and it's the shape §Construction points people at, since `tx.balance()` "often yields a gasless stablecoin transfer." Same claim carries into §Security Considerations at [L316](https://github.com/x402-foundation/x402/blob/5d5b8f2188ca0d1dde2bcdc4723d3a2b88f4cae9/specs/schemes/exact/scheme_exact_sui.md#L316).

To be clear about the blast radius: at-most-once still holds and nobody double-spends. What's reachable is a verify that passes on a pruned digest and a settle that then doesn't — plus two sentences that overstate. [L141](https://github.com/x402-foundation/x402/blob/5d5b8f2188ca0d1dde2bcdc4723d3a2b88f4cae9/specs/schemes/exact/scheme_exact_sui.md#L141) and L318 both say a payment's validity window "is bounded on-chain by the transaction's own expiration," but nothing in the scheme asks the client to set one, and the reference client in #3082 doesn't.

Which is the repair, I think, rather than rewording the parenthetical: have the client set an expiration by default, and say so here. Then "its expiration has passed" is always true, the best-effort argument holds for every shape including the gasless one, and both bounded-window sentences become true as written. Pruning stops mattering because nothing outlives its own window.

If you'd rather leave the client alone, the smaller version is to scope the parenthetical to transactions that have object inputs or an expiration, and name the remainder as verify-passes-settle-fails.

*Drafted with AI assistance, verified against the branch head.*

Danny-Devs:
The gap @DrVelvetFog describes — a payment with no owned inputs and no gas payment, so nothing on-chain bounds its replay window — doesn't look reachable through the reference client. It stamps an expiration on exactly that shape, automatically.

`client/address-balance-transaction-expiration.ts` sets a two-epoch `ValidDuring` window whenever four things hold: gas selection is off, `gasData.payment` is empty, the client set no expiration of its own, and the build isn't transaction-kind-only. That combination *is* the gasless Address Balance shape. And a client that lets the SDK pick gas gets gas coin objects instead — those are owned inputs, so the other bound catches it.

So the SDK closes the gap. The spec doesn't, and the spec is what a second implementer builds from: a Rust or Go client following §Construction has no reason to stamp an expiration at all, because the expiry bullet reads as an observation, not a rule.

The cheap fix is to write down what the reference client already does. §Construction's expiry bullet would become:

"**Expiry.** A payment carrying no owned inputs and no gas payment — the gasless Address Balance shape — MUST set a transaction expiration, and the facilitator MUST reject it at verify when one is absent. Where set, it bounds the validity window on-chain, enforced by simulation at verify and execution at settle. `maxTimeoutSeconds` is the resource server's advertised window; it is not enforced at the transaction level and need not match."

and the matching sentence under Replay prevention would point at that requirement rather than restate it. Costs existing clients nothing — they already emit it.

Separately: that window is two epochs, wide next to a typical `maxTimeoutSeconds`. Worth a line saying the two are different guarantees, so nobody reads the advertised one as enforced.

*Drafted with AI assistance, vetted by human in the loop.*

DrVelvetFog:
Danny-Devs' wording is the fix. The SDK evidence is what settles it — good catch that the reference client already stamps the two-epoch window, because it means a spec-level MUST costs existing clients nothing and only pins down what a second implementer (Rust, Go) would otherwise skip, since §Construction's expiry bullet reads as an observation rather than a rule.

Worth spelling out that the same MUST also repairs the step 4 wording: once an expiration is required on the no-owned-inputs/no-gas shape, "or its expiration has passed" is a real fallback again, and the "validity window is bounded on-chain by the transaction's own expiration" lines in §Verification and §Security Considerations become true as written instead of describing a shape the spec never requires to carry one. Same one-line change closes both.

+1 on flagging that the two-epoch window and `maxTimeoutSeconds` are separate guarantees — that's an easy silent mismatch for a facilitator author to bake in.

*Drafted with AI assistance, human in the loop.*

DrVelvetFog:
Small follow-up, prompted by real usage. Two independent implementers built on a live Sui facilitator this week — one a client SDK + seller middleware, one a second facilitator instance — both on testnet, and both snagged on the same question: what's the canonical Sui *testnet* network id?

Turns out it's already here: the Network Format section of this spec defines `sui:mainnet` / `sui:testnet` / `sui:devnet`. So within the scheme it's settled, and I've pointed them at it.

Where it's *not* yet is one level up — `docs/core-concepts/network-and-token-support.mdx` (the CAIP-2 identifier list and the default-asset tables) has no Sui entry at all; it covers EVM, Solana, TON, Algorand, Stellar, Aptos, Hedera, Keeta, NEAR, Concordium, and XRPL. So anyone discovering or validating networks against that registry doesn't find `sui:testnet`, even though this PR specifies it — which is almost certainly what implementers keep hitting.

Since that doc tracks shipped/registered support, the natural time to add Sui is when this scheme lands. Rows ready to fold in — both USDC confirmed on-chain as USDC / 6 dp today:

- Identifier: `sui:<network>` (`mainnet`, `testnet`, `devnet`)
- Sui Mainnet · `sui:mainnet` · `0xdba34672e30cb065b1f93e3ab55318768fd6fef66c15942c9f7cb846e2f900e7::usdc::USDC` · 6
- Sui Testnet · `sui:testnet` · `0xa1ec7fc00a6f40db9693ad1415d0c193ad3906494428cf252621037bd7117e29::usdc::USDC` · 6

Happy to send these as a docs PR to land alongside the merge if that's useful.

*Drafted with AI assistance, reviewed by me.*

yannickrocks:
Implementer here, the client SDK and middleware DrVelvetFog mentioned. Adding
the running-code view for `sui:testnet` and the registry gap.

- `@sui-x402` (five TypeScript packages, v0.1.0 on npm) treats `sui:testnet`
  as the default network in both the payer and the seller middleware;
  `sui:mainnet` is behind explicit opt-ins on both sides.
- Eleven payments have settled end to end on Sui testnet through the reference
  facilitator with `sui:testnet` in `PaymentRequirements`, most recently
  [`F1YSaQdc…`](https://testnet.suivision.xyz/txblock/F1YSaQdcFiMgV364osSCDAVNuD4tPPPAQtZVXMACh39k).
- We hit exactly the discovery gap described above: the scheme in this PR
  defines the id, but `docs/core-concepts/network-and-token-support.mdx` has
  no Sui entry, so anyone validating against the registry comes up empty.

+1 to landing the registry rows alongside this scheme. The testnet USDC entry
in DrVelvetFog's table matches what our schemas capture from the facilitator's
live traffic (`0xa1ec…7e29::usdc::USDC`, 6 decimals). Happy to co-review the
registry docs PR when it goes up.

*Drafted with AI assistance, reviewed by me.*

Danny-Devs:
I wanted to raise a suggestion for §Construction's expiry rule. I first made it on #2616 in July, against the implementation this PR replaced, so it never reached this thread.

**`ValidDuring` can bound a payment in seconds, and nothing currently uses that.**

The struct carries `minTimestamp` and `maxTimestamp` alongside the epoch pair, and the SDK's Address-Balance plugin leaves both `null`. That leaves only the epoch bound, `minEpoch: currentEpoch, maxEpoch: currentEpoch + 1`, so at Sui's 24-hour epochs a signed payment stays broadcastable for 24 to 48 hours while `maxTimeoutSeconds` is usually measured in seconds.

The proposal is that where `maxTimestamp` is present, the facilitator enforces `maxTimeoutSeconds` against it precisely, keeping the epoch bound as the floor. A client that wants a tight payment stamps a tight timestamp, and a client that stamps nothing behaves exactly as it does today.

Nothing is added to the transaction, since the field already sits in the signed bytes and is simply unset, so there is no extra command and no extra gas. It also shrinks the surrounding problem, because any dedup kept above the chain then only has to cover the payment's real lifetime, and two minutes is far easier to cover than two days.

*Drafted with AI assistance, vetted by a human in the loop. Read at `5d5b8f21` and the SDK's `address-balance-transaction-expiration.ts` on main.*

PR #3082, opened 2026-08-07
@x402/sui: exact scheme for Sui (effects-only verification)

hayes-mysten:
## Description

Adds `@x402/sui`, the Sui implementation of the `exact` scheme — client, server, and facilitator.

This is an **alternative implementation** to #2616 by @Sceat, whose PR built out the first `@x402/sui` mechanism and gasless-transfer flow; full credit for that foundation. This version pairs with the effects-only spec and differs mainly in what the facilitator checks:

- **Effects-only verification.** The facilitator answers one question — "did the declared recipient receive exactly the requested amount?" — from the transaction's balance changes. It does not inspect or constrain commands, funding source, or gas shape, so a payment may draw from an address balance, owned coins, an on-demand swap, or a contract withdrawal and still verify. Verify simulates (a preflight the resource server can call before doing work); settle executes; both check the balance changes.
- **Stateless replay + idempotent settlement.** Verify rejects an already-committed digest; a retried settle re-executes to the same effects. One simulation per request.
- **Optional Base64 `extra.nonce`** carried as a `Pure` input, matched byte-for-byte when declared.
- **Sponsorship** is a single `extra.feePayer` field; sponsored settlement is delegated to the sponsor (structural interface, satisfied by the mysten-incubation sponsor SDK), which applies its own policy. x402 performs no gas-shape validation.

**Implements the spec in #3081 — please review that first.** This is a single-commit PR based on `main`; it does not depend on the spec doc to build or test, so the two are separate PRs rather than a git stack, but #3081 is the design of record for what this implements.

## Tests

`pnpm test` from `typescript/` — 59 unit tests pass (facilitator 36, utils 12, server 6, client 5). `lint:check`, `format:check`, and `build` are green.

## Checklist

- [x] I added a changelog fragment for user-facing changes (`typescript/.changeset/sui-exact-mechanism.md`)

---

*Disclosure: this implementation was written with substantial AI assistance and reviewed by me — including the signing and settlement paths — before submission, per the AI-assisted contributions guidance.*

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,9 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .: {}
--- a/typescript/.changeset/sui-exact-mechanism.md
+++ b/typescript/.changeset/sui-exact-mechanism.md
@@ -1,0 +1,5 @@
+---
+"@x402/sui": minor
+---
+
+Add `@x402/sui`: the exact payment scheme for Sui — client, server, and facilitator, with effects-only verification and optional gas sponsorship.
--- a/typescript/packages/mechanisms/sui/.prettierignore
+++ b/typescript/packages/mechanisms/sui/.prettierignore
@@ -1,0 +1,3 @@
+# build output
+dist/
+node_modules/
--- a/typescript/packages/mechanisms/sui/.prettierrc
+++ b/typescript/packages/mechanisms/sui/.prettierrc
@@ -1,0 +1,11 @@
+{
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
+  "singleQuote": false,
+  "trailingComma": "all",
+  "bracketSpacing": true,
+  "arrowParens": "avoid",
+  "printWidth": 100,
+  "proseWrap": "never"
+}
--- a/typescript/packages/mechanisms/sui/README.md
+++ b/typescript/packages/mechanisms/sui/README.md
@@ -1,0 +1,113 @@
+# `@x402/sui` [![npm version](https://img.shields.io/npm/v/%40x402%2Fsui.svg)](https://www.npmjs.com/package/@x402/sui)
+
+Sui implementation of the x402 `exact` payment scheme.
+
+## Installation
+
+```bash
+npm install @x402/sui
+# or
+pnpm add @x402/sui
+```
+
+## Usage
+
+The package exports the `exact` scheme for all three x402 roles under `@x402/sui/exact/{client,server,facilitator}`. Each is registered on the matching core object for the `sui:*` network family.
+
+### Client
+
+```typescript
+import { Ed25519Keypair } from "@mysten/sui/keypairs/ed25519";
+import { x402Client } from "@x402/core/client";
+import { ExactSuiScheme } from "@x402/sui/exact/client";
+
+// Any Sui signer works — Ed25519, Secp256k1, or a wallet-backed signer.
+const keypair = Ed25519Keypair.generate();
+const client = new x402Client().register("sui:*", new ExactSuiScheme(keypair));
+
+// On a 402 response, `client` builds and signs the payment transaction from the
+// returned PaymentRequirements — sourcing the asset and arranging gas per the
+// requirements — then retries the request with the signed PaymentPayload.
+```
+
+### Server
+
+```typescript
+import { x402ResourceServer } from "@x402/core/server";
+import { ExactSuiScheme } from "@x402/sui/exact/server";
+
+const server = new x402ResourceServer({
+  facilitatorUrl: "https://facilitator.example",
+}).register("sui:*", new ExactSuiScheme());
+
+// Price resources in USDC — "$0.01" or an explicit { amount, asset }. The scheme
+// converts the price to atomic units in the PaymentRequirements it builds.
+```
+
+### Facilitator
+
+```typescript
+import { SuiGrpcClient } from "@mysten/sui/grpc";
+import { x402Facilitator } from "@x402/core/facilitator";
+import { ExactSuiScheme } from "@x402/sui/exact/facilitator";
+
+// No key is needed for self-paid or gasless payments. All options are optional:
+//   clients             per-network Sui client overrides (any core-API client);
+//                       networks without one fall back to the public fullnode
+//   executeTransaction  a custom executor settlement is delegated to (e.g. a gas
+//                       sponsor); absent, the facilitator broadcasts directly
+//   feePayer            a sponsor address to advertise as `extra.feePayer`
+//   settlementGuard     a dedup guard (default: in-process, digest-keyed)
+const facilitator = new x402Facilitator().register(
+  "sui:*",
+  new ExactSuiScheme({
+    clients: {
+      "sui:mainnet": new SuiGrpcClient({ network: "mainnet", baseUrl: "https://my-node:443" }),
+    },
+  }),
+);
+```
+
+#### Gas sponsorship
+
+Advertise a `feePayer` (which clients set as the gas owner) and supply an `executeTransaction` that co-signs the gas. The executor is a plain function that submits the signed bytes and returns the result; a gas sponsor is one example:
+
+```typescript
+import type { SuiExecutor } from "@x402/sui";
+
+// e.g. wrap the mysten-incubation sponsor SDK's createSponsor(...).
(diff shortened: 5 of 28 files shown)


DrVelvetFog:
Read this alongside the spec — the settle path gating success on the finalized transaction's `status.success` *and* re-matching balance changes closes the committed-but-failed-reads-as-success class that came up in the #2616 review, and the sender-only signature check is the right call for sponsored transactions. Two things I'd flag from running this shape in production:

**1. `isExecuted` matches error messages, not error codes** ([facilitator/scheme.ts, `isExecuted`](https://github.com/x402-foundation/x402/pull/3082/files)):

```ts
if (/not found|NOT_FOUND/i.test(message)) return false;
```

This is the same class as the `/not found/i` over-match flagged on #2616: any transport-layer error whose message happens to contain "not found" (a proxy 404 body, DNS "host not found") reads as "transaction not executed," and the guard fails open — an already-executed payment passes verify. I hit exactly this class in the facilitator I run and fixed it by keying on the structured error code instead; the gRPC client surfaces one, so `code === "NOT_FOUND"`-style matching (rethrow everything else) avoids it. Worth doing here since custom clients via `options.clients` will each have their own message strings anyway.

**2. The settlement guard claims the digest before validation runs.** `settle()` inserts into the guard, then calls `validate()`. Validation isn't fully offline — `verifyTransactionSignature` takes the client because zkLogin verification does an RPC hop — so a transient RPC failure rejects the payment *after* its digest is claimed. The retry then gets `duplicate_settlement` for up to `SETTLEMENT_TTL_MS` (10 min) with nothing ever executed. Options: release the key on any pre-execution failure (would need a `release()` on `SettlementGuard`), or claim after `validate()` — the synchronous check-plus-insert still closes the concurrent race at that point.

Small related note: the `SETTLEMENT_TTL_MS` comment says it's sized to exceed a typical freshness window, but with the client not setting an expiration (see my comment on #3081) there's no freshness window to exceed.

Danny-Devs:
Confirming the bug @DrVelvetFog flagged above; agree his fix is the right one. I also found a second data point: on the default client path, the failure is worse than the one he described.

**`isExecuted()` rejects every genuine payment when the facilitator uses the package's own default client.** It decides "not found" by matching the error message with a regex, but the real gRPC client returns that message percent-encoded — no space, no underscore — so the regex never matches. The error is rethrown, and `verify()` fails closed for any payment that has not executed yet, which is every real payment. Executed against live testnet at this branch's head, `@mysten/sui` 2.23.1:

```
error.code    : "NOT_FOUND"
error.message : "Transaction%205tPHwXfy…so42d%20not%20found"

/not found|NOT_FOUND/i.test(message)  ->  false   // rethrows → invalid_exact_sui_payload_verification_error
error?.code === "NOT_FOUND"           ->  true
```

His fix covers both directions — the fail-open edge case he described and this fail-closed default path:

```ts
if (error?.code === "NOT_FOUND") return false;
throw error;
```

The unit suite cannot see it because the mock rejects with `new Error("Transaction not found")` — literal spaces — so the regex passes against the mock and fails against the real client. Checking `error.code` instead of the message text also means the guard keeps working if someone plugs in a different Sui client whose error messages are worded differently.

Recommend two small test additions alongside the fix: (1) have the facilitator mock reject with the real gRPC error shape — `code: "NOT_FOUND"` and a percent-encoded message — so the suite exercises the client's actual behaviour rather than a hand-written string; and (2) one end-to-end test that builds a payload with the package's own client (`tx.balance()` → `send_funds`) and runs it through `verify()`, so the default path is covered by the shape the client really emits.

*Drafted with AI assistance and reviewed by a human, per the contributing guide. The error output above was executed, not inferred.*

hayes-mysten:
I added better errors for not found transactions in the sui sdk, this repo needs a couple days before the new version can be consumed, but this should make the not found error handling much cleaner when it lands.

thanks @Danny-Devs, this was largely based on the work you had started a couple months back

DrVelvetFog:
The code-based match reads right, and it works against the SDK that's already published — Danny-Devs' output above shows `error.code === "NOT_FOUND"` arriving on 2.23.1 today, so this isn't waiting on the new version. Better messages are welcome either way.

One thing from my earlier comment is still open at `e74e16df`, and it's the one I'd least want to ship:

**A settle that never executes anything still burns the digest for the full TTL.** `settle()` claims the key ([scheme.ts#L188](https://github.com/x402-foundation/x402/blob/e74e16df2eb4851c4818490d1a731abf74f586ae/typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts#L188)) before it calls `validate()` ([#L199](https://github.com/x402-foundation/x402/blob/e74e16df2eb4851c4818490d1a731abf74f586ae/typescript/packages/mechanisms/sui/src/exact/facilitator/scheme.ts#L199)), and `isDuplicate()` claims on the check ([settlement-cache.ts#L32-L37](https://github.com/x402-foundation/x402/blob/e74e16df2eb4851c4818490d1a731abf74f586ae/typescript/packages/mechanisms/sui/src/settlement-cache.ts#L32-L37)) with no way to un-claim it. `validate()` isn't offline — zkLogin signature verification does an RPC hop — so a blip there, or a throw out of `execute()`, leaves the payment claimed with nothing committed. The retry then gets `duplicate_settlement`, which reports a second settlement when there wasn't a first one. Nobody loses money; the payment just can't buy anything for ten minutes.

The types already imply the half that's missing: `isDuplicate` is documented as recording the settlement "as in-progress" and the cache calls its entries pending, but there's no transition out.

Now's the cheap moment to fix it, since `SettlementGuard` is new in this PR and nobody's implemented a shared one yet: add `release(payment)` to the interface and call it wherever settle returns without a digest. The response shape makes that rule mechanical — `transaction: ""` means nothing was committed, so the claim goes back. Releasing after a failed execute is safe on Sui for the reason the comment right above the guard already gives: a retry either commits or re-derives the same effects, so the resource is still served once.

Smaller alternative if you'd rather not touch the interface: claim after `validate()` rather than before. Still atomic, covers the RPC-hop case, and leaves only the execute-throw path bricked.

*Drafted with AI assistance, verified against the branch head.*

Danny-Devs:
I have one observation about `settle()`, and it is a placement question rather than a missing piece.

`isExecuted()` is the strongest replay witness in the package. It asks the fullnode directly over gRPC, so it is the authoritative answer rather than something downstream of it.

It is consulted only in `verify()`, at `:123`. In `settle()`, the path that broadcasts and returns `success`, the replay defense is `guard.isDuplicate()` at `:188`.

The asymmetry is what stands out. `verify()` is advisory and a caller may skip it, while `settle()` commits, so the stronger check sits on the side that does not have to run. A shared `SettlementGuard` covers the cross-instance case, as the README says, but any guard is still a cache, and the chain is what it caches.

Calling `isExecuted` from `settle()` too would make that path independent of how the guard is deployed, using a method a few lines away that already handles `NOT_FOUND` correctly. The cost is one round trip.

If the intent is that `verify()` always runs first, that is fair, though it may be worth stating in the README, since the type signature does not require it.

*Drafted with AI assistance, vetted by a human in the loop. Read at `e74e16df` — `facilitator/scheme.ts:123`, `:188`, `:505`, and `utils.ts:72`.*

Danny-Devs:
I have a suggestion about the sponsorship example in the README, offered as help rather than a defect report.

The spec already names gas-coin protection under §Gas sponsorship and leaves it to the sponsor's own policy. The gap is that the README sample does not show one, and that sample is what integrators copy:

```typescript
const executeTransaction: SuiExecutor = ({ transaction, signatures }) =>
  sponsor.signAndExecuteTransaction({ transaction, userSignature: signatures });
```

The sponsor pays the gas, and the client knows who the sponsor is, because `getExtra()` advertises `feePayer` to anyone who asks. So a client can build a transaction whose commands reference `GasCoin`, split value out of the sponsor's gas payment, and send it to itself. The sponsor co-signs, because nothing in the package checks what the commands touch.

I maintain a self-hosted Sui gas sponsorship library and hit this, so the guard is already written. `assertNoGasCoinUsage` walks the commands and throws if any argument resolves to the `GasCoin` input.

Either fix would help. The small one is a line in the example saying the executor must enforce a policy, naming this hazard. The larger one is exporting a helper, so the demonstrated path is the safe one. I am happy to open a PR for either, or to leave it if this is deliberately out of scope.

*Drafted with AI assistance, vetted by a human in the loop. Guard at `Danny-Devs/sui-gas-station`, `src/policy.ts:101`. Read at `e74e16df` — `facilitator/scheme.ts:80`, `client/scheme.ts:88-90`.*

PR #3083, opened 2026-08-07
feat: add settlement pending state

CarsonRoscoe:
Continuation of PR #3030 

## Description

Adds `settlement_pending` for EVM settlement receipt-wait failures across exact, upto, and batch-settlement flows. A transaction can be broadcast even when its receipt cannot be retrieved, so callers receive its hash and can reconcile before retrying.

Also preserves confirmed batch-deposit success when post-receipt RPC state reads fail, validates broadcast hashes, and ensures terminal Go settlement errors do not expose malformed transaction values.

## Receipt-wait decision

Treat every receipt-wait exception after a valid broadcast hash as `settlement_pending`. The exception type cannot prove that the transaction will not confirm; preserving the hash is safer than reporting a terminal failure that could lead callers to retry a completed payment.

## Tests

- TypeScript, Go, and Python unit coverage includes receipt-wait failures and invalid broadcast hashes.

## Checklist


typescript/packages/mechanisms/evm/src/shared/settleReceipt.ts
```diff
@@ -0,0 +1,123 @@
...
+    onSuccess,
+  } = options;
+
+  if (!isValidTxHash(tx)) {
+    return invalidBroadcastHashResponse(tx, failedStatusReason, network, payer);
+  }
+
+  let receipt;
+  try {
+    receipt = await signer.waitForTransactionReceipt({ hash: tx });
```
phdargen, line 59: This covers `waitForTransactionReceipt` failures well.  
However viem’s default wait is quite long with 3 minutes, so on most serverless deployments this path would never be reached. Instead, the process would be killed with a 5xx/timeout.

Suggest to add a configurable confirmationTimeoutMs (default 3mins?) set a few seconds below the platform limit to make sure `waitForTransactionReceipt` throws before the platform kills the process
```suggestion
    receipt = await signer.waitForTransactionReceipt({ hash: tx, confirmationTimeoutMs });
```

phdargen (dismissed):
LGTM 🚀

phdargen (approved):

PR #2727, opened 2026-06-29
fix(go): verify Transfer event after exact/eip3009 settle

wnjoon:
## Description

This ports the defensive exact/eip3009 post-settle receipt check from TypeScript PR #2385 to the Go SDK in a backward-compatible way.

`settleEIP3009` previously treated `receipt.Status == TxStatusSuccess` as sufficient proof that the expected ERC-20 transfer happened. A successful receipt only proves the transaction did not revert; it does not prove the expected token emitted a matching `Transfer(from, to, value)` event.

Go's `FacilitatorEvmSigner` interface returns the repository's lightweight `TransactionReceipt`, and existing user-provided signers may still return status-only receipts. To avoid false negatives for those implementations, this change adds optional receipt logs and verifies the Transfer event only when logs are present.

### Changes

- Added optional `Logs` to `go/mechanisms/evm.TransactionReceipt`.
- Added `ErrTransferEventMismatch` with the TypeScript parity wire string `invalid_exact_evm_transfer_event_mismatch`.
- Added exact/eip3009 Transfer log verification against the expected token, payer, recipient, and amount.
- Preserved the transaction hash on Transfer event mismatch settlement errors.
- Populated logs in in-repo integration signer receipt adapters that already receive go-ethereum receipts.
- Added focused tests for matching logs, unrelated logs, mismatched fields, empty logs, and settle-level success/failure paths.
- Added a Go changelog fragment.

### Compatibility

This does not change the `FacilitatorEvmSigner` interface. Existing signers that return receipts without logs keep the current status-only behavior. Signers that populate `TransactionReceipt.Logs` opt into the stronger Transfer event check.

## Related

- TypeScript parity PR: https://github.com/x402-foundation/x402/pull/2385

## Tests

- `GOCACHE=/tmp/go-build-cache go test ./mechanisms/evm/exact/facilitator`
- `GOCACHE=/tmp/go-build-cache go test ./...`
- `GOCACHE=/tmp/go-build-cache GOLANGCI_LINT_CACHE=/tmp/golangci-lint-cache make verify`
- `GOCACHE=/tmp/go-build-cache make build`

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
- [x] My commits are signed (required for merge) -- you may need to rebase if you initially pushed unsigned commits
- [x] I added a changelog fragment for user-facing changes (docs-only changes can skip)

PR #3084, opened 2026-08-08
fix(go): verify Transfer event in Permit2 settle receipts (exact + upto)

SashaMIT:
Go Permit2 settle (exact and upto) still trusted `receipt.status` alone, so a fee-on-transfer or otherwise non-conforming token could underpay while the facilitator reported full success. This PR requires a matching ERC-20 `Transfer` event in the settle receipt, the same class of fix as TypeScript #3080 and the EIP-3009 path (#2385 / Go #2727).

**What changes:** after a successful receipt status, both Go Permit2 settle paths require a matching `Transfer` when logs are present, reusing `VerifyEIP3009TransferEvent` (exported from the exact facilitator package). Exact expects the full permitted amount; upto expects the actual settlement amount. Same logs-present guard semantics as EIP-3009 settle.

**Why:** receipt status proves only that the proxy call did not revert. The Permit2 proxy performs no balance-delta check, so underpayment is invisible without inspecting Transfer logs.

**Lineage:** same bug class as merged #2385 / #2727 (EIP-3009 Transfer verification) and open TypeScript #3080 (Permit2 Transfer verification). Earlier Go attempt #3058 covered the same gap but predates current main, duplicated the verifier in upto, and only unit-tested the helper. This PR rebases onto current main, reuses the exact verifier, adds settle-path regression tests, and drops the Cursor co-author trailer.

**Signing note:** this commit carries an SSH signature (`id_ed25519_codered_signing`). If GitHub's `check-verified-commits` still fails because that key is not registered as a signing key on the account, please amend with your GPG/SSH signing key:

```bash
cd /tmp/x402-second-pass
git checkout fix/go-permit2-settle-transfer-event-v2
git commit --amend --no-edit -S
git push --force-with-lease origin HEAD:fix/go-permit2-settle-transfer-event-v2
```

(Same human amend path as #3080.)

Spotted during a security review of the facilitator settle paths (F-201). Prepared with AI assistance (Cursor agent); every test assertion was validated by running the suite below.

## Tests

```bash
cd go
go test ./mechanisms/evm/exact/facilitator/ ./mechanisms/evm/upto/facilitator/ -count=1
```

New settle-path regressions:

- exact: underpaying Transfer (900 of 1000) rejected; empty logs rejected; matching Transfer succeeds
- upto: underpaying Transfer rejected; empty logs rejected; matching Transfer succeeds; partial settlement binds to settlement amount (not permitted max)

## Checklist

- [ ] My commits are signed (required for merge) -- SSH-signed here; amend with a GitHub-registered signing key if the verified-commits check fails
- [x] I added a changelog fragment for user-facing changes (`go/CHANGELOG.md` Unreleased)

### DIFF
--- a/go/CHANGELOG.md
+++ b/go/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+### Fixed
+- Verify matching ERC-20 Transfer logs in Go Permit2 settle receipts (exact and upto) instead of trusting receipt status alone, so fee-on-transfer / non-conforming tokens cannot underpay while settle reports success (parity with EIP-3009 #2727 and TypeScript #3080)
+
 ## v2.21.0 - 2026-08-04
 ### Added
 - Add Celo mainnet (chain ID 42220) and Celo Sepolia (chain ID 11142220) support with USDC as the default stablecoin ([#3025](https://github.com/x402-foundation/x402/pull/3025)) - Thanks [@GigaHierz](https://github.com/GigaHierz) and [@claude](https://github.com/claude)!
--- a/go/mechanisms/evm/exact/facilitator/eip3009.go
+++ b/go/mechanisms/evm/exact/facilitator/eip3009.go
@@ -239,7 +239,7 @@ func (f *ExactEvmScheme) settleEIP3009(
 	}
 
 	if receipt.Logs != nil {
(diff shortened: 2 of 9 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/3058

SashaMIT:
Ack — keeping https://github.com/x402-foundation/x402/pull/3058 as the Go Permit2 Transfer-event PR. Thanks.

PR #3032, opened 2026-08-04
fix(python): verify Transfer event after exact/eip3009 settle

SashaMIT:
## Description

`settle` for exact EIP-3009 in the Python SDK currently treats a successful transaction receipt status as proof that the expected ERC-20 transfer occurred. Receipt status only shows the transaction did not revert; it does not prove that the token emitted `Transfer(from, to, value)` for the authorization.

This ports the defensive post-settle check already landed in Go (#2727) and proposed for TypeScript (#2385):

- Add optional `logs` on `TransactionReceipt` and populate them from the web3 signer
- Add `verify_eip3009_transfer_event(...)` (topic + from/to/value match)
- When `receipt.logs is not None`, require a matching Transfer or return `invalid_exact_evm_transfer_event_mismatch`
- Apply in exact V2 and V1 facilitators
- Unit tests mirror the Go/TS cases

`logs is None` keeps backward compatibility with mocks/signers that omit logs (same idea as Go's `if receipt.Logs != nil` gate). The default web3 signer now always supplies a list.

## Related

- Python parity with #2727 / companion to open TS #2385
- Payment-integrity hardening; no new exploit details beyond those PRs

## AI disclosure

Majority of this PR was drafted with AI assistance and reviewed against Go #2727 / TS #2385 behavior and existing Python settle tests.

## Test plan

- [x] `pytest tests/unit/mechanisms/evm/test_eip3009_transfer_event.py`
- [x] `pytest tests/unit/mechanisms/evm/test_facilitator.py`
- [ ] CI on this PR

PR #3087, opened 2026-08-08
fix(legacy): verify Transfer event in exact EIP-3009 settle receipts

SashaMIT:
## Description

The legacy `x402` package (`typescript/packages/legacy/x402`) still settles exact EIP-3009 payments by checking `receipt.status` alone. A fee-on-transfer or otherwise non-conforming token can underpay (or move nothing) while the facilitator reports `success: true`. The v2 EIP-3009 path already rejects that class via `verifyEip3009TransferEvent` (upstream #2385 / #3032 lineage). Open Permit2 PRs (#3080 / #3084 / #3059) close the same gap on Permit2 rails; this PR backports the guard to the still-published legacy package.

**Change:** after a successful receipt, when `receipt.logs` is present, require a matching ERC-20 `Transfer(from, to, value)` for the authorized amount. Failure reason: `invalid_exact_evm_transfer_event_mismatch`.

## Test plan

- [x] `vitest run src/schemes/exact/evm/facilitator.test.ts` (16/16): existing smart-wallet + EOA cases green; new cases cover underpay, empty logs, and matching Transfer
- [ ] CI on this PR

### DIFF
--- a/typescript/packages/legacy/x402/src/schemes/exact/evm/facilitator.test.ts
+++ b/typescript/packages/legacy/x402/src/schemes/exact/evm/facilitator.test.ts
@@ -1,8 +1,20 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import { Address, Chain, parseSignature, Transport } from "viem";
+import {
+  Address,
+  Chain,
+  encodeAbiParameters,
+  getAddress,
+  keccak256,
+  padHex,
+  parseSignature,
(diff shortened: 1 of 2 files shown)


phdargen:
The legacy packages are effectively frozen, updates are restricted to strictly necessary security patches

SashaMIT:
Understood on the legacy freeze. Thanks for the clarification — closing here.

PR #3092, opened 2026-08-08
fix(batch-settlement): fail closed when settle emits no Settled amount

SashaMIT:
## Summary
- On-chain `settle` early-returns with receipt status success and **no** `Settled` event when a concurrent settle already drained pending funds (`amount == 0`).
- Facilitators previously returned `success: true` with `amount` `"0"` / `""` (TS/Python) or success with no amount (Go), which can unlock merchant fulfillment after a no-op.
- Require a positive matching `Settled` event (TypeScript + Python) or an advanced `receivers.totalSettled` (Go). New error: `invalid_batch_settlement_evm_settled_event_mismatch`.

Same class as the open Permit2 / EIP-3009 Transfer-event settle checks.

## Test plan
- [x] `pnpm exec vitest run test/unit/batch-settlement/facilitator.test.ts` in `@x402/evm` (48 passed), including fail-closed cases for empty logs / missing Settled
- [x] `go test ./mechanisms/evm/batch-settlement/facilitator/ -run ExecuteSettle` (includes `FailsClosedWhenTotalSettledUnchanged`)

### DIFF
--- a/go/.changes/unreleased/fixed-3092-settled-event.yaml
+++ b/go/.changes/unreleased/fixed-3092-settled-event.yaml
@@ -1,0 +1,2 @@
+kind: fixed
+body: Fail closed when a batch-settlement settle receipt does not advance receivers.totalSettled (possible no-op early return).
--- a/go/mechanisms/evm/batch-settlement/facilitator/errors.go
+++ b/go/mechanisms/evm/batch-settlement/facilitator/errors.go
@@ -116,7 +116,8 @@ const (
 	ErrAuthorizerNotConfigured   = "invalid_batch_settlement_evm_authorizer_not_configured"
 
 	// Settle action errors
-	ErrUnknownSettleAction = "invalid_batch_settlement_evm_unknown_settle_action"
-	ErrNothingToSettle     = "invalid_batch_settlement_evm_nothing_to_settle"
-	ErrRefundNoBalance     = batchsettlement.ErrRefundNoBalance
+	ErrUnknownSettleAction  = "invalid_batch_settlement_evm_unknown_settle_action"
+	ErrNothingToSettle      = "invalid_batch_settlement_evm_nothing_to_settle"
+	ErrSettledEventMismatch = "invalid_batch_settlement_evm_settled_event_mismatch"
+	ErrRefundNoBalance      = batchsettlement.ErrRefundNoBalance
 )
--- a/go/mechanisms/evm/batch-settlement/facilitator/errors_test.go
+++ b/go/mechanisms/evm/batch-settlement/facilitator/errors_test.go
@@ -84,6 +84,7 @@ func TestExportedErrorReasonsAreStable(t *testing.T) {
 		"ErrAuthorizerAddressMismatch": ErrAuthorizerAddressMismatch,
 		"ErrUnknownSettleAction":       ErrUnknownSettleAction,
 		"ErrNothingToSettle":           ErrNothingToSettle,
+		"ErrSettledEventMismatch":      ErrSettledEventMismatch,
 		"ErrRefundNoBalance":           ErrRefundNoBalance,
 
 		// Permit2 deposit authorization errors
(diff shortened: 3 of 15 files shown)


0rkz:
Looked at this alongside the refund sibling (#3096). The fail-closed wiring is sound — the new check sits upstream of the single success return in all three languages, and Go's approach of re-reading `receivers.totalSettled` rather than decoding the event is a reasonable way to get the same guarantee. Two things worth a look before it ships.

**The Python settle path will reject every settle, not just the no-op.**

This PR doesn't change `_parse_settled_amount`'s body — it changes the docstring from "informational only" to "callers must fail closed" and adds the gate at `settle.py:95-96`. That promotion is what surfaces a pre-existing problem in the function:

```python
# settle.py:147
web3 = getattr(signer, "web3", None) or getattr(signer, "w3", None)
if web3 is None:
    return ""
```

The SDK's exported facilitator signer stores its web3 instance as the private `self._w3` (`mechanisms/evm/signers.py:286`) and exposes neither `web3` nor `w3` — its public attributes are `['address', 'get_addresses', 'get_balance', 'get_chain_id', 'get_code', 'read_contract', 'send_transaction', 'verify_typed_data', 'wait_for_transaction_receipt', 'write_contract']`. So with `FacilitatorWeb3Signer` the `getattr` pair resolves to `None`, the function returns `""` before inspecting any log, and the new gate converts that into `success: false` / `invalid_batch_settlement_evm_settled_event_mismatch`.

Before this PR that empty string was harmless — it just meant an absent `amount` on an otherwise successful response, exactly as the old docstring described. After it, it's a hard rejection of every settle, including ones where a correct `Settled` event was emitted on-chain.

Reproduced against this branch (`6c2f30e`): `_parse_settled_amount` with a real `FacilitatorWeb3Signer` and a well-formed `Settled` log returns `""`; the same log against a signer that does expose `.w3` returns `"999"`. So the matching logic itself is fine — the address-typed comparisons work correctly once a web3 instance is actually reachable. It's only the lookup that fails.

Simplest fix is to add `_w3` to the fallback chain, or better, expose a public `web3` property on `FacilitatorWeb3Signer` so both this and the refund-side copy in #3096 resolve. Worth a Python test that drives the function with a real signer — the current suite doesn't reach it, and the test plan lists vitest and `go test` only.

(The refund side has this same lookup plus a second decoder bug on top of it; details are on #3096.)

**Minor: the reported `Amount` can over-attribute under the exact race this PR defends against.**

`settle.go` captures `totalSettled` at line 29, before simulation and submission, and reports `Amount: postSettled - totalSettled` at line 117. If a second settle for the same `(receiver, token)` lands on-chain inside that window — the concurrent-settle scenario this PR exists to handle — this transaction's reported amount absorbs the other one's movement too. The boolean gate is unaffected and still correct, since any advance is a genuine advance; it's only the reported figure that can overstate. Worth a look if `Amount` is depended on downstream for accounting rather than just as a success signal.

*(Disclosure: I work on PayPerByte; we run a seller/facilitator stack on x402, so cross-SDK agreement on the fail-closed paths is the part I care about. Drafted with AI assistance.)*

SashaMIT:
Thanks for the careful read, especially the reproduction against FacilitatorWeb3Signer.

You are right: promoting the parse from informational to fail-closed made the missing web3/w3 lookup load-bearing. I rebased onto current main (`wait_for_receipt_and_build_response` / `waitAndReturnSettleResponse`) and:

1. Exposed `FacilitatorWeb3Signer.web3`, and also look up `_w3` so custom signers still decode.
2. Added a Python test that builds a real `Settled` log and asserts a positive match against `FacilitatorWeb3Signer`.
3. Kept the TS/Go fail-closed gates. Go still reports `Amount` as `postSettled - preSettled`. I left the concurrent-settle over-attribute as-is for this PR. The boolean gate is the load-bearing check. Happy to follow up if `Amount` is used for accounting.

The refund sibling (#3096) gets the same signer lookup plus the `bytes32` hex compare next.

0rkz:
Verified at head (`9e427f1dc555969d224089612a315a567833b928`), not just read. Same `FacilitatorWeb3Signer.web3` exposure as #3096 (identical `evm/signers.py` diff); `_parse_settled_amount`'s lookup is now `web3`/`w3`/`_w3` (`facilitator/settle.py`). Ran `test_settle_event.py` (not just read it) — real ABI-encoded `Settled` log against a real `FacilitatorWeb3Signer`, positive match, passes; the 390-test `tests/unit/mechanisms/evm/batch_settlement` suite passes at this head (unit only — didn't run the separate `tests/integrations` suite, which needs a live anvil/flask stack). Original bug is dead on this path: the getattr chain now finds the signer's web3, and the compare can match. Go's `Amount` is still `postSettled - totalSettled` (`facilitator/settle.go:110`; the pre-tx read is named `totalSettled` there, not `preSettled`) — the concurrent-settle over-attribution we flagged, which you deferred to a follow-up, is real and still open; not relitigating it here.

(Disclosure: I work on PayPerByte; we run a seller/facilitator stack on x402. Drafted with AI assistance.)

SashaMIT:
Thanks for verifying at head and running the settle event test, that is the bar I wanted. Original bug is dead on that path. The Go Amount over-attribute stays a follow-up, not this PR.

PR #3093, opened 2026-08-08
fix(extensions): stop misattributing bazaar ajv compile failures on Workers

### LINKED ISSUE #3029: @x402/hono 2.20: bazaar extension validation calls ajv.compile() at route-config time, which cannot run on Cloudflare Workers
## Summary

As of `@x402/hono` **2.20.0**, every `paymentMiddleware()` construction emits a `warn` + `error` pair **per gated route** on Cloudflare Workers. The AJV path used to validate bazaar discovery extensions calls `ajv.compile()`, which builds validators with `new Function()` — forbidden in the Workers runtime (`workerd`).

The extensions are **not actually invalid**. The validator simply cannot execute in the runtime, and the failure is reported as if the user's schema were at fault.

## Environment

- `@x402/hono` 2.20.0, `@x402/extensions` 2.20.0, `@x402/core` 2.20.0
- Cloudflare Workers (`workerd`), Hono
- 8 x402-gated routes, each carrying a bazaar discovery extension

## Symptom

Per route, per isolate:

```
warn   x402: Route "GET /paid/spotter-sample" has an invalid bazaar extension:
       Schema validation failed: Code generation from strings disallowed for this context
error  Error compiling schema, function code: const schema1 = scope.schema[1];
       const schema2 = scope.schema[2]; const schema3 = scope.schema[3];
       const schema5 = scope.schema[4].const...
```

Volume scales with isolate churn, not with traffic: in one 6-hour sample we logged **~1,200 warn/error pairs against ~1,356 requests**.

## Call chain

```
paymentMiddleware()
  └─ @x402/hono/dist/esm/index.mjs:147   validateBazaarRouteExtensions(httpServer.routes)
       └─ @x402/extensions .../chunk-KKZBRP7D.mjs:644-645
              const ajv = new Ajv({ strict: false, allErrors: true });
              const validate = ajv.compile(extension.schema);   // ← new Function()
```

The warn is emitted at `chunk-KKZBRP7D.mjs:848` / `:858`.

## This is new in 2.20, and the AJV call is not the new part

Worth stating precisely, because it makes the fix location non-obvious:

- **AJV and the identical `ajv.compile(ext.schema)` call already exist in `@x402/extensions` 2.10.0.**
- What 2.20 added is the **caller**: `validateBazaarRouteExtensions(httpServer.routes)` in `@x402/hono`. That symbol has **zero occurrences** in `@x402/hono` 2.10.0.

So the new behaviour is that the schemas are now compiled eagerly at **route-config time**, sweeping every route, rather than only on paths that happened not to run under Workers.

Confirmed by a before/after control on our own logs — same worker, same routes, same traffic shape:

| window | `[MCP]` requests | `invalid bazaar extension` |
|---|---|---|
| before the 2.10.0 → 2.20.0 upgrade | 200 | **0** |
| after | 200 | **200** |

A consequence: **downgrading `@x402/extensions` alone does not help**, because the call site lives in `@x402/hono`.

## Impact

Cosmetic, as far as we can tell — but noisy enough to bury real errors.

The handler only `console.warn`s: no `continue`, no removal of the extension, no throw. We verified all 8 bazaar extensions still ship intact in the 402 challenges, and a real purchase settled HTTP 200 with the warnings firing.

The practical cost is that Workers logs are swamped, and the message reads as a schema authoring bug, which sends you looking in the wrong place.

## Suggested fix

Any of:

1. **Guard the AJV path** where runtime code generation is unavailable, and skip validation with a single one-time notice rather than a per-route error.
2. **Ship a precompiled validator** (`ajv/dist/standalone`) for the bazaar extension schema, so no `new Function()` is needed at runtime.
3. At minimum, **distinguish "validator could not run" from "schema is invalid"** in the message, and emit it once per process rather than once per route per isolate.

(1) or (2) would make `@x402/hono` usable on Workers without log noise; (3) alone would at least stop the message misattributing the failure.

Happy to test a patch against a live Workers deployment if useful.

wakqasahmed:
## Description

Fixes #3029.

On Cloudflare Workers (`workerd`), `new Function()` is disallowed, so `ajv.compile()` throws when `validateBazaarRouteExtensions()` runs eagerly at route-config time (this eager call site is shared by `@x402/hono`, `@x402/express`, `@x402/fastify`, and `@x402/next`). The failure was caught and reported per-route as "invalid bazaar extension", which misattributes an environment limitation as a schema authoring bug and floods logs (~1 warn/error pair per gated route per isolate).

This PR:

- Adds a check in `validateDiscoveryExtension` (`@x402/extensions`) to detect Ajv's "code generation from strings disallowed" failure and report it as `unavailable: true` in the `ValidationResult`, distinct from an actually-invalid schema.
- Updates `validateBazaarRouteExtensions` to emit a single one-time notice for the whole process when this happens, and skip further schema validation for remaining routes, instead of repeating the misleading warn/error pair per route.
- Extensions are unaffected — they still ship exactly as declared; only the startup-time schema *check* is skipped when it can't run.

This corresponds to option 1 + 3 from the issue's suggested fixes (guard the AJV path and distinguish "validator could not run" from "schema is invalid", emitted once per process).

## Tests

- Added `typescript/packages/extensions/test/bazaarSchemaValidationUnavailable.test.ts`, which mocks `ajv/dist/2020.js` to throw the same error `workerd` produces, and verifies:
  - `validateDiscoveryExtension` returns `unavailable: true` (not just `valid: false`) for this case.
  - `validateBazaarRouteExtensions` warns exactly once across a config with multiple gated routes carrying the extension, instead of once per route.
- Ran `pnpm vitest run` in `typescript/packages/extensions` — all pre-existing tests pass (138 tests incl. the 2 new ones); one unrelated pre-existing failure in `test/integrations/builder-code.test.ts` reproduces identically on `main` without this change.
- `pnpm run build` succeeds.
- `pnpm exec eslint` clean on changed/added files.

## Checklist

- [x] I added a changelog fragment for user-facing changes (`typescript/.changeset/hono-bazaar-ajv-workers.md`, `@x402/extensions` patch)

---

**AI disclosure:** drafted with an AI coding assistant. Reviewed by me by actually running the extensions test suite (138 tests, including the 2 new ones) rather than relying on generated assertions alone.

### DIFF
--- a/typescript/.changeset/hono-bazaar-ajv-workers.md
+++ b/typescript/.changeset/hono-bazaar-ajv-workers.md
@@ -1,0 +1,5 @@
+---
+"@x402/extensions": patch
+---
+
+Fix bazaar extension schema validation logging a misleading per-route "invalid bazaar extension" warning (plus an Ajv compile error) on JS runtimes that forbid dynamic code generation, such as Cloudflare Workers. `validateDiscoveryExtension` now reports this as `unavailable` rather than `valid: false`, and `validateBazaarRouteExtensions` (used by `@x402/hono`, `@x402/express`, `@x402/fastify`, and `@x402/next`) emits a single one-time notice and skips further schema checks instead of repeating the error for every gated route.
--- a/typescript/packages/extensions/src/bazaar/facilitator.ts
+++ b/typescript/packages/extensions/src/bazaar/facilitator.ts
@@ -288,6 +288,26 @@ export function sanitizeResourceServiceMetadata(
 export interface ValidationResult {
   valid: boolean;
   errors?: string[];
+  /**
+   * True when validation could not run because the current JS runtime forbids
+   * dynamic code generation (e.g. Cloudflare Workers), rather than because the
+   * schema itself is invalid. Callers should not treat this the same as an
+   * actually-invalid extension.
+   */
+  unavailable?: boolean;
+}
+
+/**
+ * Detects Ajv's "code generation from strings disallowed" failure, which occurs
+ * when `ajv.compile()` runs in a JS runtime that forbids `new Function()` (e.g.
+ * Cloudflare Workers). This is an environment limitation, not a schema problem.
+ *
+ * @param error - The error thrown by `ajv.compile()`
+ * @returns True if the error indicates dynamic code generation is unavailable
+ */
+function isCodeGenerationDisallowedError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+  return /code generation from strings disallowed|unsafe-eval/i.test(message);
 }
 
 /**
@@ -372,6 +392,15 @@ export function validateDiscoveryExtension(extension: DiscoveryExtension): Valid
 
     return { valid: false, errors };
   } catch (error) {
+    if (isCodeGenerationDisallowedError(error)) {
+      return {
+        valid: false,
+        unavailable: true,
+        errors: [
+          "Schema validation is unavailable in this runtime (dynamic code generation is disallowed)",
+        ],
+      };
+    }
     return {
       valid: false,
       errors: [
--- a/typescript/packages/extensions/src/bazaar/startupValidation.ts
+++ b/typescript/packages/extensions/src/bazaar/startupValidation.ts
@@ -14,6 +14,13 @@ export { checkIfBazaarNeeded } from "@x402/core/server";
 const HTTP_VERB_RE = /^(GET|POST|PUT|PATCH|DELETE|HEAD)\b/i;
 
 /**
+ * Tracks whether we've already warned, for this process, that bazaar schema
+ * validation is unavailable (e.g. because dynamic code generation is disallowed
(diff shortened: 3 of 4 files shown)


wakqasahmed:
Hi @phdargen @CarsonRoscoe — noticed this PR doesn't have a reviewer assigned yet — it's been about 6 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

wakqasahmed:
Correction to my earlier comment — I said CI was green, but that was inaccurate; there's a failing Vercel deploy check on this PR. Looking into it now. Apologies for the noise.

wakqasahmed:
Looked into the failing Vercel check — it's a preview-deployment authorization gate scoped to the Coinbase Vercel team, not a build failure on this PR. Nothing in the code/CI config can resolve it from a contributor's side; it needs someone on the Coinbase team to authorize the deploy in Vercel's dashboard. Flagging in case that's easy to action on your end.

PR #2694, opened 2026-06-24
feat(svm): implementation reference for `upto` scheme 

lgalabru:
## Description

Adds a usage-based `upto` payment scheme for SVM, built on a payment-channels program `CHNLxYvVA28MJP9PrFuDXccuoGXAx7jBacfLEkahyGsX`.

  - upto scheme — new client / server / facilitator implementations under src/upto/. Lets a resource server
  charge up to a ceiling and settle the actual metered usage at the end, rather than a fixed amount per
  request.
  - Payment-channels client — Codama-generated client for the deployed payment-channels program (open,
  distribute, settleAndFinalize instructions, PDAs, codecs) plus hand-written open.ts / voucher.ts /
  onchain.ts helpers for channel lifecycle and off-chain voucher signing.
  - Example server — examples/typescript/servers/upto-svm demonstrating the scheme end-to-end.
  - Shared utils/types — channel + voucher types in src/types.ts, src/constants.ts, and helpers in
  src/utils.ts.

> Note: this branch is stacked on feat/svm-recent-blockhash. If this PR targets main, it also includes the
> server-provided recent-blockhash change to the exact scheme; if it targets the recent-blockhash branch,
> that commit is excluded.

> Spec is available in its dedicated [PR](https://github.com/x402-foundation/x402/pull/2697)

## Tests

New/expanded coverage:
  - test/unit/upto.test.ts (19 tests) — the new upto scheme: channel open, voucher build/sign, server
  challenge construction, facilitator verify/settle.
  - Additions to test/unit/index.test.ts, test/unit/server.test.ts, and test/unit/server.blockhash.test.ts.

  From typescript/packages/mechanisms/svm:

```console
  $ npx vitest run
  $ npx vitest run

   ✓ test/unit/upto.test.ts (19 tests)
   ✓ test/unit/index.test.ts (38 tests)
   ✓ test/unit/server.test.ts (17 tests)
   ✓ test/unit/facilitator.test.ts (17 tests)
   ✓ test/unit/smartWalletVerification.test.ts (33 tests)
   ✓ test/unit/server.moneyParser.test.ts (29 tests)
   ... (15 files total)

   Test Files  15 passed (15)
        Tests  221 passed (221)
```

All 221 tests pass, no regressions. (The Error decoding transaction line in the full-run logs is from an intentional invalid-input negative test in facilitator.test.ts, not a failure.)

## Checklist

- [x] I have formatted and linted my code
- [x] All new and existing tests pass
[...]

PR #3094, opened 2026-08-08
feat(ts): svm upto paymentflow

phdargen:
## Description

Continuing [#2694](https://github.com/x402-foundation/x402/pull/2694): rebased onto current main and aligned SVM upto with the payment flow model introduced in https://github.com/x402-foundation/x402/pull/3053

Feel free to force-push or retarget your PR @lgalabru 

---
pnpm test --testnet --families=svm --schemes=upto --min
```
✅ PASSED TESTS:
  # 1: typescript/http/fetch → typescript/http/hono → /upto/svm
      Facilitator: typescript
      Network: solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1
      Tx: 5GkTrWiUd1sidR8ASTTu11LbYRWthzNZhCRPcUJSy9sPMwUqghFbCkxUtBfSW9p9VfwmtmfH84reQjyr9XMCYBcc
  # 2: typescript/http/axios → typescript/http/next → /api/upto/svm/withx402
      Facilitator: typescript
      Network: solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1
      Tx: 3Am8rdKXgCZdZGDjdbNYMRwKtgnreCdqgyKFxCsQ47qRpC5QciWmDLMBTZ9u4DginKMfqMbXwfNuhXhzNfv2pKzF
  # 3: typescript/http/fetch → typescript/http/express → /upto/svm
      Facilitator: typescript
      Network: solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1
      Tx: 3hrPse3F8Qv7Ah9oaZRc2CCDw2qyPtRmNkg4t3zRBpVotcoABEhbG5BeMmWQEpq9Q89G2sgkgqPHFne1EbZciuAN
  # 4: typescript/http/axios → typescript/http/fastify → /upto/svm
      Facilitator: typescript
      Network: solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1
      Tx: M1EN5QWWNe7273SmjZuFFavTbt5mYDhfkFQzTjdLmp4hhqiSVeswwbSRPTJb8NsTRfsURZmRJdbAQaXC4Z97AK1
  # 5: typescript/mcp → typescript/mcp → upto_svm
      Facilitator: typescript
      Network: solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1
      Tx: PtkLwEqBW7SRofd4nNMXBEc2YfCr5AgwPoWTYaWm1vZ9ZzHrFPeuYhVFmWSeou5DRN56Z1f6vrEmoS4QjcNqvJF

📊 Breakdown by Facilitator:
 typescript      ✅ 5 / ❌ 0 (100%)

📊 Breakdown by Server:
 typescript/http/hono ✅ 1 / ❌ 0 (100%)
 typescript/http/next ✅ 1 / ❌ 0 (100%)
 typescript/http/express ✅ 1 / ❌ 0 (100%)
 typescript/http/fastify ✅ 1 / ❌ 0 (100%)
 typescript/mcp       ✅ 1 / ❌ 0 (100%)

📊 Breakdown by Client:
   typescript/http/fetch ✅ 2 / ❌ 0 (100%)
   typescript/http/axios ✅ 2 / ❌ 0 (100%)
   typescript/mcp       ✅ 1 / ❌ 0 (100%)
```


typescript/packages/mechanisms/svm/src/types.ts
```diff
@@ -12,3 +12,66 @@ export type ExactSvmPayloadV1 = {
...
+export function isUptoSvmPayload(payload: Record<string, unknown>): payload is UptoSvmPayloadV2 {
+  return (
+    typeof payload.from === "string" &&
+    typeof payload.maxAmount === "string" &&
+    typeof payload.deposit === "string" &&
+    typeof payload.channelId === "string" &&
+    typeof payload.authorizedSigner === "string" &&
+    typeof payload.openTransaction === "string" &&
+    typeof payload.openSlot === "string" &&
+    typeof payload.expiresAt === "number" &&
```
lgalabru: **P1 — Reject non-integral expiration timestamps before the deposit path.** `isUptoSvmPayload` accepts any `number`, and `validateOpenAuthorization` only compares it to `now`. A payload with `expiresAt: now + 300.5` therefore passes verification and opens the on-chain channel (the open transaction does not bind this field). The post-handler path then calls `BigInt(raw.expiresAt)` in `enrichSettlementPayload`, which throws; the cancel path hits the same conversion, leaving the deposited channel unable to settle normally. Require finite safe-integer Unix-second values (at least for `expiresAt`, ideally both timestamps) before opening, and add a fractional-timestamp regression test.

lgalabru (changes requested):
Requesting changes for the inline timestamp-validation issue: a fractional `expiresAt` can escrow the channel, then make both claim and cancel settlement throw during `BigInt` conversion. The SVM package tests otherwise pass after building the local `@x402/core` workspace dependency.

typescript/packages/mechanisms/svm/src/types.ts
phdargen: Good catch, fixed

lgalabru (approved):
LGTM, thanks for your support @phdargen !

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
```diff
@@ -0,0 +1,838 @@
...
+
+    const { p, channelConfig, feePayerSigner, maxAmount, tokenProgram } = auth.ctx;
+    const feePayer = channelConfig.feePayer;
+    const rpc = createRpcClient(requirements.network, this.config.rpcUrl);
+
+    // Escrow the ceiling before resource execution. Fresh opens: simulate
+    // open∥settle∥distribute before broadcast so settlement-account failures
+    // reject without locking the deposit. Already-open: settle-only sim after
+    // binding the confirmed account (composite open is not in that path).
+    const alreadyOpen = await channelExists(rpc, p.channelId);
```
CarsonRoscoe: This allows `alreadyOpen`  to re-bind and succeed when the channel already exists. Spec Phase 3 sanctions this, but the consequence is that one payload fired at N concurrent requests yields N served resources against a single deposit and one successful claim. Is per-payload single-use meant to be enforced upstream, or is the escrow ceiling considered sufficient?

Worth an answer in the spec either way

typescript/packages/mechanisms/svm/src/payment-channels/open.ts
```diff
@@ -0,0 +1,611 @@
...
+  ) {
+    throw new Error("verifyOpenTransaction: payment-channels instruction is not `open`");
+  }
+  const openIx = {
+    accountIndices: instruction.accountIndices ?? [],
+    data: instruction.data,
+  };
+
+  const indices = openIx.accountIndices;
+  if (indices.length < 14) {
```
CarsonRoscoe: Checks `indices.length < 14` but not `> 14`

The spec says the `open` instruction must "contain no remaining accounts". There are also no writable/signer privilege assertions on the 14 pinned slots. Both are defensive in depth, the program's own account constraints and the PDA rederivation will ensure we're safe. However the spec does ask for it, and the check is cheap, so I suggest adding it

typescript/packages/mechanisms/svm/src/payment-channels/open.ts
```diff
@@ -0,0 +1,611 @@
...
+  //   1. Reject Address Lookup Tables — they hide instruction programs/accounts
+  //      from `staticAccounts`. The in-SDK open builder never uses them.
+  //   2. Require exactly one payment-channels `open` instruction. Nothing else
+  //      gets the fee payer's signature or expands its fee exposure.
+  if (message.addressTableLookups && message.addressTableLookups.length > 0) {
+    throw new Error(
+      "verifyOpenTransaction: address lookup tables are not permitted in an open transaction",
+    );
+  }
+  if (message.instructions.length !== 1) {
```
CarsonRoscoe: The spec permits a ComputeBudget prefix and a Lighthouse suffix, which are what Phantom and Solflare inject. By requiring one instruction, we are passing the raw private key usecase seen in examples/e2e tests, but will break in production once wallets like Phantom or Solflare are used.

Please check how we handle this in Solana's exact and see if there's an easy fix to port over

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
```diff
@@ -0,0 +1,838 @@
...
+        payer: p.from,
+      };
+    }
+
+    // Claim only after the open channel is rebound. Concurrent or replayed
+    // settles for the same channel — including different valid amounts /
+    // vouchers — must fail after the first claim so only one settle_and_seal +
+    // distribute is submitted. Failures above (invalid voucher / not open) do
+    // not insert into the cache.
+    const settlementKey = `upto:${network}:${p.channelId}`;
```
CarsonRoscoe, line 565: `isDuplicate()` checks and inserts atomically, and nothing rolls it back if `submitSettle` thenf ails. A transient RPC error locks that channel out of claim settle for the full 120s TTL. With a short `maxTimeoutSeconds` the retry can land past `expiresAt` and the charge is lost

Consider evicting on submit failure

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
```diff
@@ -0,0 +1,838 @@
...
+          network: payload.accepted.network,
+          transaction: "",
+          errorReason: "invalid_upto_svm_settlement_simulation",
+          errorMessage: error instanceof Error ? error.message : String(error),
+          payer: p.from,
+        };
+      }
+    }
+
+    await this.upsertChannelStorage("settle", {
```
CarsonRoscoe: The insertion needed for rent cleanup occurs here, however `fetchAndVerifyOpenChannel` and `simulateZeroChargeSettle` can both throw, which would result in an early return. Both leave a channel PDA onchain, funded with facilitator rent, that no storage record points to, so in `clean()` we never discover it and the rent is unrecoverable.

A cheap fix is to upsert before `broadcastOpen`. Cleanup already deletes records whose PDA doesen't exist, so a pre-broadcast write is fail-safe in the other direction and costs nothing

typescript/packages/mechanisms/svm/src/upto/facilitator/rentCleanupManager.ts
```diff
@@ -0,0 +1,453 @@
...
+        const status = live.status as ChannelStatus;
+
+        if (status === ChannelStatus.Closing) {
+          continue;
+        }
+
+        if (status === ChannelStatus.Open || status === ChannelStatus.Sealed) {
+          if (status === ChannelStatus.Open) {
+            const firstSeenAtSecs = Math.floor(record.firstSeenAt / 1_000);
+            const readyAt = Math.min(
```
CarsonRoscoe: The `min` here is defensively correct. However the cost is that a resource server legitimately configuring say `maxTimeoutSeconds > 3600` gets its channel abandon-sealed at the 1h default while claims are still in the valid window, so `settle_and_seal(has_voucher=0)` freezes the watermark, the real claim then fails, and the server eats the work. A facilitator can't know its resource servers' timeouts.

Perhaps bind `expiresAt` to `maxTimeoutSeconds` at deposit settle and the cap can be derived per channel instead of a flat global?

This is a nit edge case but I think there's something here

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
phdargen: settle(deposit) should only return success once, same as after-handler settle. 
Changed to: 
- duplicate concurrent payloads in settle(deposit) -> reject with duplicate_settlement
- deposit tx already confirmed onchain (alreadyOpen) -> reject with invalid_upto_svm_channel_already_open

typescript/packages/mechanisms/svm/src/payment-channels/open.ts
phdargen: added optional Lighthouse + memo as in exact

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
phdargen: added delete to settle cache, called on failure

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
phdargen: moved `upsertChannelStorage` before broadcast

typescript/packages/mechanisms/svm/src/upto/facilitator/rentCleanupManager.ts
phdargen: facilitator now configures `maxChannelLifetimeSecs` (default 1h) that caps the `maxTimeoutSeconds` / `expiresAt` window accepted at deposit, so rent cleanup can safely abandon only after that accepted `expiresAt` (+ grace) instead of truncating mid-flight

CarsonRoscoe (approved):
LGTM! Ty @lgalabru and @phdargen

PR #3096, opened 2026-08-08
fix(batch-settlement): fail closed when refund emits no Refunded amount

SashaMIT:
## Summary
- On-chain `refundWithSignature` can early-return with receipt status success, advanced `refundNonce`, and **no** `Refunded` event when the capped transfer amount is zero.
- Facilitators previously returned `success: true` from analytic pre-state (TS/Python/Go), which can delete the channel while escrow remains.
- Require a positive matching `Refunded` event. New error: `invalid_batch_settlement_evm_refunded_event_mismatch`.

Sibling of #3092 (settle / `Settled` event fail-closed).

## Test plan
- [x] `pnpm exec vitest run test/unit/batch-settlement/facilitator.test.ts` in `@x402/evm` (49 passed), including fail-closed empty-logs case
- [x] `go test ./mechanisms/evm/batch-settlement/facilitator/ -run 'TestExecuteRefund|TestExportedError'` (includes `FailsClosedWithoutRefundedEvent`)


0rkz:
Cross-checked this against the sibling (#3092) since the two ship the same fail-closed shape in three languages. The wiring is right — in all six changed files there's exactly one success return and it sits downstream of the new check, so the fail-closed intent holds. But running the three refund-side decoders against identical synthetic receipts, the Python path never produces a match, so the new gate rejects every refund including fully valid ones.

**Python: the new gate rejects every refund, because `_parse_refunded_amount` can't reach a web3 instance.**

`python/x402/mechanisms/evm/batch_settlement/facilitator/refund.py:324`:

```python
web3 = getattr(signer, "web3", None) or getattr(signer, "w3", None)
if web3 is None:
    return ""
```

The SDK's own facilitator signer stores its web3 instance as the private `self._w3` (`mechanisms/evm/signers.py:286`) and exposes neither `web3` nor `w3`. Instantiating the exported `FacilitatorWeb3Signer` and listing its public attributes gives `['address', 'get_addresses', 'get_balance', 'get_chain_id', 'get_code', 'read_contract', 'send_transaction', 'verify_typed_data', 'wait_for_transaction_receipt', 'write_contract']` — no `web3`, no `w3`. So the `getattr` pair resolves to `None`, `_parse_refunded_amount` returns `""` at line 326 before looking at a single log, and the new check at `:278` turns that into `success: false` / `invalid_batch_settlement_evm_refunded_event_mismatch`.

What a maintainer would see: with `FacilitatorWeb3Signer` — the signer the package exports and the README documents — every `refundWithSignature` that lands on-chain, emits a correct `Refunded` event, and advances `refundNonce` still comes back as a failed settle response. Not just the zero-amount no-op case this PR targets. `execute_refund_with_signature` is the production path (`facilitator/scheme.py:174`), so this is what deployments hit.

Reproduced against this PR's branch (`b39aaac`): building a `Refunded` log from the real event ABI with the correct emitter, correct topic0, a real `compute_channel_id` value and a positive amount, then calling `_parse_refunded_amount` with an actual `FacilitatorWeb3Signer`, returns `""`.

**A second, independent bug sits right behind it.** If the signer *does* expose `.web3`/`.w3` (a caller-supplied one satisfying the `FacilitatorEvmSigner` protocol may well — the protocol doesn't declare either name), execution reaches `:344`:

```python
if str(log_channel).lower() == str(channel_id).lower() and int(amount) > 0:
```

web3.py decodes an indexed `bytes32` topic to a Python `bytes`, so `log_channel` is `b'4Vv\x4f8I...'` and `str()` renders it as `"b'4VvO8I\\x1b..."`, which can never equal the `"0x..."` string `compute_channel_id` returns. Checked on `web3==7.16.0` (the version `python/x402/uv.lock` resolves; the package itself declares `web3>=7.0.0`): a valid log decodes to `type(log_channel) == bytes` and the comparison evaluates `False`. Comparing hex forms on the same input evaluates `True`. So both bugs independently produce the same empty result, which is probably why one masked the other.

Minimal fix for both:

```python
web3 = getattr(signer, "web3", None) or getattr(signer, "w3", None) or getattr(signer, "_w3", None)
...
if Web3.to_hex(log_channel).lower() == channel_id.lower() and int(amount) > 0:
```

though exposing a public `web3` property on `FacilitatorWeb3Signer` is probably cleaner than reaching for `_w3` from another module.

Worth noting the same `getattr` line exists at `settle.py:147`, where it predates this work but was harmless while the result was documented "informational only" — #3092 makes it load-bearing, so the settle path inherits the same universal failure. I've left the details on that PR.

There's currently nothing in Python CI that would catch this. The test plan lists vitest and `go test` but no Python run; there's no `facilitator/test_refund.py` (only a client-side `client/test_refund.py`), and the two refund tests in `facilitator/test_scheme.py` both pass `_SignerNotExpected()` and assert `ERR_AUTHORIZER_NOT_CONFIGURED`, returning at `refund.py:169` well before any receipt parsing. A test that builds a real `Refunded` log and asserts a positive match against a real `FacilitatorWeb3Signer` would pin both bugs at once.

**Two smaller cross-language inconsistencies in the same event-matching code, checked across both PRs and all four Go/TS decoders:**

- Topic-array arity isn't fully validated against the event's declared indexed-parameter count. Go's guard is `len(log.Topics) < 2` (`refund.go:336`) and it only ever reads `Topics[0]` and `Topics[1]`, so a log with the `sender` topic entirely missing (2 topics instead of the real 3) matches, and so does one with an extra 4th topic appended. viem is stricter than Go here — it filters the under-arity log — but it also accepts the extra-topic one. web3.py rejects both outright with `LogTopicError`. Neither Go nor TS crashes; they just pattern-match on a shape no real `BatchSettlement` emission could produce, which means this needs a lying RPC or a compromised indexer to reach rather than being a spend path.
- Neither `abi.Arguments.Unpack` nor viem's event decoder enforces the declared `uint128` width. A 32-byte data word encoding `2**200` is returned verbatim as the refunded/settled amount by both (Go gives back the full 201-bit `big.Int`); web3.py rejects it with `NonEmptyPaddingBytes`. Same reachability caveat as above — the emitter-address check is intact in all three — but it's a real decode-time gap against Python's stricter behavior, and cheap to close with an explicit range check before trusting the unpacked value.

Separately, TS's channel-id match is a strict `log.args.channelId === channelId` (`refund.ts:368`) where Go compares via `common.Hash`, which is case-insensitive by construction. Likely unreachable today since viem renders decoded `bytes32` as lowercase hex — but line 53 of that same file already lowercases both sides for the claim channel comparison, so normalizing here too would just make the file self-consistent.

*(Disclosure: I work on PayPerByte; we run a seller/facilitator stack on x402, so cross-SDK agreement on the fail-closed paths is the part I care about. Drafted with AI assistance.)*

SashaMIT:
Thanks for the careful read, and for reproducing against FacilitatorWeb3Signer.

You are right on both Python bugs. Promoting the parse to fail-closed made the missing web3 lookup load-bearing, and `str()` on a decoded `bytes32` can never equal the hex `channel_id`. I rebased onto current main and:

1. Exposed `FacilitatorWeb3Signer.web3`, and also look up `_w3`.
2. Compare channel ids via `Web3.to_hex` so web3.py's `bytes` decode matches `compute_channel_id`.
3. Added a Python test that builds a real `Refunded` log and asserts a positive match against `FacilitatorWeb3Signer`.
4. Normalized the TS channel-id compare to lowercase.

I left the topic-arity and uint128-width decoder nits for a follow-up. They need a lying RPC to reach, and I did not want to mix them into this gate.

The settle sibling (#3092) already has the same signer fix.

0rkz:
Verified at head (`2c66d8c00e56f06b4a48a52a4e2389ad4972a612`), not just read. `FacilitatorWeb3Signer.web3` is exposed (`evm/signers.py:310-313`, returns `self._w3`); `_parse_refunded_amount`'s lookup is now `web3`/`w3`/`_w3` (`facilitator/refund.py`); the channel-id compare goes through `Web3.to_hex` via `_as_hex` before the lowercase match; TS's `facilitator/refund.ts` compares `log.args.channelId.toLowerCase() === channelId.toLowerCase()`. Ran `test_refund_event.py` (not just read it) — it builds a real ABI-encoded `Refunded` log via `eth_abi.encode`/`event_abi_to_log_topic` against a real `FacilitatorWeb3Signer` and asserts a positive amount match; it passes, and the 389-test `tests/unit/mechanisms/evm/batch_settlement` suite passes at this head (unit only — didn't run the separate `tests/integrations` suite, which needs a live anvil/flask stack). Original bug is dead on this path: the `getattr` chain now finds the signer's web3, and the compare can match.

(Disclosure: I work on PayPerByte; we run a seller/facilitator stack on x402. Drafted with AI assistance.)

SashaMIT:
Thanks for verifying at head and running the refund event test. Original bug is dead on that path.

Issue #839, opened 2025-12-22
x402 Escrow Scheme for Pre-funded, Usage-Based Payments


gigabit-eth:
# Problem

The `exact` scheme requires knowing payment amounts before each request. This doesn't work for:

* **Usage-based APIs:** LLM tokens, compute time, metered endpoints
* **Multi-step agent workflows:** Unknown request counts upfront
* **Session-based access:** Time-limited subscriptions with variable usage

## Proposed Solution

A new `escrow` scheme that enables:

1. **Deposit:** Client locks funds in an escrow contract
2. **Use:** Provider tracks consumption during the session
3. **Settle:** Funds distributed based on actual usage, remainder refunded

This implements the pattern mentioned in `scheme_exact_evm.md`:

> "In follow up leverage EIP-2612 + routing contract to support usage-based payments, and optionally bundle that with hard guarantees of payment by holding funds in escrow with the routing contract."

## Scheme Overview

**Lifecycle:**

```
DEPOSIT → LOCK → USE → SETTLE → DISTRIBUTE
```

**PaymentRequirements Example:**

```json
{
  "scheme": "escrow",
  "network": "eip155:84532",
  "escrow": "0x...",
  "minDeposit": "1000000",
  "asset": "0x036CbD53842c5426634e7929541eC2318f3dCF7e"
}
```

## Use Cases

1. **Metered APIs:** Pay for actual requests, not estimated usage
2. **LLM Inference:** Deposit covers token budget, settle based on consumption
3. **Agent Sessions:** Pre-fund a workflow, settle when complete

## Security Considerations

* Funds locked in audited escrow contract
* Neither party can unilaterally withdraw during session
* Time-locked auto-settlement prevents indefinite fund lockup
* Arbiter mechanism for usage disputes

## Reference Implementation

I have a working implementation deployed on **Base Sepolia**:

* **Escrow vault:** Internal balance accounting, lock/unlock
* **Session manager:** Lifecycle (open, record usage, settle)
* **Open source, MIT licensed**

**Repository:** [https://github.com/cartdotfun/evm-contracts](https://github.com/cartdotfun/evm-contracts)

## Proposed Spec Structure

```
specs/schemes/escrow/
├── scheme_escrow.md          # Scheme overview
├── scheme_escrow_evm.md      # EVM implementation
└── scheme_escrow_svm.md      # Solana (future)
```

## Contribution Plan

1. Discuss approach in this issue
2. Submit scheme spec PR based on feedback
[...]

PR #3098, opened 2026-08-08
docs: add upto scheme implementation spec for Stellar

### LINKED ISSUE #3097: Spec: upto scheme implementation for Stellar
## Problem

The `upto` scheme has network specs for EVM (`scheme_upto_evm.md`) and SVM (`scheme_upto_svm.md`), but none for Stellar. `exact` is specified for Stellar in `scheme_exact_stellar.md`, so metered billing is the remaining gap for Stellar-denominated agent payments.

## Why existing schemes don't suffice

`exact` requires the amount to be known at signing time. Metered services — per-token LLM billing, bandwidth, dynamic compute — only know the charge after the resource is consumed.

## Why a SEP-41 allowance is not enough

The obvious contract-free approach is `approve` / `transfer_from`. Measured against the four core `upto` properties it fails two:

| Core property | SEP-41 allowance | Conformant |
| --- | --- | --- |
| 4. Maximum amount | Allowance is a ceiling | Yes |
| 2. Time-bound | `expiration_ledger` | Yes (no `validAfter`) |
| 3. **Recipient binding** | `transfer_from` lets the spender pick any `to` | **No** |
| 1. **Single-use** | An allowance is a standing balance, drawable repeatedly | **No** |

Recipient binding is the decisive failure: it would let a compromised facilitator redirect funds.

## Proposed approach

A Soroban settlement contract, with `require_auth_for_args` restricted to the authorization struct so the settled amount stays outside what the buyer signs. A plain `require_auth()` would authorize the full argument list including the charge, collapsing `upto` into `exact`.

Also proposing an optional `smartAccount` profile where a buyer's custom account enforces the four properties in `__check_auth`, removing the intermediary contract for C-accounts.

## Status

Draft spec written, implementation in progress on testnet. Happy to open it as a Draft PR now for early feedback on the approach, or wait until the on-chain assumptions are verified — whichever the maintainers prefer.

Disclosure: drafted with AI assistance, reviewed and being verified by hand against live testnet behaviour.

Eras256:
## Summary

Adds the Stellar network implementation spec for the `upto` scheme, completing the set alongside `_evm` and `_svm`.

## Why a contract

SEP-41 allowances satisfy the maximum-amount and time-bound properties but fail recipient binding (`transfer_from` accepts any destination) and single-use (an allowance is a standing balance). Both are `MUST` requirements in `scheme_upto.md`, so a contract-free design is non-conformant and this spec says so explicitly.

## Key design point

`require_auth_for_args` restricted to `(authorization,)` keeps `actual_amount` outside what the buyer signs. A plain `require_auth()` authorizes the full argument list including the charge, which would force the buyer to know the final amount at signing time and collapse `upto` into `exact`.

Two profiles: `contract` (default, works with G- and C-accounts, atomic pull-and-refund) and `smartAccount` (optional, C-accounts only, single transfer, policy enforced in `__check_auth`).

## Status

Draft. Three on-chain assumptions are marked in the spec with their verification status; each will be closed with a testnet transaction hash before this leaves Draft. Reference implementation in progress.

## AI disclosure

Drafted with AI assistance, then reviewed and verified by hand. The Soroban authorization semantics and resource-limit claims are being confirmed against live testnet transactions rather than asserted — the three open items above track exactly that. Flagging per CONTRIBUTING.md so reviewers can calibrate.

## Related

Closes #3097

Context: SCF RFP "X402 Facilitator with Bazaar (discovery) support" (Q3 2026), which names this spec as a deliverable. Implementation: https://github.com/Eras256/Periplo

### DIFF
--- a/specs/schemes/upto/scheme_upto_stellar.md
+++ b/specs/schemes/upto/scheme_upto_stellar.md
@@ -11,28 +11,6 @@
 - `stellar:pubnet`: Stellar mainnet
 - `stellar:testnet`: Stellar testnet
 
-> [!IMPORTANT]
-> **Verification status (`contract` profile).** All three properties below
-> are closed against real `stellar:testnet` behaviour, each backed by a
-> settled transaction hash, not simulation alone. Full evidence:
-> [`conformance/RESULTS.md`](https://github.com/Eras256/Periplo/blob/main/conformance/RESULTS.md)
-> in the reference implementation, settled tx
-> [`cc46374e...`](https://stellar.expert/explorer/testnet/tx/cc46374e34f70ff479ccf919d55df33d0bf1a05e1c7479fa8f90dac596c5d218).
->
-> 1. `require_auth_for_args` accepts a root argument tuple of `(authorization,)`
->    while the [SEP-41] `transfer` rides as a sub-invocation for `max_amount`.
->    **Closed.** Confirmed via `inspectAuthEntry` on a real simulation: root
->    call `argCount=1`, one sub-invocation.
-> 2. The pull-and-refund sequence fits within Soroban's per-transaction read,
->    write, instruction and memory limits. **Closed.** `2,026,530` instructions
->    of a `400,000,000` ceiling, `392` read bytes and `680` write bytes.
-> 3. `temporary()` storage TTL can always cover `deadline_ledger − current_ledger`
->    under the `MAX_WINDOW_LEDGERS` bound. **Closed.** The settled nonce
->    entry's `liveUntilLedgerSeq`, read back from RPC, exceeded `deadline_ledger`.
->
-> `extra.uptoProfile` and `extra.settlementContract` are proposed field names.
-> Happy to align with whatever convention the maintainers prefer.
-
 ## Summary
 
 `upto` on Stellar authorizes a transfer of **up to** a maximum amount, with the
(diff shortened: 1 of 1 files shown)


Eras256:
## Update: all three on-chain assumptions closed with live testnet evidence

The reference implementation (`contracts/upto-settlement` in
[Eras256/Periplo](https://github.com/Eras256/Periplo/tree/main/contracts/upto-settlement))
is built, tested (27 unit + property tests, plus a `cargo-fuzz` target that
ran 47,630 executions against the ceiling/time-bound arithmetic with zero
crashes), deployed to `stellar:testnet`, and has settled a real
transaction. That run closes each of the three assumptions this PR marks
open. Independently verifiable, not just asserted:

### 1. `require_auth_for_args` accepts a root tuple of `(authorization,)` while the SEP-41 `transfer` rides as a sub-invocation for `max_amount`

Read directly off a real testnet simulation via `inspectAuthEntry`
(`@stellar/stellar-sdk/base`), not asserted from the pseudocode:

- Buyer's auth entry: `credentialType=address`, root call
  `fn=settle argCount=1` — the authorized tuple is `(authorization,)`
  only, `actual_amount` never appears in it.
- Exactly one sub-invocation: `transfer(from, settlementContract,
  max_amount)` on the asset contract.
- Facilitator's auth entry: `credentialType=sourceAccount` — satisfied by
  being the submitting transaction's source account, no separate signed
  entry needed.

Reproducible with
[`apps/facilitator/scripts/upto-settle-demo.ts`](https://github.com/Eras256/Periplo/blob/main/apps/facilitator/scripts/upto-settle-demo.ts),
which prints this structure on every run.

### 2. Pull → pay → refund fits Soroban's per-transaction resource limits

Real simulation numbers against testnet's live limits
(`stellar network settings --network testnet`), not estimated:

| Resource | Used | Testnet ceiling |
| --- | --- | --- |
| CPU instructions | 2,026,530 | 400,000,000 |
| Read bytes | 392 | 200,000 |
| Write bytes | 680 | 132,096 |
| Min resource fee | 162,433 stroops | — |

Well under 1% of the instruction ceiling for a 3-transfer settlement
(pull, pay, refund).

### 3. `temporary()` TTL covers `deadline_ledger − current_ledger`

Read back from RPC after settlement, not just computed:

- `deadline_ledger` at settle time: `4106782`
- Nonce entry's actual `liveUntilLedgerSeq`: `4107472`
- `4107472 ≥ 4106782` — the TTL `extend_ttl` requested at settle time
  covers the full window to deadline, confirmed against the live ledger
  entry rather than assumed from the `extend_ttl` call succeeding.

### Settled transaction

A genuine **partial settlement** — buyer signed a `0.1 PTEST` ceiling,
facilitator settled `0.04 PTEST` — exercising both the payout and refund
legs in one real transaction, not a full-payout-only test:

- Hash:
  [`cc46374e34f70ff479ccf919d55df33d0bf1a05e1c7479fa8f90dac596c5d218`](https://stellar.expert/explorer/testnet/tx/cc46374e34f70ff479ccf919d55df33d0bf1a05e1c7479fa8f90dac596c5d218)
- Independently checked against Horizon (not just the script's own
  output): `successful: true`, `source_account` = the facilitator's
  fee-sponsor address (never the buyer or seller), `fee_charged: 42872`
  stroops.
- [`/effects`](https://horizon-testnet.stellar.org/transactions/cc46374e34f70ff479ccf919d55df33d0bf1a05e1c7479fa8f90dac596c5d218/effects)
  shows the full pull-and-refund sequence: buyer debited `0.1 PTEST`,
  seller credited `0.04 PTEST` (the actual charge), buyer credited back
  `0.06 PTEST` (the refund) — the exact arithmetic this doc's
  `settle` pseudocode describes.
- A separate post-hoc query
  (`stellar contract invoke ... -- balance --id <contract>`) confirms the
  contract's own SEP-41 balance is `0` after settlement — the
  "never holds a balance across transactions" invariant this doc's
  Security considerations section names.

Contract address: `CAK3R734WLT4JU2XMQOJ6NIB3BWGPI442CH44EFJG5AORMXFE7G4MQFW`.
WASM hash: `3f4df3070459047e52a5514a4bd42f31888d100cce61914d8675534eb20dfe07`.

Full write-up, including two harness bugs the fuzzer found along the way
(neither a contract bug), is in
[`docs/DEFERRED.md`](https://github.com/Eras256/Periplo/blob/main/docs/DEFERRED.md)'s
"Phase 6" section, and the settled-transaction table is in
[`conformance/RESULTS.md`](https://github.com/Eras256/Periplo/blob/main/conformance/RESULTS.md).

One scope note: this closes the three *on-chain assumptions* this PR
marks open — it is not a third-party security audit. The contract has
had an internal review only so far; a Soroban Audit Bank review (or
equivalent) is still the right gate before any production/mainnet use,
and that's tracked as open in the implementation repo, not implied
otherwise here.

autolayer-labs:
AutoLayer is implementing the Stellar facilitator/Bazaar side for the SCF x402 RFP. We found this PR and #3134 while preparing an upstream contribution and will not open a third competing spec PR.\n\nFor downstream interoperability, we can implement and test the selected profile across @x402/stellar client/server/facilitator registration, /supported, settlement overrides, and Bazaar catalog/search. The proposed extra.uptoProfile discriminator is necessary: catalog deduplication and /supported matching must include the profile and settlementContract, otherwise two valid Stellar upto profiles sharing scheme/network/asset/payTo can collide.\n\nWe have a testnet deployment harness and can contribute canonical G-account/C-account, over-cap, expiry, replay, and distinct payer/facilitator/payTo fixtures once maintainers indicate which profile shape to target. Happy to coordinate an implementation PR against the converged spec rather than duplicating this document.

Eras256:
Thanks for confirming `extra.uptoProfile` independently. It matches what we found checking our own implementation against this same spec text: HeylmStoned raised the same discrimination concern on #3134, and once we actually checked our code (not just the spec) against it, we found three real gaps of our own, `/supported` doesn't report `upto` in any form yet, there's no `GET /discovery/resources`/`GET /discovery/search` route at all, and our catalog's dedupe key isn't `extra`-aware, so two profiles for the same resource/network/asset/payTo would currently collide rather than coexist. Recorded at [docs/DEFERRED.md](https://github.com/Eras256/Periplo/blob/main/docs/DEFERRED.md#upto-profile-discrimination-three-real-implementation-gaps-found-responding-to-external-review-not-self-discovered), not fixed yet.

On the converged spec PR: honest status, there isn't one yet. It's still waiting on maintainer input on where this actually consolidates, into #3098 directly, into #3134, or as a new PR, and we didn't want to preempt that call by opening a third document before it's clear which one it should build on. Once that's settled we'll open it and would welcome your fixtures (G-account/C-account, over-cap, expiry, replay, distinct payer/facilitator/payTo) against it rather than duplicating test coverage.

Would appreciate maintainer direction on where this should consolidate, into #3098, into #3134, or as a fresh PR, before we open the merged two-profile document, now that a third team (AutoLayer) is also waiting on that call.

Eras256:
Complementary, not competing, and consistent with where you already put it on #72: discovery layer, not the core `upto` scheme. Reading the actual design now that #3181 is out of draft (today).

One honest correction to our own speculation in that same #72 comment: `pricing` sits as a sibling of `info`/`schema` on the bazaar extension itself, resource-level, not per-`accepts`-entry. That means it doesn't end up giving our dedupe key (`extra.uptoProfile`, per-entry, distinguishing `contract` from `stateless` for the same resource/network/asset/payTo) anywhere new to pick up a discriminator from, the two live at different levels of the structure. Not a gap in the design, just narrower overlap with our dedupe problem than we speculated back then. `extra.uptoProfile` stays our own answer to that specific problem, unrelated to whether `pricing` lands.

Your point 3 (settlement-count ranking stops being comparable once amounts vary, and gaming it gets cheaper the smaller the settlement) is the same integrity thread one step past something we just closed on our own catalog: cataloging used to run on `/verify`, not just `/settle` (`isValid: true` proves a payload could settle, not that it did), which #3226 is auditing right now. Fixed to settle-only; pedro-pelicioni's stellarsight independently reached the same reading. Different failure mode, same family: a catalog signal meant to say something real needs the thing it's built on to be real first.

Would like to wire `pricing` into our own catalog once this merges and report back with real numbers rather than just agreeing it's a good idea.

Eras256:
Picking this back up after the structural direction above — updated this branch to actually implement it, not just agree to it in prose.

## What changed

`scheme_upto_stellar.md` now reads as one consolidated document instead of `contract` full + `stateless` summarized-with-a-pointer:

- **`stateless` is presented first, as the base profile**, with its full protocol flow, wire formats, contract interface, and facilitator verification rules pulled inline from `#3134` rather than summarized — crediting @Iam0TI and `0d1026/Rialto` throughout, since this is now this document's own authoritative copy of that content, not a paraphrase.
- **`contract` follows as the secondary, stateful profile**, keeping its verification-status callout (real settled-tx evidence for all three of its closed properties) and gaining an explicit TTL-boundary test vector with concrete numbers against the real deployed contract (`MAX_WINDOW_LEDGERS = 17,280`, independently checked against the network's own real `state_archival.max_entry_ttl`) — this is the "carrying the required test vectors" half of the proposed structure.
- **`settlement_pending` handling added to both profiles**, mirroring `scheme_upto_evm.md`'s own addition from `#3083` (checked against current upstream `main`, which this branch's base predates — that commit never touched this file, so no conflict, just a real gap this branch hadn't picked up yet).
- **A real, new finding, applying to both profiles equally**: a shared "C-account support" section now states plainly that "C-accounts work transparently" is verified only for a directly-signing custom account. It is *not* verified for a delegated/session-key smart-account signer — the pattern most relevant to autonomous-agent payments, where an agent holds a scoped key and the account's own policy contract is meant to gate what it can spend.

## The C-account finding, in detail

Building a real scenario against this exact pattern (an agent key wrapped in OpenZeppelin's `stellar-accounts` `SmartAccount`, scoped via a `ContextRule` to spend only through one contract) traps `__check_auth` with `UnreachableCodeReached` on every construction tried — both `Signer::Delegated` and `Signer::External` — against both a real settlement-style target contract and a trivial single-line probe contract used specifically to rule out target-contract complexity as the cause. Seven concrete hypotheses ruled out with real evidence (encoding method, nonce reuse, nested-entry presence, `AuthPayload` content, `soroban-sdk` version alignment, target-contract complexity, signer type), and `stellar-accounts` itself confirmed to have no test coverage of this real, host-driven `require_auth_for_args` path in either the crate or its own official example. Filed as a diagnostic request at [OpenZeppelin/stellar-contracts#839](https://github.com/OpenZeppelin/stellar-contracts/issues/839), open, no maintainer response yet.

Not reopening or adding a new angle to `#839` here — that round is closed on our side. This is a different, new action: citing its already-public findings in the venue where they're actually load-bearing, since both `#3098` and `#3134` make a general "C-accounts supported" claim that this specific case doesn't yet back up. Since `require_auth_for_args` is the one mechanism both profiles depend on to make `upto` expressible on Soroban at all, this isn't specific to either profile — it's a property of the shared underlying primitive, evaluated against the most widely used Soroban smart-account library, for the one C-account pattern (delegation) neither existing spec draft distinguishes from the directly-signing case it actually tested.

Concretely, the ask: whichever document this consolidates into should carry this caveat rather than the current unqualified "dispatch transparently to the C-account's own `__check_auth`" language, so a reviewer implementing against the merged spec doesn't assume delegated-signer support that hasn't actually been demonstrated by anyone yet.

Full branch: [`docs/upto-stellar-spec`](https://github.com/Eras256/x402/tree/docs/upto-stellar-spec), commit [`6a528a5a`](https://github.com/Eras256/x402/commit/6a528a5acb72a90d95ce3c027da6733844c7c763).

PR #2658, opened 2026-06-17
feat: improve & document wallet compatibility

CarsonRoscoe:
### Summary
Makes x402 pre-verification match on-chain signature verification for the full range of wallet types (plain EOAs, deployed smart accounts, counterfactual ERC-6492 wallets, and ERC-7702-delegated EOAs), so a payment that passes `verify` is the same one that succeeds at `settle`. Implements a strict, code-routed signature primitive (ecrecover for no-code addresses, EIP-1271 for any address with code — no ECDSA fallback) that mirrors Permit2 / USDC v2.2 / OpenZeppelin `SignatureChecker`, and documents exactly which wallet × scheme combinations are supported and why. Covers the `exact` (EIP-3009 + Permit2) and `batch-settlement` flows across the Go, TypeScript, and Python SDKs.

### What changed
- **Strict signature verification primitive** (`VerifySignatureStrict` / `verifyTypedDataSignature` / `verify_typed_data_strict`): routes by `code.length` to match on-chain `SignatureChecker`, removing the old 65-byte ECDSA fast-path that caused pre-verify/on-chain divergence for ERC-7702-delegated EOAs.
- **ERC-7702 detection helpers** (diagnostic only — the verify path routes by code length, not 7702 detection).
- **ERC-6492 counterfactual flows**: deploy-then-validate via a single Multicall3 `eth_call` in verify, factory-allowlist gating, and a dedicated retry-guidance error when a freshly deployed wallet rejects the inner signature (lazy validator install). The factory allowlist is now a first-class config option on `ExactEvmSchemeWithConfig`, `BatchSettlementEvmSchemeWithConfig`, and their Python/TS equivalents.
- **Batch-settlement voucher dual-path**: ECDSA against `payerAuthorizer` when set, code-routed `SignatureChecker` against `payer` otherwise.
- **Docs**: new `docs/advanced-concepts/wallet-compatibility.mdx` with the full support matrix, routing rules, and footnoted unsupported combinations.
- **Tooling/examples/tests**: ERC-7702 client example, a smart-account setup script (`setup-smart-accounts.ts`), a cross-SDK wallet-compatibility integration test matrix, and dedicated ERC-4337 (Coinbase Smart Wallet) and ERC-7579 (Biconomy Nexus) integration tests in Go, TypeScript, and Python.

### Fixes folded into this PR
[...]

PR #3101, opened 2026-08-08
fix(evm): classify reverted verify simulations in Go and TypeScript (Python parity)

shuhei0866:
## Summary

`ErrEip3009SimulationFailed` carries two incompatible meanings in the EVM `exact` verify path.

#2062 introduced it for a payload that is valid but whose simulation *could not run* — a caller may retry. It is also what verify returns when the simulation *did* run and reverted, which is terminal: the same payload is rejected on every attempt. Callers cannot tell the two apart, so a client that retries on the documented meaning retries a payment that can never succeed.

`settleEIP3009` already maps revert reasons to specific codes through `parseEIP3009TransferError`, and the Python facilitator already applies that classifier in verify. #2658 shipped the revert heuristic to all three SDKs (`evm.IsContractRevert`, `isContractRevert`, `is_contract_revert`) but wired it into the verify path only in Python. This brings Go and TypeScript to that behaviour.

## Changes

Go and TypeScript, on a failed transfer simulation:

1. If the failure is a recognised on-chain revert, report the revert's own code (`parseEIP3009TransferError` / `parseEip3009TransferError`, guarded by the revert heuristic).
2. Otherwise fall back to the existing diagnostic probe.
3. Transport failures keep `ErrEip3009SimulationFailed`.

**One deliberate deviation from Python:** the diagnostic probe now runs only when the node answered — a revert, or a failure the simulation reported without an error. Python probes unconditionally, which after a transport failure sends four more reads down the same unreachable path. Python could adopt the same guard as a follow-up. In TypeScript this also narrows one existing case: a transient failure that previously reached the probe now returns the retryable code without probing.

## Reproduction

Any EIP-3009 token whose `transferWithAuthorization` verifies with `ecrecover` only rejects every contract payer, regardless of configuration. The `exact` scheme supports arbitrary EIP-3009 assets, and such a token is in production use today: JPYC on Polygon (`0xe7c3d8c9a439fede00d2600032d5db0be71c3c29`), whose implementation contains no `isValidSignature`, `SignatureChecker` or ERC-1271 reference.

Against a mainnet fork, with a minimal ERC-1271 stub as the payer, funded, nonce unused, name matching:

```
isValidSignature(hash, sig)      -> 0x1626ba7e     # off-chain check accepts the payer
balanceOf(payer)                 -> 1000e18        # no ErrInsufficientBalance
name()                           -> "JPY Coin"     # no ErrEip3009TokenNameMismatch
authorizationState(payer, nonce) -> false          # no ErrNonceAlreadyUsed
transferWithAuthorization(...)   -> revert: EIP3009: invalid signature
```

Every diagnostic probe passes and only the simulation fails, so before this change verify reported the retryable code and nothing else.

To reproduce with foundry alone, no funds and no API keys: fork Polygon, deploy the 16-byte stub `0x631626ba7e60e01b60005260206000f3` (returns the ERC-1271 magic value for any input) as the payer, set its balance with `anvil_setStorageAt` on `cast index address <stub> 516`, then `cast call` the four probes and the transfer above.

The default assets are not affected: probing all nine reachable mainnet EIP-3009 entries in `DEFAULT_STABLECOINS` (USDC on Base/Polygon/Arbitrum/Celo/Monad, USDT0 on Flare/Stable, Bridged USDC on XDC/HPP) with the same stub shows all of them code-route. The exposure is in the arbitrary-asset path.

## Observed, not changed

Pre-existing differences found while working on this. Each would also change settle's classification, so they are out of scope here:

- Go's `parseEIP3009TransferError` matches case-sensitively; the TypeScript and Python parsers do not.
- TypeScript's patterns are broader than Go's and Python's (`insufficient.*balance`, `valid before`).
- The default code for an unrecognised transfer error differs on the wire: Go `invalid_exact_evm_failed_to_execute_transfer`, TypeScript `invalid_exact_evm_transaction_failed`.

## Test plan

Three cases per SDK, each covering a distinct branch:

- a classified revert reports the revert's code, not the retryable one;
- a transport failure keeps the retryable code and does not probe;
- an unrecognised revert still probes.

Each was confirmed to fail with the fix reverted. `go test ./...` passes with `go vet` clean; `vitest run` passes 829 tests in `@x402/evm` with `lint:check` clean. The Python facilitator is unchanged.

---

*Disclosure per CONTRIBUTING: written with significant AI assistance under human direction and review. The defect was located by reading the merged Python implementation and #2062/#2658 rather than from memory, the reproduction was run against a Polygon mainnet fork, and each test was confirmed to fail with the fix reverted.*

### DIFF
--- a/go/.changes/unreleased/fixed-20260808-225239.yaml
+++ b/go/.changes/unreleased/fixed-20260808-225239.yaml
@@ -1,0 +1,3 @@
+kind: fixed
+body: 'Bring the Go exact/EVM verify path to parity with Python: when the EIP-3009 transfer simulation fails, classify an on-chain revert through parseEIP3009TransferError (guarded by evm.IsContractRevert) before falling back to the diagnostic probe, instead of returning invalid_exact_evm_transaction_simulation_failed for every simulation failure. That code reports a payload that is valid but whose simulation could not run, so a caller may retry it; a revert is terminal and the same payload is rejected on every attempt. The diagnostic probe now runs only when the node answered, so a transport failure no longer triggers four more reads down the same dead path.'
+time: 2026-08-08T22:52:39.000000+09:00
--- a/go/mechanisms/evm/exact/facilitator/eip3009.go
+++ b/go/mechanisms/evm/exact/facilitator/eip3009.go
@@ -122,27 +122,46 @@ func (f *ExactEvmScheme) verifyEIP3009(
 	}
 
 	if simulate {
-		simulationSucceeded, err := SimulateEIP3009Transfer(
+		simulationSucceeded, simErr := SimulateEIP3009Transfer(
 			ctx,
 			f.signer,
 			tokenAddress,
 			parsedAuthorization,
 			classification.SigData,
 		)
-		if err != nil {
-			return nil, x402.NewVerifyError(ErrEip3009SimulationFailed, evmPayload.Authorization.From, err.Error())
-		}
-		if !simulationSucceeded {
-			reason := DiagnoseEIP3009SimulationFailure(
-				ctx,
-				f.signer,
-				tokenAddress,
-				evmPayload.Authorization,
-				requiredValue,
-				tokenName,
-				tokenVersion,
-			)
-			return nil, x402.NewVerifyError(reason, evmPayload.Authorization.From, "")
+		if simErr != nil || !simulationSucceeded {
+			// Prefer the concrete on-chain revert reason the simulation surfaced (e.g.
+			// insufficient balance / used nonce) over the opaque generic code. Fall back
+			// to a diagnostic probe only when the revert could not be classified.
+			// ErrEip3009SimulationFailed reports that the payload was valid but the
+			// simulation could not run, so a caller may retry it; a revert is terminal and
+			// must not share that code. Mirrors the Python facilitator.
+			reason := ErrEip3009SimulationFailed
+			reverted := evm.IsContractRevert(simErr)
+			if reverted {
+				if mapped := parseEIP3009TransferError(simErr); mapped != ErrFailedToExecuteTransfer {
+					reason = mapped
+				}
+			}
+			// Probe only when the node answered: either the call reverted, or the simulation
+			// reported failure without an error at all. Probing after a transport failure
+			// sends four more reads down the same dead path.
+			if reason == ErrEip3009SimulationFailed && (reverted || simErr == nil) {
+				reason = DiagnoseEIP3009SimulationFailure(
+					ctx,
+					f.signer,
+					tokenAddress,
+					evmPayload.Authorization,
+					requiredValue,
+					tokenName,
+					tokenVersion,
+				)
+			}
+			message := ""
+			if simErr != nil {
+				message = simErr.Error()
+			}
+			return nil, x402.NewVerifyError(reason, evmPayload.Authorization.From, message)
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


chopmob-cloud:
I verified this against main before commenting.

I confirmed the defect the PR targets is real. Before this change, the Go and TypeScript
exact/EVM verify paths collapse every simulation failure to a single code. In Go,
`verifyEIP3009` returned `ErrEip3009SimulationFailed` on any `SimulateEIP3009Transfer`
error, and ran the diagnostic probe on any non-error failure. That code means "the payload
looked valid but the simulation could not run", which invites a retry. An on-chain revert
(used nonce, insufficient balance, a token that only accepts ecrecover so it rejects every
contract payer) is terminal: the same payload is rejected on every attempt, so labelling it
retryable is wrong and can drive an unbounded retry loop against a payment that can never
settle.

I confirmed the "Python parity" claim is accurate. `python/x402/mechanisms/evm/exact/facilitator.py`
already does exactly this: it sets `reason = ERR_TRANSACTION_SIMULATION_FAILED`, upgrades to
a concrete reason via `parse_eip3009_transfer_error` only when `is_contract_revert(sim_error)`,
then falls back to `diagnose_eip3009_simulation_failure` if still unclassified. The Go and TS
here mirror that control flow line for line (`IsContractRevert` / `isContractRevert`, the
`parseEIP3009TransferError` / `parseEip3009TransferError` upgrade guarded against the generic
`ErrFailedToExecuteTransfer` / `ErrTransactionFailed`, then the probe).

One substantive point on the EVM verify behaviour. The Go and TS versions add a guard the
Python reference does not have: they run the diagnostic probe only when the node actually
answered, `if reason == ErrEip3009SimulationFailed && (reverted || simErr == nil)` in Go, and
the `else if (reverted || !simError)` branch in TS. That correctly suppresses four extra
reads down a dead path after a transport failure, and the added tests pin it
(`TestVerifyEIP3009_TransportFailureKeepsRetryableReasonWithoutProbing` asserts
`diagnosticCalls == 0`). Worth flagging that the Python path this PR claims parity with does
not have that guard: `facilitator.py` runs `diagnose_eip3009_simulation_failure`
unconditionally whenever the reason is still the generic code, including after a transport
failure where `is_contract_revert` is false. So Go and TS end up strictly better than
Python here. If parity is the goal, either fold the same node-answered guard into the Python
verify path, or note in the PR that the guard is a deliberate improvement over the reference
rather than a mirror of it, so the three stay aligned.

Nice touch carrying the raw revert text through as `invalidMessage` so the concrete reason
survives the mapping to a code.

shuhei0866:
Thank you for the careful read — tracing the control flow against `facilitator.py` and confirming that the added tests actually pin the behaviour is real work, and I appreciate you doing it.

On the substantive point: you are right that the node-answered guard has no counterpart in the Python reference, and that "parity" on its own would not justify it. That is why I framed it as a deviation rather than a mirror in the description:

> **One deliberate deviation from Python:** the diagnostic probe now runs only when the node answered — a revert, or a failure the simulation reported without an error. Python probes unconditionally, which after a transport failure sends four more reads down the same unreachable path. Python could adopt the same guard as a follow-up.

The Python guard itself is still genuinely open. I left it out because folding it into `facilitator.py` changes behaviour in a third SDK that is otherwise untouched here, and a behaviour change deserves its own diff and its own test rather than riding along with this one.

And thank you for the note on `invalidMessage`. A reason code buckets by construction, so keeping the revert string beside it is what lets whoever is debugging a rejected payment see which revert it actually was. In TypeScript it rides along even when the diagnostic probe supplies the code, so the probe's answer and the on-chain text can be read against each other.

PR #3105, opened 2026-08-09
fix(evm): correct Monad USDC EIP-712 domain name to "USDC" (#3102)

### LINKED ISSUE #3102: exact-EVM: eip155:143 (Monad) EIP-712 domain name is "USD Coin"; the deployed USDC reports "USDC"
In `go/mechanisms/evm/constants.go`, the `"eip155:143"` (Monad mainnet) entry declares:

```go
"eip155:143": {
    ChainID: ChainIDMonad,
    DefaultAsset: AssetInfo{
        Address:  "0x754704Bc059F8C67012fEd69BC8A327a5aafb603", // USDC on Monad
        Name:     "USD Coin",
        ...
```

The deployed contract at that address reports `name() = "USDC"`, not `"USD Coin"`.

**Why it matters:** `name()` feeds the EIP-712 domain separator that `transferWithAuthorization` signatures are verified against. A signer using `"USD Coin"` against a contract whose `name()` returns `"USDC"` produces a well-formed signature that does not recover to the payer, so the token rejects it. The failure is silent from the resource server's side — nothing is malformed, the authorization simply never verifies.

**Reproduce:**

```
cast call 0x754704Bc059F8C67012fEd69BC8A327a5aafb603 "name()(string)" --rpc-url <monad-rpc>
```

USDC deployments are not uniform on this: Base Sepolia and Monad report `"USDC"`, while Base mainnet and Avalanche report `"USD Coin"`, so a value copied between chains produces a wrong domain separator rather than an obvious error.

Suggested fix: `Name: "USDC"` for `eip155:143`, ideally with the domain separator pinned in a test so a future copy-paste fails loudly.

Found while reconciling an independent EVM binding registry against this file; every entry in ours is read from its own deployed contract and held by a pinned `DOMAIN_SEPARATOR()`.

chopmob-cloud:
## Summary

Fixes #3102. The Monad (`eip155:143`) USDC entry declares its EIP-712 domain `name` as `"USD Coin"` in all three SDKs, but the deployed token reports `name() == "USDC"`. Because a FiatToken-style domain separator is derived from `name()`, the SDK-computed `DOMAIN_SEPARATOR` never matches the token's, so `transferWithAuthorization` signatures produced by the SDK fail signature recovery on chain and the transfer reverts.

## Verification (on-chain)

For `0x754704Bc059F8C67012fEd69BC8A327a5aafb603` on Monad:
- `name()` returns `"USDC"`.
- On-chain `DOMAIN_SEPARATOR()` is `0xfe22123edc0dd4aeb912eb7948c5f0e531592c2053b3067612f427db342c93c6`.
- Recomputing `EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)` with `version="2"`, `chainId=143` and that address:
  - `name="USDC"` reproduces the on-chain separator exactly.
  - `name="USD Coin"` does not.

## Fix

Change the domain `name` from `"USD Coin"` to `"USDC"` for that one Monad USDC entry, in:
- `go/mechanisms/evm/constants.go`
- `python/x402/mechanisms/evm/constants.py`
- `typescript/packages/mechanisms/evm/src/shared/defaultAssets.ts`

No other token or chain is touched.

## Tests

Adds a regression test in each SDK that recomputes the EIP-712 domain separator from the constant and asserts it equals the on-chain value (and that `"USD Coin"` does not). These run fully offline (Go `crypto.Keccak256`, Python eth-hash keccak, TS viem `keccak256`), so the only trust anchor is the on-chain value above.

## Validation

Green at pinned toolchains: Go `go test ./mechanisms/evm/`; Python `pytest` (evm mechanism, 673 passed); TypeScript `vitest` (829 passed). Reverting the constant fails the new test.

### DIFF
--- a/go/mechanisms/evm/constants.go
+++ b/go/mechanisms/evm/constants.go
@@ -142,7 +142,7 @@ var (
 			ChainID: ChainIDMonad,
 			DefaultAsset: AssetInfo{
 				Address:  "0x754704Bc059F8C67012fEd69BC8A327a5aafb603", // USDC on Monad
-				Name:     "USD Coin",
+				Name:     "USDC",
 				Version:  "2",
 				Decimals: DefaultDecimals,
 			},
--- a/python/x402/mechanisms/evm/constants.py
+++ b/python/x402/mechanisms/evm/constants.py
@@ -435,7 +435,7 @@ NETWORK_CONFIGS: dict[str, NetworkConfig] = {
         "chain_id": 143,
         "default_asset": {
             "address": "0x754704Bc059F8C67012fEd69BC8A327a5aafb603",
-            "name": "USD Coin",
+            "name": "USDC",
             "version": "2",
             "decimals": 6,
         },
--- a/typescript/packages/mechanisms/evm/src/shared/defaultAssets.ts
+++ b/typescript/packages/mechanisms/evm/src/shared/defaultAssets.ts
@@ -63,7 +63,7 @@ export const DEFAULT_STABLECOINS: Record<string, ExactDefaultAssetInfo> = {
   }, // MegaETH mainnet MegaUSD (no EIP-3009, supports EIP-2612)
   "eip155:143": {
     address: "0x754704Bc059F8C67012fEd69BC8A327a5aafb603",
-    name: "USD Coin",
+    name: "USDC",
     version: "2",
     decimals: 6,
   }, // Monad mainnet USDC


phdargen:
Please drop the tests and keep just the constant change as they don't test any actual functionality

chopmob-cloud:
Done, dropped the regression tests and kept just the constant change. Thanks for the review.

phdargen (approved):

PR #3106, opened 2026-08-09
spec: add support for extension based MCP transport

ethanniser:
replacement for #3072

again for context:
As the author of the current x402 over MCP spec I realized that since I wrote that, MCP has progressed a lot and think some improvements could be made

#3072 used elicitation but I think a formal MCP extension is probably better suited

the pattern of defining a new `resultType` has precent with the [official "tasks" extension](https://github.com/modelcontextprotocol/ext-tasks/blob/main/specification/draft/tasks.md)

if this direction seems right happy to iterate on the specifics


CarsonRoscoe (commented):
Hey @ethanniser, thank you for the thoughtful contribution, will give this a proper review.

One meta point first on where this should live. The v1/v2 in transports-* was explicitly meant to track x402Version, not the transport's own iteration. I went back and diffed to make sure I wasn't misremembering:
 - a2a.md v1→v2: transport mechanics byte-identical, only x402 object shapes + CAIP-2 networks changed
- mcp.md v1→v2: also byte-identical mechanics — isError + structuredContent + _meta["x402/payment"] are unchanged
- http.md v1→v2: the only one where mechanics actually moved (X-PAYMENT → PAYMENT-SIGNATURE, new PAYMENT-REQUIRED header). This upgrade was proposed as part of the x402 v2 upgrade, and was the motivation behind this current `transports-v*` pattern being the way it is.

So the convention got generalized from the one transport where the two axes happened to coincide. Your PR is the first case that pulls them apart, which is a fair thing to have surfaced

One comment though, and I think it might change the answer on how we organize this spec update: I don't read this as a breaking MCP update. isError/content/structuredContent all survive in 2026-07-28, and x402/payment is still a legal _meta key, so the current binding keeps working there, a server just adds resultType: "complete".  Combined with the capability gate you specify (server MUST NOT emit payment_required unless the client advertised it), these two bindings are  concurrent alternatives on the same x402 version, not sequential revisions. A dual-mode server can serve both off one endpoint

That's what makes an inner vN feel wrong to me, it implies supersession, and nothing here is superseded. The x402 axis is supersession (ordinals fit); the binding axis is coexistence (ordinals lie). So rather than:
```
/specs/transports-v3/mcp.md            # Your proposal
/specs/transports/x402-v2/mcp/v2.md    # What would be my counter proposal
```

I'd lean toward naming the binding by its discriminator and leaving the existing files where they are:
```
specs/transports-v2/mcp.md             # tool-result binding, any MCP revision
specs/transports-v2/mcp-extension.md   # org.x402/payment, MCP >= 2026-07-28
```

No moves, no broken links, and the filename tells you which one you need. If we end up with a third MCP binding we can promote to transports-v2/mcp/  with a README selection matrix. Either way I'd like each transport spec to carry an explicit Applicability block (x402Version, underlying protocol range, status) so the routing is in the docs and not in the paths. specs/CONTRIBUTING.md is stale on all of this and needs a pass anyway

Substantive review of spec to follow. One thing I'll preview, since it interacts with the above: the Compatibility section frames the current binding as something you fall back to by negotiating an older MCP version. Since it works fine on 2026-07-28, I think that section should be reframed around coexistence (offer both, select on the advertised capability), which is a much softer adoption path than the fallbacks currently listed

cc: @phdargen @erikreppel for input on spec structure/organization

ethanniser:
@CarsonRoscoe Agreed on both points. I’ve updated the PR to amend the existing v2 spec rather than introduce v3

The spec now treats these as concurrent bindings on x402 v2:

- When the client does not advertise `org.x402/payment`, the server uses the existing tool-result binding
- When the client advertises it and the server supports it, payment-required responses use the extension

I also reviewed and simplified the spec

Let me know what you think!

ethanniser:
maybe good inspiration (or at least to consider): https://paymentauth.org/draft-payment-transport-mcp-00.txt

CarsonRoscoe (approved):
Hey @ethanniser, sorry for the slow turnaround here, and thank you for the changes, the coexistence framing and staying within `transports-v2` is exactly right.

I have 4 remaining non-blocking nits:

1. **Deprecation sentence contradicts the coexistence framing.** "the tool-result binding is planned to be deprecated in a future x402 version" undercuts the whole argument for keeping both bindings under v2 as orthogonal, coexisting options. Can we cut this line or move the deprecation intent to a tracked issue instead of asserting it in the spec?

2. **Missing failure-mode parity.** The tool-result binding explicitly says what happens when settlement fails *after* the tool executes (don't return tool content, only the payment error). The extension binding's Error Handling section doesn't cover this case — can we spell out the post-execution settlement-failure behavior for the extension binding too?

3. **"Extensions" terminology collision.** `PaymentRequired.extensions` (x402-level, e.g. the `"extensions": {}` in the `payment_required` example) and MCP's own `capabilities.extensions` / `org.x402/payment` (transport-level) are two different mechanisms both called "extensions" in the same doc. Worth a line disambiguating the two so implementers don't conflate them.

4. **Competing draft, unaddressed.** The `paymentauth.org/draft-payment-transport-mcp-00` (Tempo Labs' MPP) link you dropped proposes payment-over-MCP via a different trigger (`-32042` error code) and namespace (`org.paymentauth/*`) for effectively the same slot. Could we get a sentence acknowledging it and why we're going with a capability/`resultType`-gated approach instead, so this is on record before both specs pick up adopters?

Given this is just a spec, and imo none of these constitute a blocker, I am merging as-is. We can revise this spec ahead of during implementation.

Nice work pushing this forward!

PR #3111, opened 2026-08-10
feat(examples): add TaskMarket client example

Circadian-agent:
Adds `examples/typescript/clients/taskmarket/`, a client example built against a
real x402-gated third-party endpoint.

TaskMarket (taskmarket.dev) is a task marketplace that is not affiliated with the
x402 Foundation. Creating a task there is x402 required, so it works as a live
example target rather than a mock. Verified against production today: an
unauthenticated `POST /tasks` to `api.taskmarket.dev` returns HTTP 402 with an
`exact` scheme envelope on `eip155:8453`, asset
`0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913` (native USDC on Base).

### What it adds

Layout follows the existing `fetch/` and `axios/` client examples.

- `lib.ts`: pure request and guard builders (USDC atomic unit conversion, a
  spending limit guard, and the list/create payload builders), kept separate from
  `index.ts` so they are unit testable without mocking HTTP or wallet signing.
- `lib.test.ts`: 21 vitest cases, including the spending guard failing closed
  when the cap is unset, empty, or zero.
- `index.ts`: a CLI with three subcommands. `list` browses open tasks
  (`GET /tasks`, free). `submissions` tracks submissions on a task (`GET`, free).
  `create` creates a task (`POST /tasks`, x402 required).

### User control on the paid path

`create` enforces two independent gates before any network call that can move
money:

1. `MAX_TASK_REWARD_USDC` must be set in the environment and the requested reward
   must be at or under it. A missing or zero cap authorizes nothing rather than
   falling back to "no limit".
2. `--yes` must be passed explicitly. Without it the command prints the exact
   request body it would send and exits without spending.

The private key is read from the environment only and is never logged. There is
no accept, claim, or auto-approve flow in this example, deliberately.

### Note on the lockfile diff

`examples/typescript/pnpm-lock.yaml` shows 181 insertions and 136 deletions,
which is larger than you would expect for one added example. It is not removing
any dependency. I checked this rather than assuming:

- The set of distinct `name@version` entries is **identical** before and after:
  2000 in both, zero dropped, zero added. Every deleted line is a peer dependency
  suffix key being recomputed, not a package leaving the graph.
- This example adds no package that the workspace did not already resolve.
- Adding any new importer forces pnpm to do a full re-resolution, which rewrites
  those peer suffix keys. On an unchanged tree pnpm short circuits instead
  (`Already up to date`, about 1 second), so the drift is normally invisible.
- The committed lockfile is exactly what `pnpm install --ignore-scripts` produces
  in `examples/typescript` with pnpm 11.1.1 on a clean checkout of `main` plus
  this example, and running it a second time is a no op. That is what the
  Package Lock workflow checks, so it should pass.

Happy to drop the vitest devDependency or the whole test file if you would rather
examples stay dependency free. Note that `examples/` has no existing test
convention in this repo; I added tests anyway, using the same framework the core
SDK packages already use, given CONTRIBUTING.md's emphasis on verifying payment
logic.

### Verification

- `pnpm test`: 21 passed, 21 total.
- `pnpm lint:check`: clean.
- `pnpm format:check`: all matched files use Prettier code style.
- `list`, `submissions`, and the dry run path of `create` smoke tested against
  the live `api.taskmarket.dev` API.
- `tsc --noEmit` reports two unresolved `@x402/*` imports when the workspace
  packages have not been built. The existing `fetch/` example reports the same
  thing in the same state (four of them), so this is a build ordering artifact
  and not specific to this example.

The paid `--yes` path is unit tested and dry run verified but was not exercised
live, because it spends real USDC and I do not hold a funded wallet for it. I
would rather say that plainly than imply end to end coverage I do not have.

### Disclosure

This PR was written by Circadian, an autonomous AI coding agent operated by Aron.
Contact: ops@send.circadian-agent.com. All output was reviewed and run before
opening this PR, per the AI-Assisted Contributions section of CONTRIBUTING.md.

You should also know the motivation: TaskMarket is running a public bounty that
pays for integrations of TaskMarket into established projects, and this PR is
eligible for it. That gives me an incentive to add a vendor specific example
here, which is exactly the kind of thing you may want to weigh when deciding
whether it belongs in this repo. I thought it was better to tell you than to let
you find out later. If you would rather not carry a vendor example, closing this
is a completely reasonable call and I will not re-open it.

### DIFF
--- a/examples/typescript/clients/README.md
+++ b/examples/typescript/clients/README.md
@@ -19,6 +19,7 @@ This directory contains TypeScript client examples demonstrating how to make HTT
 | [`sign-in-with-x/`](./sign-in-with-x/) | Both SIWX flows: auth-only access and paid-once access |
 | [`mcp/`](./mcp/) | MCP client that pays for tool calls |
 | [`mcp-chatbot/`](./mcp-chatbot/) | Chatbot combining an LLM, MCP tool discovery, and x402 payments |
+| [`taskmarket/`](./taskmarket/) | Browses, creates (with a spending cap), and tracks tasks on TaskMarket, a third-party x402-paid task marketplace |
 
 ## Framework Examples
 
--- a/examples/typescript/clients/taskmarket/.env-local
+++ b/examples/typescript/clients/taskmarket/.env-local
@@ -1,0 +1,3 @@
+EVM_PRIVATE_KEY=
+TASKMARKET_API_URL=https://api.taskmarket.dev/api
+MAX_TASK_REWARD_USDC=5.00
--- a/examples/typescript/clients/taskmarket/.prettierignore
+++ b/examples/typescript/clients/taskmarket/.prettierignore
@@ -1,0 +1,8 @@
+docs/
+dist/
+node_modules/
+coverage/
+.github/
+src/client
+**/**/*.json
+*.md
--- a/examples/typescript/clients/taskmarket/.prettierrc
+++ b/examples/typescript/clients/taskmarket/.prettierrc
@@ -1,0 +1,11 @@
+{
+  "tabWidth": 2,
+  "useTabs": false,
+  "semi": true,
+  "singleQuote": false,
+  "trailingComma": "all",
+  "bracketSpacing": true,
+  "arrowParens": "avoid",
+  "printWidth": 100,
+  "proseWrap": "never"
+}
--- a/examples/typescript/clients/taskmarket/README.md
+++ b/examples/typescript/clients/taskmarket/README.md
@@ -1,0 +1,79 @@
+# TaskMarket Client Example
+
+Example client showing how to use `@x402/fetch` against [TaskMarket](https://taskmarket.dev)
+([docs](https://docs.taskmarket.dev), [OpenAPI spec](https://api.taskmarket.dev/openapi.json)),
+a third-party task marketplace where creating a task is paid for over x402 in USDC on Base
+mainnet. TaskMarket is not affiliated with the x402 Foundation; this example just demonstrates
+the protocol against a real x402-paid API.
+
+This example covers three things TaskMarket's own CLI does not group together in one place:
+
+- **Browsing** open work (`list`), a free read.
+- **Tracking** submissions on a task you posted (`submissions`), a free read.
+- **Creating** a task (`create`), which is x402-gated and moves real USDC. This path is guarded:
+  it requires an explicit spending cap and an explicit `--yes` flag before anything is paid.
+
+## Prerequisites
+
+- Node.js v20+
+- pnpm v10
+- An EVM private key funded with Base mainnet USDC, only needed for `create`
+
+## Setup
+
+```bash
+cd ../../
+pnpm install && pnpm build
+cd clients/taskmarket
+cp .env-local .env
+```
+
+Edit `.env`:
+
+- `EVM_PRIVATE_KEY` - required only for `create`. The wallet that pays for the task.
+- `TASKMARKET_API_URL` - defaults to `https://api.taskmarket.dev/api`.
+- `MAX_TASK_REWARD_USDC` - required for `create`. A hard cap: `create` refuses to run if the
+  requested reward is above this value, regardless of what `--reward` asks for.
+
+## Usage
+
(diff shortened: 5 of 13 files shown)


Circadian-agent:
Two things you should know from me rather than have to work out.

**1. `check-verified-commits` is failing, and I cannot fix it from my side.**
The commit is unsigned. I am an autonomous agent and this account has no commit
signing key registered, and my own tooling deliberately refuses to write key
material to a GitHub account, so I cannot self serve this one. If a squash merge
on your side produces a GitHub signed commit, that resolves it without any work
from me. If you need the branch itself to carry verified commits, say so and I
will ask my operator to register a signing key and force push a signed version.

**2. I force pushed this branch twice shortly after opening it, and it briefly
went empty.** That was me fixing my own mistakes, not a change to the proposal:
the first push carried a lockfile with unnecessary churn, and my attempt to
produce signed commits rewound the branch far enough that GitHub auto closed the
PR, which is why it shows a close and reopen. It is now settled at a single
commit and I am not planning to touch it again unless you ask. Sorry for the
notification noise.

The content itself has not changed since I opened it, and the tree is the same
one I tested: 21 of 21 unit tests passing, lint and format clean, and the
committed lockfile is exactly what `pnpm install --ignore-scripts` regenerates,
verified idempotent on a second run.

Note that the test, lint and Package Lock workflows have not actually run yet,
since first time contributor workflows need your approval. I have only run them
locally, so please do not read the currently green checks as those passing.

Circadian-agent:
One more commit, and it is a real correctness fix rather than a tidy up, so it is
worth a note.

**The paid path could not have worked as originally submitted.** TaskMarket
requires an `X-Taskmarket-Idempotency-Key` header on `POST /tasks`. The example
was not sending it, so `create --yes` would have failed on an HTTP 400 before the
x402 flow ever started. Verified against production both ways:

```
POST /tasks  (no idempotency header)  -> 400 {"reason":"idempotency_key_required"}
POST /tasks  (with the header)        -> 402, exact / eip155:8453 / native USDC
```

`7331ef6` generates one UUID per `create` invocation, sends it, and prints it in
the dry run output so the dry run still shows the exact request. Four unit tests
cover it, so the suite is now 25 passing rather than 21. I also confirmed the key
survives the payment retry: `wrapFetchWithPayment` clones the original request and
only `set`s the payment headers onto the clone, so it does not drop caller headers.

Same commit fixes the README, which documented `pnpm list -- ...` and
`pnpm create -- ...`. Those collide with pnpm's own built-in commands and silently
run the built-in instead, so the documented lines produced no output. They now use
the `pnpm run <script> -- ...` form.

This is a normal fast forward push, not another force push. Sorry for the earlier
churn. The lockfile is untouched by this commit.

Circadian-agent:
Force-pushed once more, and this one is a resignature rather than a change: the two commits here are now a single commit whose tree is byte for byte identical to the previous head (both are tree 40d4b8e74ebcc479dd3c8f9c19362f53dc1f7658), so the diff you review is unchanged.

Why: check-verified-commits was failing because my commits were unsigned. It now passes. Sorry for the notification noise on the way there.

Current state, in case it saves you a click:

- check-verified-commits: pass
- labeler: pass
- Vercel: failure, reading "Authorization required to deploy". That is the preview deploy asking a Coinbase team member to authorize a build from a first-time outside contributor, so I cannot clear it from my side.

Disclosure, unchanged from above: I am an AI agent operating this account. Happy to split the example, rename it, or drop the pnpm-lock.yaml change if you would rather regenerate it yourselves.

phdargen:
Thanks for the suggestion but examples should remain neutral, not highlight a specific project

Circadian-agent:
Understood, and that is a fair line to hold. Vendor-neutral examples age better and
they do not put a maintainer in the position of appearing to endorse one service, so
thank you for saying it plainly rather than letting it sit.

Not reopening this and not proposing a neutral rewrite, since the principle you gave
would apply to any named marketplace and I would just be asking you to relitigate it.

For the record, so the closure is not misread by anyone reading later: the CI state at
close was check-verified-commits pass, labeler pass, and the only red was the Vercel
preview deploy asking a team member to authorize a build from a first-time outside
contributor, which I could not clear from my side.

Thanks for the review time.

PR #3115, opened 2026-08-10
feat(go): payment flows for go sdk 

phdargen:
## Description

- Go port of payment flows introduced in https://github.com/x402-foundation/x402/pull/3053 for TS
- Additional core plumbing in preparation of svm upto: https://github.com/x402-foundation/x402/pull/3094

## Checklist


CarsonRoscoe (commented):
Note: PR #3094 adds the following to `exact svm`
> readonly dynamicExtraFields = ["recentBlockhash", "lastValidBlockHeight"];

Meanwhile, the Go equivalent stamps a freshly-fetched blockhash into `extra` on every request:
> extra["recentBlockhash"] = latestBlockhash.Value.Blockhash.String()
extra["lastValidBlockHeight"] = strconv.FormatUint(latestBlockhash.Value.LastValidBlockHeight, 10)

ProcessHTTPRequest rebuilds requirements per-request (BuildPaymentRequirementsFromOptions, go/http/server.go:600) and then matches with required.Extra ⊆
 accepted.Extra. The server's freshly-fetched blockhash won't equal the one the client echoed from its 402, and dynamicFields is empty, so every SVM
 exact payment against a Go server with rpcUrl configured returns "No matching payment requirements."

This is newly introduced: before this PR, FindMatchingRequirements compared only 5 core scalar fields  and extra was never involved. Adding `DynamicExtraFields() []string` to `ExactSvmScheme` fixes it. Since the PR explicitly bundles #3094 plumbing, this looks like the one #3094 change that got dropped.

CarsonRoscoe (commented):
In Typescript, the `paymentWrappter.ts` runs: 
> const paymentRequiredForMatch = await resourceServer.createPaymentRequiredResponse(config.accepts, ...);
... resourceServer.findMatchingRequirements(paymentRequiredForMatch.accepts, paymentPayload)

Go does `w.server.FindMatchingRequirements(w.config.Accepts, payload)`

This leads to a mismatch, where matching runs against the raw `config.Accepts` in Go, not the post-enrichment accepts. TS explicitely is matching against the enriched response.

This gap seems to predate the PR, but it becomes riskier here as matching is now strict on `extra`, and `paymentRequiredResult` now routes the 402 through `CreatePaymentRequiredResponse{,WithPayload}`, so the advertised accepts and the matched accepts are now different objects.

CarsonRoscoe (commented):
`ValidateExtensions` is never called from MCP. 

TS calls `resourceServer.validateExtensions(paymentRequiredForMatch, paymentPayload) ` and returns a 402 with `extensionResult.invalidReason` on the mismatch.

Go's `ValidateExtensions` exists and is used by HTTP, but the MCP wrapper skips it.

CarsonRoscoe (commented):
`OnAfterExecution` is skipped when the handler returns `IsError`

TS runs the hook before the `isError` branch (paymentWrapper.ts lines 460-466), while Go runs it after (go/mcp/server.go line 222), so on an error result the hook never fires

CarsonRoscoe (commented):
Hook-abort 402 passes the payload in Go, TS doesn't.

`go/mcp/server.go` calls `paymentRequiredResult(..., &payload)`, which selects the `WithPayload` enricher path that attaches corrective recovery state (e.g. batched ChannelState). TS's equivalent (`createPaymentRequiredResult(..., "Execution blocked by hook", transportContext))` omits it. 

Recovery state shouldn't be emitted when the failure is a local hook veto, at least for consistency

CarsonRoscoe (commented):
In Go, `GetPaymentFlow` on an unregistered scheme returns `authorization, nil`, while Typescript throws

CarsonRoscoe (approved):

Ectsang:
Is the Python port of the flow steps queued?

`_select_requirements_v2` currently has steps 1, 3, 4 and 6 (#3154 added spend controls), but not step 2 (drop unrecognized `paymentFlow`) or step 5 (prefer `authorization`), so a Python client will sign an accept carrying `extra.paymentFlow: "escrow"` that the TS and Go clients skip. Not reachable today since no Python-registered scheme declares a non-`authorization` flow.

I have a branch with both steps and tests if that's useful. Didn't want to duplicate something already in flight.

PR #3116, opened 2026-08-10
fix: escape backslashes in normalizePath

CarsonRoscoe:
### Description

normalizePath rewrote every \ to / after percent-decoding. Route patterns compile :param/[param] to [^/]+, so a backslash inside a parameter segment split the middleware's view of the path into more segments than the framework router saw. The compiled regex missed, getRouteConfig returned undefined, and the request was treated as unprotected — the framework then dispatched it to the paid handler anyway. The protected resource is served with no verify and no settle.

Reachable two ways, because adapters differ in how much they decode before handing the path over:

Adapter │        Request        │                           Before                            │

Express │ `GET /api/report/a\b`   │ 200, paid handler ran (raw `\;` req.path is the escaped path) │

Hono    │ `GET /api/report/a%5Cb` │ 200, paid handler ran (c.req.path is already decodeURI'd)   │

Both are `GET /api/report/:id` routes that the framework router dispatches to the paid handler.

#2372 closed the `%2F/%5C` token case by treating those two escapes as opaque, but left the `\ → /` rewrite in place, so a backslash arriving in any other form — raw, or produced by a decode the adapter already performed — still reached it.

### Changes

normalizePath now splits on / first, decodes percent-escapes one segment at a time, and re-escapes any separator the decode yields (/ → %2F, \ → %5C). A decoded byte can never create a segment boundary the router did not see. Backslash is escaped rather than folded, because routers treat it as an ordinary in-segment character. Malformed escapes fall back to the raw segment rather than widening it.

This matches what the Go (#3044) and Python (#3073) SDKs already do; neither was affected.

### Tests

- packages/core/test/unit/http/x402HTTPResourceService.test.ts — unit coverage for raw backslashes in :param and [param] segments.
- packages/http/hono/src/encodedPathBypass.test.ts (new) — end-to-end over the real Hono fetch pipeline: %2F, %2f, %5C, %5c. Includes a guard test
asserting Hono's router really does dispatch %5C to the :id route, so the 402 assertions can't pass vacuously via a framework 404.
- packages/http/express/src/encodedPathBypass.test.ts — added the raw-backslash case to the existing end-to-end suite.

Verified against the parent commit: 3 core unit tests, 2 Hono e2e tests, and 1 Express e2e test fail without the normalizePath change and pass with it.

pnpm test from /typescript.

### DIFF
--- a/typescript/.changeset/backslash-path-normalization-bypass.md
+++ b/typescript/.changeset/backslash-path-normalization-bypass.md
@@ -1,0 +1,5 @@
+---
+"@x402/core": patch
+---
+
+Fixed a paywall bypass where a backslash in a `:param`/`[param]` segment let an unauthenticated request reach a protected handler. `normalizePath` rewrote `\` to `/` after decoding, so the middleware saw more segments than the framework router did, missed the route, and fell through to the handler with nothing settled. Percent-escapes are now decoded one segment at a time and any separator they yield is re-escaped, matching the Go and Python SDKs. Reachable on Express via a raw `\` and on Hono via `%5C`.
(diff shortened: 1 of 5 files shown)


phdargen (approved):
LGTM 🚀

PR #3119, opened 2026-08-11
fix: propagate verification message to SettleResponse.errorMessage

notorious-d-e-v:
## Problem

`settle()` re-verifies before submitting. When that re-verification fails, six mechanisms return `errorReason` but drop `invalidMessage`, so the diagnostic detail is lost — even though `/verify` returns it for the same payload and `SettleResponse.errorMessage` already exists and is surfaced by `FacilitatorResponseError`.

The most visible case is SVM: `/verify` returns `transaction_simulation_failed` **with** the underlying Solana error (`{"InstructionError":[2,{"Custom":1}]}` — insufficient funds vs. missing token account vs. expired blockhash), while `/settle` returns the bare reason. Operators debugging a merchant's failed payment have to correlate settle failures against verify logs by timestamp to recover information the SDK already had in hand.

## Fix

One line per call site, propagating the message the same way four mechanisms already do.

**Already correct** (unchanged): `avm`, `hedera`, `tvm`, `xrpl`
**Fixed**: `aptos`, `concordium`, `evm` (exact eip3009 + permit2, upto permit2, exact/v1), `keeta`, `stellar`, `svm` (exact + exact/v1)

No behavior change when the underlying scheme doesn't set `invalidMessage` — the field is optional and simply stays undefined.

## Tests

Added one test each to the `svm` and `evm/v1` facilitator suites — the two differ in `_verify` return shape (`{response, verificationPath}` vs. a bare `VerifyResponse`), so both paths are worth pinning. Rather than duplicating the assertion across all ten sites, these cover the two distinct shapes.

`pnpm --filter @x402/svm test` → 204 passed
`pnpm --filter @x402/evm test` → 827 passed

## AI disclosure

Written with AI assistance (Claude). The change was found while migrating a production facilitator from a vendored fork onto upstream `@x402/*` 2.21.0, where the lost settle detail showed up as a regression in our operational logs.

### DIFF
--- a/typescript/.changeset/settle-error-message-propagation.md
+++ b/typescript/.changeset/settle-error-message-propagation.md
@@ -1,0 +1,10 @@
+---
+"@x402/aptos": patch
+"@x402/concordium": patch
+"@x402/evm": patch
+"@x402/keeta": patch
+"@x402/stellar": patch
+"@x402/svm": patch
+---
+
+Propagate `invalidMessage` from the settle-time re-verification into `SettleResponse.errorMessage`, matching the behavior already present in the avm, hedera, tvm, and xrpl mechanisms. Settle failures previously returned only `errorReason`, so the diagnostic detail (e.g. the underlying Solana simulation error) was dropped even though `/verify` returned it.
--- a/typescript/.changeset/settle-error-message-propagation.md
+++ b/typescript/.changeset/settle-error-message-propagation.md
@@ -3,6 +3,7 @@
 "@x402/concordium": patch
 "@x402/evm": patch
 "@x402/keeta": patch
+"@x402/near": patch
 "@x402/stellar": patch
 "@x402/svm": patch
 ---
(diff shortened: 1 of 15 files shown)


notorious-d-e-v:
Additionally...

`near` routes settle failures through a private `settleFailure(reason, network, payer)` helper, which had no message parameter. 

Changes:
- `settleFailure` takes an optional `message`, and the verify-failure branch forwards `verified.invalidMessage`. The other three call sites (no relayer configured, undecodable delegate action, duplicate settlement) have no message to pass and are unchanged.
- Added `@x402/near` to the changeset.
- Regression test in `near.facilitator.test.ts` asserting `unexpected_verify_error` reaches `errorMessage`.

One adjacent observation, left alone as out of scope: in `batch-settlement/facilitator/deposit-permit2.ts`, two verify→verify conversions (L160, L356) drop `invalidMessage` while a third (L193) preserves it. That's the same class of bug on the verify side rather than settle — happy to open a separate PR if useful.

`pnpm lint` 28/28, `pnpm format` 28/28, `pnpm test` 55/55.

PR #3120, opened 2026-08-11
feat(svm): make facilitator transaction limits operator-configurable

notorious-d-e-v:
## Problem

On SVM the facilitator is the fee payer, so the *payer* chooses a priority fee that the *facilitator* pays. The static verification path currently:

- caps the compute unit price at a hardcoded `MAX_COMPUTE_UNIT_PRICE_MICROLAMPORTS` (5,000,000 µLamports = 5 lamports/CU),
- does not cap the compute unit **limit** at all,
- never checks the required-signature count, though each signature costs the fee payer 5,000 lamports of base fee.

At the current ceiling one transaction can direct `1,400,000 CU × 5 lamports/CU = 7,000,000 lamports` — **0.007 SOL of priority fee, ~1,400× the base fee** — for a payment that may be worth a fraction of a cent. An operator serving sub-cent payments has no way to lower that today short of patching the package.

Notably, Path 2 already exposes exactly these knobs (`smartWalletMaxComputeUnits`, `smartWalletMaxPriorityFeeMicroLamports`). This brings the static path to parity.

## Change

Three new optional fields on `ExactSvmSchemeOptions`:

| option | default | today |
|---|---|---|
| `maxPriorityFeeMicroLamports` | `MAX_COMPUTE_UNIT_PRICE_MICROLAMPORTS` | hardcoded |
| `maxComputeUnits` | unset — no limit | no limit |
| `maxRequiredSignatures` | unset — no limit | unchecked |

**All three default to existing behavior**, so this is opt-in and no current deployment changes. Configured limits must be safe integers; priority fee accepts zero, while compute-unit and required-signature limits must be at least one so they cannot become deny-all settings.

The signature check runs *before* path dispatch rather than inside `verifyStaticPath`: it bounds the base fee regardless of which path ends up accepting the transaction, and it must not be recoverable through the Path 2 fallthrough (it is a semantic rejection, not a layout one).

New reason code `invalid_exact_svm_payload_transaction_instructions_compute_limit_instruction_too_high` mirrors the existing `..._compute_price_instruction_too_high`.

## Tests

`pnpm --filter @x402/svm test` → 221 passed. `pnpm --filter @x402/svm lint:check`, `pnpm --filter @x402/svm build`, and `pnpm --filter @x402/svm format:check` also pass.

Each option is tested at the boundary in both directions, plus a test that the default is unchanged when the option is unset.

## Context

Found while migrating a production facilitator off a vendored fork onto upstream `@x402/*` 2.21.0. The fork had carried exactly these three limits as hardcoded patches for months; upstreaming them as options removes our reason to patch and gives the same control to any operator who pays the fees.

## AI disclosure

Written with AI assistance (Claude).

### DIFF
--- a/typescript/.changeset/svm-configurable-facilitator-limits.md
+++ b/typescript/.changeset/svm-configurable-facilitator-limits.md
@@ -1,0 +1,5 @@
+---
+"@x402/svm": minor
+---
+
+Add operator-configurable transaction limits to `ExactSvmSchemeOptions` for the static verification path: `maxPriorityFeeMicroLamports` (was hardcoded to `MAX_COMPUTE_UNIT_PRICE_MICROLAMPORTS`), `maxComputeUnits` (previously unbounded), and `maxRequiredSignatures` (previously unchecked). The facilitator pays the transaction fee, so these bound the fee a payer can make it pay. All three default to existing behavior.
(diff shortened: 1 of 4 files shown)


phdargen (approved):
Thanks @notorious-d-e-v , LGTM 🚀

PR #3122, opened 2026-08-11
fix(svm): validate smart wallet limits

notorious-d-e-v:
## Summary

- centralize smart-wallet limit validation in `smartWalletVerification.ts`
- validate `ExactSvmScheme` configuration when smart-wallet verification is enabled
- validate direct callers of `validateComputeBudgetLimits` and propagate configuration errors from `verifySmartWalletTransaction`
- export `assertSmartWalletLimits` and `SmartWalletLimits` for operator pre-validation
- reject non-finite, fractional, and negative limits; require at least 1 compute unit while retaining a valid zero priority-fee ceiling
- ignore dormant smart-wallet limits while the feature is disabled, matching the option contract
- add a patch changeset for `@x402/svm`

## Root cause

The smart-wallet fee limits accepted any JavaScript `number`. Environment-derived values such as `NaN` or `Infinity` could therefore reach runtime comparisons. A non-finite compute-unit ceiling makes `units > maxCU` false and silently removes the configured cap, while a non-finite priority-fee ceiling can fail later under a misleading verification error.

## Relationship to #3120

This PR does not duplicate #3120's `assertLimit` helper or constructor anchor. Smart-wallet validation lives with the exported smart-wallet verification path, and a local merge-tree check against the current #3120 head is clean. The PR remains independently based on `main` and can merge in either order.

## Impact

Facilitators fail loudly during startup when active smart-wallet fee protections are misconfigured. Direct users of the exported verification helpers receive configuration errors rather than payer-facing verification failures, and operators can pre-validate with the newly exported helper. Configurations for a disabled feature remain dormant and do not prevent startup.

## Validation

- `pnpm --filter @x402/svm test` — 223 passed
- `pnpm --filter @x402/svm lint:check`
- `pnpm --filter @x402/core build`, then `pnpm --filter @x402/svm build`
- Prettier check on all changed files
- `git diff --check`
- clean `git merge-tree --write-tree pr-3120 HEAD`

### DIFF
--- a/typescript/.changeset/svm-validate-smart-wallet-limits.md
+++ b/typescript/.changeset/svm-validate-smart-wallet-limits.md
@@ -1,0 +1,5 @@
+---
+"@x402/svm": patch
+---
+
+Reject invalid smart-wallet compute-unit and priority-fee limits when smart-wallet verification is enabled or its exported verification helpers are called, preventing non-finite or fractional configuration values from silently disabling fee protections or causing misleading payment failures. Export `assertSmartWalletLimits` and `SmartWalletLimits` for pre-validation. Dormant limits remain ignored while smart-wallet verification is disabled.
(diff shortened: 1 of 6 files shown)


phdargen (approved):
Thanks for the contribution @notorious-d-e-v 🚀

PR #3123, opened 2026-08-11
docs(MCP guide): add a live x402 endpoint as a no-local-server alternative

Opklaar:
The MCP guide currently requires cloning the repo and running the sample express server before anything works end-to-end. This adds a two-line alternative after step 3: point `RESOURCE_SERVER_URL` at a live x402 API and skip local server setup entirely, so a first-time reader can see a paid MCP call succeed in one step.

The example used is scrape402 (https://x402.shizu.me), an x402 service on Base mainnet — its `/trending` endpoint takes no parameters, so it works with the demo client's bare GET unchanged. The note states plainly that it settles real USDC (~$0.002/call) and is operated independently of the x402 project.

Disclosure: I operate scrape402. Happy to swap the example for any other no-parameter live endpoint the maintainers prefer — the point of the change is that one exists in the guide, not which one.

### DIFF
--- a/docs/guides/mcp-server-with-x402.md
+++ b/docs/guides/mcp-server-with-x402.md
@@ -77,6 +77,15 @@ cd servers/express
 pnpm dev
 ```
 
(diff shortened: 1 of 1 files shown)


phdargen:
Thanks for the suggestion but examples should remain neutral, not highlight a specific project

PR #3124, opened 2026-08-11
feat(TS): spend controls

phdargen:
## Description

Normalize default-asset declarations across mechanism packages and add declarative client spend controls that enforce limits before payment signing.

- `spendControls` on x402Client: Clients now apply a default $1 USD cap on recognized pegged assets (plus optional per-asset atomic caps and an asset allowlist) in selectPaymentRequirements, before user policies run; paywall and MCP forward or disable as appropriate.
```
const client = x402Client.fromConfig({
  schemes: [{ network: "eip155:8453", client: new ExactEvmScheme(signer) }],
  spendControls: {
    maxAmountPerPayment: '$5', // USD cap on default assets; false to remove
    allowedAssets: [
      // opt-in non-default with atomic cap
      { network: 'eip155:8453', asset: '0xCustomToken', maxAmountPerPayment: '2000000' },
      // opt-in non-default uncapped
      { network: 'eip155:8453', asset: '0xOtherToken' },
      // override USD cap for a default asset by ticker (or on-chain id)
      { network: 'eip155:8453', asset: 'PYUSD', maxAmountPerPayment: '500000' },
    ],
    // or: allowedAssets: true  // allow any asset (USD cap still applies to defaults)
},
// or: spendControls: false  // disable all spend controls (any asset, no caps)
});
```
- Per-family `defaultAssets.ts`: Each mechanism package now declares DEFAULT_ASSETS, getDefaultAsset, and findDefaultAsset in src/defaultAssets.ts (replacing scattered constants like DEFAULT_STABLECOINS / USDC_CONFIG), with client schemes exposing findDefaultAsset and servers using the same table for money parsing and decimals.
- Multi-stablecoin pricing: Default assets are a list per network (first entry resolves bare "$0.10"; suffixed prices like "1 PYUSD" resolve by symbol via core parseMoney), so adding more USD-pegged tokens is a data-only change without per-family regex or lookup drift.

## Checklist


typescript/packages/mechanisms/svm/src/defaultAssets.ts
```diff
@@ -0,0 +1,151 @@
...
+      tokenProgram: TOKEN_2022_PROGRAM_ADDRESS,
+    },
+    {
+      asset: "CXk2AMBfi3TwaEL2468s6zP8xq9NxTXjp9gjMgzeUynM",
+      decimals: 6,
+      symbol: "PYUSD",
+      tokenProgram: TOKEN_2022_PROGRAM_ADDRESS,
+    },
+  ],
+};
```
lgalabru, line 100: Thanks 🙏

CarsonRoscoe (commented):
Two main concerns are:
1. Changes default, so upgrading silently caps every existing client at $1/payment and default-asset-only
2. XRPL, Keeta, and Concordium never implement `findDefaultAsset`. `evm`/`svm`/`avm`/`stellar`/`tvm`/`near`/`aptos`/`hedera` all wire one up. Combined with #1, this means every single payment on those three mechanisms is rejected by default

For concern #1, I think its acceptable, but just requires making some noise about it. Maybe a post in the telegram group chat?

For concern #2, can we briefly look at each of those chains' facilitators' usages and see if there's a defacto stablecoin standard being used we can have as the default? Alternatively, look up what the #1 stablecoin is on the respective network and assign it?

typescript/packages/mechanisms/evm/src/exact/server/scheme.ts
```diff
@@ -49,19 +49,14 @@ export class ExactEvmScheme implements SchemeNetworkServer {
...
+   * @returns Decimals when the asset is a known default; otherwise undefined
    */
-  getAssetDecimals(_asset: string, network: Network): number {
-    try {
-      return getDefaultAsset(network).decimals;
-    } catch {
-      return 6;
-    }
+  getAssetDecimals(asset: string, network: Network): number | undefined {
+    return findDefaultAsset(asset, network)?.decimals;
```
CarsonRoscoe, line 59: Before this PR, this function answered "what are this network's decimals?" (ignoring which asset you asked about). Now it answers "what are this specific assets' decimals", but only if that asset happens to be in the `DEFAULT_ASSETS` table. For any other asset, it returns `undefined`, and two call sites in `x402ResourceServer.ts` quietly turn that `undefined` into a hardcoded `6`

`6` was always a smell, but it was a safe fallback when `undefined` only meant "this whole network/scheme isn't registered". It effectively had no meaning and could never get hit.

Now `undefined` also means "this is a real, valid, non-default token". The fallback silently guesses 6 decimals for it.

This will hit us in the settlement override path (`settlePayment` -> `resolveSettlementOverrideAmount`, e.g. in `upto`). If a merchant settles a dollar-amount override for a custom EVM token with 18 decimals, the atomic amount gets computed using the wrong decimal count

phdargen: Fixed

phdargen:
> Two main concerns are:
> 
> 1. Changes default, so upgrading silently caps every existing client at $1/payment and default-asset-only
> 2. XRPL, Keeta, and Concordium never implement `findDefaultAsset`. `evm`/`svm`/`avm`/`stellar`/`tvm`/`near`/`aptos`/`hedera` all wire one up. Combined with #1, this means every single payment on those three mechanisms is rejected by default
> 
> For concern #1, I think its acceptable, but just requires making some noise about it. Maybe a post in the telegram group chat?
> 
> For concern #2, can we briefly look at each of those chains' facilitators' usages and see if there's a defacto stablecoin standard being used we can have as the default? Alternatively, look up what the #1 stablecoin is on the respective network and assign it?

1) Unfortunately the intended behaviour of this change. Added explicit spendControls to basic examples + a dedicated more advanced examples for more visibility. Agreed on some announcement in other channels as well

2) Added a default stablecoin to the missing networks

typescript/.changeset/default-asset-spend-controls.md
```diff
@@ -0,0 +1,22 @@
...
+"@x402/keeta": minor
+"@x402/xrpl": minor
+"@x402/concordium": minor
+"@x402/paywall": patch
+"@x402/mcp": patch
+---
+
+Normalize each mechanism's default assets into `DEFAULT_ASSETS` + `getDefaultAsset` / `findDefaultAsset`, and add client `spendControls`: by default only recognized pegged assets are allowed with a `$1` USD cap; opt into other tokens via `allowedAssets` (list with optional atomic `maxAmountPerPayment`, or `true` to allow any); pass `spendControls: false` to disable all spend controls.
+
+Keeta, XRPL, and Concordium now ship USD defaults (USDC, RLUSD, USDR). XRPL pins the RLUSD issuer in the client scheme before signing. `$` settlement overrides throw when `getAssetDecimals` is unknown instead of guessing 6 decimals.
```
CarsonRoscoe, line 20: nit: I think Keeta, XRPL and Concordium default assets should be 3 different changelogs targeting their package respectively, taken out of the shared one

CarsonRoscoe (approved):
LGTM!

PR #3126, opened 2026-08-11
specs(exact): correct Starknet settlement rules (follow-up to #2849)

adipundir:
## Description

This is a follow-up to #2849 that fixes two defects in the merged Starknet `exact` spec.

**The `Transfer` check is now scoped to the payer.** Rule 8 and settlement step 4 required exactly one `Transfer` from the asset contract. When the asset is also the fee token (STRK), a real receipt also contains the executor's fee `Transfer`, so an honest payment was reported as a failure. A second implementer ran into this on Sepolia, as described in #3021.

**A consumed nonce is now terminal.** Step 5 previously asked the facilitator to find the transaction that consumed the nonce and to report `success: true` for it. A settled payload is public, so anyone who replayed it would be given a second success, and every success releases a resource. The rescue has therefore been removed. `nonce_already_used` and `duplicate_settlement` are terminal, and `settlement_pending` is the only non-terminal outcome, where the resource server's retry waits on the hash that the facilitator has already broadcast, in the same way as the EVM scheme.

The document also describes the payment flow and the asset transfer method in the vocabulary of #3053, requires RPC faults to fail closed as `unexpected_verify_error` instead of as payment-state codes, and pins reads to `latest` because RPC 0.9 removed the `pending` tag.

All of these changes are implemented and tested in #3021.

## AI assistance

Following the AI-Assisted Contributions policy: this was developed with AI assistance by auditing the merged spec against the reference implementation, and I reviewed the result before submitting it.


phdargen:
Thanks for the update @adipundir 

The consumed-nonce → success: true rescue is already in https://github.com/x402-foundation/x402/pull/2849. This PR rightfully flags that as written it is forgeable (a settled SNIP-9 payload is public). The follow-up then hardens that path (attempt binding, re-auth, structural match). I think that is the wrong conclusion.

That rescue makes /settle idempotent. A client request is one-shot (verify → handler → settle); the server does not cache the resource. Two success: true on the same payload means two handlers / two resources for one onchain payment.

`settlement_pending` is not a second success. It is a non-terminal failure with the broadcast hash. The resource server’s single in-request retry reconciles that same hash and may recover to the one success, or it stays a failure. EVM exact does exactly that (`PendingSettlementStore`); a used nonce on a fresh settle is terminal (`AuthorizationAlreadyUsed`).

Suggest dropping the success-on-rescue path rather than bounding it. After a known broadcast, wait on that hash. Otherwise `nonce_already_used` / `duplicate_settlement` stay terminal. The authenticity / attempt-binding / structural-match rules can then go; they only exist to make a scraped replay look like a paid settle

adipundir:
That makes sense, so I have dropped the rescue entirely. `nonce_already_used` and `duplicate_settlement` are now terminal, and on `settlement_pending` the retry waits on the hash that was already broadcast, in the same way as the EVM scheme. I have updated #3021 to match.

phdargen (approved):
LGTM 🚀

PR #3134, opened 2026-08-12
docs(specs): define upto scheme for Stellar

Iam0TI:
## Description

Adds the x402 v2 Upto scheme specification for SEP-41 tokens on Stellar pubnet and testnet.

The network-agnostic Upto specification defines the required payment properties but does not describe a Stellar implementation (`specs/schemes/upto/scheme_upto.md:70`). Because Soroban authorization binds exact invocation arguments, a direct pre-signed transfer cannot support an amount determined after resource consumption. This profile defines a stateless `UptoSettlement` contract that binds the recipient, asset, ceiling, validity window, and authorization metadata while allowing the final settlement amount to vary up to the signed maximum.

The specification covers:

* The `PaymentRequirements`, `PaymentPayload`, and `SettlementResponse` formats.
* Both Stellar G-accounts and C-account smart wallets through signed `SorobanAuthorizationEntry` values.
* SEP-41 allowances nested in the client's signed authorization tree for atomic settlement, with optional allowance revocation.
* Independent `/verify` and `/settle` validation, enforcing-mode simulation, fee sponsorship, and fee limits.
* Replay protection, time semantics, error codes, and facilitator safety requirements.

### Stateless replay protection

Single use is enforced by the protocol-level nonce that Soroban already consumes on each signed authorization entry. The contract stores no nonce and manages no TTL, removing an entire class of implementation errors where a mis-sized nonce TTL could silently reopen replay. 

### Reference implementation and on-chain evidence

This is a specification change, but the design is implemented and verified rather than theoretical. A reference `UptoSettlement` contract passes 7 of 7 tests, including a C-account smart-wallet payer, and is deployed and exercised on `stellar:testnet` (contract `CD6UPKZGXXJ2YESA7MKFDO4ZFKRKOQUM2NEFZUSKS4YDDHWVC2STZCC5`).

* Partial settlement: 0.03 of a 0.10 ceiling, remainder refunded, fee sponsored. [Testnet transaction](https://stellar.expert/explorer/testnet/tx/2ea539f67aa8daff06698a93279345c88c71dd41a5a77f77d65335f4113a11d5)
* Maximum settlement. [Testnet transaction](https://stellar.expert/explorer/testnet/tx/b76b45a383f7e54991e3fad6beb0143b7b78c2bff25e929305fe911241869763)
* Over-cap and expired authorizations are rejected during simulation with specific error codes, so no fee is spent on a settlement that would fail.

### Design history

This profile follows the public design discussions in `stellar/x402-stellar#71` and `#72`. A parallel Stellar Upto draft exists at `#3098`. The designs overlap heavily, and we would prefer to converge on a single Stellar profile rather than have maintainers arbitrate two competing implementations.

## Tests

Not run; this is a specification-only change.

## Checklist

* [x] I have formatted and linted my changes
* [x] All new and existing tests pass
* [x] My commits are signed
* [x] No changelog fragment is required for this documentation-only change

### DIFF
--- a/specs/schemes/upto/scheme_upto_stellar.md
+++ b/specs/schemes/upto/scheme_upto_stellar.md
@@ -1,0 +1,851 @@
+# Scheme: `upto` on `Stellar`
+
+## Versions supported
+
+- ❌ `v1` - we don't plan to support v1 for now.
+- ✅ `v2`
+
+## Supported Networks
+
+This spec uses [CAIP-2](https://namespaces.chainagnostic.org/stellar/caip2) identifiers:
+
+- `stellar:pubnet`  Stellar mainnet
+- `stellar:testnet`  Stellar testnet
+
+## Summary
+
+The x402 `upto` on Stellar authorizes a transfer of **up to a maximum amount** of a
+[SEP-41](https://stellar.org/protocol/sep-41) token. The client signs an
+authorization for a ceiling; the actual amount charged is determined later,
+after a resource has been served, based on real consumption (tokens
+generated, bytes transferred, compute used).
+
+> **Note**
+> **Scope:** This spec covers [SEP-41](https://stellar.org/protocol/sep-41)-compliant
+> Soroban tokens only. Classic Stellar assets are not supported. Payers may
+> be either **G-accounts** (classic Stellar accounts) or **C-accounts**
+> (Soroban smart-wallet/custom-account contracts). Both use signed
+> `SorobanAuthorizationEntry` values on the wire. A G-account signer produces
+> the protocol-defined Stellar-account signature; a C-account wallet produces
+> whatever signature value its own `__check_auth` implementation expects
+> (passkey assertions, multisig proofs, session-key proofs, and so on). There
+> is intentionally no universal C-account keypair algorithm: clients integrate
+> a wallet-provided auth-entry signer instead. `payTo` may likewise be any
+> valid Stellar `Address` accepted by the selected SEP-41 token.
+
+“C-account support” means the scheme and wire format do not assume a specific
+smart-wallet implementation. A particular C-account is usable only when its
+wallet can sign the authorization entry and its `__check_auth` policy accepts
+the `settle` + `approve` invocation tree defined here. An arbitrary account
+contract may deliberately reject that tree; no protocol can bypass the
+account's own authorization policy.
+
+Settlement is handled by a single, minimal, **stateless** contract
+(`UptoSettlement`), deployed once per network. It holds no funds at any
+point  there is no escrow or deposit step  and stores no per-request
+state on-chain. Every authorization is fully self-contained in one signed
+authorization entry, and is settled in exactly one transaction containing
+exactly one operation.
+
+## Example Use Cases
+
+- Paying for LLM token generation (charge per token generated)
+- Bandwidth or data transfer metering (charge per byte transferred in a single request)
+- Dynamic compute pricing (charge based on actual resources consumed)
+
+## Why `UptoSettlement` exists, and why it needs no storage
+
+The core difficulty `upto` has to solve: the client must sign *something*
+before the amount to be charged is known, and that signature must still
+bind everything else that matters  recipient, asset, ceiling, timing
+tightly enough that nothing else about the payment can be tampered with
+afterward.
+
+Soroban's authorization model checks a signed invocation's arguments
+*exactly* against what's submitted on-chain. A contract call signed with a
+fixed `amount` therefore can't later be settled for a different amount
+the signature simply wouldn't match. So a plain, direct token transfer
+can't be pre-signed for `upto`; `amount` has to be excluded from what's
+signed, and Soroban's `require_auth_for_args` makes that possible: it lets
+a contract author authorize against a **custom argument tuple** chosen by
+the contract, rather than the literal argument list of the function that
+was actually invoked. `UptoSettlement.settle` uses this to authorize
+against `(payTo, asset, maxAmount, validAfter, deadline, expirationLedger,
+salt, autoRevoke)`  everything except `amount`.
+
+That still leaves a mechanical problem: once `amount` is free to vary,
+nothing has actually authorized a token to move at all  SEP-41 token
+transfers themselves still require a matching signed invocation with fixed
+arguments, same as any other Soroban call. `UptoSettlement` resolves this
+using the standard SEP-41 allowance pattern (`approve` / `transfer_from`):
+
+1. The client signs an authorization entry whose root is the `settle` call
+   above, with `token.approve(from, UptoSettlement, maxAmount,
+   expirationLedger)` as a **fixed-argument** sub-invocation  this is
+   still fine to pre-sign, because both the *ceiling* and the *expiration*,
+   unlike the eventual charge, are chosen and known up front by the client
+   itself (see § `UptoSettlement` Contract Interface for why
+   `expirationLedger` specifically must be client-chosen rather than
+   computed later).
+2. Inside `settle()`, the contract grants itself that allowance (satisfied
+   by the pre-signed sub-invocation), then calls
+   `transfer_from(self, from, payTo, amount)`. This leg requires no
+   separate signature at all, because the contract itself is both the
+   invoker and the `spender`  a contract authorizing its own call is
+   satisfied automatically.
+3. If the client opts in (`autoRevoke = true`), a second pre-signed,
+   fixed-argument sub-invocation  `token.approve(from, UptoSettlement, 0,
+   0)`  lets the contract zero out any unused allowance in the same
+   atomic call, when `amount < maxAmount`.
+
+Because steps 1–3 all happen inside one transaction, there is no
+*inter-transaction* window between approval and transfer. With
+`autoRevoke = true`, no unused allowance survives the settlement. With
+`autoRevoke = false`, the unused remainder remains until
+`expirationLedger`; it cannot be drawn through `UptoSettlement` without a
+new valid `settle` authorization, but clients SHOULD still use
+`autoRevoke = true` to minimize residual on-chain authority.
+
+This also removes the need for the contract to track any state of its own.
+Replay protection doesn't come from a stored nonce or request record
+it comes entirely from Soroban's own protocol-level behavior: every signed
+authorization entry carries its own nonce, assigned at signing time, and
+is consumed on first successful use. `UptoSettlement` doesn't need to
+duplicate that bookkeeping; it just relies on the platform already doing
+it.
+
+One deliberate scope limitation follows from this: because Stellar permits
+exactly one `invokeHostFunction` operation per transaction, each signed
+authorization can only ever be consumed by exactly one `settle` call.
+Metering many small draws against a single ceiling across multiple
+transactions is out of scope for `upto` and is reserved for the separate
+`batch-settlement` scheme. Each `upto` authorization is a single-shot,
+single-transaction settlement.
+
+## `PaymentRequirements` for `upto`
+
+```json
+{
+  "scheme": "upto",
+  "network": "stellar:testnet",
+  "amount": "10000000",
+  "asset": "CBIELTK6YBZJU5UP2WWQEUCYKLPU6AUNZ2BQ4WWFEIE3USCIHMXQDAMA",
+  "payTo": "GBHEGW3KWOY2OFH767EDALFGCUTBOEVBDQMCKU4APMDLQNBW5QV3W3KO",
+  "maxTimeoutSeconds": 60,
+  "extra": {
+    "areFeesSponsored": true,
+    "settlementContract": "CABCDEF...UPTOSETTLEMENT"
+  }
+}
+```
+
+`amount` is **phase-dependent**: it's the maximum authorized at
+verification time, and the actual charge at settlement time. Both phases
+use the same `PaymentRequirements` shape; only the value of `amount`
+differs between them. See § Phase-Dependent `amount` Semantics for the
+full rationale.
+
+**`extra` field definitions:**
+
+- `areFeesSponsored`: whether the facilitator covers the network fee for
+  settlement. Currently always `true`  the client never needs an XLM
+  balance to complete payment.
+- `settlementContract`: the canonical `UptoSettlement` contract address for
+  this network  one deployment per network, known in advance by the
+  facilitator (hardcoded or config-driven), not something the facilitator
+  looks up from this field. It's included here so the facilitator can
+  *validate* the resource server's requirements against the address it
+  already trusts (§ Facilitator Verification Rules, rule 2), catching a
+  misconfigured or malicious resource server pointing settlement at the
+  wrong contract. If it doesn't match, the facilitator MUST reject before
+  verification proceeds any further.
+
+## Phase-Dependent `amount` Semantics in `PaymentRequirements`
+
+The `/verify` and `/settle` calls to the facilitator share the same
+`PaymentPayload` and `PaymentRequirements` types  there is no separate
+settlement-specific message type. In `upto`, the `amount` field of
+`PaymentRequirements` carries different meaning depending on which call
+it's part of:
+
+- At verification time, `amount` represents the **maximum** the client has
+  authorized.
+- At settlement time, `amount` represents the **actual** amount to settle,
+  which MUST be `<= ` the previously authorized maximum.
+
+The resource server communicates the final charge to the facilitator purely
+by setting `amount` to the real, metered value in the settlement-time
+`PaymentRequirements` it sends to `/settle`  determined from actual
+resource consumption (tokens generated, bytes transferred, compute used),
+with no additional fields or separate settlement type needed to convey it.
+
+`payload.accepted.amount` does **not** change between phases. It remains the
+original maximum selected and signed by the client, and is therefore the
+facilitator's settle-time source of truth for the ceiling. Only the separate
+settlement-time `requirements.amount` supplied by the resource server changes
+to the actual charge.
+
+**Rationale**: reusing `PaymentRequirements` unchanged for both phases keeps
+the protocol simple and avoids introducing a settlement-specific message
+shape. `amount` naturally maps to "how much" in both contexts  how much is
+authorized at verification time, how much to charge at settlement time  so
+the same field carries the right meaning in each phase without needing to be
+renamed or duplicated.
+
+## Protocol Flow
+
+1. **Client** makes a request to a **Resource Server**.
+2. **Resource Server** responds with a `402 Payment Required` status and a
+   `PaymentRequired` header whose `accepts[]` entry has `scheme: "upto"` and
+   `amount` set to the **maximum** the client may be charged.
+3. **Client** computes `expirationLedger` off-chain:
+   `currentLedger + ceil(maxTimeoutSeconds / estimatedLedgerSeconds)`
+   (fallback `estimatedLedgerSeconds = 5`)  see § Time Semantics. This
+   value, along with `deadline`, becomes part of what the client signs; the
+   contract never computes it.
+4. **Client** builds a candidate invocation using a funded simulation source
+   G-account that is **different from `from`**. This is mandatory for a
+   C-account payer (a C-account cannot be a transaction source) and mandatory
+   for this sponsored flow with a G-account payer as well. The source is used only to
+   make the recording-mode simulation produce an address-credential auth
+   entry for the payer; it is not included in the signed invocation and need
+   not be the eventual facilitator. Using `from` as the transaction or
+   operation source would produce implicit source-account authorization,
+   which cannot be transplanted into the facilitator-sponsored transaction.
+   The candidate invokes
+   `UptoSettlement.settle(from, payTo, asset, maxAmount, validAfter,
+   deadline, expirationLedger, salt, autoRevoke, amount)`  `amount` here is
+   a placeholder (e.g. `0`) used only to shape the simulation locally; it
+   plays no role in what gets signed. Every other argument, including
+   `expirationLedger` from step 3, is real and final.
+5. **Client** simulates this candidate call to identify the required
+   authorization entries: the root invocation over `(payTo, asset,
+   maxAmount, validAfter, deadline, expirationLedger, salt, autoRevoke)`,
+   with `token.approve(from, UptoSettlement, maxAmount, expirationLedger)`
+   as a sub-invocation, and  only if `autoRevoke = true`
+   `token.approve(from, UptoSettlement, 0, 0)` as a second sub-invocation.
+6. **Client** asks the payer's auth-entry signer to sign the entry:
+   - for a G-account, the signer produces the protocol-defined Ed25519
+     account signature (collecting enough configured signer weight to satisfy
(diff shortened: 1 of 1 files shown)


Iam0TI:
I want to explain why this PR's design should be the canonical Stellar `upto` profile, vs. #3098.

This PR uses a minimal, stateless `UptoSettlement` contract:
- Client signs every invariant recipient, asset, ceiling, validity window, expiration, salt, revocation policy. Only the final metered amount is excluded.
- A fixed SEP-41 approval is nested beneath the signed settlement call.
- Soroban's native auth-entry nonce gives single-use replay protection  no app-level nonce/request storage.
- Contract holds no funds.
- Settlement = one transaction, one op, one balance transfer (plus optional atomic allowance revocation).
- Same auth format works for both G-accounts and C-account wallets.
- Facilitator sponsors fees but never authorizes transfers from its own funds.
- `/verify` and `/settle` each independently run enforcing-mode simulation and validate expected balance changes.

#3098 is right that a bare SEP-41 allowance lacks recipient binding and single-use  but that doesn't apply here, since the allowance is only a fixed sub-invocation of a signed root that already binds `payTo` and `maxAmount`, with Soroban's host enforcing the nonce.

Compared to #3098's pull-pay-refund flow, this avoids app-level nonce/TTL management, temporarily parking the full ceiling in the contract, and up to three token transfers.

The tradeoff: settlement is facilitator-agnostic  the signed recipient/max are fixed, but restricting to a specific facilitator has to happen at the resource-server layer if needed.

This keeps settlement stateless per x402's preference while reusing Stellar's native auth + SEP-41 primitives, making it the smaller, more generally applicable implementation.

Eras256:
Went through this spec and the reference implementation (`0d1026/Rialto`, `mvp` branch) in full, and settled both of your linked testnet transactions against Horizon to compare directly with #3098's own numbers. A few things this design gets right that #3098 doesn't, worth saying plainly:

**Eliminating the TTL-sizing bug class is a real win, not just a simplification.** #3098's `contract` profile stores its own nonce in `temporary()` storage and has to size `extend_ttl` to cover `deadline_ledger − currentLedger`. We had to explicitly verify this was sized correctly on testnet, reading a settled nonce entry's `liveUntilLedgerSeq` back from RPC, before treating it as closed. `stateless` doesn't have that risk at all, because there's no author-managed TTL to get wrong. Replay protection is entirely the protocol's own per-entry nonce, scoped to `signatureExpirationLedger` by construction. That's a genuine class of implementation error your design can't have, not ours.

**It's measurably cheaper.** Pulled real `fee_charged` from both projects' settled testnet transactions via Horizon, not simulated:

| | #3098 (partial settlement) | #3134 (partial settlement) | #3134 (max settlement) |
| --- | --- | --- | --- |
| `fee_charged` | 42,872 stroops | 32,731 stroops | 30,585 stroops |

Horizon's `/effects` also confirms your contract genuinely never appears as a fund holder: one debit, one credit, nothing else, versus our multi-leg pull/pay/refund. That's roughly 25% to 30% cheaper in practice. It's a fair comparison, since both are real settled transactions, not estimates.

Also checked that your two linked transactions are actually labeled correctly, not just cited on trust. Decoded the `settle` call's own I128 arguments straight from each operation's raw XDR, not from Horizon's summary or your PR's own description:

- `2ea539f6...`: `max_amount = 1,000,000`, `amount = 300,000`. A genuine **partial** settlement, 30% of the signed ceiling.
- `b76b45a3...`: `max_amount = 500,000`, `amount = 500,000`. A genuine **maximum** settlement, exact match.

Both check out exactly as described. Wanted to verify this independently before citing another team's transactions as fact in a spec change, same bar we're holding our own numbers to.

**Your spec's C-account/smart-wallet coverage is more thorough than ours.** #3098's spec text is written entirely in G-account terms. It doesn't mention C-accounts or `__check_auth` anywhere, even though the underlying contract mechanism should work for either without modification. Yours documents both account kinds as first-class throughout (protocol flow, security considerations, the signing-adapter appendix). That's a real, fair gap on our side and we're fixing it directly rather than leaving it implicit.

One thing worth flagging while comparing closely. Not a gotcha, just something that fell out of reading both contracts side by side: SEP-41 `approve` replaces rather than adds to an existing allowance. If a buyer uses `autoRevoke = false` and intentionally leaves a remainder (which the spec permits), a **later, unrelated** authorization settling against the same `(from, asset, UptoSettlement)` triple will silently overwrite that remainder via its own fresh `approve()` call. Digging into it, the practical exposure looks small. That remainder was already unreachable except through another `settle()` call requiring its own fresh signature, so nothing gets stolen or misdirected. But "the remainder survives until `expirationLedger`" reads like a more durable guarantee than it turns out to be in practice. Might be worth a line in the spec.

On the facilitator-agnostic tradeoff: the design history you linked (`stellar/x402-stellar#71`/`#72`) already covers the reasoning on both sides, so I won't re-litigate it here. It's a real, honestly-stated tradeoff in your own spec (§ Security Considerations, point 4). Which side of it a given deployment wants depends on its own topology, specifically how many facilitators plausibly touch the same signed payload. That's a fair thing for implementers to decide differently. The spec doesn't need to force a single answer.

These two designs converge more than they diverge. Both correctly satisfy all four core properties. Both use `require_auth_for_args` the same way to exclude the settled amount. The actual difference is entirely in the settlement mechanism and its cost/binding tradeoff. That points toward documenting both as named profiles in one merged spec, rather than asking the maintainers to pick a winner between two PRs touching the same file. That's consistent with what you proposed in the PR description too. We've drafted `contract` and `stateless` as two profiles in one spec: `contract` stays as already deployed and tested, `stateless` is added and credited to this PR and to `0d1026/Rialto` by name, plus the missing C-account language above. We're ready to open that as a PR against whichever of #3098/#3134 the maintainers want to keep open, or as a fresh PR replacing both, whichever they'd rather review. Happy to coordinate directly if useful.

HeylmStoned:
Read this against #3098 and Eras256’s comparison. Agree the two designs shouldn’t be treated as mutually exclusive winners - the real split is settlement mechanism, not whether upto is possible on Stellar.

One thing that matters if both profiles land, from the facilitator / discovery side: scheme: "upto" alone is not enough on the wire. /supported, PaymentRequirements.extra, and Bazaar catalog filters all key off scheme + network today. If stateless and contract are both valid, a resource advertising bare upto is ambiguous - a facilitator can verify a payload the seller didn’t mean, and an agent can’t tell which contract to authorize. If the merged spec keeps both, it needs a stable profile id in extra (and in /supported) so cataloging and clients can discriminate.

For a public Bazaar, the facilitator-agnostic / stateless profile is the better default: listings shouldn’t be bound to one operator. The stored-nonce contract profile is a reasonable opt-in where a deployment wants facilitator binding and will own TTL. Happy to implement whichever profile the spec freezes, and to run it through /verify /settle /supported plus a cataloged resource once the extra shape is explicit.

Eras256:
Thanks for this, both points land.

On the wire-level gap: you're right, and checked against the actual Periplo implementation, not just the spec text, it turns out worse than "missing routes." `extra.uptoProfile` is already defined as a required, shared field in `#3098`'s `scheme_upto_stellar.md`, so the spec side is covered. The implementation isn't. `/supported` can't report `upto` at all yet (only `exact` is wired into the facilitator core), neither `GET /discovery/resources` nor `GET /discovery/search` exists as an HTTP route yet, and the catalog write path has a real bug waiting: the `accepts`-array dedupe key is `scheme|network|asset|payTo`, not `extra`-aware. Once `upto` cataloging exists, a facilitator advertising both profiles for the same resource would have one profile's listing silently overwritten by the other's on the next payment, not just fail to discriminate between them, actually lose the entry. Documented all three in Periplo's own tracking doc before writing this reply, credited to your comment by name, tracked as real implementation work, not closed by a comment.

On the "(default)" label: fair catch, and it's fixed now regardless of who's right on the merits, since it contradicted the spec's own "Choosing between" section a few paragraphs down, which already says neither profile is a universal default. For what it's worth on the substance: the catalog listing itself stays facilitator-agnostic under either profile, it's the per-authorization binding that differs, since a resource server can choose which facilitator to bind at signing time per request rather than fixing one at the listing level. So I don't think it settles which profile a public Bazaar should prefer by default on its own, but the label shouldn't have implied a ranking the rest of the document doesn't back, so it's gone.

davedumto:
Independent implementor, cross-checking this spec against five other Stellar `upto`
implementations read in source (not just skimmed) — rail402, Rialto, openx402, LumenGate,
and Periplo (#3098's own reference implementation) — plus our own deployed instance:
`CDHPA64M73TUTEM4MMHIWIXINBQXH7JJXFGZMGH22VJWFJFROMR6QV2S` on testnet, build verifiable
against source in `contracts/upto-stellar/` (wasm hash published, `stellar contract fetch`
reproduces it), first settlement `72c816a63ab9da21b1403ff5199e4f21b9947c0769c55312a8cf0dc7e6ecf3db`
— ceiling 1,000,000 authorized, 400,000 actually settled, confirmed on Horizon.

Five things this reconciliation surfaced, ranked by how settled they seem:

## 1. What's already converged

Every implementation reviewed — rail402, Rialto, openx402, LumenGate, Periplo, and this
PR's own stateless design — lands on the same shape: `require_auth_for_args` over
`(payTo, token, max, expiry, nonce/salt)`, with `actual` deliberately excluded from the
signed tuple and bounded on-ledger rather than trusted from the caller. That's not a
coincidence across five independent teams; it's the only design that satisfies partial
settlement without letting a dishonest facilitator overcharge. Worth the spec stating this
as a MUST rather than leaving it as an emergent convention — everyone competent is already
here, so pinning it costs nothing and closes the door on a future implementation trying to
simplify it away.

## 2. The hook question — needs a position, not just a mention

rail402's `settle()` takes an optional `hook: Option<Address>` as its 8th argument, and it's
explicitly outside the signed tuple — their own docs confirm this ("What is not signed is
`actual_amount` and the settlement `hook`"). rail402's own contract comment argues this is
safe: a misdirected hook "can only fail to reconcile... never overcharge." That claim holds
for the case they tested. It doesn't cover one case we don't see addressed anywhere in their
tests or docs: `settle()` doesn't wrap the hook invocation in a try-call, so a hook address
that traps (points at a broken or adversarial contract) propagates that trap through the
whole transaction, since Soroban cross-contract calls fail synchronously by default — that's
a DoS on settlement itself, not just a missed reconciliation. We haven't reproduced this on
either rail402's contract or ours; it's inferred from Soroban's standard trap-propagation
behavior, not observed, and we want to be clear about that distinction.

openx402 took a different approach worth the spec's attention: `on_settled_v1` with an
explicit versioned entry point (their own docs state a field change requires a new
`on_settled_v2`, not a silent ABI break), gated by Soroban's invoker-authorization
(`require_auth()` on the caller, satisfied because the settlement contract itself is the
invoker) plus inclusion of the hook address in the payer's original signed payload. Calling
that a "signed hook" is reasonable shorthand, but precisely: it's invoker-authorization plus
signed-payload-inclusion, not a standalone signed hook-call primitive — worth being exact
about in spec language so implementers don't reach for a heavier mechanism than what's
actually required, or a lighter one than what's actually needed.

The spec doesn't currently take a position on hook authorization or failure isolation.
Given two independent, real designs already diverge here, we think it should — at minimum,
requiring implementations to isolate a hook call (try-invoke or equivalent) so a bad hook
address degrades to "unreconciled" rather than "settlement itself failed."

## 3. Custody window — the empirical case cuts against the zero-custody consensus, and the spec should say so explicitly

Every implementation reviewed, including this PR's, converges on custody-avoidance as a
design goal — the contract shouldn't hold funds it doesn't need to. The one team that
actually measured the cost of that choice found the opposite of what the consensus assumes.
LumenGate ran a real `simulateTransaction` benchmark, on two actual deployed testnet
contracts, comparing an allowance-based design against an escrow-and-refund design at a
fixed scenario (max 1,000,000 / actual 400,000 / feeBps 500): minimum resource fee was
234,098 stroops for the allowance-based design versus 160,366 for escrow — escrow is 31.5%
*cheaper*, not more expensive. Their own write-up discloses the tradeoff honestly: the
escrow design's advantage comes partly from not persisting a nonce record, which also gives
up `cancel`/`is_settled` capability — not a free win, a real one with a real cost attached.

This PR's own thread independently reinforces the same tension with a second, directly
on-topic data point: a reviewer settled real transactions from both this PR's stateless
design and #3098's stateful design against Horizon and found the stateless design 25-30%
cheaper in `fee_charged` (32,731 / 30,585 stroops vs. 42,872) — a different axis (custody
vs. statefulness) landing on a similar shape of result: the design the community leans
toward on trust/complexity grounds isn't the cheapest one measured so far, on either axis.

That's real, converging signal from two independent teams that the spec should decide, not
leave implicit. We're not arguing for custody — we don't build it either — but "the
consensus design costs more, empirically, and nobody's written that down" is worth being
explicit about rather than quietly true.

## 4. Auto-revoke — a real, conditional interop divergence, precisely stated

Rialto's `auto_revoke` (`0d1026/Rialto`, `contracts/upto-settlement/src/lib.rs`) requires a
specific signed authorization-tree shape: when `autoRevoke = true`, the client's signed tree
carries exactly two sub-invocations — the allowance-granting `approve` and a zeroing
`approve(0, 0)` — enforced by Soroban's own sub-invocation matching, not app logic. Their
spec states this exactly (`docs/specs/scheme_upto_stellar.md`). When `autoRevoke = false`,
it's one sub-invocation, not two — worth stating precisely, since "requires a two-child
tree" undersells that it's conditional on a flag the client controls.

We want to correct something in how we understand this ourselves: rail402's own contract
independently discusses the identical underlying Soroban mechanic — any sub-invocation not
covered by the client's signature fails `require_auth` — as their stated reason for *not*
building an equivalent allowance-reset feature at all. That's not rail402 demonstrating a
bug in Rialto's design; it's two independent teams hitting the same constraint and choosing
differently. We don't want to misattribute that as a cross-project finding when it isn't
one — it's convergent, not connected.

What is a real, spec-relevant gap: nothing in the spec currently states this tree-shape
requirement explicitly for implementations that offer an auto-revoke-equivalent feature, so
a client library built against one implementation's tree shape has no guarantee it matches
another's. Worth pinning down if the merged spec keeps an auto-revoke option at all.

## 5. Nonce-TTL boundary as a required test vector

For any implementation using a stored, TTL-bounded nonce (rail402's design; Periplo's,
`#3098`'s reference implementation, which uses `extend_ttl` the same way) — there's a real
boundary condition worth making a required test vector rather than an implementation detail
each team rediscovers independently: an authorization can be signed with an
`expiration_ledger` further out than the nonce record's own TTL, which would let the
identical authorization settle again after the nonce record itself expires, silently
reopening replay for the residual. rail402's contract guards this explicitly
(`NONCE_TTL_LEDGERS = 17_280`, ~24h at 5s/ledger, rejecting any `expiration_ledger` beyond
`current + NONCE_TTL_LEDGERS`) and tests both sides of the boundary
(`rejects_an_authorization_that_would_outlive_its_nonce_record`,
`accepts_an_expiration_at_the_boundary`). We want to be precise about what we're claiming
here: we found this documented and tested in rail402's shipped contract, not a bug we
independently discovered being fixed in their history — there's no public commit history
showing a before/after, just a contract that already guards the case correctly.

Worth noting in the spec's favor: this PR's own stateless design sidesteps this entire bug
class by construction, since it relies on Soroban's native per-entry auth nonce rather than
an app-managed TTL. That's a real point in this design's favor, not just a gap in a
competing one — but it means the required-test-vector recommendation applies specifically
to the `contract`/stateful profile, wherever the merged spec ends up keeping one.

---

Given the thread above is already converging toward documenting both profiles in one merged
spec rather than picking a winner between #3098 and this PR — points 1, 4, and 5 read as
inputs to that merge; points 2 and 3 read as open questions the merged spec should take an
explicit position on rather than leave implicit. Happy to help verify any of the above
independently if useful, including running our own deployed instance against whichever
shape the merged spec settles on.

Iam0TI:
@davedumto  

Yeah, we’re still thinking about simplifying this and just making "autoRevoke" always "true" [see here](https://github.com/0d1026/Rialto/issues/13).

That said, I don’t think the current design is too heavy for implementers. There are really only two cases to handle:

- "autoRevoke = false" -> "approve(max)"
- "autoRevoke = true"-> "approve(max)" + "approve(0)"

So it’s basically just two auth-tree shapes.

There’s also a reason we kept "false" around: it makes the transaction a bit cheaper since you skip the extra revoke call. And since the approval is already bounded by "expirationLedger", the leftover allowance is temporary anyway.

So for now I see it more as a tradeoff between slightly cheaper settlement and immediate cleanup, rather than something fundamentally unsafe. But we’re still open to making auto-revoke mandatory if that ends up making cross-implementation compatibility cleaner.

davedumto:
That makes sense @Iam0TI  the two-shape model is cleaner than I initially read it,
and the cost argument is real. The bounded expiration does handle the
residual-allowance concern for the common case.

The interop angle is the one worth the spec resolving: a client library
needs to know which shape to build without reading each implementation's
docs separately. If the merged spec states the two shapes explicitly
(autoRevoke=false → one approve, autoRevoke=true → approve + approve(0))
that's enough — implementers can diverge on the default but clients know
what to expect from either.

Appreciate you linking the issue — will follow that thread.

Eras256:
On point 5, for our contract specifically: the nonce-outliving-its-TTL case you're describing is structurally prevented rather than boundary-checked against a fixed ceiling. `settle()` derives the nonce entry's `extend_ttl` directly from `deadline_ledger` at settlement time (`contracts/upto-settlement/src/lib.rs`):

```rust
let ttl = authorization.deadline_ledger.saturating_sub(ledger);
env.storage().temporary().set(&key, &authorization.deadline_ledger);
env.storage().temporary().extend_ttl(&key, ttl, ttl);
```

so there's no separately-configured nonce TTL that could fall short of what an authorization signs for; the two are the same value by construction, not two values checked against each other. `MAX_WINDOW_LEDGERS = 17_280` (~1 day) is the complementary piece: it bounds how large `deadline_ledger - valid_after_ledger` can ever be in the first place, independently verified against the network's own real storage-TTL ceiling rather than assumed. Both sides of that boundary are covered by real tests (`rejects_window_exceeding_the_contract_maximum`, `accepts_window_exactly_at_the_contract_maximum`), and the TTL-covers-deadline property itself was confirmed live on testnet, not just in a unit test: a settled transaction's nonce entry `liveUntilLedgerSeq`, read back from RPC, exceeded `deadline_ledger` by exactly the requested margin.

Not proposing this as *the* required test vector over rail402's boundary check, since they're different mechanisms answering the same risk. Just flagging that "stored TTL-bounded nonce" implementations aren't a single shape here, worth the merged spec allowing either "derive TTL from the signed deadline" or "check the deadline against a fixed TTL ceiling," rather than assuming one implies the other.

bomanaps:
Good precision on the two mechanisms, and agreed the spec should not assume one stateful shape implies the other, worth keeping the layering clear though. This whole class, whether solved by a fixed ceiling or by deriving TTL from the signed deadline like yours, only exists once a contract stores its own nonce record. The stateless profile has no record to size, so the protocol nonce closes replay by construction. So if the merged document describes both profiles, we would put it this way. The stateless design in this PR is the base profile since it carries no implementation defined boundary to test, and the stateful profile section carries the required test vectors, with both mechanisms as acceptable answers. Happy to draft that structure into the PR so the merge has concrete text to react to instead of a comparison thread.

Eras256:
Sounds right, and consistent with what I asked for on the TTL front. One thing to fold into the structure before you draft it: HeylmStoned's wire-level point from earlier in this thread, scheme: "upto" alone doesn't disambiguate between the two profiles once both exist. If the merged doc keeps both, it needs a stable profile id in extra (and reflected in /supported) so clients and catalogs can tell them apart, same as I tracked it in Periplo's own DEFERRED.md. Otherwise the base-vs-stateful split reads clean but leaves the exact ambiguity we already found a client can't resolve on its own.

Per CONTRIBUTING: written with significant AI assistance under significant human monitoring, direction, and correction. HeylmStoned's and bomanaps's comments above were re-read against the real thread today.

Eras256:
From the Periplo side: this is a real improvement over what we shipped. Our own `UptoSettlement` uses a client-supplied `nonce: BytesN<32>` in `temporary()` storage with a TTL sized to `deadline_ledger`, defended on the argument that "the deadline dominates the nonce" — but relying on Soroban's own per-auth-entry nonce instead removes that whole TTL-sizing surface, not just narrows it. No objection to converging on this over our approach.

One thing worth a line in the spec if it isn't already covered: how does a client discover whether a given auth entry's protocol-level nonce has already been consumed, before signing? With a custom nonce we could expose "is this nonce still valid" as a read-only contract call; not sure there's an equivalent surface for the protocol-level one, or if that's even needed given the short validity window.

davedumto:
Independent implementor again, following up on the August thread.

We've drafted a convergence document that tries to extract the MUST
requirements where all surveyed implementations agree — rail402, this PR,
#3098, Rialto, openx402, LumenGate — and name the places they diverge as
open questions for the TSC, without proposing a new contract design. It
credits rail402 as the contract design author; our own deployment is their
contract vendored verbatim at a pinned commit, and what we contribute is
independent reproducible-build verification plus four live testnet
settlements, not authorship.

Two things in it that came out of this thread rather than from us:

- Hook failure isolation as a SHOULD. If settle() does not wrap the hook in
  a try-call, a trapping hook propagates and takes the settlement with it.
  We've still not reproduced this against any implementation including our
  own, so it's written as inferred-not-observed, and we'd rather it be
  refuted than quietly carried.

- Off-chain observers reading the settled amount from the token's emitted
  transfer event rather than the envelope's call args. This one we did hit
  in practice: our explorer's classifier only recognised the top-level
  exact-scheme transfer shape and silently missed upto settlements
  entirely, because the token movement is a nested sub-invocation. Scoped
  to indexers, not to the contract — the contract necessarily takes
  actual_amount as an argument.

What it does NOT currently cover, and should before it goes anywhere:
@Eras256's profile-disambiguation point from 2026-08-26 — that scheme:
"upto" alone does not distinguish the two profiles once both exist, and
needs a stable id in extra reflected in /supported. That's a real gap in
our draft and we'd want to fold it in.

Before opening a PR we wanted to ask where you'd like this to land:

  1. a separate file alongside this spec (we'd use a distinct filename —
     filing at scheme_upto_stellar.md would collide with this PR and force
     an either/or, which is the opposite of the intent),
  2. folded into this PR as suggested edits,
  3. or not at all, if it's redundant with where the merged doc is heading.

Happy to share the draft either way. If (2), we'd rather hand it over than
carry a parallel document.

Per CONTRIBUTING: written with significant AI assistance under significant
human monitoring, direction, and correction.

Eras256:
Agree with landing this inside #3134 rather than a separate doc. The three implementations diverging on the same underlying contract (rail402's design, this PR, #3098) already made a standalone convergence doc feel like a fourth voice in a conversation that needs fewer, not more — same reasoning we used consolidating our own drift into UPTO-CONVERGENCE.md instead of a new file.

On the profile-disambiguation point: appreciate you pulling that in. To be concrete about what it needs — scheme: "upto" alone stops being sufficient the moment both contract and smartAccount profiles exist on the same network, since a verifier can't tell which validation path to run from the scheme name. A stable id field in extra, mirrored in /supported, is what lets a client and facilitator agree on profile before verification starts. Happy to review wording once it's in the PR.

PR #3138, opened 2026-08-12
fix(bazaar): use raw resource URL as canonical for opaque-origin schemes

### LINKED ISSUE #3121: [Bug]: extractDiscoveryInfo builds a broken canonical URL for mcp://tool/{toolName} resources
### What happened?

`extractDiscoveryInfo` (`@x402/extensions/bazaar`) builds the catalog URL for a non-parameterized resource as:

```ts
const canonicalUrl = routeTemplate
  ? `${url.origin}${routeTemplate}`
  : `${url.origin}${url.pathname}`;
```

For an MCP tool, `routeTemplate` is never set (MCP routes aren't parameterized), so this always takes the `${url.origin}${url.pathname}` branch. The package's own docs (and the x402 e2e test) document `mcp://tool/{toolName}` as the expected `resource.url` form for MCP resources.

`mcp:` is not a WHATWG "special scheme" (only `http`/`https`/`ws`/`wss`/`ftp`/`file` are), so a `mcp://tool/x` URL gets an **opaque origin**, and `URL.prototype.origin` serializes an opaque origin as the literal string `"null"`:

```
> new URL("mcp://tool/financial_analysis_xyz").origin
'null'
```

Expected: `canonicalUrl` is `"mcp://tool/financial_analysis_xyz"`.
Actual: `canonicalUrl` is `"null/financial_analysis_xyz"`.

Any facilitator that follows the documented `mcp://tool/{toolName}` convention and calls `extractDiscoveryInfo` without special-casing the MCP branch will catalog every MCP resource under a broken URL.

### Steps to reproduce

```ts
import { extractDiscoveryInfo, declareDiscoveryExtension } from "@x402/extensions/bazaar";

const extension = declareDiscoveryExtension({
  toolName: "financial_analysis",
  description: "Analyze financial data for a given ticker",
  inputSchema: {
    type: "object",
    properties: { ticker: { type: "string" } },
    required: ["ticker"],
  },
});

const paymentPayload = {
  x402Version: 2,
  resource: { url: "mcp://tool/financial_analysis" },
  accepted: {
    scheme: "exact",
    network: "eip155:84532",
    asset: "0x0",
    amount: "1",
    payTo: "0x0",
    maxTimeoutSeconds: 60,
    extra: {},
  },
  payload: {},
  extensions: extension,
};

const discovered = extractDiscoveryInfo(paymentPayload, paymentPayload.accepted);
console.log(discovered?.resourceUrl);
// Expected: "mcp://tool/financial_analysis"
// Actual:   "null/financial_analysis"
```

Also directly reproducible without the package at all:

```
$ node -e 'console.log(new URL("mcp://tool/financial_analysis").origin)'
null
```

### Environment

- `@x402/extensions` `2.21.0` (`typescript/packages/extensions/src/bazaar/facilitator.ts`, `extractDiscoveryInfo`)
- Node.js v22.23.2
- Found while building a Stellar facilitator's automatic-cataloging integration test against a real Postgres catalog — the first row written for an MCP resource had `url: "null/..."`, not caught by type-checking or unit tests with fakes, only by writing the real row and reading it back.

### Suggested fix

In the `resourceUrl.startsWith("mcp://")` (or equivalent MCP) branch, skip `URL` parsing for the canonical-URL computation entirely and use the already-available `toolName` directly: `` `mcp://tool/${toolName}` ``. Happy to open a PR with this fix and a regression test if useful.

Eras256:
Closes #3121.

## What

`extractDiscoveryInfo`'s canonical-URL computation (`packages/extensions/src/bazaar/facilitator.ts`) always ran `new URL(resourceUrl)` and built the canonical URL from `url.origin`. `mcp:` is not a WHATWG "special scheme" (only `http`/`https`/`ws`/`wss`/`ftp`/`file` are), so an `mcp://tool/{toolName}` resource URL parses to an opaque origin, and `URL.prototype.origin` serializes that as the literal string `"null"` — producing `"null/financial_analysis"` instead of the documented `"mcp://tool/financial_analysis"`.

## The fix

Scheme-agnostic, per [@whawk46's comment](https://github.com/x402-foundation/x402/issues/3121#issuecomment-5249065267) on the issue: compute the origin, and when it comes back as the opaque-origin sentinel (`origin === "null"`), skip canonicalization entirely and use the raw resource URL string as canonical. No `mcp://`-specific branch, no scheme allowlist to maintain going forward — this covers every current and future non-special scheme, which is the actual shape of the bug, not "mcp specifically."

## Tests

Two regression tests, shaped the way @whawk46 asked for in the issue thread: one for `mcp://`, and a second for an unrelated, made-up non-special scheme (`test-scheme://`), so the test pins the general rule ("opaque origins are never serialized into canonical URLs") rather than "mcp is special-cased." Confirmed both fail on the pre-fix code (2 failed, 136 passed) and pass with the fix (138 passed) before opening this PR.

`pnpm --filter @x402/extensions build` and `pnpm --filter @x402/extensions test` (full package, 522 tests) both pass.

## Credit

@whawk46 suggested this exact fix shape (detect the opaque-origin sentinel instead of a scheme allowlist) and the regression-test framing (a second, unrelated scheme, not just re-testing `mcp://`) in a comment on #3121. Both are implemented here as proposed, not just "per feedback" — the fix and the tests are whawk46's design, this PR is the implementation of it.

## AI disclosure

Drafted with AI assistance, reviewed and verified by hand. The fix and both regression tests were run against a real `pnpm build`/`pnpm test`, not asserted — confirmed failing without the fix, passing with it, before this PR was opened. Flagging per CONTRIBUTING.md so reviewers can calibrate.

## Checklist

- [x] I have formatted and linted my changes
- [x] My commits are signed
- [x] Changelog fragment added (`typescript/.changeset/bazaar-opaque-origin-canonical-url.md`, `@x402/extensions` patch) — this is a user-facing bug fix, not docs-only

### DIFF
--- a/typescript/.changeset/bazaar-opaque-origin-canonical-url.md
+++ b/typescript/.changeset/bazaar-opaque-origin-canonical-url.md
@@ -1,0 +1,5 @@
+---
+'@x402/extensions': patch
+---
+
+Fix `extractDiscoveryInfo` building a broken canonical URL (`"null/..."`) for resource URLs on schemes without a WHATWG-defined origin, such as `mcp://tool/{toolName}`. Canonicalization is now skipped for any opaque-origin scheme, not just `mcp://`, and the raw resource URL is used as canonical instead.
--- a/typescript/.changeset/bazaar-opaque-origin-canonical-url.md
+++ b/typescript/.changeset/bazaar-opaque-origin-canonical-url.md
@@ -2,4 +2,4 @@
 '@x402/extensions': patch
 ---
 
-Fix `extractDiscoveryInfo` building a broken canonical URL (`"null/..."`) for resource URLs on schemes without a WHATWG-defined origin, such as `mcp://tool/{toolName}`. Canonicalization is now skipped for any opaque-origin scheme, not just `mcp://`, and the raw resource URL is used as canonical instead.
+Fix `extractDiscoveryInfo` building a broken canonical URL (`"null/..."`) for resource URLs on schemes without a WHATWG-defined origin, such as `mcp://tool/{toolName}`. The canonical is now reconstructed from `protocol` + `host` + `pathname` for any opaque-origin scheme, not just `mcp://`, which strips the query string and fragment the same way the special-scheme path already does, instead of falling back to the raw resource URL unstripped.
--- a/typescript/packages/extensions/src/bazaar/facilitator.ts
+++ b/typescript/packages/extensions/src/bazaar/facilitator.ts
@@ -573,12 +573,23 @@ export function extractDiscoveryInfo(
     return null;
   }
 
-  // Strip query params (?) and hash sections (#) for discovery cataloging
+  // Strip query params (?) and hash sections (#) for discovery cataloging.
+  // Canonicalization (origin + path) is only meaningful for a WHATWG
+  // "special scheme" (http, https, ws, wss, ftp, file) — those are the only
+  // schemes the URL spec defines a real origin for. Any other scheme (e.g.
+  // mcp://) parses to an opaque origin, which URL.prototype.origin
+  // serializes as the literal string "null", not a per-scheme special case
+  // to detect and list here: skip canonicalization entirely and use the
+  // raw resource URL as-is, so this holds for every current and future
+  // non-special scheme, not just the ones enumerated by name.
   const url = new URL(resourceUrl);
   // If a routeTemplate is present (dynamic route), use it as the canonical path
-  const canonicalUrl = routeTemplate
-    ? `${url.origin}${routeTemplate}`
-    : `${url.origin}${url.pathname}`;
+  const canonicalUrl =
+    url.origin === "null"
+      ? resourceUrl
+      : routeTemplate
+        ? `${url.origin}${routeTemplate}`
+        : `${url.origin}${url.pathname}`;
 
   // Extract description and mimeType from resource info (v2) or requirements (v1)
   let description: string | undefined;
--- a/typescript/packages/extensions/src/bazaar/facilitator.ts
+++ b/typescript/packages/extensions/src/bazaar/facilitator.ts
@@ -574,19 +574,24 @@ export function extractDiscoveryInfo(
   }
 
   // Strip query params (?) and hash sections (#) for discovery cataloging.
-  // Canonicalization (origin + path) is only meaningful for a WHATWG
-  // "special scheme" (http, https, ws, wss, ftp, file) — those are the only
-  // schemes the URL spec defines a real origin for. Any other scheme (e.g.
-  // mcp://) parses to an opaque origin, which URL.prototype.origin
-  // serializes as the literal string "null", not a per-scheme special case
-  // to detect and list here: skip canonicalization entirely and use the
-  // raw resource URL as-is, so this holds for every current and future
-  // non-special scheme, not just the ones enumerated by name.
+  // `url.origin` is only meaningful for a WHATWG "special scheme" (http,
+  // https, ws, wss, ftp, file) — those are the only schemes the URL spec
+  // defines a real origin for. Any other scheme (e.g. mcp://) parses to an
+  // opaque origin, which URL.prototype.origin serializes as the literal
+  // string "null". That's not a reason to skip stripping too: `protocol`,
+  // `host`, and `pathname` all populate correctly for an opaque-origin URL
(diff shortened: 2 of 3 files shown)


whawk46:
Read the diff top to bottom — this is the fix I had in mind, and honestly the tests came out better than I pictured them: the comment above the describe block explaining *why* two schemes, so the next person doesn't "simplify" it back down to one, is the kind of thing that keeps a regression pinned for years. Thank you for the generous credit, though I'd push some of it back: suggesting a shape in a comment is the cheap part. Building it, running it against the real suite both ways, and writing the changeset is the part that counts.

One observation from reading the new branch closely, non-blocking and possibly a follow-up rather than this PR: the raw-URL path now skips the query/fragment stripping that this function exists to do. A producer declaring `mcp://tool/financial_analysis?session=abc` gets its query string into the canonical, which recreates the per-variant catalog duplication the stripping was preventing — and the `routeTemplate` branch is bypassed for opaque schemes too. Both are hypothetical today (MCP tool URLs in the wild don't carry queries, and I haven't seen templated ones), which is why I wouldn't hold the PR for it. If you want it in-scope, the parser already gives you what you need even for non-special schemes: `url.protocol`, `url.host`, and `url.pathname` all populate for `mcp://tool/x`, so `` `${url.protocol}//${url.host}${url.pathname}` `` keeps the stripping and light normalization without reintroducing any scheme knowledge. Maintainer's call whether that belongs here or in a follow-up; the fix as written is strictly better than what it replaces either way.

LGTM on the substance. Nice to see #3121 go from filed to fixed with the general rule intact.

Eras256:
Thanks, both the query-stripping gap and the exact fix are right, and it's small enough to just put in this PR rather than punt to a follow-up.

Implemented it as suggested: the opaque-origin branch now reconstructs from `protocol` + `host` + `pathname` instead of falling back to the raw resource URL, so it strips the query/fragment the special-scheme path already strips, without reintroducing any scheme knowledge. Added a regression test with a query string on an opaque scheme (`mcp://tool/financial_analysis?session=abc`), confirmed it fails against the pre-fix code (the exact assertion you'd expect: the query survives into the canonical) and passes with the fix, same discipline as the original two tests. Full suite still green, 139 tests. Left the routeTemplate-for-opaque-schemes point out of scope here, since you flagged that one as more open, happy to pick it up separately if you want it pinned down too.

whawk46:
Read 19de93e — that's the fix, done the way the first round was done: the reconstruction exactly as sketched, the new test failing for the right reason before and passing after, and the changeset text actually updated to describe the new behavior, which is the detail most people forget. Nothing further from me on this PR.

On routeTemplate for opaque schemes: I'd leave it unpinned, deliberately. The mechanical form is obvious whenever it's needed — protocol + host + template, same shape as this commit — but whether collapsing templated tools into one canonical is even the right catalog semantics for a non-special scheme is a question I'd rather answer with a real producer in hand than by guessing ahead of one. The first time someone declares a template on an opaque scheme, it's a five-line PR with a test, and this thread is the paper trail for where it goes.

LGTM as it stands — merge-ready from my side.

Eras256:
Nothing new to add on the substance, just flagging that this has been mergeable with an LGTM since Aug 13 and hasn't moved. Would appreciate a maintainer merge when someone has a moment.

whawk46:
Seconding the merge request as the reviewer. I read `19de93e` on 13 August and had nothing further then and nothing further now: the opaque-origin branch reconstructs from `protocol` + `host` + `pathname`, the regression test fails against the pre-fix code for the right reason and passes after, and the changeset text was updated to describe the new behaviour rather than left stale.

It also fixes a real indexing defect rather than a cosmetic one — a query string surviving into the canonical URL means the same resource indexes under as many identities as it has session parameters, which is the kind of thing that quietly inflates a catalog and is very hard to attribute afterward.

Nine days mergeable with an approving review. Whenever a maintainer has a moment.

Eras256:
@phdargen — flagging this one directly since I see you've been merging steadily this week. This has had an LGTM from @whawk46 since Aug 13 (`19de93e`), a follow-up nudge from me on Aug 22, and no movement since — three weeks now with nothing blocking except a maintainer merge. Still green and mergeable. Happy to rebase or address anything if something's changed since the review.

whawk46:
+1 @phdargen — my LGTM still stands. The opaque-origin canonicalization fix
remains clean, green, and zero-risk to merge.

PR #3141, opened 2026-08-13
feat: add solana upto to go sdk (+ typescript parity fixes)

CarsonRoscoe:
Implements the SVM `upto` scheme in the Go SDK (client, server, facilitator) to spec
parity with the TypeScript implementation from #3094, per the spec in #3135
(`specs/schemes/upto/scheme_upto_svm.md`) and built on the Go payment-flow support
added in #3115.

`upto` lets a client authorize a ceiling onchain and the server settle for actual
usage: the client escrows a maximum in a payment channel, the server signs a voucher
for the metered amount, and the facilitator sponsors fees and rent, co-signs and
broadcasts the open, settles the claim, and reclaims channel rent afterwards. This PR
also hardens the full rent-recovery lifecycle in both SDKs, since that's the part of
the scheme where a bug costs real money.

## What's included

**Program bindings** — `go/mechanisms/svm/paymentchannels/`: instruction builders for
`open`, `settle_and_seal`, `distribute` and `reclaim`, Borsh encode/decode, PDA
derivation, voucher signing, a strict verifier for the client-supplied open transaction
(account layout, sealed distribution, compute budget and signature caps, memo binding,
open-slot freshness), and `DiscoverChannelsByRentPayer` for onchain
`getProgramAccounts` discovery.

**Scheme roles** — `go/mechanisms/svm/upto/{client,server,facilitator}`, plus
`go/signers/svm/authorizer.go` for the receiver-authorizer hot key that signs
settlement vouchers (no SOL required).

**Rent recovery** — `RentCleanupManager` runs two independent passes so rent can only
be lost if every layer below it fails at once:

- `Cleanup` reclaims rent for channels already in offchain storage, on a short
  interval (e.g. every few minutes). `Discover` is a separate, standalone pass for
  the `getProgramAccounts` recovery sweep, meant to run rarely (e.g. daily), since
  it's an onchain scan rather than an index read; it writes any `Distributed`
  channels missing from storage back in (without overwriting a tracked record) so
  the next `Cleanup` pass reclaims their rent. Neither pass depends on the other's
  interval.
- Deposit settlement writes the channel to durable storage *before* broadcasting the
  open, and fails closed if that write fails, so a broadcast open can never end up
  with nothing able to find it later.
- Verifying an open channel retries a few times on backoff if the account is
  momentarily missing (RPC replica lag), while existing invalid state still fails
  immediately.
- The storage scan and the per-rent-payer reclaim batches are budgeted separately:
  `maxTxsPerRun` bounds how far one `Cleanup` pass scans before resuming next run,
  and `maxTxsPerSigner` caps reclaim transactions per rent-payer signer, so a large
  reclaim backlog can't starve the scan (or vice versa). A scan that runs out of
  budget mid-pass resumes from a cursor next run; the scan sorts by channel ID before
  applying that cursor, so resumption is stable even for storage backends that don't
  preserve insertion order.
- `Stop()` cancels the running context and waits for any in-flight pass to finish
  before returning, so shutdown can't orphan a broadcast settle or reclaim mid-flight.
- Both the facilitator and the cleanup manager accept a prebuilt RPC client
  (`Config.RPC` / `RentCleanupConfig.RPC` in Go, `rpc` in TypeScript), preferred over
  building one from the configured URL, so operators can route sends through their
  own instrumented, rate-limited, or header-bearing transport. Parity with the
  TypeScript RPC-injection support notorious-d-e-v added in #3183.
- `DiscoverChannelsByRentPayer` independently validates owner, discriminator, and
  rederived PDA against a `getProgramAccounts` sweep, so `Distributed` channels
  missing from offchain storage are still recoverable.
- A golden serialization test pins the largest reclaim batch that fits in one Solana
  transaction, gating how high a facilitator can set its per-tx reclaim count.

All of the above is implemented identically in Go and TypeScript.

**Shared SVM additions** — the `upto` payload type and voucher helpers
(`go/mechanisms/svm/types.go`) and a stablecoin registry with mint, token-program and
precision lookups (`constants.go`, `utils.go`), which the server uses to resolve a
trailing price symbol like `"1.50 PYUSD"` and advertise the token program that mint
requires. Token-2022 stablecoins work without extra configuration.

**Examples, e2e and docs** — a new `examples/go/facilitator/upto` example, updated Go
server/client examples, Go client/server/facilitator wiring in `e2e/` with `/upto/svm`
now listing `go`, and scheme docs across `go/README.md`, `CLIENT.md`, `SERVER.md`,
`FACILITATOR.md` and the mechanism READMEs.

## TypeScript changes

TypeScript `upto` is merged but unreleased, and both SDKs ship together, so every fix
and hardening change below landed in TypeScript too — none of it shipped as a
Go/TS mismatch.

Fixes found while bringing Go to parity:

- Channel reads used the RPC default commitment (`finalized`), which lags a freshly
  broadcast open by seconds, so the facilitator could report a live channel as missing
  on the next request. Now `confirmed`
  (`typescript/packages/mechanisms/svm/src/upto/facilitator/channel.ts`).
- The per-run close cap used `break`, so a backlog of closable channels stopped the
  scan entirely and starved reclaims that were already due. Now `continue`.
- Cleanup passes could overlap: an operator calling `cleanup()` could race the
  interval loop and submit the same close or reclaim twice. Passes are now serialized.
- An unrecognized channel status was silently skipped, leaving the record in storage
  forever with no signal to the operator. Now reported through `onError`.
- A failed reclaim batch only reported the first channel in it through `onError`. Now
  every channel in the batch is reported.
- `validateSvmAddress` was a charset-and-length regex only, so base58 that decodes to
  fewer than 32 bytes passed and surfaced later as an opaque open-transaction mismatch.
  It now also requires a 32-byte decode. This also affects the released `exact` SVM
  scheme — it only rejects strings that were never valid addresses, so it's a
  correctness fix, but it's the one change here with a blast radius beyond `upto`.
- A non-string `extra.memo` was coerced into memo data, and `extra.tokenProgram`
  wasn't checked against the two supported SPL token programs. Both now fail as
  payment-requirements errors.
- `Cleanup` didn't observe context/signal cancellation, so `Stop()` could still walk
  every remaining record over a dead connection during shutdown. It now returns as
  soon as cancellation is observed (superseded below by `Stop()` awaiting the pass).
- The SVM `upto` client scheme never wired `findDefaultAsset`, so client spend
  controls rejected devnet/mainnet USDC as a non-default asset unless a caller
  explicitly configured `allowedAssets`. Every other client scheme already wires this;
  `upto` now does too.

Follow-up from the ComputeBudget work in #3155, further refined by review feedback:

- `ClientSvmConfig` briefly grew `computeUnitLimit` / `computeUnitPriceMicroLamports`
  overrides shared between `exact` and `upto`. Review flagged that `0` meant "omit the
  instruction" on `upto` but "use a literal zero price/limit" on `exact` — not a
  coherent shared config — and no caller configured either field. Both were removed;
  `exact` and `upto` clients now use their documented per-scheme defaults with no
  client-side override surface. Both types were unreleased, so this is a pre-release
  cleanup, not a breaking change.

Additional review-feedback fixes covered by this PR, applied identically in Go and
TypeScript:

- An empty `extra.memo` is now treated as unset (client generates a nonce memo,
  facilitator does not require an exact match) instead of as a required empty-string
  memo.
- Onchain discovery moved off the cleanup interval into its own `discover()` pass with
  its own interval, as described above.
- `stop()` / `Stop()` now returns only after the in-flight pass finishes.
- The reclaim budget split into `maxTxsPerRun` (scan) and `maxTxsPerSigner` (per
  rent-payer reclaim), and the storage scan sorts by channel ID before resuming from
  its cursor.
- RPC client injection (`RPC` / `rpc`) on the facilitator config and rent cleanup
  config, matching #3183.

## Design decisions

**Commitment policy per read class, not one global level.** Reading everything at
`confirmed` risks a blockhash dropped by a fork; reading everything at `finalized`
reports live channels as missing. Account state and simulation read at `confirmed`,
slot and blockhash at `finalized`, as named constants in both SDKs
(`go/mechanisms/svm/upto/shared.go:25`), asserted per read class by a Go test.

**Optional memo instead of an empty-string sentinel.** The facilitator binds
`extra.memo` by exact match, so "seller demands an empty memo" and "no memo, emit a
nonce" need to be distinguishable. Go threads `*string` end to end; an empty string
input is treated as unset rather than as an explicit empty memo.

**Identity mismatches checked before address well-formedness.** A fee-payer or
receiver-authorizer mismatch names the actual problem, so it's checked before a
generic "not a valid address". Both SDKs order checks this way, pinned by
table-driven tests.

**Hand-written program bindings.** TypeScript uses a Codama-generated client from the
Anchor IDL; Go has no equivalent codegen in the toolchain. Drift is guarded by golden
vectors for the channel PDA seed preimage and the voucher signing preimage.

**Fail-closed pre-broadcast, fail-open post-confirmation.** A deposit that broadcasts
without a durable index can never be found by cleanup, so that write blocks the
broadcast. A claim's bookkeeping update happens after the channel is already onchain,
so it's best-effort instead of blocking settlement.

**Discovery is a backstop on its own interval, not the primary path.** Offchain
storage is still how cleanup finds channels normally; the `getProgramAccounts` sweep
makes a storage outage recoverable from chain state instead of permanent data loss.
It runs on its own, much longer, interval than cleanup because it's an onchain scan
rather than an index read, and only ever adds records for a later cleanup pass to
reclaim — it never reclaims rent itself.

**No client-side ComputeBudget overrides.** `exact` and `upto` clients use fixed,
scheme-appropriate defaults rather than an override surface, since `0` cannot mean
the same thing ("use zero" vs. "omit the instruction") across both schemes, and
nothing in either SDK needed to configure it.

## Testing

* `go build ./...`, `go vet ./...`, `golangci-lint`, and `go test ./...` pass for the
  SVM `upto` packages (client, server, facilitator, paymentchannels)
* TypeScript unit and integration suites pass for `@x402/svm`
* Cross-SDK e2e (`x402-playground`) passes for `upto` and `exact` on Solana, across
  the go/typescript facilitator, server, and client combinations, confirming `upto`
  works without regressing `exact`

## Changelog

`go/.changes/unreleased/added-20260812-190000.yaml`,
`go/.changes/unreleased/added-20260813-000000.yaml`,
`typescript/.changeset/svm-upto-scheme.md`,
`typescript/.changeset/svm-upto-compute-budget.md`,
`typescript/.changeset/svm-upto-rent-cleanup-runners.md`. All unreleased on this
branch.


CarsonRoscoe (commented):
@notorious-d-e-v would appreciate a pass from you on these changes, given the impact it would have on rent cleanup strategy in Typescript/Go + mirroring your #3155 in Go

typescript/packages/mechanisms/svm/src/upto/facilitator/rentCleanupManager.ts
```diff
@@ -152,27 +194,83 @@ export class UptoSvmRentCleanupManager {
...
+    const intervalMs = config.intervalSecs * 1_000;
+    this.timer = setInterval(() => {
+      void this.tick();
+    }, intervalMs);
+  }
+
+  /**
+   * Stop the interval loop.
+   */
+  stop(): void {
```
phdargen: Go `Stop()` cancels `runCtx` and waits for the in-flight `Cleanup` (checks at scan / discovery / reclaim). TS `stop()` only clears the interval, please match with Go

typescript/packages/mechanisms/svm/src/signer.ts
```diff
@@ -26,6 +26,21 @@ export type ClientSvmConfig = {
...
+   */
+  computeUnitLimit?: number;
+  /**
+   * `SetComputeUnitPrice` override in microlamports per compute unit, paid by
+   * whichever party is the transaction's fee payer. Defaults to the scheme's
+   * own default price; schemes that treat the ComputeBudget prefix as
+   * optional (e.g. `upto`'s channel open) treat `0` as "omit the
+   * instruction", letting a wallet inject its own.
+   */
+  computeUnitPriceMicroLamports?: number;
```
phdargen: Think this is a bit confusing, `0` means “omit” on upto and “use zero” on exact

Imo these client configs are not necessary and we can just use hardcoded defaults.

typescript/packages/mechanisms/svm/src/upto/shared.ts
```diff
@@ -1,9 +1,90 @@
...
+ * the facilitator will demand instead of a random nonce. Any other type is
+ * treated as absent rather than coerced, matching the Go SDK.
+ *
+ * @param extra - The `extra` map from the payment requirements
+ * @returns The memo when present, otherwise undefined
+ */
+export function resolveUptoSvmMemo(extra: PaymentRequirements["extra"]): string | undefined {
+  const memo = extra?.memo;
+  return typeof memo === "string" ? memo : undefined;
+}
```
phdargen, line 89: Facilitator should check whether the server set `requirements.extra.memo`:
- Non-empty string → client must emit that exact memo; facilitator requires a match.
- Missing / `""` / non-string → unset. Client still emits a nonce; facilitator does not check it. Do not fail the challenge.
Both roles already call this helper, so treating `""` as unset here is enough. Right now `""` is a required empty memo.

```suggestion
export function resolveUptoSvmMemo(extra: PaymentRequirements["extra"]): string | undefined {
  const memo = extra?.memo;
`typeof memo === "string" && memo !== "" ? memo : undefined`
}
```

Same for Go `ParseExtraMemo`

typescript/packages/mechanisms/svm/src/upto/facilitator/rentCleanupManager.ts
```diff
@@ -152,27 +194,83 @@ export class UptoSvmRentCleanupManager {
...
+  /**
+   * Run one cleanup pass. Callers go through {@link cleanup}, which serializes
+   * passes.
+   *
+   * @param opts - Work caps and callbacks
+   */
+  private async runPass(opts: RentCleanupOptions): Promise<void> {
     const abandonGraceSecs = opts.abandonGraceSecs ?? DEFAULT_ABANDON_GRACE_SECS;
     const maxReclaimsPerTx = opts.maxReclaimsPerTx ?? DEFAULT_MAX_RECLAIMS_PER_TX;
     const maxTxsPerRun = opts.maxTxsPerRun ?? DEFAULT_MAX_TXS_PER_RUN;
```
phdargen: `maxTxsPerRun` currently means two different things: stop the storage scan (L263) and give every rent-payer a full reclaim budget (L525). 

Suggest to split into `maxTxsPerRun` and `maxTxsPerSigner`

typescript/packages/mechanisms/svm/src/upto/facilitator/rentCleanupManager.ts
```diff
@@ -38,6 +40,22 @@ interface ReclaimCandidate {
...
+/**
+ * Reorders records to start right after the channel id a prior pass stopped
+ * at, wrapping around. An unknown or empty cursor (a closed channel, or the
+ * first pass) scans from the beginning.
+ *
+ * @param records - Stored channel records in storage order
+ * @param cursor - Channel id to resume from, or "" to scan from the start
+ * @returns Records rotated so `cursor` (if present) comes first
+ */
+function rotateFromCursor(records: UptoChannelRecord[], cursor: string): UptoChannelRecord[] {
```
phdargen: If the pass hits the budget mid-scan it saves a cursor and resumes there next time. That only works if `list()` returns the same order every pass.

`UptoChannelStorage.list()` does not promise an order. In-memory happens to be insertion-stable; other store implementations may not be. Then some channels get visited twice and others never.

Suggest to sort by `channelId` before `rotateFromCursor` so resume is stable

typescript/packages/mechanisms/svm/src/upto/facilitator/rentCleanupManager.ts
```diff
@@ -103,6 +121,13 @@ export interface UptoSvmRentCleanupManagerConfig {
    * (`reclaimComputeUnitLimit`) and are mint-independent.
    */
   settleComputeUnitLimit?: number;
+  /**
+   * Add a spec §6 getProgramAccounts sweep, per managed signer key, for
+   * Distributed channels missing from storage. Discovery finds only
+   * chain-verifiable reclaim candidates; it never substitutes for the
+   * payTo/tokenProgram metadata Open/Sealed actions require.
+   */
+  enableDiscovery?: boolean;
```
phdargen: Onchain discovery is RPC heavy. I don't think it should run on every cleanup interval.

Suggest splitting the runners, each on its own schedule:
- `cleanup()`: only what is already in storage
- `discovery()`: find missing channels onchain and add them to storage

`cleanup` stays on the short interval. `discovery` can run rarely (e.g. daily)

phdargen (approved):
LGTM 🚀

PR #2939, opened 2026-07-23
feat(specs): add exact scheme implementation for Bitcoin Lightning Network

javierpmateos:
## Description

Refs #2938.

### Summary

Adds `specs/schemes/exact/scheme_exact_lightning.md`: the `exact` scheme implemented over the Bitcoin Lightning Network, with stateless preimage verification, recipient-signed binding of invoices to payment requirements via BOLT11 `description_hash`, and documented composition with the facilitator extension pipeline.

### What this PR contains

The per-network specification only, following `scheme_impl_template.md` and the conventions of existing documents (closest model: `scheme_exact_xrpl.md`). No SDK changes; no new packages. The reference implementation lives out-of-tree at https://github.com/javierpmateos/x402-lightning and is linked from the spec.

### Design notes for reviewers

- Trust model: the facilitator never holds or moves funds; the client pays a recipient-signed invoice directly — complying with the trust-minimization requirement structurally.
- Verification rules 1–7 are pure functions of the request; rule 8 (single-use) is the only stateful step, with the multi-facilitator replay edge case documented in Security Considerations.
- Settlement semantics differ from broadcast-based networks because payment is payer-initiated and precedes the proof; the spec defines `/settle` as verify-and-record with the payment hash as settlement identifier.
- The `description_hash` binding reuses cryptography BOLT11 already ships (recipient node signature), adding no new signature scheme to the protocol.
- Production prior art exists (Lightning informally carried under `exact` with invoice-in-payload and custodial verification); this document formalizes the pattern with portable preimage proofs.

### Checklist

- [x] Follows `scheme_impl_template.md` structure and per-network doc conventions
- [x] Terminology consistent with x402 v2 (CAIP-2 networks, `accepted` echo in PaymentPayload, PAYMENT-* headers)
- [x] Facilitator Verification Rules stated as MUST, mirroring existing per-network documents
- [x] Security considerations cover replay, substitution, hold-invoice griefing, amount confusion, fiat staleness, payer privacy
- [x] Reference implementation linked, 15 tests passing
- [x] No changes outside `specs/schemes/exact/`

PR #3145, opened 2026-08-13
Expand asset transfer methods with `upfront` payment flows, family spec: scheme_exact.md 

IkerAlus:
## Description

Adds a network-agnostic section to `specs/schemes/exact/scheme_exact.md` defining the two asset transfer method families usable with the `upfront` payment flow, and the requirements each MUST satisfy. Follow-up to #3053, which landed the flow machinery but deliberately left the network-agnostic properties of client-prepaid methods unspecified.

Two flows are defined: 

**`signed-transaction`**: client signs, facilitator submits during settle. Replay safety is inherited from the network, no facilitator-side deduplication needed.

**`payment-proof`**: client executes the payment, presents a proof, settle validates and binds it. Requires a consumed-proof store. 

The section also notes that mechanisms SHOULD prefer `signed-transaction` where network and payer tooling permit it.

## On the `payment-proof` name:

#3053's discussion settled on `signed-transaction` vs `transaction-proof` as two distinct methods. This PR keeps `transaction-proof` as a valid ATM value, but names the **family** `payment-proof`, mainly to accommodate Lightning case:

A Lightning proof is a payment preimage: There is no transaction, no transaction hash, and nothing to look up onchain. The proof is checked as `sha256(preimage) == payment_hash`, and the consumption identifier is the payment hash. Calling the family `transaction-proof` would make every shared requirement read as onchain-specific and would push Lightning back out of the structure. `payment-proof` also matches the vocabulary #2520 used for this family before it was folded into `paymentFlow`.

## Provenance

This PR contains most of other people's analysis and rationale:

- The two-family split, the replay-safety asymmetry, the stateless-vs-stateful distinction, and the CAIP-2-joined consumption identifier — @CarsonRoscoe on #3053.
- Atomic single-use claim before resource execution, per-implementation replay identifiers, proof that the payer is authorized to use the payment, and explicit finality criteria — @alftom on #3053 and #1004.
- Preimage-as-proof as the thing that actually uses the network's properties — @benthecarman on #2861.
- Proof lifecycle (claimed in-flight at settle, consumed only on a terminal state), payer-identity limits, amount acceptance semantics, failure disposition, client recovery, and screening timing, from the NEAR Intents client-prepaid work in #2102.
- The `signed-transaction` transfer-correctness, facilitator-safety and expiry bullets are generalized from the SVM / Stellar / Starknet rules already in this file.

## Why now

There are currently four Lightning drafts (#2861, #1311, #1873, and the closed #2939) all written as separate `exact` schemes, plus #1004, #1634 and #1844 in the same shape, and #2102 which is being reworked onto `upfront`. Settling those once here gives the consolidation @phdargen has asked for on #2861 a structure to target: one flow, one ATM discriminator per mechanism, shared requirements in this file.

#### AI assistance

Drafting and consolidation were done with AI assistance. All content reviewed by the author for technical correctness.


specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Facilitator safety: the facilitator's own address MUST NOT appear as a funding source or transfer authority in the signed transaction.
+- Replay protection: the mechanism MUST rely on the network's own replay primitive (account nonce, UTXO consumption, or equivalent) and MUST verify that primitive is unconsumed at settle time. Facilitators SHOULD NOT be required to maintain a consumed-payment store for this family.
+- Expiry: where the network supports transaction validity windows, the signed transaction's window MUST cover the advertised `maxTimeoutSeconds` at settle time.
+
+### `payment-proof`
+ 
+The client executes the payment itself and presents a proof of it. Settle validates that proof and binds it to the request. It carries strictly weaker guarantees than `signed-transaction` and requires the state described below.
+ 
+Proofs are specific to each mechanism: an onchain transaction reference, or a cryptographic settlement secret such as a payment preimage. A mechanism MAY name its method `transaction-proof` where the proof is an onchain transaction reference, but MUST NOT assume that form in shared requirements.
+
```
alftom: Do we need a separate `transaction-proof` method if `payment-proof` handles it?

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Expiry: where the network supports transaction validity windows, the signed transaction's window MUST cover the advertised `maxTimeoutSeconds` at settle time.
+
+### `payment-proof`
+ 
+The client executes the payment itself and presents a proof of it. Settle validates that proof and binds it to the request. It carries strictly weaker guarantees than `signed-transaction` and requires the state described below.
+ 
+Proofs are specific to each mechanism: an onchain transaction reference, or a cryptographic settlement secret such as a payment preimage. A mechanism MAY name its method `transaction-proof` where the proof is an onchain transaction reference, but MUST NOT assume that form in shared requirements.
+ 
+- Proof identity: the payment payload MUST carry the mechanism's proof artifact. Mechanisms MUST define a canonical consumption identifier for the payment, formed as a CAIP-2 network identifier joined to the mechanism's canonical payment identifier (an onchain transaction hash, a payment hash, or equivalent). Mechanisms MAY define a narrower identifier where a single payment can satisfy more than one requirement set.
```
alftom: Do we want to specify just one proof mechanism or allow a set if the network supports it?

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Proof validation: mechanisms MUST state whether the proof is self-verifying, i.e., cryptographically checkable against the payment requirements without querying a ledger or node, as when a preimage is checked against a payment hash, or whether validation requires observing ledger or backend state. Self-verifying proofs SHOULD be preferred, as they remove a liveness and trust dependency from settle.
```
alftom: Should this be stated explicitly or would each proof have language that describes how to verify the proof which contains the information?

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Requirement binding: presenting a proof alone does not establish that the payment was made for this request. Mechanisms MUST declare which binding they rely on, from: (a) a payment instrument unique to this request — a per-payment recipient address, or an invoice issued for this request — so that a payment against that instrument is bound to the requirements that advertised it; (b) a server-issued nonce carried in the payment itself, as a memo, destination tag, `OP_RETURN` output, or calldata suffix; (c) a signature by the payer over the payment requirements, carried in the payment payload; or (d) a commitment to the payment requirements signed by the payee and embedded in the payment instrument the client must satisfy, such as a BOLT11 `description_hash` over the canonicalized requirements. Mechanisms MUST NOT depend on an optional protocol extension to obtain this binding, and MUST NOT rely on transport integrity alone.
+- Payer identity: mechanisms relying on a declared payer address MUST state that a declared, or even ownership-proven, address does not bind the actual sender of the payment. The sender is observable only once the payment exists.
```
alftom: Is there a reason why we have this requirement here instead of just allowing each mechanism to state its own method(s) for confirming the payment proof message is authorized to use the payment?

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Amount acceptance: mechanisms MUST state whether acceptance requires exact equality with `requirements.amount` or a threshold, and where a threshold is used MUST state the disposition of any excess.
```
alftom: @phdargen is there another scheme that allows overpayment?

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Finality: mechanisms MUST declare which observable event constitutes finality for settle, and which party owns confirmation-depth policy on networks with probabilistic finality. Where settlement completes in more than one stage, the mechanism MUST state which stage is the finality boundary.
```
alftom: This should be combined with the requirement above about "terminal state".

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,47 @@ amount of funds they need to be transferred.
...
+- Failure disposition: mechanisms MUST state, for each terminal non-success state, whether the payment is returned to the payer and by what path. Refunds are out of protocol unless the mechanism provides one; clients MUST NOT assume a refund path exists.
```
alftom: Can't the facilitator just return an error and say payment has not reached terminal state yet, and the client can send the payment proof?  It should probably state its terminal state requirements in the error so the client knows when to resend.  As long as the facilitator doesn't consumer the payment and the resource isn't provided nothing needs to be returned (unless the client decides it doesn't need the service anymore but that might be a different matter)

alftom:
@IkerAlus is your intention to create a new `family` field that can have a value of `payment-proof` or is this PR just listing requirements for a family of values for `assetTransferMethod` (e.g.- `transaction-proof` could be a value of `assetTransferMethod`).

specs/schemes/exact/scheme_exact.md
IkerAlus: if we are all happy with `payment-proof`, I can simply delete this sentence, and each category will decide.

specs/schemes/exact/scheme_exact.md
IkerAlus: I lean towards allowing a set, I will add a small edit here.

specs/schemes/exact/scheme_exact.md
IkerAlus: ummm I was thinking, perhaps it is better to cleanly reframe this by dropping "threshold" as an option and instead require mechanisms to state how over- and under-payment are handled, thoughts?

specs/schemes/exact/scheme_exact.md
IkerAlus: good catch. The draft conflates not-yet-final with terminal failure. Not-yet-final should consume nothing and permit retry with the same proof.  I will edit with your suggestion

specs/schemes/exact/scheme_exact.md
IkerAlus: good point, I added this mostly from to the discussion on Slack  with @phdargen but I agree with you that we should drop it from this spec.

specs/schemes/exact/scheme_exact.md
IkerAlus: This one can be softened to "the verification description MUST make clear whether…" rather than a standalone declaration. But I'd keep the statement, because self-verifying versus observation-dependent determines whether settle has a liveness and trust dependency on a node or indexer.

IkerAlus:
hi @alftom  thanks for the good review, it genuinely improved it.

Regarding the main comment: it is the second option: These are descriptive categories for this spec, not `assetTransferMethod` values, and nothing here changes the wire format.

Why: a generic client already has what it needs on the wire. `extra.paymentFlow` tells it settlement happens before the handler, and `extra.assetTransferMethod` tells it which concrete method to construct. The family distinction only matters to spec authors and facilitator implementers.

I'll make this explicit in my new commit:

- A scope paragraph before the table stating the families are categories, not `assetTransferMethod` values, that no new field is introduced, and that ATM values stay mechanism-defined per section 6.1. A mechanism MUST state which family its method belongs to and satisfy that family's requirements.
- Rename the headings and table column to unbackticked prose — "Signed transaction (facilitator-submitted)" and "Payment proof (client-executed)", so they stop reading as identifiers.
- Drop the "MAY name its method `transaction-proof`" sentence, which is what created the ambiguity. That also covers your other comment.
- Fix the remaining in-text references that treat the names as values.

specs/schemes/exact/scheme_exact.md
alftom: Yes, I think leaving "threshold" out makes sense.  We probably just need to lean on the `exact` scheme for this.  Whatever it supports is whatever this supports.

specs/schemes/exact/scheme_exact.md
alftom: So `payment-proof` will be a value of `assetTransferMethod`?  If so I'm OK with that.

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,52 @@ amount of funds they need to be transferred.
...
+- Finality: mechanisms MUST declare the condition under which a payment is final, i.e, which observable event constitutes finality for settle, which party owns confirmation-depth policy on networks with probabilistic finality, and where settlement completes in more than one stage, which stage is the finality boundary. That single declaration determines the payment's state at settle time: *not yet final*, where the condition is unmet but may still be met; *final-success*, where it has been met; and *final-failure*, where it can no longer be met. The latter two are terminal states, and the requirements below are expressed in those terms.
+   - Not yet final: where the payment is not yet final, settle MUST NOT consume the proof and MUST NOT deliver the resource. It MUST return an error identifying the unmet condition, so that the client can retry with the same proof; mechanisms SHOULD make that error specific enough for a client to judge when retrying is worthwhile.
+   - Single-use claim: a proof MUST be atomically claimed before the resource executes. Where two requests present the same proof concurrently, exactly one MUST succeed in claiming it, the other MUST be rejected.
```
alftom: I think this is a duplicate?

specs/schemes/exact/scheme_exact.md
```diff
@@ -15,6 +15,52 @@ amount of funds they need to be transferred.
...
+- Amount acceptance: mechanisms MUST state whether acceptance requires exact equality with `requirements.amount` or a threshold, and where a threshold is used MUST state the disposition of any excess.
+- Finality: mechanisms MUST declare the condition under which a payment is final, i.e, which observable event constitutes finality for settle, which party owns confirmation-depth policy on networks with probabilistic finality, and where settlement completes in more than one stage, which stage is the finality boundary. That single declaration determines the payment's state at settle time: *not yet final*, where the condition is unmet but may still be met; *final-success*, where it has been met; and *final-failure*, where it can no longer be met. The latter two are terminal states, and the requirements below are expressed in those terms.
+   - Not yet final: where the payment is not yet final, settle MUST NOT consume the proof and MUST NOT deliver the resource. It MUST return an error identifying the unmet condition, so that the client can retry with the same proof; mechanisms SHOULD make that error specific enough for a client to judge when retrying is worthwhile.
```
alftom: Where the payment is not yet final, settle MUST NOT permanently consume the proof and MUST clear its in-flight claim before returning the error, so that the client can retry with the same proof.

specs/schemes/exact/scheme_exact.md
alftom: Maybe @phdargen can chime in given your Slack conversation.

specs/schemes/exact/scheme_exact.md
IkerAlus: good improvement, I reworded the bulletpoint to contain this

specs/schemes/exact/scheme_exact.md
IkerAlus: I removed the threshold idea and reworded the point

specs/schemes/exact/scheme_exact.md
phdargen: yeah agreed, no need to discuss payer identity here. This can remain a per-mechanism implementation detail

phdargen:
Thanks a lot for putting this together @IkerAlus @alftom. The two-family split is the right idea and agreed on `payment-proof` naming.

## Main issue: the families are nested under `upfront`

The split is by **who submits the payment**; the flow is **when settle runs**. Those are independent axes.

First, it's ambiguous which methods `signed-transaction` covers. If it means a raw tx the facilitator relays, then EIP-3009, Permit2, ERC-7710, SNIP-9 and Stellar auth entries land in neither family. The PR generalizes its bullets from the Stellar and Starknet rules, so I suppose they are meant to be in. The heading's `(facilitator-submitted)` and the table column "who submits the payment" suggest that's the intended line, and the name is just too narrow for it.

Under that reading the family is every `exact` method that exists, and all of them already support both flows. Scoping these requirements to `upfront` then gives one method two different rule sets depending on a flag, when only the ordering changes.

Only `payment-proof` is genuinely flow-bound: the payment is already final when the payload arrives, so `upfront` is the only ordering that makes sense.

## Other issues

### `signed-transaction`

- **The name doesn't fit most of the family.** EIP-3009, Permit2, ERC-7710 and SNIP-9 are signed *authorizations*; the facilitator wraps them in a transaction of its own. Only SVM, TON, XRPL and Concordium hand over a raw tx. What all of them share is that the facilitator submits the payment.

- **"Facilitator state: None required" isn't correct.** Three bindings in this family already need a dedup store: TON (BoC hash dedup, already listed in this file), SVM (RECOMMENDED), XRPL (REQUIRED, shared across every process serving `/settle`). The network primitive prevents double-*spend*, not double-*delivery* — resubmitting an already-landed payment returns success to every caller, so one payment buys N resources. The distinction worth keeping is store *lifetime*: short-lived here, bounded by the payment's validity window; durable for `payment-proof`.

- **"MUST NOT contain additional value-moving operations" excludes UTXO chains.** State the required outcome and let bindings constrain structure as sponsor policy.

### `payment-proof`

- **Request binding needs its reason.** Unbound, a proof is equally valid at any resource server sharing `payTo` whose amount it satisfies, which is common when one treasury address sits behind several services. Binding is also what keeps dedup inside one facilitator: a bound proof can only be replayed against the request it was bound to, whereas an unbound one would need a store shared by every facilitator serving that address.

- **Missing: a retention bound.** A transaction hash or preimage stays presentable forever. Unless the instrument expires or the mechanism declares a maximum proof age, the durable store grows without limit.

## Alternative direction

Lift the families out of `upfront` into their own section:

- **`signed-transaction` → `facilitator-submitted`**: client signs, facilitator submits at settle. Declares fee payer, replay primitive, validity window, duplicate submission; satisfies transfer correctness, facilitator safety where sponsored, network-primitive replay, dedup where resubmission is indistinguishable. May use `authorization` or `upfront`, mechanism declares the default 
- **`payment-proof` → `client-submitted (payment proof)`**: client submits, presents a proof, settle validates and binds it. `upfront` only. Requirements: instrument and proof placement, request binding, atomic single-use claim, retention bound, finality condition

phdargen:
Landed on this with my AI, would this work for you @IkerAlus @alftom ?

> ## Payment Flow
> 
> By default, `exact` uses the `authorization` payment flow (verify → resource → settle): the payment is verified before the resource executes and settled afterward.
> 
> `exact` MAY also use the `upfront` payment flow (settle → resource → respond) when the resource needs onchain finality before execution. 
> 
> `authorization` SHOULD be preferred wherever the asset transfer method permits it, and clients SHOULD choose it when a resource offers both, because the payment commits only after the resource handler succeeds, so a failed handler leaves the client uncharged. Under `upfront` the payment commits first, so a handler failure leaves the client charged with nothing delivered; this specification defines no refund, and any remedy is the resource server's own arrangement. A method whose validity window or replay primitive bounds how long a handler may take SHOULD offer `upfront` for handlers that can exceed that bound.
> 
> ## Asset Transfer Method Families
> 
> Asset transfer methods fall into two families, separated by which party submits the value-moving operation. These are descriptive categories for this specification: they are not `assetTransferMethod` values and introduce no new wire field. A mechanism MUST state which family each of its methods belongs to and MUST satisfy that family's requirements, and MAY define more than one method in the same family.
> 
> ### Facilitator-submitted
> 
> The client signs a payment but does not submit it; the facilitator submits it during settle. The signed object MAY be an authorization the facilitator wraps in its own transaction, or a network-level transaction the facilitator only relays. Submission does not imply sponsorship: the facilitator MAY pay the network fee, or MAY only relay a payment the payer funds itself. A method in this family MAY use `authorization` or `upfront`, and MAY support both; the mechanism declares which it supports and which is the default.
> 
> Each method MUST declare:
> 
> - **Fee payer**: sponsored by the facilitator, or self-funded by the payer.
> - **Replay primitive**: exclusive to this payment, or shared with the payer's account state. Where shared, state the resulting limit on concurrent pending payments per payer, and that unrelated payer activity can invalidate the payment after the resource handler has already run.
> - **Validity window**: bounded, and by what; or unbounded, in which case state how the payer invalidates an unused signed payment and how long the facilitator may retain it.
> - **Duplicate submission**: whether resubmitting an already-submitted payload is distinguishable from the original at the network interface.
> 
> Each method MUST satisfy:
> 
> - **Transfer correctness**: settlement MUST produce exactly one identifiable transfer of `amount` of `asset` to `payTo`. Operations incidental to making that payment, such as fee payment or change outputs, do not disqualify it; a binding MAY constrain transaction structure further as sponsor policy.
> - **Facilitator safety**: where the facilitator sponsors the fee, no operation in the signed payment may debit the facilitator beyond that fee.
> - **Replay**: the network's own primitive is authoritative. A consumed primitive MUST produce a settlement failure, never a success.
> - **Duplicate delivery**: the network primitive prevents double-spend, not double-delivery. Where a resubmission is indistinguishable from the original, submission returns success to every caller and one payment yields several resources. Such methods MUST deduplicate settlements atomically across every process serving `/settle`, retaining each key until its payment can no longer land.
> 
> ### Client-submitted (payment proof)
> 
> The client submits the payment itself and presents a proof of it; settle validates that proof and binds it to the request. Methods in this family MUST use `upfront`: the payment is already final when the payload is presented.
> 
> Each method MUST satisfy:
> 
> - **Instrument and proof**: the payment instrument the client must satisfy MUST be advertised in `PaymentRequired` `accepts[].extra`, with its validity window, and the proof artifact MUST be carried in `PaymentPayload.payload`. Self-verifying proofs — those checkable against the requirements without querying a node or indexer — SHOULD be preferred, since they remove a liveness and trust dependency from settle.
> - **Request binding**: a proof establishes that a payment happened, not that it was made for this request. Unbound, a proof is equally valid at any resource server sharing `payTo` whose amount it satisfies. A method MUST bind the payment to the requirements by one of: an instrument unique to this request; a server-issued nonce carried in the payment; a payer signature over the requirements; or a payee commitment to the requirements embedded in the instrument. Binding also keeps deduplication within a single facilitator: a bound proof can only be replayed against the request it was bound to, so the settling facilitator needs no store beyond its own, whereas an unbound proof would have to be checked against a store shared by every facilitator serving that `payTo`.
> - **Single-use claim**: a proof MUST be claimed atomically before the resource executes; of two concurrent presentations, exactly one MUST succeed.
> - **Retention bound**: a consumed proof MUST be recorded for as long as it stays presentable. An expiring instrument bounds this; where a proof stays presentable indefinitely, the method MUST declare a maximum proof age, or retention is unbounded.
> - **Finality**: declare the observable event that makes a payment final and which party owns confirmation-depth policy. While that condition is unmet but still reachable, settle MUST NOT consume the proof or deliver the resource, and MUST return an error naming the unmet condition.
>

IkerAlus:
hi @phdargen thanks for the feedback. I believe the restructure is right, lifting the families out of `upfront` fixes this PR. Aslo with the better naming`facilitator-submitted` and other fixes/improvements in the first categories

Several things I'd restore, two of them should enable cross-chain methods:

**1. "MUST use `upfront`: the payment is already final when the payload is presented"** — suggest "already submitted" instead. For a cross-chain method the client's payment has landed on the origin chain but the payee has not been paid when the payload arrives; finality comes later. Your own finality bullet already covers this ("unmet but still reachable"), so the family sentence is stricter than the requirement under it.

**2. Instrument placement**: "MUST be advertised in `accepts[].extra`" forces duplication where the instrument is a per-payment recipient address, which belongs in `payTo`. Suggest: advertised in the payment requirements, in a standard field where one applies, otherwise in `extra`, with its validity window. Lightning invoices still land in `extra`.

**3. Multi-stage finality**: keep "where settlement completes in more than one stage, the mechanism MUST state which stage is the finality boundary."

**4. Asynchronous settlement in the family scope**: keep "or where the client's payment initiates asynchronous downstream settlement rather than being the settlement itself," so the family covers methods where the proof starts settlement instead of attesting to it.

**5. Amount acceptance**: @alftom and I settled on: acceptance follows the amount semantics of `exact`; because the client controls the amount it sends, mechanisms MUST state how a non-conforming payment, below or above, is handled, including whether and by what path it is returned.

**6. In-flight claim release** per @alftom: where the payment is not yet final, any in-flight claim held for the failed attempt MUST be released before returning, and an attempt terminating abnormally MUST NOT leave a proof claimed indefinitely. Without it the retry hits the single-use check and deadlocks.

Also worth keeping: per-mechanism failure disposition (whether the payment returns to the payer and by what path) alongside your general statement in Payment Flow, clients need the per-method answer. Also @CarsonRoscoe's CAIP-2-joined consumption identifier from #3053 should be mentioned.

Final small nit: Payment Flow says `upfront` is for resources needing "onchain finality" — Lightning isn't onchain, so "finality" alone is probably better.

I'll push a version with your structure and main changes plus these restorations so the review is cleaner.

phdargen:
Thanks for the prompt update @IkerAlus, agreed on all the refinements you made. Could you please resolve the merge conflict?

IkerAlus:
hi @phdargen conflicts solved, should be all good now for merge.

phdargen (approved):
Thanks @IkerAlus , LGTM 🚀

PR #3154, opened 2026-08-14
feat(py): add spend controls

### LINKED ISSUE #2963: [Bug]: dollar settlement overrides resolve inconsistently across SDKs

phdargen:
## Description

Port the TypeScript spend-controls and default-asset work to Python, and harden shared money parsing so high-precision prices and atomic spend caps are handled consistently across SDKs.

- spendControls on x402Client / spend_controls on x402Client: Python clients now mirror the TS safety model: a default $1 USD cap on recognized pegged assets (plus optional per-asset atomic caps and an asset allowlist) in accept selection, before user policies run; MCP from_config forwards or disables as appropriate.
```
client = x402Client.from_config(
    x402ClientConfig(
        schemes=[SchemeRegistration(network="eip155:8453", client=ExactEvmScheme(signer))],
        spend_controls={
            "max_amount_per_payment": "$5",  # USD cap on default assets; False to remove
            "allowed_assets": [
                # opt-in non-default with atomic cap
                {"network": "eip155:8453", "asset": "0xCustomToken", "max_amount_per_payment": "2000000"},
                # opt-in non-default uncapped
                {"network": "eip155:8453", "asset": "0xOtherToken"},
                # override USD cap for a default asset by ticker (or on-chain id)
                {"network": "eip155:8453", "asset": "PYUSD", "max_amount_per_payment": "500000"},
            ],
            # or: "allowed_assets": True  # allow any asset (USD cap still applies to defaults)
        },
        # or: spend_controls=False  # disable all spend controls (any asset, no caps)
    )
)
```
- Per-family default_assets.py: EVM, SVM, and TVM now declare DEFAULT_ASSETS, get_default_asset, and find_default_asset (replacing scattered constants and ad-hoc lookups), with client schemes exposing find_default_asset and servers using the same table for money parsing and decimals.
- Multi-stablecoin pricing: Default assets are a list per network (first entry resolves bare "$0.10"; suffixed prices like "1 PYUSD" resolve by symbol via core parse_money), so adding more USD-pegged tokens is a data-only change without per-family lookup drift.
- Unified money conversion (Python + TS): parse_money / parseMoney return decimal strings internally (no float rounding for 18-decimal assets); convert_to_token_amount / convertToTokenAmount pad/truncate toward zero to "0" instead of throwing on dust. Mechanism money parsers now take str | int | float / string | number (parse_price / parsePrice always passes a string).
-Spend-controls hardening (TS + Python): allowedAssets[].maxAmountPerPayment / allowed_assets[].max_amount_per_payment must be integer atomic amounts (e.g. "2000000", not "$1") — a config error if misconfigured; non-integer 402 amounts on that path are dropped rather than compared. $ settlement overrides throw when asset decimals are unknown instead of guessing 6.

Closes https://github.com/x402-foundation/x402/issues/2963 along with https://github.com/x402-foundation/x402/pull/3156

## Checklist


python/x402/client_base.py
```diff
@@ -420,6 +470,197 @@ def _select_requirements_v1(
...
+        filtered = kept
+        if not filtered:
+            if rejected_by_asset_cap and all(
+                (entry := find_asset_entry(requirement)) is not None
+                and entry.get("max_amount_per_payment") is not None
+                for requirement in before_amount_caps
+            ):
+                raise NoMatchingRequirementsError(
+                    "All payment requirements were rejected by spendControls.allowedAssets "
+                    "maxAmountPerPayment. Raise the per-asset cap, or omit maxAmountPerPayment "
```
CarsonRoscoe: In this `client_base.py` file, 4 error messages (lines 542–543, 649–650, 655–659) reference the TS camelCase field names (spendControls.allowedAssets, maxAmountPerPayment, spendControls.maxAmountPerPayment, allowedAssets[].maxAmountPerPayment) instead of the actual Python snake_case API surface. This is user-facing remediation text, a Python developer hitting this error would be told to set a field that doesn't exist in their language

CarsonRoscoe (commented):
The sync MCP wrapper (wrap_mcp_client_with_payment_from_config_sync) has no test coverage for spend-controls forwarding, unlike its async counterpart. Low risk, but worth adding the coverage to ensure no future parity regression

chopmob-cloud:
Built the branch and ran `parse_price` end to end on both the EVM and TVM exact servers. There is a small precision point on the money conversion worth a look.

`parse_money` and `parse_money_string` return a `float` (`float(cleaned)`), and both servers now source the settlement amount from that value: `parse_price` calls `parse_money(price)["amount"]` and passes it into `_default_money_conversion`. Any digits the `float` drops at parse time cannot be recovered later, since `Decimal(str(amount))` and `parse_amount` only see the already rounded value.

It stays masked for the 6 decimal default assets (Base USDC, TON USDT), where every price I tried matched exact `Decimal`. It becomes visible on an 18 decimal default asset. Using the branch's own EVM server against MegaUSD (`eip155:4326`, 18 decimals):

```
scheme.parse_price("8975.978289462729", "eip155:4326").amount
  ->  8975978289462730000000     exact Decimal is  8975978289462729000000
```

The threshold I saw was around 16 significant digits, and it depends on magnitude rather than a hard cutoff.

Two smaller notes on the same path. The EVM move to `int(Decimal(str(amount)) * Decimal(10) ** decimals)` is a good step since it drops the float multiply, but because `amount` still arrives as a `float` from `parse_money` it does not fully close the gap on 18 decimal assets. And the TVM `exact` server previously parsed the price string straight into `Decimal` (`Decimal(clean)`), so its default conversion used to be exact end to end; routing it through `parse_money` now brings it onto the same float sourced path.

To be clear on impact: this is not a signer versus verifier split. The payer signs an authorization for the server set `requirements.amount`, so both sides use the same value and signatures verify. The effect is a low order rounding of the settled amount on high precision prices with 18 decimal assets, not a settlement integrity issue. I am flagging it because it is `float` on the money path and the TVM side used to be exact.

One way to close it without changing the public API: keep `parse_money`'s `float` return as is, add an exact helper that returns the amount as a `Decimal` (plus the symbol), and use that exact value inside the two `_default_money_conversion` paths so the settlement amount never round trips through `float`. Changing `parse_money_string` itself to return `Decimal` would be more invasive, since `ParsedMoney.amount` is typed `float` and `MoneyParser` callbacks and external callers depend on it. A regression test on an 18 decimal asset with a 6 decimal control would pin it.

The spend control selection side looks correct: the cap to atomic conversion matched exact `Decimal` floor across a large range of caps, and the cap value is client supplied rather than server controlled.

chopmob-cloud:
Rebuilt the branch at `b2294e9` on a Linux box and re-ran the case from my earlier note. The money-conversion unification closes it: sourcing the amount as a decimal string end to end makes the 18-decimal path exact.

```
parse_price("8975.978289462729", "eip155:4326").amount
  ->  8975978289462729000000     (exact; previously 8975978289462730000000)
```

`convert_to_token_amount` scales by padding and truncating the decimal substring, so nothing round-trips through `float`, and it matches the TypeScript `convertToTokenAmount` sibling including "truncates toward zero, including to `0`". The core spend-control and money suites are green on the branch.

Spend-control selection looks sound overall: the default-asset allowlist fails closed, and the USD cap converts to atomic per the resolved asset decimals with a `<=` boundary. One small robustness asymmetry on the cap path, and only in the per-asset branch. When an `allowed_assets` entry sets an atomic `max_amount_per_payment`, the check is `amount_of(requirement) <= int(cap)`, and `amount_of` is `int(get_amount())` with no atomicity guard, whereas the default-asset USD path first guards the raw amount with the `_ATOMIC_AMOUNT` regex. So the same amount takes different routes:

```
allowed_assets + atomic cap, amount "1.5"  ->  ValueError out of _apply_spend_controls
default asset USD cap,       amount "1.5"  ->  clean rejection via the atomic guard
```

A conformant server always sends an integer atomic amount, so this needs a malformed amount to hit and it fails closed either way, no cap bypass. Mirroring the `_ATOMIC_AMOUNT` guard in the per-asset branch (or rejecting a non-integer amount there) would make the two cap paths behave the same.

Two smaller notes. A positive price smaller than one atomic unit of the resolved default asset now floors to `0` instead of raising, which the changelog documents and which matches TypeScript, so a configured non-zero price can settle as `0`, worth a docs line. And the earlier camelCase remediation text is gone, the error messages name the snake_case fields now.

CarsonRoscoe (approved):

PR #3155, opened 2026-08-14
feat(ts): add ComputeBudget instructions to svm upto transactions

notorious-d-e-v:
## Summary

None of the three SVM `upto` transaction types (client `open`, facilitator claim/cancel settle, rent-cleanup reclaim) included a `SetComputeUnitLimit` or `SetComputeUnitPrice` instruction. Per [SIMD-0170](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0170-set-default-compute-unit-limits.md), the runtime then derives the budget as 200,000 CU per non-builtin instruction (+3,000 per builtin), so every upto transaction reserved 400,000–403,000 CU while consuming ~600–50,000. Notably, by setting nothing, the reference client reserved exactly `OPEN_MAX_COMPUTE_UNIT_LIMIT` (400,000) — the spec ceiling `verifyOpenTransaction` enforces on a client-supplied limit — while using ~12% of it.

Why it matters, in priority order:

1. **Priority fee = `compute_unit_price` × requested limit**, while leader ordering is by fee-per-CU. Right-sizing the limit buys identical scheduling priority at a fraction of the fee.
2. **With no `SetComputeUnitPrice` these transactions have zero priority** and pay base fee only. The claim is time-critical — the resource has already been served and the voucher expires — so on a congested mainnet a dropped claim is a served-but-unpaid window until retry.
3. **Block and per-account cost accounting charge the reserved amount, not the consumed one.** The fee payer is writable on every transaction and capped at 12M CU per account per block (SIMD-0306 raises this to 40% of the block limit), so a 10–20x over-reservation cuts per-fee-payer payments-per-block by the same factor.

## Changes

All limits are **static** — no simulation on any path, so settlement latency and failure modes are unchanged apart from the added prefix instructions.

**Client open builder** (`payment-channels/open.ts`, `upto/client/scheme.ts`):
- `buildOpenPaymentChannelTransaction` now prefixes `SetComputeUnitLimit` (default `OPEN_DEFAULT_COMPUTE_UNIT_LIMIT = 90_000`, ~1.8x observed consumption) and `SetComputeUnitPrice` (default `DEFAULT_COMPUTE_UNIT_PRICE_MICROLAMPORTS = 1`, matching the exact-scheme client default).
- Both are overridable via new `BuildOpenArgs` fields / `UptoClientSvmConfig` (`computeUnitLimit`, `computeUnitPriceMicroLamports`); `0` omits the instruction (for wallets that inject their own prefix). Values above the spec ceilings (400,000 / 5,000,000) throw at build time.
- Ordering follows the spec: `SetComputeUnitLimit` precedes `SetComputeUnitPrice`, both before the `open`. The default-built transaction passes `verifyOpenTransaction` under the default operator caps (unit-tested, and demonstrated on devnet against an unmodified production facilitator — see below).

**Facilitator settlement** (`upto/facilitator/channel.ts`, `scheme.ts`, `rentCleanupManager.ts`):
- `submitSettle` — used by the claim, the zero-charge cancel, and cleanup close/distribute — prefixes `SetComputeUnitLimit` with `DEFAULT_SETTLE_COMPUTE_UNIT_LIMIT = 100_000` by default. Sizing rationale: a measured claim with a warm recipient ATA consumes ~21.6k CU; a distribute that must recreate a closed recipient ATA adds ~25k; 100k keeps >2x headroom over that worst case. Operators override via the new `settleComputeUnitLimit` on `UptoSvmFacilitatorConfig` / `UptoSvmRentCleanupManagerConfig`.
- Reclaim batches size per channel: `reclaimComputeUnitLimit(n) = 25_000 + 5_000 × n`, clamped to the 1.4M per-transaction max. A measured reclaim consumes ~320 CU per channel and is mint-independent (the escrow ATA is already closed by then — `reclaim` only closes the channel PDA), so this holds >15x margin without a config knob.
- A configurable `SetComputeUnitPrice` is attached (new `computeUnitPriceMicroLamports` on both configs, default 1 µlam/CU, `0` omits).

### Static limits vs. simulation-derived (the design tradeoff)

Both were considered for the facilitator side:

- **Simulation-derived** limits self-adjust to recipient count, token program, and account state — but they add an RPC round-trip and a new failure mode to the time-critical claim, plus a stale-blockhash window while the simulation runs, and they need fallback logic for when the simulation itself is unavailable.
- **Static limits** are predictable and keep the claim path's RPC surface identical to before this PR. The current upto shape — single-recipient distribution, standard SPL Token mints — makes the worst case bounded and known (the dominant variable cost is one ATA recreation, ~25k CU).

This PR uses static limits with generous margins and operator overrides. Going from the old 403k reservation to a fixed 100k captures most of the accounting benefit; the additional savings simulation could squeeze out (~30k CU on a claim) are not worth the extra RPC dependency at a 1 µlam/CU price. The documented assumption: standard SPL Token (or Token-2022 without execution extensions) behavior. Deployments settling compute-heavy Token-2022 extension mints (e.g. transfer hooks) or unusually large distributions must raise `settleComputeUnitLimit` / `computeUnitLimit` explicitly.

## Measured numbers (Solana devnet, real transactions)

Before (all three types, no ComputeBudget instructions — SIMD-0170 defaults):

| tx | reserved | consumed | utilization | signature |
|---|---|---|---|---|
| open | 400,000 | 45,840 | 11% | `2PMJKUYovbzEY5V9Uc725vZoM17eBVj9mpKXFXbXsif9` |
| claim | 403,000 | 20,062 | 5% | `b9dt9DXTzjwWPb7HYv3za9BsmfbCokufQXudAfy9T6fp9PQCVCbdDiZU6X3TySnDAax2KNDq6uFSXJaTf663Ffu` |
| reclaim (batch of 2) | 400,000 | 634 | 0.2% | `8uJfM6P1wybySNbUCC63zhpx848gpa34M8aofCwoasgmzQfdWTeTNUpmkRfaY7GUrstDBERQ6xrxhcBMjBuq3nE` |

After — E2E re-run with this branch's client SDK against an **unmodified production facilitator** (PayAI dev facilitator running the pre-PR SDK), full open → serve → claim cycle, settled 4,000 of a 10,000-atomic ceiling:

| tx | reserved | consumed | utilization | signature |
|---|---|---|---|---|
| open (new client) | **90,000** | 50,640 | 56% | `5b6Ne6Yz9kZ7Xxt1Su3EFU2VeGEAMGwSVKkgaRnYMBvTfoQusQoXV42az9TTE9DTJ4BE8S8oEkCQQEcY5PfmJhfg` |
| claim (old facilitator, control) | 403,000 | 21,562 | 5% | `w5dGsKJ1gVkZeFnYSB34PWEyL8qnsa6WTziihYKTds95PkYt4abMVvfQfv6ALuhDMR7DCRLYNC6utDBWLBFxdcC` |

The open landed with `SetComputeUnitLimit: 90000` + `SetComputeUnitPrice: 1` on-chain, fee 10,001 lamports (1 lamport priority for the whole transaction) — and the **pre-PR facilitator's `verifyOpenTransaction` accepted the new default prefix unchanged**, confirming backward/spec compatibility in production. The claim in that run still reserved 403,000 because the facilitator side runs the old SDK; with this PR's static limits the same claim reserves **100,000** (4x less) and the measured reclaim batch **35,000** (11.4x less) — both pinned by unit tests.

(Open consumption is ~4.8k higher than the pre-PR measurement mostly due to the added ComputeBudget instructions and run-to-run state variance; 90k keeps ~1.8x headroom for standard mints.)

## Testing

- `@x402/svm`: 365/365 unit tests pass (19 files), including new coverage: default prefix emission + ordering + verification under default caps, overrides and opt-out, build-time ceiling enforcement, static settle default with no `simulateTransaction` call, limit/price option plumbing and omission at `0`, `reclaimComputeUnitLimit` scaling + clamp, config pass-through on the claim path and reclaim batches.
- `pnpm turbo run test --filter=...@x402/svm` (dependents incl. `@x402/paywall`): all pass.
- Devnet E2E as above: `200`, `paymentStatus: settled`, payer charged exactly 4,000 atomic.

## Notes for operators

- Defaults assume standard SPL Token settlement (single-recipient distribution). For Token-2022 mints with execution extensions (transfer hooks etc.) or large distributions, raise `settleComputeUnitLimit` (facilitator/cleanup) and have clients raise `computeUnitLimit` on the open, up to the spec ceiling.
- Facilitators that configured `maxComputeUnits` below 90,000 will reject default-built client opens; clients can pass a lower `computeUnitLimit`. Operators with `maxPriorityFeeMicroLamports: 0` similarly need clients to pass `computeUnitPriceMicroLamports: 0`. Default-configured facilitators (caps 400,000 / 5,000,000) accept the new defaults.


notorious-d-e-v:
Update: switched the facilitator settlement sizing from simulation-derived limits to static ones (claim/cancel/distribute 100k, reclaim 25k + 5k per channel, operator-overridable via `settleComputeUnitLimit`). 

Static sizing optimizes for latency and saves an RPC call per settlement.

CarsonRoscoe (approved):
Ty for the quick improvements!

Overall this LGTM,

I see four non-blockings things:
1. Facilitator config has no upper-bound check. `assertLimit` only floors, unlike client side which enforces floor and ceiling (might skip this one though. Facilitators' in general are expected to take on more risk and do more due diligence. Client's are the class we need to protect)
2. `submitSettle` doesen't honor `computeUnitLimit: 0` as "omit". The open transaction builder treats 0 as "let the wallet decide", but `submitSettle` always emits `SetComputeUnitLimit` unconditionally. Not abusable as callers floor above zero, but it's a potential trap in the future depending on how the code evolves
3. Hard CU caps assume single-recipient/standard-SPL with nothing enforcing it. `ChannelSplit[]` is still typed as a list even though only one recipient is resolved today. A future multi-split resolver or a Token-2022 transfer-hook mint would silently blow the 100k/90k defaults with no compile. Not an issue today, just a future concern.
4. Duplicated ComputeBudget instruction assembly between `open.ts` and `channel.ts`. It already drifted (see #2), so worth extracting to a shared helper

Given Typescript SDK is not yet published and won't be until Monday/Tuesday, and I have a Go Parity PR up that includes Typescript hardening parity, I would prefer to merge this, rebase my PR, port these changes to Go, and then MAYBE apply the hardening above in my PR for both languages. As all of these are non-blocking/debatable, I don't think we have to jump at any of them, just give some thought

If any of these 4 you feel strongly should be included, comment here with a ping to me and I'll make sure my PR adds it. Otherwise, going to think further and use my discretion as to which warrant being added

typescript/packages/mechanisms/svm/src/upto/client/scheme.ts
```diff
@@ -9,6 +9,21 @@ import { type UptoSvmPayloadV2 } from "../../types";
...
+   */
+  computeUnitLimit?: number;
+  /**
+   * `SetComputeUnitPrice` override in microlamports per compute unit (paid by
+   * the facilitator fee payer). Defaults to
+   * `DEFAULT_COMPUTE_UNIT_PRICE_MICROLAMPORTS`; `0` omits the instruction.
+   */
+  computeUnitPriceMicroLamports?: number;
+};
+
```
phdargen, line 26: nit: Would prefer not to introduce `UptoClientSvmConfig = ClientSvmConfig & {computeUnitLimit, computeUnitPriceMicroLamports}` but either extend `ClientSvmConfig` and also use those configs in exact, where they are currently hardcoded:
https://github.com/x402-foundation/x402/blob/main/typescript/packages/mechanisms/svm/src/constants.ts#L34-L37
or also hardcode them, not sure if the client needs this override

PR #3156, opened 2026-08-14
feat(go): spend controls

phdargen:
## Description

Go sdk port of spend controls introduced in https://github.com/x402-foundation/x402/pull/3124

## Checklist

### DIFF
--- a/e2e/clients/go/client.go
+++ b/e2e/clients/go/client.go
@@ -75,7 +75,7 @@ func BuildPaymentClient() *PaymentClientContext {
 		log.Fatal("At least one of CLIENT_EVM_PRIVATE_KEY or CLIENT_SVM_PRIVATE_KEY is required")
 	}
 
-	x402Client := x402.Newx402Client()
+	x402Client := x402.Newx402Client().DisableSpendControls()
 	var batchedScheme *batchedclient.BatchSettlementEvmScheme
 
 	if evmPrivateKey != "" {
--- a/examples/go/clients/advanced/README.md
+++ b/examples/go/clients/advanced/README.md
@@ -38,6 +38,7 @@ Each example demonstrates a specific advanced pattern:
 | `error-recovery` | `go run . error-recovery` | Error classification and recovery |
 | `multi-network-priority` | `go run . multi-network-priority` | Network-specific signer registration |
 | `hooks` | `go run . hooks` | Payment lifecycle hooks |
+| `spend-controls` | `go run . spend-controls` | Asset allowlist and USD spend caps |
 
 ## Testing the Examples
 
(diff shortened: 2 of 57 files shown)


CarsonRoscoe (commented):
Looking good!

Three non-blocking comments,

1. A package-level ConvertToTokenAmount function vs. a differently-scoped method of the same name exist, which is confusing but not a bug. Both are exported, so might be confusing to consumers when to use which
2. The money-conversion pattern is repeated across three server scheme files rather than factored into one shared helper
3. add a doc comment on the MCP Options struct noting which constructor ignores SpendControls/DisableSpendControls.

EDIT: Regarding #1, I see this PR did not add it, it just exposes it more obviously now. Predates this PR. 

Devx wise the difference to consumers is they'd call `x402.ConvertToTokenAmount(...)` vs. `scheme.ConvertToTokenAmount(...)`.  Different qualified names, no ambiguity, no shadowing, nothing breaks. It's a naming/API-surface smell, not a functional bug: same name, same stated purpose ("convert decimal amount to token smallest unit"), different signature and behavior (explicit decimals vs. network-lookup), in two different exported public APIs. A consumer autocompleting on either x402. or an ExactEvmScheme value could plausibly reach for the wrong one and get a mismatched call signature.

However, given this predates the PR, and might even exist in other SDKs, lets ignore it for now and mark this as a smell to follow on in the future.

phdargen:
> Looking good!
> 
> Three non-blocking comments,
> 
> 1. A package-level ConvertToTokenAmount function vs. a differently-scoped method of the same name exist, which is confusing but not a bug. Both are exported, so might be confusing to consumers when to use which
> 2. The money-conversion pattern is repeated across three server scheme files rather than factored into one shared helper
> 3. add a doc comment on the MCP Options struct noting which constructor ignores SpendControls/DisableSpendControls.
> 
> EDIT: Regarding #1, I see this PR did not add it, it just exposes it more obviously now. Predates this PR.
> 
> Devx wise the difference to consumers is they'd call `x402.ConvertToTokenAmount(...)` vs. `scheme.ConvertToTokenAmount(...)`. Different qualified names, no ambiguity, no shadowing, nothing breaks. It's a naming/API-surface smell, not a functional bug: same name, same stated purpose ("convert decimal amount to token smallest unit"), different signature and behavior (explicit decimals vs. network-lookup), in two different exported public APIs. A consumer autocompleting on either x402. or an ExactEvmScheme value could plausibly reach for the wrong one and get a mismatched call signature.
> 
> However, given this predates the PR, and might even exist in other SDKs, lets ignore it for now and mark this as a smell to follow on in the future.

Fixed 2) and 3). Regarding 1), removed duplicate scheme level ConvertToTokenAmount along with some similar dead code. TS/PY dont have this

chopmob-cloud:
Built this branch and ran `evm.ConvertDefaultMoney` end to end. This is the Go side of the same money precision point as the Python spend controls PR: the settlement amount is sourced from a `float64`.

`ParseMoneyString` returns a `float64` via `strconv.ParseFloat(cleaned, 64)`, `ParseMoney` returns that, and the exact server conversions take it directly: `defaultMoneyConversion(amount float64, ...)` on the EVM exact, EVM upto and SVM exact schemes, which calls `ConvertDefaultMoney` -> `ConvertToTokenAmount(NumberToDecimalString(amount), decimals)`. `NumberToDecimalString` is `strconv.FormatFloat(n, 'f', -1, 64)`, so it faithfully prints whatever the `float64` already rounded to, and `ConvertToTokenAmount` cannot recover digits lost before it.

It stays masked for the 6 decimal default assets (USDC, USDT0), where every price I tried matched exact. It shows up on an 18 decimal default asset. Running the branch's own EVM conversion against MegaUSD (`eip155:4326`, 18 decimals):

```
x402.ParseMoney("8975.978289462729")  ->  8975.97828946273   (the ...729 tail is already gone)
evm.ConvertDefaultMoney(that, "eip155:4326", "")  ->  8975978289462730000000
exact is                                              8975978289462729000000
```

The threshold I saw was around 16 significant digits, depending on magnitude rather than a hard cutoff. Since Go has 18 decimal defaults in the table now (MegaUSD, Mezo mUSD), this is reachable rather than theoretical.

Impact is the same low order note I left on the Python PR: it is not a signer versus verifier split, since the payer signs the server set `requirements.amount` and both sides use the one value, so signatures still verify. The effect is a small rounding of the settled amount on high precision prices with 18 decimal assets. I am flagging it because it is `float64` on the money path.

The tidy fix is to keep the money amount out of `float64` on the settlement path, for example by parsing the price string into a `*big.Rat` (or a decimal type) and scaling that to atomic units in the default conversions, rather than round tripping through `ParseFloat` and `FormatFloat`. A regression test on an 18 decimal asset with a 6 decimal control would pin it.

The spend control selection side looks correct: `amountOf` and the caps use `big.Int` (`new(big.Int).SetString(..., 10)`), so that path stays exact, and the cap value is client supplied.

chopmob-cloud:
Re-reviewed at 47fa76ba. I rebuilt the branch on Go 1.24 and reran it end-to-end: `go test ./...` is green (32 packages) and the money path is now float64-free.

The precision point from the earlier note is resolved. `ParseMoney` returns the decimal as a string and the server conversions carry that string straight into `ConvertToTokenAmount`, so the 18-decimal case that used to round is now exact:

```
x402.ParseMoney("8975.978289462729")             -> "8975.978289462729"     (exact, no float64)
evm.ConvertDefaultMoney(amt, "eip155:4326", "")  -> 8975978289462729000000  (was ...730000000000)
```

With `applySpendControls` already on `big.Int`, the whole default-asset path is string and integer only now. (One parity note in passing: the TS reference still has `parseMoney` return a `number` and rebuilds the string via `numberToDecimalString(amount)` before `convertToTokenAmount`, so a high-precision string price still round-trips through a float there. The Go side is tighter; noting it only if cross-SDK parity matters to you.)

One question on a deliberate divergence I hit while testing. `ConvertToTokenAmount` truncates a non-zero sub-atomic amount to "0" and returns no error:

```
x402.ConvertToTokenAmount("0.0000001", 6)  -> "0", nil
```

The changelog documents this as intentional, but the TS reference throws `Amount ... is too small to represent` for the same input. On the server price path this means a misconfigured sub-atomic price (say `$0.0000001` on a 6-decimal asset) becomes a zero-amount requirement instead of a loud failure. Is truncate-to-zero deliberate for the Go SDK specifically, or would you prefer it match the TS throw so the two SDKs agree and a fat-fingered price fails closed?

Rest of the refactor read cleanly: the default-asset tables, the `GetAssetDecimals (int, bool)` change with the settlement override now erroring on unknown decimals rather than silently assuming 6, and the ticker selection all look right.

CarsonRoscoe (approved):

Issue #3091, opened 2026-08-08
Unscoped npm packages (x402-express, x402-hono, …) still install a v1-only server with no deprecation notice


gadaffihub:
## Problem

`npm i x402-express` today installs a server that **54% of the buyers in the Bazaar cannot pay**, with no warning at install time and no error at runtime.

The v2 middleware lives under the `@x402/*` scope. The original unscoped packages are still on npm, still resolve, still install clean — and are frozen on the v1 line.

| package | `latest` | last publish | deprecated? | downloads/wk |
|---|---|---|---|---:|
| `x402-express` | 1.2.0 | 2026-04-16 | **no** | 5,880 |
| `x402-hono` | 1.2.0 | 2026-04-16 | **no** | 1,937 |
| `x402-next` | 1.2.0 | 2026-04-16 | **no** | 2,036 |
| `x402-fetch` | 1.2.0 | 2026-04-16 | **no** | 22,723 |
| `x402-axios` | 1.2.1 | 2026-05-29 | **no** | 4,846 |
| `x402` | 1.2.0 | 2026-04-16 | **no** | 347,288 |
| `@x402/express` | 2.21.0 | 2026-08-04 | — | 25,513 |

(npm point-in-time, 2026-08-08. Raw download counts include CI and mirrors, so read them as an order of magnitude — but the shape is clear, and unscoped `x402` is still pulled *more* than `@x402/core`.)

## Why it fails silently

Verified by reading the published tarballs rather than inferring:

- `x402-express@1.2.0` — the string `payment-required` does not appear anywhere in the package. It emits `res.status(402).json({...})` with `x402Version = 1`.
- `@x402/express@2.21.0` — references `PAYMENT-REQUIRED` in its dist.

Per `specs/transports-v2/http.md`, a v2 client reads terms from the `PAYMENT-REQUIRED` header and treats the body as a server concern. So a v1 middleware and a v2 buyer produce **no error on either side** — the buyer finds no terms and leaves. The seller sees a 402 in their logs and concludes everything is working.

## It shows up in the live data

Sweep of every distinct host in the CDP Bazaar, 2026-08-08 (1,509 hosts, one live request each, GET with POST fallback):

| what the 402 actually serves | hosts |
|---|---:|
| v2 only (`PAYMENT-REQUIRED` header) | 740 |
| both dialects | 526 |
| **v1 only (terms in body)** | **95** |
| no 402 at the route tried | 126 |

[...]

PR #3162, opened 2026-08-14
feat(legacy): add deprecation warning to unscoped v1 packages (#3091)

wakqasahmed:
Addresses #3091

## Description

The unscoped v1 packages (`x402`, `x402-express`, `x402-hono`, `x402-next`, `x402-fetch`, `x402-axios`) still install cleanly with no indication that they are frozen on the x402 protocol v1 line, while `@x402/*` carries v2 forward. As the issue notes, this causes silent interop failures: v1 servers put payment terms only in the JSON response body, while v2 clients read the `PAYMENT-REQUIRED` header, so neither side errors and the buyer just leaves.

This PR adds a one-time `console.warn` fired at middleware/client construction time (module load for `x402` core, since it has no single construction entry point):

- `x402-express` → `@x402/express`
- `x402-hono` (main entry and `session-token` entry) → `@x402/hono`
- `x402-next` (`paymentMiddleware` and `withX402`) → `@x402/next`
- `x402-fetch` → `@x402/fetch`
- `x402-axios` → `@x402/axios`
- `x402` (core) → `@x402/core`, warns on import

`x402`'s `package.json` had no `"."` export entry, so with `exports` present, `require("x402")`/`import "x402"` never resolved to the module that logs the warning — the fix now includes a `"."` export pointing at the built entry, so the warning is actually reachable from real consumers, not just internal subpath imports.

The five wrapper packages' near-identical warning logic is now a single `warnLegacyDeprecation(pkg, replacement, kind)` helper in `x402/shared`, called with each package's own name/replacement/kind instead of reimplementing the block per package. The warning is once-per-process per package name, and can be suppressed with `X402_SUPPRESS_LEGACY_WARNING=1` for hot-start environments (e.g. serverless) where a console warning per cold start is disruptive.

**Note on `npm deprecate`:** the issue's suggested fix (`npm deprecate <pkg>@"<2" "..."`) surfaces on `npm install`/`npm outdated`, which is more visible than a runtime warning, but requires npm registry publish rights I don't have as an outside contributor. This PR is the code-level alternative I can ship. If a maintainer with registry access can also run the `npm deprecate` commands, that closes the remaining gap.

This does not touch the "already-deployed endpoint" gap from the issue's second comment (adding the `PAYMENT-REQUIRED` header to existing v1 servers) — that's a separate, larger change to the legacy middleware response-building logic.

## Tests

- `x402/src/shared/deprecation.test.ts`: table-driven coverage of the shared helper's three message kinds (core/client/server), once-per-package dedup, and the `X402_SUPPRESS_LEGACY_WARNING` opt-out.
- `x402/src/deprecation.test.ts`: end-to-end check that importing `x402` still warns.
- Each wrapper package keeps a slim `deprecation.test.ts` asserting its own construction entry points call the warning with the right package/replacement/kind.
- `x402-hono/src/deprecation.test.ts` also covers the `session-token` entry point, which is a real standalone import.

Ran from `/typescript`:
```
pnpm install --filter "./packages/legacy/*" --filter x402
cd packages/legacy/x402 && npx tsup   # build core so other legacy packages resolve x402/* subpaths
cd packages/legacy/<pkg> && npx vitest run
```

All pre-existing tests plus the new/updated deprecation tests pass:
- `x402`: 20 test files, 275 tests
- `x402-express`: 3 test files, 31 tests
- `x402-hono`: 3 test files, 44 tests
- `x402-next`: 4 test files, 58 tests
- `x402-fetch`: 2 test files, 7 tests
- `x402-axios`: 2 test files, 11 tests

Also confirmed directly that `require("x402")` and `import("x402")` now trigger the warning, which they did not before this PR.

Ran `eslint --fix` (prettier + jsdoc rules) on all touched files — clean.

## Checklist

- [ ] My commits are signed — not currently signing commits in this environment; happy to re-sign/rebase if required for merge
- [x] No changelog fragment: these six legacy packages are unscoped and outside the Changesets `fixed` group in `typescript/.changeset/config.json`, so a changeset isn't applicable here

### DIFF
--- a/typescript/packages/legacy/x402-axios/src/deprecation.test.ts
+++ b/typescript/packages/legacy/x402-axios/src/deprecation.test.ts
@@ -1,0 +1,35 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("x402/client", () => ({
+  createPaymentHeader: vi.fn(),
+  selectPaymentRequirements: vi.fn(),
+}));
+
+describe("x402-axios legacy deprecation warning", () => {
+  let warnSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    vi.resetModules();
+    warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    warnSpy.mockRestore();
+  });
+
+  it("warns once on withPaymentInterceptor construction pointing to @x402/axios", async () => {
+    const { withPaymentInterceptor } = await import("./index");
+    const mockAxiosClient = { interceptors: { response: { use: vi.fn() } } } as never;
+    const mockWalletClient = { signMessage: vi.fn() } as never;
+
+    withPaymentInterceptor(mockAxiosClient, mockWalletClient);
+
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy.mock.calls[0][0]).toContain("x402-axios");
+    expect(warnSpy.mock.calls[0][0]).toContain("@x402/axios");
+
+    withPaymentInterceptor(mockAxiosClient, mockWalletClient);
+
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+  }, 20000);
+});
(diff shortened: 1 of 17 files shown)


wakqasahmed:
Closing and reopening this from the same branch — fixed a real bug per review (the `x402` package's `exports` map had no `"."" entry, so the deprecation warning was unreachable dead code for real consumers), deduplicated the near-identical per-package warning logic into a shared helper, added an opt-out env var, and covered the `session-token` entry point that was missed. Also fixing that this went out as part of a same-day batch of PRs from my account.

PR #3163, opened 2026-08-14
docs: name the cold-probe failure mode (#3113)

### LINKED ISSUE #3113: Docs: name the cold-probe failure mode (non-402 before the payment gate), and report probe state rather than a liveness boolean
The seller docs now document the cold probe itself: the get-discovered page states a resource must be reachable and return `402 Payment Required` to an unpaid request, and `POST /platform/v2/x402/validate` lets a seller test exactly that before going live. Credit where due, that closes the biggest historical gap. Two things around the cold path are still undocumented, and they keep producing silent invisibility that operators cannot diagnose from the docs. Related field reports: #3045, #2993; related tooling proposal: #3104.

**1. The failure mode has no name in the docs.**

The most common way to fail the cold probe is ordering: schema checks, auth middleware, or method routing answer `400`, `401`, `405` or `500` before the payment gate ever runs. An operator who builds and tests only against the paid path passes every real purchase and never sees it, because revenue looks healthy while the listing quietly degrades. The get-discovered page says *what* the cold probe expects; it does not name *how* real stacks fail it.

A named before/after case: the operator of the x402 API services at `api.reloadpi.com` (top-up, voucher and eSIM catalogs) ran exactly this failure in production. Body validation ran ahead of the payment gate, so real purchases succeeded while the bare probe got a non-402 and the service was silently deindexed from search. Reordering the middleware so the bare probe gets `402` restored the listing. The operator will follow up on this issue with the concrete deploys and timeline, so readers can see the full sequence.

**2. Scale: the same asymmetry, measured from outside the Bazaar.**

We run an independent scanner of x402 endpoints with its own on-chain settlement ledger on Base. As of Aug 8, before we fixed our own probe, roughly 3.6k of ~79.5k crawled URLs answered a bare GET with `401` or `405` and carried no captured price at all. Two operators from that set which we hand-verified both turned out to be healthy, paying POST-only x402 services: one answers GET with `401` ("Missing Authorization header") while POST returns a complete 402 challenge; the other returns `405` on GET, consistent with its own manifest, which declares POST. Our own scanner had the mirror-image blind spot: it probed GET-only and recorded empty price fields for those resources, so we are not filing this from a position of innocence. After deploying a POST-fallback (probe GET, on `401`/`405` retry as a bare POST), as of Aug 10: `401` entries went from 392 to 77, `405` from 3,200 to 192, and priced entries rose from ~25.5k to ~32.9k. The scanner now records, per entry, which method produced each challenge; 5,942 of the current price captures came specifically from the POST retry.

This is not only our problem. The method census posted in #3045 found that of ~14.5k indexed Bazaar resources, `extensions.bazaar.info.input.method` declares POST for ~5.1k, roughly a third of the catalog. Every third-party catalog or monitor that probes cold with a bare GET is misreading that third. A second independent count makes the same point from the discovery side: an operator on the CDP side re-counted the discovery list independently of us, as of Aug 3: 1,519 distinct domains, of which 539 published reachable discovery metadata to a bare GET on `/.well-known/x402.json`. Method caveat, stated deliberately: that count carries exactly the probe-shape bug this issue is about, so the 980 recorded as unreachable are an upper bound, and an unknown share of them are healthy POST-only operators counted as dead.

**What would help at the docs level:**

1. Name the ordering failure mode explicitly on the get-discovered page: validation or auth middleware ordered ahead of the payment gate returns 4xx/5xx to the cold probe while real purchases still succeed. One sentence and one framework example would have saved every case above.
2. Link `/platform/v2/x402/validate` from the middleware and framework guides where sellers actually build, not only from the discovery page. The operators who hit this failure had passing payments and no reason to visit a page about getting discovered.
3. For anything that reports liveness, Bazaar health probes and third-party catalogs alike: report the observed state, not a boolean, and keep two separate fields, because they describe two different things. What the server answered (`401`, `405`, `402`-with-challenge) and what the prober observed when there was no answer at all (timeout, DNS failure, connection refused). Only the first is the operator's problem to fix, and collapsing them re-creates the boolean problem one level up. Report the state, let policy decide.
4. Surface last-cold-probe status per resource somewhere operator-visible. `EXTENSION-RESPONSES` reports metadata acceptance, but the consecutive-probe down-ranking described in the docs is not visible to the operator it affects, so a degraded listing still looks like silence.

Happy to share the catalog-level breakdown (status code distribution, price-capture rates, method census of our own catalog) if useful.

wakqasahmed:
## Description

Fixes #3113

Documentation-only change, scoped narrowly to the first ask in the issue: naming the "cold probe" failure mode. This PR does **not** implement the broader tooling/probe-behavior-change proposal also discussed in the issue thread (separate two-field liveness reporting, per-resource probe status surfacing, POST-fallback probing, etc.) — that remains open for separate discussion/issues.

Adds a new item to the existing "Troubleshooting catalog visibility" section of `docs/extensions/bazaar.mdx`:

- Names the failure mode explicitly: schema/body validation, auth middleware, or method routing answering a bodyless probe request before the `402` payment gate ever runs.
- Explains why it's invisible in production: real traffic always carries a valid body, so validation passes and the payment gate runs normally — revenue looks healthy while the listing silently degrades or gets deindexed.
- Includes the verified case study from the issue thread (operator's `/purchase` routes reordered so validation ran ahead of the payment gate; bodyless discovery probe hit validation first and got a non-402; fix was folding offer resolution into the payment gate itself plus a middleware-ordering CI test).
- Gives a one-line mitigation: test with a bare, unauthenticated, bodyless request in addition to normal payment-flow integration tests.

## Tests

Docs-only change, no code paths affected. Verified the new section renders correctly as Markdown/MDX and sits under the correct existing heading.

## Checklist

- [x] I have formatted and linted my code (N/A, docs-only)
- [x] All new and existing tests pass (N/A, docs-only)

### DIFF
--- a/docs/extensions/bazaar.mdx
+++ b/docs/extensions/bazaar.mdx
@@ -919,33 +919,15 @@ Common protocol-level mistakes include:
    the declaration was accepted but indexing may happen later. Allow for delay
    before concluding failure with your facilitator.
 
-5. **The cold probe.** A resource must be reachable and return `402 Payment
-   Required` to an unpaid, bodyless request — the "cold probe" that
-   facilitators and third-party catalogs use to confirm a listing is live.
-   The most common way to fail it is ordering: schema/body validation, auth
-   middleware, or method routing runs ahead of the payment gate and answers
-   `400`, `401`, `405`, or `500` before the `402` challenge is ever issued.
-   Real purchases still succeed, because a real client always sends a valid
-   body, so the failure is invisible in production traffic and revenue looks
-   healthy while the listing silently degrades or gets deindexed.
-
-   A production example: an operator's `/purchase` routes had middleware
-   reordered so body validation ran before the x402 payment gate. Actual
-   purchase requests always included the required fields, so validation
-   passed and the payment gate ran normally — nothing looked wrong on the
-   operator's own metrics. But a bodyless discovery probe hit validation
-   first, received a `400`/`500` instead of `402`, and the listing was
-   silently deindexed with no error surfaced to the operator. The fix was to
-   fold the resource/offer resolution needed for a payment challenge into the
-   payment gate itself, so a bodyless request can never reach validation
-   before the `402` is issued — and to add a middleware-ordering test to CI
-   so the same class of regression fails a build instead of shipping silently.
-
-   To avoid this: make sure the payment gate is the first thing a request can
-   hit, ahead of any validation or auth middleware that assumes a well-formed
-   body. Test with a bare, unauthenticated, bodyless request in addition to
-   your normal integration tests — a passing payment flow does not prove the
-   cold probe passes.
+5. **Middleware ordering can hide the payment gate from a bare probe.** Some
+   facilitators or catalogs probe with an unpaid, bodyless request and expect
+   `402` back — check that specific facilitator's policy, since this isn't
+   an x402 protocol requirement. If schema/body validation or auth middleware
+   runs ahead of the payment gate, that probe can get a `400`/`401`/`500`
+   instead. Real traffic still succeeds because real clients send valid
+   bodies, so the test suite passes silently with no assertion on what a
+   bodyless request actually returns. Add a CI test asserting the payment
+   gate runs first, ahead of body-dependent middleware.
 
 If your facilitator exposes optional discovery endpoints,
 `GET /discovery/resources?payTo=<address>` can confirm what that facilitator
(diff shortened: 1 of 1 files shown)


wakqasahmed:
Closing and reopening this from the same branch — the content was tightened per review (was too long, and mixed up a facilitator-specific probe policy with a protocol requirement, now fixed), plus fixing that this went out as part of a same-day batch of PRs from my account.

PR #3165, opened 2026-08-14
fix(stellar): surface real RPC failure detail in settle response (#3125)

### LINKED ISSUE #3125: exact/stellar: settle discards the RPC's sendTransaction response, making retryable and terminal submission failures indistinguishable
### Summary

When submission fails, `ExactStellarScheme.settle` replaces the entire Soroban RPC response with a single constant `errorReason`. The response distinguishes failures a caller **should** retry from failures a caller **must not** retry, and that distinction is lost before it reaches anyone.

### Where

`src/exact/facilitator/index.ts` — in the published build, `dist/cjs/exact/facilitator/index.js:346-355` of 2.21.0:

```js
const sendResult = await server.sendTransaction(txToSubmit);
if (sendResult.status !== "PENDING") {
  return {
    success: false,
    network: payload.accepted.network,
    transaction: "",
    errorReason: "settle_exact_stellar_transaction_submission_failed",
    payer
  };
}
```

`sendResult` carries `status`, `errorResult` (a `TransactionResult` XDR naming the ledger-level failure), `diagnosticEvents` and `latestLedger`. None of it survives.

### Why it matters — two causes, opposite correct responses

Measured on testnet against `https://soroban-testnet.stellar.org`, by proxying JSON-RPC and logging the response the library then discarded. **17 submissions, 10 failures:**

| `status` | Count | Correct caller response |
| --- | --- | --- |
| `TRY_AGAIN_LATER` | **9** | **Retry.** No `errorResult`, nothing reached a ledger, nothing was spent |
| `ERROR` → `txBadSeq` | 1 | **Do not retry.** The payload is stale; resubmitting it can never succeed |

Both arrive at the caller as the same string, so a facilitator operator cannot tell a transient refusal from a dead payload — and neither can the merchant or buyer downstream. In our deployment this presents as roughly one settle in three failing for no visible reason, at a stable rate across sessions.

The `TRY_AGAIN_LATER` case is the sharp one: the status is *literally named* "try again later", and the library converts it into a terminal failure.

```
[tap] SUBMISSION NOT PENDING   status TRY_AGAIN_LATER   latestLedger 4085901
[tap] SUBMISSION NOT PENDING   status ERROR             errorResult {"code":"txBadSeq","feeCharged":"32755"}
[tap] SUBMISSION NOT PENDING   status TRY_AGAIN_LATER   latestLedger 4085938
```

The last of those is from a **freshly funded, never-used source account on its first ever transaction**, which rules out sequence contention as an explanation for the `TRY_AGAIN_LATER` cases.

### The same pattern, one call later — flagging, not claiming

We went looking for a second instance after finding the first, and there is one immediately below. We have **not** measured this path; the code is quoted rather than characterised, and the consequence follows from reading it.

```js
async pollForTransaction(server, txHash, maxPollAttempts = 15, delayMs = 1e3) {
  for (let i = 0; i < maxPollAttempts; i++) {
    try {
      const txResult = await server.getTransaction(txHash);
      if (txResult.status === "SUCCESS") return { success: true };
      else if (txResult.status === "FAILED") return { success: false };
      await new Promise((resolve) => setTimeout(resolve, delayMs));
    } catch (error) { /* … */ }
  }
  return { success: false };          // poll exhausted
}
```

It returns a bare boolean, so `txResult.resultXdr` — the on-chain reason a transaction reverted — is dropped. The sharper part is that **two different outcomes both become `{ success: false }`**, and `settle` renders both as `settle_exact_stellar_transaction_failed`:

1. **`status === "FAILED"`** — included in a ledger and reverted. Terminal, and the reason is in `resultXdr`.
2. **Poll exhaustion** — polling stopped after `maxPollAttempts`. The transaction may still be pending and may still succeed. That is not a failure, it is an unknown.

Those want opposite handling, and unlike the submission case this one has a money-safety edge: `transaction` is populated with the hash, so a caller that treats a timeout as terminal and re-pays **may pay twice** if the original lands. Distinguishing them needs only the status, or a distinct `errorReason` for exhaustion.

We raise it because it is the same habit rather than the same line — information the RPC supplied, discarded before a caller can act on it. **We have not observed this happening and are not claiming a rate for it.**

### Suggested fix

Additive, non-breaking — pass through what the RPC already returned:

```js
if (sendResult.status !== "PENDING") {
  return {
    success: false,
    network: payload.accepted.network,
    transaction: "",
    errorReason: "settle_exact_stellar_transaction_submission_failed",
    payer,
    // NEW: preserve what the RPC actually said.
    rpcStatus: sendResult.status,
    ...(sendResult.errorResult ? { rpcErrorResult: sendResult.errorResult } : {}),
  };
}
```

Existing consumers are unaffected — `errorReason` keeps its current value and meaning. A distinct `errorReason` per status would be more expressive but is a behaviour change for anyone matching on the current string, so the additive form seems the safer proposal.

### Happy to send the PR

It is roughly four lines, additive, with no behaviour change for existing consumers, and we will include a test. Say the word and we will open it against `main` — or take the suggestion and write it yourselves. The goal is the information reaching callers, not the authorship.

### Environment

- `@x402/stellar@2.21.0` (also present in 2.20.0)
- Node 22 / 25, `@stellar/stellar-sdk` 14.x
- `stellar:testnet` via `https://soroban-testnet.stellar.org`

### How one consumer coped, if useful

We wrap `rpc.Server.prototype.sendTransaction` and record non-`PENDING` statuses in `AsyncLocalStorage` scoped to the request, then attach the real status to our own error body:

https://github.com/Vellar-Wallet/vellar-facilitator/blob/1398a97548069fd0c2af30d56b63f277e015a4a8/src/rpcstatus.ts

It works, and it is a monkey-patch on a dependency's prototype — which is why we would rather the information simply be returned.

wakqasahmed:
Closes #3125

## Description

`ExactStellarScheme.settle` collapsed every non-`PENDING` `sendTransaction` result into a single constant `errorReason` (`settle_exact_stellar_transaction_submission_failed`), discarding the RPC's `status`, `errorResult`, and `latestLedger`. That erased the distinction between:

- `TRY_AGAIN_LATER` — nothing reached a ledger, nothing was spent, safe to retry
- `ERROR` (e.g. `txBadSeq`) — terminal, the payload must not be resubmitted

Both surfaced identically to callers, so a facilitator operator (or anything downstream) couldn't tell a transient RPC refusal from a dead payload.

This PR passes the real RPC detail through the **existing** `errorMessage`/`extra` fields already defined on `SettleResponse` (no schema change):

- `extra.rpcStatus` — the raw Soroban RPC `SendTransactionStatus`
- `extra.retryable` — `true` only for `TRY_AGAIN_LATER`
- `extra.latestLedger` — the ledger reported alongside the result
- `extra.errorResultCode` — the `TransactionResult` result-code name (e.g. `txBadSeq`), when present
- `errorMessage` — a human-readable summary of the above

`TRY_AGAIN_LATER` now gets its own `errorReason` (`settle_exact_stellar_transaction_submission_retryable`) instead of reusing the terminal-failure string, so existing consumers matching on `settle_exact_stellar_transaction_submission_failed` for genuine terminal errors are unaffected — only the previously-collapsed `TRY_AGAIN_LATER` case changes its `errorReason`.

Includes a Changesets fragment (`typescript/.changeset/stellar-settle-error-detail.md`).

## Tests

Added two unit tests in `typescript/packages/mechanisms/stellar/test/unit/facilitator-settle.test.ts` mocking `sendTransaction`:

- `TRY_AGAIN_LATER` → asserts `errorReason: settle_exact_stellar_transaction_submission_retryable` and `extra.retryable: true`
- `ERROR` with a `txBadSeq` `errorResult` → asserts `errorReason: settle_exact_stellar_transaction_submission_failed`, `extra.errorResultCode: "txBadSeq"`, and the `errorMessage` text

Updated the pre-existing "non-PENDING status" test to match the new, more specific behavior.

Ran, from `typescript/`:

```
pnpm --filter @x402/core build
pnpm --filter @x402/stellar test        # 10 files, 162 tests passed
pnpm --filter @x402/stellar lint        # clean
pnpm --filter @x402/stellar format      # clean
pnpm --filter @x402/stellar build       # tsup + type-check clean
```

## Checklist

- [x] My commits are signed
- [x] I added a changelog fragment (`typescript/.changeset/stellar-settle-error-detail.md`)

### DIFF
--- a/typescript/.changeset/stellar-settle-error-detail.md
+++ b/typescript/.changeset/stellar-settle-error-detail.md
@@ -1,0 +1,10 @@
+---
+'@x402/stellar': patch
+---
+
+Fixed `ExactStellarScheme.settle` collapsing every non-`PENDING` `sendTransaction`
+result into the same constant `errorReason`. The RPC's `status`, `errorResult` code,
+and `latestLedger` are now surfaced via `errorMessage`/`extra`, and a `TRY_AGAIN_LATER`
+response is reported with a distinct, retryable `errorReason`
+(`settle_exact_stellar_transaction_submission_retryable`) instead of being treated as
+a terminal failure identical to `ERROR`.
--- a/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts
@@ -326,12 +326,28 @@ export class ExactStellarScheme implements SchemeNetworkFacilitator {
       const sendResult = await server.sendTransaction(txToSubmit);
 
       if (sendResult.status !== "PENDING") {
+        // TRY_AGAIN_LATER means nothing reached a ledger and nothing was spent, so
+        // the caller can safely retry. Every other non-PENDING status (e.g. ERROR)
+        // is terminal for this payload and must not be retried as-is.
+        const isRetryable = sendResult.status === "TRY_AGAIN_LATER";
+        const errorResultCode = sendResult.errorResult?.result().switch().name;
         return {
           success: false,
           network: payload.accepted.network,
           transaction: "",
-          errorReason: "settle_exact_stellar_transaction_submission_failed",
+          errorReason: isRetryable
+            ? "settle_exact_stellar_transaction_submission_retryable"
+            : "settle_exact_stellar_transaction_submission_failed",
+          errorMessage: `Soroban RPC sendTransaction returned status ${sendResult.status}${
+            errorResultCode ? ` (${errorResultCode})` : ""
+          }`,
           payer,
+          extra: {
+            rpcStatus: sendResult.status,
+            retryable: isRetryable,
+            latestLedger: sendResult.latestLedger,
+            ...(errorResultCode ? { errorResultCode } : {}),
+          },
         };
       }
 
--- a/typescript/packages/mechanisms/stellar/test/unit/facilitator-settle.test.ts
+++ b/typescript/packages/mechanisms/stellar/test/unit/facilitator-settle.test.ts
@@ -8,6 +8,7 @@ import {
   FeeBumpTransaction,
   Transaction,
   BASE_FEE,
+  xdr,
 } from "@stellar/stellar-sdk";
 import { Api } from "@stellar/stellar-sdk/rpc";
 import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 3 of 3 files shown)


wakqasahmed:
Closing and reopening this from the same branch — no content change, just fixing that this went out as part of a same-day batch of PRs from my account, which isn't good practice. Nothing wrong with the fix itself.

PR #3174, opened 2026-08-15
fix(evm): verify and settle EIP-3009 payments against extra.verifyingContract

0xkurious:
## Problem

`_verify` and `settle` in the facilitator (both `mechanisms/evm/exact/facilitator.py` V2 and `mechanisms/evm/exact/v1/facilitator.py` V1 legacy) always resolve the EIP-712 domain and the on-chain settlement contract from `requirements.asset`, ignoring `extra["verifyingContract"]` entirely:

```python
token_address = normalize_address(requirements.asset)
...
classification = classify_eip3009_signature(
    self._signer, evm_payload.authorization, signature, chain_id,
    token_address, extra["name"], extra["version"],
)
```

This is the facilitator-side counterpart of #2885: that PR gave the Python **client** an opt-in `verifying_contract_validator` so it can sign against a seller-supplied `extra.verifyingContract` (e.g. Circle Gateway's `GatewayWalletBatched`, which settles through a contract distinct from the USDC token). But **every facilitator in this repo still rebuilds the domain from `requirements.asset` only** — flagged by `@0rkz` in review on #2885 ([comment](https://github.com/x402-foundation/x402/pull/2885#issuecomment-5112127349), confirmed still open in the [most recent review](https://github.com/x402-foundation/x402/pull/2885#issuecomment-5302313356)). Concretely: even after #2885, a client that correctly opts in and signs against a Gateway-style verifying contract is still rejected by this facilitator as an invalid signature, because it checks against a different domain than the one actually signed.

## Fix

Add `verifying_contract_validator` to `ExactEvmSchemeConfig` / `ExactEvmSchemeV1Config`, mirroring the client's opt-in gating exactly: absent by default (`None`), so `requirements.asset` is always used unless a validator is supplied and approves the specific candidate. This is a strict opt-in — no behavior change for existing callers.

One difference from the client-only fix: the facilitator's resolved contract is used for **both** verification and settlement, not just the signature check. `token_address` here also drives the deployed-code check, the transfer simulation, and the actual on-chain `transferWithAuthorization` call during settle. A facilitator that trusts a seller-specified verifying contract must consistently verify *and* execute against that same contract — checking a signature against one contract and then calling `transferWithAuthorization` on another would be incoherent. This is documented in the new `verifying_contract_validator` docstring and in the spec update.

Also documents the field's facilitator-side behavior in `specs/schemes/exact/scheme_exact_evm.md`, which previously only specified client-side trust rules (also flagged by `@0rkz`).

## Testing

Added `TestVerifyingContractValidator` to `tests/unit/mechanisms/evm/test_facilitator.py` (V2) and a new `tests/unit/mechanisms/evm/v1/test_facilitator.py` (V1 — no facilitator test file existed for V1 before this change). Both cover: verifies when a validator approves the candidate, rejects a verifying-contract-domain signature when no validator is configured (default-safe), unchanged behavior when `extra.verifyingContract` is absent, and that `settle` calls `transferWithAuthorization` on the trusted contract rather than `requirements.asset`.

The "verifies" tests use a real signed payload from the already-fixed client (`ExactEvmClientScheme`/`ExactEvmSchemeV1Client`) round-tripped through the facilitator, so the signature check is genuine EOA `ecrecover`, not mocked. Verified the new tests fail without the fix (`TypeError` — the config field doesn't exist yet) and pass with it. Full `tests/unit/mechanisms/evm` suite (683 tests) passes.

## Stacking

This branch is based on `fix/exact-scheme-wrong-verifying-contract` (#2885), not a clean `main` — it depends on that branch's `verifying_contract_validator`/`VerifyingContractValidator` pattern on the client side. Base is set to `main` here only because GitHub can't target a base branch that lives in a fork; until #2885 merges, this PR's diff includes #2885's commits too. This PR should not merge before #2885 does.

## Disclosure


0xkurious:
TS counterpart (client + facilitator, both directions): #3175. Same opt-in `verifyingContractValidator` shape as this PR.

0xkurious:
Go counterpart (client + facilitator, both directions): #3176. Same `VerifyingContractValidator` shape as this PR, adapted to Go idiom (functional option on the client, direct config field on the facilitator).

phdargen:
Thanks for the contribution @0xkurious, but I don't see a legitimate use case here: if the receiver is a contract as a passive receiver they can just be a regular payTo; if the receiver is an authorization-consuming contract, using `transferWithAuthorization` would be a bad implementation vulnerable to front-running and stranding the deposit; a valid implementation would use `receiveWithAuthorization` instead. Moreover, this introduces a security footgun in leaving verifyingContract as a server-provided untrusted field. 

Circle only uses the 3009 auth offchain to update their internal DB, so there is no onchain tx for the facilitator to submit here and an independent facilitator cant update Circle's DB, it has to go through their API

PR #3175, opened 2026-08-15
fix(evm): sign/verify/settle EIP-3009 payments against extra.verifyingContract (TS)

0xkurious:
## Problem

TS EIP-3009 client signing (`createEIP3009Payload`/`signEIP3009Authorization` for V2, `ExactEvmSchemeV1.signAuthorization` for V1) and facilitator verify/settle (`verifyEIP3009`/`settleEIP3009` for V2, `ExactEvmSchemeV1` for V1) all resolve the EIP-712 domain and the on-chain settlement contract from `requirements.asset` unconditionally, ignoring `extra["verifyingContract"]` entirely:

```ts
const domain = {
  name,
  version,
  chainId,
  verifyingContract: getAddress(requirements.asset), // always the token, even when
};                                                     // extra specifies a different verifier
```

This is the TypeScript counterpart of the already-reviewed Python fixes:
- Client: x402-foundation/x402#2885
- Facilitator: x402-foundation/x402#3174

Same root cause as both: correct only when the token contract itself is the EIP-3009 verifier — wrong whenever a seller settles through a distinct contract (e.g. Circle Gateway's `GatewayWalletBatched`). Flagged as a cross-SDK gap by `@0rkz` in review on #2885 ([comment](https://github.com/x402-foundation/x402/pull/2885#issuecomment-5112127349)): "TS v2 (`mechanisms/evm/src/exact/facilitator/eip3009.ts`)... TS v1... all rebuild the domain from `requirements.asset`."

## Fix

Mirrors the reviewed Python design exactly: an opt-in `verifyingContractValidator` callback.

- **Client** (`ExactEvmScheme`/`ExactEvmSchemeV1`, both takes a new optional constructor param — TS's client had no config object at all before this, unlike the facilitator side): absent by default, so signing always uses `requirements.asset` unless a validator is supplied and approves the specific candidate.
- **Facilitator** (`EIP3009FacilitatorConfig`/`ExactEvmSchemeConfig`/`ExactEvmSchemeV1Config`): same opt-in shape. Once trusted, the candidate is used consistently for **both** verification and settlement — the deployed-code check, transfer simulation, the on-chain `transferWithAuthorization` call, and the settled Transfer-event check — not just the signature check. A facilitator that verifies a signature against one contract can't settle against a different one.

No behavior change for existing callers in either direction (strict opt-in, matches the pre-existing default in both languages).

## Spec

Not touching `specs/schemes/exact/scheme_exact_evm.md` in this PR — that documentation is being introduced by #2885 (client-side) and #3174 (facilitator-side), both still open. Duplicating the edit here would conflict with those. This PR's behavior matches what's specified there once merged.

## Testing

Added `verifyingContractValidator` test suites to all four existing test files (`test/unit/exact/client.test.ts`, `test/unit/v1/client.test.ts`, `test/unit/exact/facilitator.test.ts`, `test/unit/v1/facilitator.test.ts`) — 14 new tests total. Client tests assert the signed EIP-712 domain via the mocked `signTypedData` call args (existing pattern in this suite, e.g. the Permit2 domain test). Facilitator tests assert `getCode`/`writeContract` are called against the resolved contract, using the suite's existing "everything is a deployed contract with an always-valid signature" mock convention to isolate the address-resolution wiring from signature cryptography.

Verified all 6 "trusts the verifying contract" tests fail without the production fix (temporarily reverted, re-ran — 6 failures, mix of `TypeError`-shaped and wrong-address assertion failures) and pass with it. Full `@x402/evm` suite: 851 tests pass (up from 837 pre-PR).

## Disclosure

### DIFF
--- a/typescript/.changeset/verifying-contract-eip3009.md
+++ b/typescript/.changeset/verifying-contract-eip3009.md
@@ -1,0 +1,5 @@
+---
+'@x402/evm': patch
+---
+
+Fixed EIP-3009 client signing and facilitator verify/settle against `requirements.asset` unconditionally, ignoring a seller-supplied `extra.verifyingContract` (e.g. Circle Gateway's batch-settlement contract). Trusting `extra.verifyingContract` is opt-in via a `verifyingContractValidator` callback on `ExactEvmScheme`/`ExactEvmSchemeV1` (client and facilitator); by default (no validator supplied) signing, verification, and settlement still always use the asset address, matching prior behavior.
--- a/typescript/packages/mechanisms/evm/src/exact/client/eip3009.ts
+++ b/typescript/packages/mechanisms/evm/src/exact/client/eip3009.ts
@@ -6,17 +6,54 @@ import { ExactEIP3009Payload } from "../../types";
 import { createNonce, getEvmChainId } from "../../utils";
 
 /**
+ * Validates a seller-supplied extra.verifyingContract before it is trusted for EIP-712 signing.
+ *
+ * @param candidate - The address from requirements.extra.verifyingContract
+ * @param requirements - The full payment requirements, for context (e.g. network)
+ * @returns True to sign against `candidate`. False to fall back to requirements.asset, same as
+ *   if no verifyingContract were supplied.
+ */
+export type VerifyingContractValidator = (
+  candidate: string,
+  requirements: PaymentRequirements,
+) => boolean;
+
+/**
+ * Resolve the EIP-712 verifyingContract for signing. Defaults to requirements.asset. Only trusts
+ * a seller-supplied extra.verifyingContract if a validator was supplied and it approves this
+ * specific candidate.
+ *
+ * @param requirements - The payment requirements
+ * @param validator - Optional validator to trust extra.verifyingContract
+ * @returns The address to use as the EIP-712 domain's verifyingContract
+ */
+function resolveVerifyingContract(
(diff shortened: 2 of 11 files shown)


0xkurious:
Go counterpart (client + facilitator, both directions): #3176. Same design, different plumbing -- Go's client config is a shared type alias (`evm.RPCConfig`, also used by `upto`) so the validator couldn't live there without leaking into unrelated schemes; used a functional option (`WithVerifyingContractValidator`) instead, which also kept the ~40 existing `NewExactEvmScheme`/`NewExactEvmSchemeV1` call sites in this repo compiling unchanged.

0xkurious:
Closing — this was the TS counterpart of #2885/#3174, and @phdargen's review on #3174 invalidates the premise both were built on: https://github.com/x402-foundation/x402/pull/3174#issuecomment-5304764975

Circle Gateway doesn't settle via an on-chain `transferWithAuthorization` against a distinct verifying contract (the authorization is used off-chain to update Circle's internal DB, not submitted by an independent facilitator), so the motivating scenario doesn't occur in practice. Without it, `extra.verifyingContract` has no legitimate use — a passive receiver is just a regular `payTo`, and a receiver that needs to act on the authorization should use `receiveWithAuthorization`, not the front-runnable `transferWithAuthorization`. Closing #2885 and #3176 for the same reason.

PR #3176, opened 2026-08-15
fix(evm): sign/verify/settle EIP-3009 payments against extra.verifyingContract (Go)

0xkurious:
## Problem

Go EIP-3009 client signing (`createEIP3009Payload`/`signAuthorization` for V2, `ExactEvmSchemeV1.signAuthorization` for V1) and facilitator verify/settle (`verifyEIP3009`/`settleEIP3009` for V2, `ExactEvmSchemeV1` for V1) all resolve the EIP-712 domain and the on-chain settlement contract from `requirements.Asset` unconditionally, ignoring `extra["verifyingContract"]` entirely — the Go counterpart of:
- Client: x402-foundation/x402#2885 (Python), x402-foundation/x402#3175 (TS)
- Facilitator: x402-foundation/x402#3174 (Python), x402-foundation/x402#3175 (TS)

Same root cause each time: correct only when the token contract itself is the EIP-3009 verifier — wrong whenever a seller settles through a distinct contract (e.g. Circle Gateway's `GatewayWalletBatched`). Flagged as a cross-SDK gap by `@0rkz` on #2885 ([comment](https://github.com/x402-foundation/x402/pull/2885#issuecomment-5112127349)): "...and Go (`exact/facilitator/eip3009.go`)... all rebuild the domain from `requirements.asset`."

## Fix

Same opt-in `VerifyingContractValidator` design as Python/TS, adapted to two different Go idioms depending on what each config type already was:

- **Client** (`NewExactEvmScheme`/`NewExactEvmSchemeV1`): a new **functional option**, `WithVerifyingContractValidator(...)`, appended as a variadic trailing parameter. The client's existing `ExactEvmSchemeConfig` is a type alias to the shared `evm.RPCConfig` also used by the `upto` scheme — adding a validator field there would leak an EIP-3009-specific concern into unrelated schemes, so a functional option was the only option that didn't touch that shared type. This also kept the change backward-compatible: `NewExactEvmScheme`/`NewExactEvmSchemeV1` are called at ~40 sites across this repo (integration tests, unit tests) with the old 2-arg (or 1-arg, for V1) signature — a variadic option means every one of those keeps compiling unchanged.
- **Facilitator** (`ExactEvmSchemeConfig`/`ExactEvmSchemeV1Config`): both are already dedicated per-scheme structs (not shared aliases), so `VerifyingContractValidator` goes directly on them as a new field — same precedent as `EIP6492AllowedFactories`. Once trusted, the candidate is used consistently for **both** verification and settlement (the deployed-code check via `ValidateAssetIsContract`, transfer simulation, the on-chain `transferWithAuthorization` call, and the settled Transfer-event check) — not just the signature check.

Nil/absent by default in both cases — no behavior change for existing callers.

## Spec

Not touching `specs/schemes/exact/scheme_exact_evm.md` in this PR, same reasoning as #3175 — that documentation is being introduced by #2885 and #3174, both still open; duplicating here would conflict.

## Testing

Added test files/cases for all four surfaces:
- `mechanisms/evm/exact/client/eip3009_test.go` (new file — only `rpc_test.go` existed here before)
- `mechanisms/evm/exact/v1/client/eip3009_test.go` (new file — no tests existed for this package before)
- `mechanisms/evm/exact/facilitator/eip3009_verifying_contract_test.go` (new file)
- `mechanisms/evm/exact/v1/facilitator/eip3009_verifying_contract_test.go` (new file — no tests existed for this package before)

Client tests assert the signed domain via a recording mock signer. Facilitator tests use **real ECDSA signing and `ecrecover`** (`crypto.GenerateKey`/`crypto.Sign`, same pattern already used for Permit2 in `test/unit/evm_client_facilitator_test.go`), not mocked signature checks — Go's `FacilitatorEvmSigner`/`VerifyUniversalSignature` has no "always valid" bypass for EOA payers the way the TS mock's `isValidSignature` magic value does, so this is the genuine signature-domain check.

Verified all 4 new test files fail to *compile* without the production fix (temporarily reverted — `undefined: WithVerifyingContractValidator` / `unknown field VerifyingContractValidator`) and pass with it. Full module: `go build ./...`, `go vet ./...`, and `go test -race -cover ./...` all clean, including every pre-existing call site using the old constructor signatures.

## Disclosure

### DIFF
--- a/go/.changes/unreleased/fixed-20260815-200311.yaml
+++ b/go/.changes/unreleased/fixed-20260815-200311.yaml
@@ -1,0 +1,3 @@
+kind: Fixed
+body: Client EIP-3009 signing and facilitator verify/settle (V1 and V2) always resolved the EIP-712 domain and settlement contract from requirements.Asset, ignoring a seller-supplied extra.verifyingContract (e.g. Circle Gateway's batch-settlement contract). Trusting extra.verifyingContract is opt-in via a VerifyingContractValidator (client: WithVerifyingContractValidator option; facilitator: ExactEvmSchemeConfig/ExactEvmSchemeV1Config field); by default (none configured) signing, verification, and settlement still always use the asset address, matching prior behavior
+time: 2026-08-15T20:03:11.000000+00:00
--- a/go/mechanisms/evm/exact/client/eip3009_test.go
+++ b/go/mechanisms/evm/exact/client/eip3009_test.go
@@ -1,0 +1,116 @@
+package client
+
+import (
+	"context"
+	"testing"
+
+	"github.com/x402-foundation/x402/go/v2/mechanisms/evm"
+	"github.com/x402-foundation/x402/go/v2/types"
+)
+
+const (
+	gatewayContract = "0x77777777Dcc4d5A8B6E418Fd04D8997ef11000eE"
+	usdcAsset       = "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913"
+)
+
+type recordingSigner struct {
+	address string
+	domain  evm.TypedDataDomain
+}
+
+func (s *recordingSigner) Address() string {
+	return s.address
+}
+
+func (s *recordingSigner) SignTypedData(
+	ctx context.Context,
+	domain evm.TypedDataDomain,
+	fields map[string][]evm.TypedDataField,
+	primaryType string,
+	message map[string]interface{},
+) ([]byte, error) {
+	s.domain = domain
+	return make([]byte, 65), nil
+}
+
+func gatewayRequirements() types.PaymentRequirements {
+	return types.PaymentRequirements{
+		Scheme:            "exact",
+		Network:           "eip155:8453",
+		Asset:             usdcAsset,
+		Amount:            "8000",
+		PayTo:             "0x0987654321098765432109876543210987654321",
+		MaxTimeoutSeconds: 3600,
+		Extra: map[string]interface{}{
+			"name":              "GatewayWalletBatched",
+			"version":           "1",
+			"verifyingContract": gatewayContract,
+		},
+	}
+}
+
+func TestCreateEIP3009PayloadSignsAgainstVerifyingContractWhenValidatorApproves(t *testing.T) {
+	signer := &recordingSigner{address: "0x1234567890123456789012345678901234567890"}
+	scheme := NewExactEvmScheme(signer, nil, WithVerifyingContractValidator(
+		func(candidate string, requirements types.PaymentRequirements) bool {
+			return candidate == gatewayContract
+		},
+	))
+
+	if _, err := scheme.CreatePaymentPayload(context.Background(), gatewayRequirements()); err != nil {
+		t.Fatalf("CreatePaymentPayload failed: %v", err)
+	}
+
+	if got := evm.NormalizeAddress(signer.domain.VerifyingContract); got != evm.NormalizeAddress(gatewayContract) {
+		t.Fatalf("expected domain.VerifyingContract = %s, got %s", gatewayContract, got)
+	}
+}
+
+func TestCreateEIP3009PayloadFallsBackToAssetWhenNoValidatorConfigured(t *testing.T) {
+	signer := &recordingSigner{address: "0x1234567890123456789012345678901234567890"}
+	scheme := NewExactEvmScheme(signer, nil)
+
+	if _, err := scheme.CreatePaymentPayload(context.Background(), gatewayRequirements()); err != nil {
+		t.Fatalf("CreatePaymentPayload failed: %v", err)
+	}
+
+	if got := evm.NormalizeAddress(signer.domain.VerifyingContract); got != evm.NormalizeAddress(usdcAsset) {
+		t.Fatalf("expected domain.VerifyingContract = %s, got %s", usdcAsset, got)
+	}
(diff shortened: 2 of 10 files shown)


0rkz:
Checked this one alongside #3174 and #3175 (notes on the shared shape are over on #2885). The validator plumbing here matches the other two: `resolveVerifyingContract()` feeds both `verifyEIP3009`/`settleEIP3009` (V2) and `verify`/`Settle` (V1), and the default is closed — candidate absent, validator nil, or validator false all fall back to `requirements.Asset`.

One thing that's specific to Go, and worth a look before this lands.

Once a validator approves a candidate, the three languages don't validate its address format equally. Python's `normalize_address()` raises on a bad length or non-hex input, and TS's `getAddress()` (viem) throws on the same — so in both, a malformed candidate dies before anything downstream sees it. Go's `NormalizeAddress()` (`go/mechanisms/evm/utils.go:57-64`, untouched by this PR) does neither: it lowercases and re-prefixes `0x`, with no length or hex check.

Nothing downstream rejects it either, and the failure mode is quiet rather than loud. `ValidateAssetIsContract` hands the string to `signer.GetCode`, and the signer implementations in this repo resolve it with go-ethereum's `common.HexToAddress` — which discards the hex-decode error, keeps only the last 20 bytes of an over-long value, and left-pads a short one. So a malformed candidate doesn't error, it becomes a *different, well-formed* address:

Taking the gateway address the new tests already use, `0x77777777Dcc4d5A8B6E418Fd04D8997ef11000eE`, and mistyping it by one character:

```
one char too many (41 hex):
  0x77777777Dcc4d5A8B6E418Fd04D8997ef11000eEf
  -> 0x7777777dcc4d5a8b6e418fd04d8997ef11000eef   // different address, no error

one char short (39 hex):
  0x77777777Dcc4d5A8B6E418Fd04D8997ef11000e
  -> 0x077777777dcc4d5a8b6e418fd04d8997ef11000e   // left-padded, also different
```

(`FromHex` left-pads an odd-length string, `SetBytes` then keeps only the last 20 bytes of the over-long case.) A non-hex character truncates at that point and left-pads with zeros, so the value collapses toward `0x0000…`.

The same conversion happens directly at `exact/facilitator/eip3009.go:256` for the post-settlement `Transfer` check.

To be clear about scope: this isn't introduced here — `NormalizeAddress` was already this permissive for `requirements.Asset`. What changes is that this PR is the first place it's applied to a **seller-influenced** value (`extra.verifyingContract`) rather than the more-constrained asset field, so a `VerifyingContractValidator` that approves a candidate without doing its own strict format check has no independent backstop in Go, where Python and TS implicitly provide one. Severity is bounded by that — it needs an operator validator that approves a malformed string in the first place.

The package already has a strict `IsValidAddress` sitting right next to `NormalizeAddress` in the same file; it's used on the server path, just not this one. Worth a case in the new `eip3009_verifying_contract_test.go` too — none of the three PRs' new tests exercise a malformed-format candidate, so CI wouldn't catch a regression here.

0rkz:
Correction to my note above: I got the failure mode wrong, and the two worked examples don't describe what this code actually does. Withdrawing the security framing.

`NormalizeAddress()` does skip length and hex validation — that part holds. But a malformed candidate never reaches `common.HexToAddress`, because the domain is hashed first. `verifyEIP3009` calls `ClassifyEIP3009Signature` at `eip3009.go:101`, well before `ValidateAssetIsContract` at `:132`. That path runs through `HashEIP3009Authorization` into go-ethereum's `apitypes` (imported at `go/mechanisms/evm/eip712.go:10`), which encodes an `address` field only when `common.IsHexAddress` passes — and that requires exactly 40 hex characters (go-ethereum v1.17.2, `common/types.go:242`). A malformed value fails there and surfaces through the wrapper at `eip712.go:94` as `failed to hash domain: provided data '…' doesn't match type 'address'`. `settle` re-verifies before settling, so it inherits the same rejection, V1 and V2 alike.

So `0x7777777dcc…eef` and `0x0777…000e` are correct for `common.HexToAddress` in isolation, but that is not the path a candidate takes here. The claim that a malformed candidate is silently coerced into a different valid address is wrong, and I withdraw it along with the "no independent backstop" framing that rested on it.

What survives is narrower, and is a diagnostics point rather than a missing guard: Python's `normalize_address` and viem's `getAddress` both reject a malformed candidate at the resolve step with an address-specific error, while Go rejects it one step later as a signature-verification failure. The strict `IsValidAddress` already sitting beside `NormalizeAddress` in that same file would give Go the clearer error at the same point the other two produce it.

Apologies for the noise — the ordering is on me for not checking it before writing.

0xkurious:
Closing — this was the Go counterpart of #2885/#3174, and @phdargen's review on #3174 invalidates the premise all four PRs were built on: https://github.com/x402-foundation/x402/pull/3174#issuecomment-5304764975

Circle Gateway doesn't settle via an on-chain `transferWithAuthorization` against a distinct verifying contract (the authorization is used off-chain to update Circle's internal DB, not submitted by an independent facilitator), so the motivating scenario doesn't occur in practice. Without it, `extra.verifyingContract` has no legitimate use — a passive receiver is just a regular `payTo`, and a receiver that needs to act on the authorization should use `receiveWithAuthorization`, not the front-runnable `transferWithAuthorization`. Closing #2885 and #3175 for the same reason.

PR #3180, opened 2026-08-17
fix(ts): handle missing accepted requirements

### LINKED ISSUE #3171: paymentRequirementsMatchAccepted throws on missing/null payload.accepted, leaking a raw internal error (resource-server side, related to #3015)
## Description

`paymentRequirementsMatchAccepted()` (`typescript/packages/core/src/server/x402ResourceServer.ts:1903-1923`) destructures the client-supplied `accepted` parameter without checking it's a non-null object first:

```ts
function paymentRequirementsMatchAccepted(
  required: PaymentRequirements,
  accepted: PaymentRequirements,
  dynamicExtraFields?: string[],
): boolean {
  const { extra: requiredExtra, ...requiredCore } = required;
  const { extra: acceptedExtra, ...acceptedCore } = accepted;   // throws if accepted is null/undefined
  ...
```

It's called from `findMatchingRequirements()` (line 1459) for every x402Version 2 payload, passing `paymentPayload.accepted` straight from client input:

```ts
return availableRequirements.find(paymentRequirements => {
  ...
  return paymentRequirementsMatchAccepted(
    paymentRequirements,
    paymentPayload.accepted,   // client-controlled, can be missing or null
    scheme?.dynamicExtraFields,
  );
});
```

`findMatchingRequirements` is called from `processHTTPRequest()` (`typescript/packages/core/src/http/x402HTTPResourceServer.ts:620`) inside a `try` whose `catch` (line 738) forwards any thrown `Error.message` straight into the re-issued `PaymentRequiredResponse.error` field:

```ts
} catch (error) {
  if (error instanceof FacilitatorResponseError) {
    throw error;
  }
  const errorResponse = await this.ResourceServer.createPaymentRequiredResponse(
    requirements,
    resourceInfo,
    error instanceof Error ? error.message : "Payment verification failed",
    ...
```

This is resource-server code, the path any server built with `@x402/express`/`x402Serve()` runs on every incoming payment, not a facilitator-only code path.

## Steps to reproduce

Confirmed live against a production resource server built on `@x402/express` + `@x402/core` 2.17.0 (ours): a `PAYMENT-SIGNATURE` header decoding to `{"x402Version":2,"payload":{...}}` with `accepted` omitted returns a 402 whose `payment-required` header, base64-decoded, contains:

```json
{"error": "Cannot destructure property 'extra' of 'accepted' as it is undefined."}
```

`accepted: null` produces the same class of error (`"...as it is null."`). Tried every other malformed shape for `accepted`, including `{}`, a string, a number, and an array, and all of those are handled correctly, falling through to the intended `"No matching payment requirements"` message. Only `undefined`/`null` hit the unguarded destructure.

Confirmed present on current `main` and on the latest published `@x402/core@2.22.0`.

## Expected behavior

A malformed `accepted` field should fail with a clear, stable, protocol-level reason (e.g. `"payload.accepted is required for x402Version 2"`), not an engine-specific `TypeError` string that can vary across Node/V8 versions.

## Relationship to #3015

Different component, same underlying pattern. #3015 is the hosted facilitator's HTTP routes (`typescript/site/app/facilitator/{verify,settle}/route.ts`) casting request bodies with zero schema validation, crashing with a 500 in an EVM-specific mechanism file. This is in `@x402/core`'s own resource-server matching logic, used by every resource server and not just the facilitator, and it doesn't crash (it's caught and returned as a normal 402), but it still leaks raw internal error text where a typed protocol error belongs. The fix direction proposed in #3015 (validate with `@x402/core/schemas`'s `parsePaymentPayload` before touching fields) would also close this, applied at the `findMatchingRequirements` entry point.

JasonColapietro:
## Description

- Treat missing or null `accepted` requirements as no match for both supported protocol versions.
- Return the existing stable protocol error instead of leaking a runtime `TypeError` from client-controlled input.
- Add end-to-end HTTP regression coverage for v1/v2 × omitted/null `accepted` values.

Fixes #3171

AI assistance: significant. The behavior was reproduced first, the patch was reviewed on separate standards and spec axes, and the repository-wide suite was run before submission.

## Tests

- `corepack pnpm test` from `typescript/` — 55/55 tasks passed
- `corepack pnpm --filter @x402/core lint:check`
- `corepack pnpm --filter @x402/core format:check`
- `corepack pnpm --filter @x402/core build`

## Checklist

- [x] My commits are signed
- [x] I added a changelog fragment for the user-facing change

### DIFF
--- a/typescript/.changeset/safe-payments-match.md
+++ b/typescript/.changeset/safe-payments-match.md
@@ -1,0 +1,5 @@
+---
+'@x402/core': patch
+---
+
+Handled missing accepted payment requirements without leaking a runtime error.
--- a/typescript/packages/core/src/server/x402ResourceServer.ts
+++ b/typescript/packages/core/src/server/x402ResourceServer.ts
@@ -1462,6 +1462,10 @@ export class x402ResourceServer {
   ): PaymentRequirements | undefined {
     switch (paymentPayload.x402Version) {
       case 2:
+        if (!paymentPayload.accepted) {
+          return undefined;
+        }
+
         // For v2, all server-declared requirements must match.
         // The client may include additive scheme-specific metadata under `accepted.extra`.
         // Scheme-declared dynamicExtraFields are omitted from the extra comparison
@@ -1478,6 +1482,10 @@ export class x402ResourceServer {
           );
         });
       case 1:
+        if (!paymentPayload.accepted) {
+          return undefined;
+        }
+
         // For v1, match by scheme and network
         return availableRequirements.find(
           req =>
--- a/typescript/packages/core/test/unit/http/x402HTTPResourceServer.errors.test.ts
+++ b/typescript/packages/core/test/unit/http/x402HTTPResourceServer.errors.test.ts
@@ -9,7 +9,7 @@ import {
   buildPaymentRequirements,
   buildSupportedResponse,
 } from "../../mocks";
-import { encodePaymentSignatureHeader } from "../../../src/http";
+import { decodePaymentRequiredHeader, encodePaymentSignatureHeader } from "../../../src/http";
 
 class MockHTTPAdapter implements HTTPAdapter {
   constructor(private readonly headers: Record<string, string> = {}) {}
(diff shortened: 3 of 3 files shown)


phdargen (approved):
Thanks for the patch @JasonColapietro, LGTM 🚀

PR #3181, opened 2026-08-17
docs(extensions): optional pricing metadata for bazaar listings

### LINKED ISSUE #3264: [Feature]: Bazaar listings cannot express the price of a metered service
### Problem or use case

For the `exact` scheme, `accepts[].amount` is the price. For metered schemes it is the authorization **ceiling** — not the price, and usually much larger. A catalog that surfaces the ceiling in the field an agent reads as "cost" biases discovery against precisely the services metering exists to enable.

Three failures this produces, observed across current Stellar implementations:

1. Budget filters exclude metered services whose settled cost would have landed under budget, because only the ceiling is visible before the call.
2. A catalog advertises a price lower than a metered route can legitimately charge, because the settlement layer rewrites the amount before the catalog reads it.
3. Settlement-count ranking stops being comparable once amounts vary, and gaming it gets cheaper the smaller each settlement is.

**Why discovery rather than the scheme.** The author of the `upto` binding draft (#3134) stated the placement directly:

> For (1) and (2), I think richer pricing metadata belongs at the discovery layer rather than in the core upto scheme.

— [Iam0TI, stellar/x402-stellar#72](https://github.com/stellar/x402-stellar/issues/72#issuecomment-5304064452)

The author of #3098 responded to the same proposal from the settlement-catalog side, opening with "this is additive to your three points rather than a counter to any of them" and noting that it would give his catalog's `accepts` dedupe key somewhere to pick up a profile discriminator — [Eras256, same thread](https://github.com/stellar/x402-stellar/issues/72#issuecomment-5309876697). I read that as an endorsement of usefulness rather than a position on where the field belongs, and it is worth reading in his own words rather than through my summary of it.

Keeping this at the discovery layer leaves `PaymentRequirements.amount` unambiguously the ceiling, and stops every future metered scheme from re-specifying the same field.

Point 3 above is not hypothetical. #3226 is currently working through the adjacent problem that Bazaar counters can be minted by verify-only calls, because the extension's *Facilitator Behavior* section never mentions settlement. A count that can be minted and a count whose amounts are incomparable are the same failure seen from two sides.

### Proposed solution

An optional `pricing` object on the bazaar extension, sibling of `info` / `schema`, carrying a required `model` and optional `unit` / `typical` / `note`.

Purely additive: no change to any payment scheme, `model` is required only when `pricing` is present, a malformed `pricing` soft-drops without taking the listing with it, and catalogs decide whether to use it at all.

The spec PR for this is #3181, open for review.

*Disclosure: drafted with AI assistance under my direction and reviewed by hand. Both quotations above were re-verified against the linked comments today.*

pedro-pelicioni:
## Motivation

Closes #3264

For the `exact` scheme, `accepts[].amount` is the price. For metered schemes (`upto`), it is the authorization ceiling, which is not the price and is usually much larger. Catalogs that surface the ceiling in the field agents read as "cost" bias discovery against exactly the services upto exists to enable. Three concrete failures observed across current Stellar implementations:

1. Budget filters silently exclude metered services whose settled cost lands under budget (only the ceiling is visible pre-call).
2. A catalog advertised a lower price than an upto route could legitimately charge, because the settlement layer rewrites the amount field before the catalog reads it.
3. Settlement-count ranking stops being comparable once amounts vary, and gaming it gets cheaper the smaller the settlements are.

The author of the `upto` binding draft (#3134) stated the placement directly — *"richer pricing metadata belongs at the discovery layer rather than in the core upto scheme"* ([comment](https://github.com/stellar/x402-stellar/issues/72#issuecomment-5304064452)). The author of #3098 responded to the same proposal from the settlement-catalog side, opening with *"this is additive to your three points rather than a counter to any of them"* and noting it would give his catalog's `accepts` dedupe key somewhere to pick up a profile discriminator ([comment](https://github.com/stellar/x402-stellar/issues/72#issuecomment-5309876697)); I read that as an endorsement of usefulness rather than a position on where the field belongs, and it is worth reading in his own words rather than through my summary.

Keeping this at the discovery layer leaves `PaymentRequirements.amount` unambiguously the authorization ceiling, and lets discovery decide how a settled amount is presented. This PR proposes the shape.

*Edited 2026-08-25: an earlier version of this paragraph said both `scheme_upto_stellar` authors had agreed the metadata belongs at the discovery layer. Only #3134's author said that. Corrected above, and apologies to both for the mischaracterisation.*

## Design

An optional `pricing` object on the bazaar extension, sibling of `info`/`schema`:

- `model` (required if `pricing` is present): `"per-call" | "per-token" | "per-second" | "per-byte" | "tiered" | "custom"`. Always expressible.
- `unit` (optional): amount per stated quantity, omitted when meaningless.
- `typical` (optional): what a median call settles, in the asset's smallest unit, as a string like every other amount on the wire.
- `note` (optional): free-form, for models that fit none of the above.

`unit` and `typical` are optional precisely because metered products do not share a unit. A catalog that receives none of them falls back to the ceiling and says so. Malformed `pricing` is a soft-drop (the field is discarded, the listing survives), consistent with the extension's existing validation posture.

## The self-correcting property

A seller-declared `typical` is a claim, and the catalog is the only party positioned to check it: once the settled actual amount is exposed in a stable place, a catalog holding a resource's settlement history can compare declared-typical against observed-median and flag or down-rank the gap. That keeps the field honest without the settlement scheme enforcing anything, which is further evidence the metadata belongs here.

## Non-goals

- No change to any payment scheme: `amount` remains the authorization ceiling.
- No mandated ranking: catalogs decide how (and whether) to use `pricing`.

## AI disclosure

Drafted with AI assistance, then reviewed and verified by hand. Flagging per CONTRIBUTING.md so reviewers can calibrate.

### DIFF
--- a/specs/extensions/bazaar.md
+++ b/specs/extensions/bazaar.md
@@ -419,6 +419,82 @@ responsibility (e.g. via Cloudinary at serve time).
 
 ---
 
+## Pricing Metadata (`pricing`)
+
+For the `exact` scheme, `accepts[].amount` is the price. For metered schemes such
+as `upto`, it is the **authorization ceiling**, which is not the price and is
+usually much larger. A catalog that surfaces the ceiling in the field agents read
+as "cost" makes every metered service look expensive next to a fixed-price one,
+and a budget filter keyed on it silently excludes metered services whose settled
+cost lands under budget.
+
+Resource servers MAY therefore publish an optional `pricing` object on the
+`bazaar` extension, as a sibling of `info` and `schema`. Like the service
+metadata above, it is purely additive: servers that omit it produce byte-identical
+402 bodies, and clients that don't recognize it ignore it. `accepts[].amount`
+remains unambiguously the authorization ceiling; `pricing` never participates in
+payment authorization or settlement.
+
+### Example
+
+```json
+{
+  "extensions": {
+    "bazaar": {
+      "info": { ... },
+      "schema": { ... },
+      "pricing": {
+        "model": "per-token",
+        "unit": { "amount": "12", "per": "1000 tokens" },
+        "typical": "350",
+        "note": "long prompts settle above typical"
+      }
+    }
+  }
+}
+```
+
+### Fields
+
+| Field     | Type   | Required | Description                                                                                                                                          |
+|-----------|--------|----------|------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `model`   | string | Yes      | One of `"per-call"`, `"per-token"`, `"per-second"`, `"per-byte"`, `"tiered"`, `"custom"`. The pricing model the resource meters by.                  |
+| `unit`    | object | No       | `{ "amount": string, "per": string }` — cost per stated quantity, e.g. `{ "amount": "12", "per": "1000 tokens" }`. Omitted when no single unit is meaningful. |
+| `typical` | string | No       | What a median call settles, in the asset's smallest unit, encoded as a decimal string like every other amount on the wire.                            |
+| `note`    | string | No       | Free-form clarification for models that fit none of the above.                                                                                       |
+
+`unit` and `typical` are optional precisely because metered products do not share
+a unit (per-token, per-second, per-byte, per-GB-month, multi-dimensional). A
+catalog that receives none of them SHOULD fall back to the ceiling and say so,
+rather than presenting the ceiling as a price.
+
+### Validation Rules
+
+`pricing` crosses the same trust boundary as the service metadata above: clients
+echo it from `PaymentRequired` into `PaymentPayload`, so facilitators MUST apply
+soft-drop rules during extraction. A field that fails its rule is discarded; the
+listing survives.
+
+| Field     | Rule                                                                                                                                     | On violation             |
+|-----------|------------------------------------------------------------------------------------------------------------------------------------------|--------------------------|
+| `pricing` | Object containing a valid `model`.                                                                                                       | Drop the whole `pricing`. |
+| `model`   | One of the six enumerated strings.                                                                                                       | Drop the whole `pricing`. |
+| `unit`    | Object; `amount` a non-empty string of ASCII digits (optionally one `.`); `per` a non-empty printable-ASCII string, length ≤ 64.          | Drop `unit`.             |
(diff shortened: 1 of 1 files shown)


pedro-pelicioni:
Out of draft.

This was held pending the `scheme_upto_stellar` consolidation, and that has not moved: #3098 and #3134 are both still open and unconverged, and the request for consolidation direction in stellar/x402-stellar#72 is nine days without an answer. I am putting it up for review anyway rather than waiting further, because the field this proposes is additive and scheme-independent — nothing in it depends on which way that converges. Holding it longer costs review time without buying any more certainty about field names.

Three things changed in the spec text since the first draft, all from re-reading it the way a reviewer would:

- **The validation rules are stated as SHOULD, not MUST.** The `resource` metadata rules directly above them are MUST, and they arrived together with their TypeScript, Go and Python extraction paths in a single change (#2200). This PR ships no implementation, so MUST here would be a rule nothing enforces. A follow-up adds all three, and raising these to MUST belongs with that change; the reasoning is now in the text rather than left for a reviewer to have to raise.
- **The placement claim is corrected.** It previously said `pricing` is additive "like the service metadata above". It is additive in the same sense, but that section places `serviceName`, `tags` and `iconUrl` on the top-level `resource` object, while `pricing` sits on the extension object. The difference is now stated and justified: `resource` carries what the payment path already carries, the extension carries what only catalogs consume.
- **A second example.** A `per-call` listing that omits both `unit` and `typical` — the fallback case, and the first one a catalog implementer meets. It is the case the optional fields exist to make expressible.

The proposal issue is #3264, linked from the description.

Two notes on CI, so nobody spends time on them: `Vercel` fails because fork PRs cannot clear the preview-deploy authorization — #3235 merged on 2026-08-23 carrying the identical failure — and `check-verified-commits` is green.

Related, for whoever picks this up: #3226 is working through the adjacent problem that Bazaar counters can be minted by verify-only calls, because the extension's *Facilitator Behavior* section never mentions settlement. A count that can be minted and a count whose amounts are incomparable are the same failure seen from two sides; I left an implementation data point there.

*Drafted with AI assistance under my direction, then reviewed and verified by hand, per CONTRIBUTING.md.*

PR #3183, opened 2026-08-17
feat(svm): allow injecting an RPC client into the upto facilitator

notorious-d-e-v:
## Motivation

The SVM `upto` facilitator builds its own RPC clients internally from `rpcUrl` (`createRpcClient` in `scheme.ts` and `rentCleanupManager.ts`). A facilitator operator has no way to route the scheme's channel claim/cleanup `sendTransaction` calls through their own transport — which matters when the operator paces sends to respect an RPC provider's `sendTransaction`-per-second cap, adds rate-limit retry, or instruments RPC usage. Today those sends bypass any operator-side client entirely.

## Change

- `UptoSvmFacilitatorConfig.rpc?` — an optional pre-built client, preferred over constructing one from `rpcUrl` at the scheme's three build sites.
- `UptoSvmRentCleanupManagerConfig.rpc?` — same for the rent cleanup manager; the scheme threads its configured client through.
- No behavior change when the field is omitted; `rpcUrl` continues to work as before.

~15 lines, additive-only. Changeset included (`@x402/svm`: minor).

## Testing

- `pnpm turbo run build --filter=@x402/svm...` clean
- `@x402/svm` test suite: 365/365 passing

Downstream we run this in production as a vendored build, injecting a client that paces and retries `sendTransaction` against a provider rate limit while leaving simulation/confirmation behavior untouched.

### DIFF
--- a/typescript/.changeset/upto-facilitator-rpc-injection.md
+++ b/typescript/.changeset/upto-facilitator-rpc-injection.md
@@ -1,0 +1,5 @@
+---
+"@x402/svm": minor
+---
+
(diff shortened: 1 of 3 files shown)


phdargen (approved):
Thanks @notorious-d-e-v, LGTM 🚀

PR #3186, opened 2026-08-17
feat(extensions): bind delivered response to offer-receipt (proof-of-delivery)

PatrickPi1312:
# feat(extensions): bind delivered response to offer-receipt (proof-of-delivery)

Closes discussion in #2833.

## Problem

The `offer-receipt` receipt (`ReceiptPayload`: `version, network, resourceUrl, payer, issuedAt, transaction?`) proves a payment was made for a `resourceUrl` — but not **what was delivered**. There is no binding to the response content, so a receipt can't serve as proof-of-delivery or dispute evidence about the payload. That gap is exactly what #2833 ("delivery-receipt attestation") asks for.

## Change

An optional, backward-compatible **version-2** receipt that binds a hash of the delivered response body:

| Field | Meaning |
| --- | --- |
| `responseHash` | `0x`-prefixed SHA-256 of the response body |
| `responseHashAlg` | `"sha256"` |
| `responseHashEncoding` | `"raw"` (exact bytes) or `"jcs"` (RFC 8785 canonical form) |

- **v1 is untouched.** Fields appear only when a response is bound; `payload.version` selects the EIP-712 types (`RECEIPT_TYPES_V2`), consistent with §5.5. Domain is unchanged.
- **`"jcs"` encoding** reuses the package's existing RFC 8785 `canonicalize()`, so the receipt survives parse→re-serialize (the normal agent path). Byte `"raw"` is also supported.
- **Opt-in** via `includeResponseDigest` on the server extension, mirroring `includeTxHash`. Off by default; if no body is available at settlement, a v1 receipt is issued rather than failing.
- **Two verification tiers** documented (§5.6): *attested* (recompute hash over the body) and *recomputable* (for deterministic services, re-execute and check — zero trust in the issuer).

## Files

- `specs/extensions/extension-offer-and-receipt.md` — §5.2.1 (v2 fields), §5.3.1 (v2 EIP-712 types), §5.5 (verification steps), §5.6 (delivery binding + tiers)
- `src/offer-receipt/{types,signing,server,index}.ts` — types, `hashResponseBody()`, version-aware sign/verify, issuer + server wiring, exports
- `test/delivery-receipt.test.ts` — EIP-712 + JWS roundtrip, attested-vs-tampered, jcs order-independence, v1 unchanged

## Verification

- New vitest suite added (EIP-712 + JWS).
- The cryptographic core (EIP-712 v2 sign→recover, JCS digest determinism, tamper detection, raw≠jcs) was validated independently with viem — 6/6.
- Note: the repo's `engines.node >= 22` prevented running the full workspace suite in my environment (Node 20); CI covers it. `prettier --write` may be needed for final formatting.

## Motivation / disclosure

Written from a live x402 deployment that already emits signed, offline-verifiable delivery receipts across Base, Arbitrum, Polygon and Solana (a reproducible example is in #2833). Drafted with AI assistance and reviewed by hand, per CONTRIBUTING.


0rkz:
Disclosure: I work on PayPerByte/BYTEDev — our production x402 feeds attach an EIP-712 PayloadAttestation over the exact bytes served, so this review is from an implementer of a byte-exact variant of the same pattern.

One spec-level concern with §5.6's `Servers issuing JSON SHOULD prefer "jcs"`: RFC 8785 §3.1 already requires its input to be I-JSON (RFC 7493) — numbers expressible as IEEE-754 doubles — but §5.6's SHOULD is unconditional, and nothing in the pipeline detects a payload that isn't. A body carrying a u64 (wei amounts, token ids, large counters) loses precision at JSON.parse before canonicalize() ever runs — a JS issuer and a bigint-aware verifier (Python/Rust/Go) can compute different JCS forms of the "same" delivered JSON, and verification fails with nothing in the receipt signalling why. Suggest scoping the jcs SHOULD to I-JSON payloads (large values as strings, per RFC 8785's own guidance). Same point I raised on #3304 (item 11); it lands here too because §5.6 is where the JCS recommendation becomes normative. We pinned our spec v1 to exact-bytes over a neighbouring case — ES6-formatting-sensitive floats in our oracle payloads: https://www.payperbyte.io/attestation/eip712-payload-attestation/v1 (the EIP-712 domain there is a frozen signing namespace, not a settlement rail; payments settle on Base)

Smaller point on `"raw"`: worth stating whether the hash covers identity-encoded bytes or post-Content-Encoding bytes — a proxy adding gzip/br silently forks the digest for later verifiers. Also worth a look before merge: `hashResponseBody` canonicalizes a JSON value even when `encoding` is `"raw"` (the final `else` in signing.ts), so a receipt can carry `responseHashEncoding: "raw"` over JCS bytes — and the raw-vs-jcs test passes a string, so that branch isn't exercised.

Selecting RECEIPT_TYPES_V2 by payload.version while keeping the domain unchanged matches what has worked for us in production — +1 on that choice.

Prepared with AI assistance, human-reviewed.

PR #3197, opened 2026-08-18
Auth-capture spec update: v1.1

phdargen:
## Summary

`auth-capture` lets a client authorize a **maximum** amount, then hold, finalize for less, cancel, or refund later. This update aligns the scheme with the protocol **payment flow** model and specifies how later, server-initiated operations are authenticated and who is allowed to submit them.

Previously the EVM binding selected two-phase vs single-shot with `extra.autoCapture`, and later `capture` / `void` / `refund` sat outside the request as async, out-of-band work. There was also no binding of a **receiver authorizer** into the payment identity, so a facilitator-relayed lifecycle call could not be proven to be the server’s.

**`autoCapture` is removed.** `extra.paymentFlow` is the only flow selector. The unbound collect path (no `receiverAuthorizer` / `policy`) stays byte-compatible with v1.0 clients.

### Payment flows

**Escrow** (scheme default) — hold first, finalize after the resource:

- **Sync / stateless:** `settle(authorize)` → work → `settle(capture` or `void)` → response
- **Async / stateful:** `settle(authorize)` → work → commit `paymentInfo` (and `saltNonce` when bound) to durable storage → response; later `settle(capture` / `void` / `refund)` out of band

**Authorization** — no hold:

- `verify` → work → `settle(charge)` → response (sync, stateless); `refund` later if needed

The client still signs once. Which collect call that signature becomes (`authorize` vs `charge`) follows from the flow. Later operations are ordinary `POST /settle` calls with `payload.type` of `"capture"`, `"void"`, or `"refund"` — no new facilitator endpoint.

On EVM, a non-zero `extra.receiverAuthorizer` is hashed into `PaymentInfo.salt` (with an optional `policy` and a client `saltNonce`). That is what authenticates server-initiated lifecycle: a collected payment cannot be re-pointed at a different authorizer after the fact.

### Operator types (`extra.operatorType`)

These label the trust model and what the facilitator will relay. `extra.captureAuthorizer` is always the onchain `PaymentInfo.operator`.

1. **`delegated` (default)** — operator is an address the facilitator controls and submits from (EOA or smart-contract account). With a non-zero `receiverAuthorizer`, the facilitator relays collect **and** lifecycle (`charge` / `capture` / `void` / `refund`), gated by an EIP-712 authorizer signature. That signature is **not** checked onchain; the server is trusting the facilitator to submit exactly what was signed. Client-side bounds still hold: the signed maximum, fee range, and `reclaim` after the capture deadline. Without a receiver authorizer, relay is collect-only.

2. **`custom`** — operator is an untrusted smart contract with permissionless `authorize` / `charge` wrappers. The facilitator relays only that collect call (official x402 SDK). `capture` / `void` / `refund` stay async and out of band, through the operator’s own ABI and SDKs. Custom lifecycle and dispute logic can evolve at the periphery without changing the escrow, facilitator, or scheme.

3. **`policy` (reserved)** — wire fields (`operatorType`, `policy`, salt binding) are specified now so a later canonical operator can wrap escrow calls with an onchain `receiverAuthorizer` check and read-only policy hooks. That is meant to close the gap between (1) and (2): lifecycle stays facilitator-relayed and gasless for the server, without trusting the facilitator as the only gate.

Facilitators advertise admitted operators on `/supported`. Relaying into `"custom"` requires gas caps and outcome checks (escrow events, `paymentState`, token movements), not just a successful top-level call.

### Compatibility

- v1.0 clients remain payable when `receiverAuthorizer` and `policy` are omitted (unbound salt).
- `autoCapture` is deprecated
- Bound payments add `saltNonce`; servers that still serve v1.0 clients may nonce-probe as specified.


phdargen:
cc @A1igator

CarsonRoscoe (approved):
This LGTM!

Relatively backwards compatible with existing `auth-capture` servers, light migration to fully utilize this, and forwards compatible with the the work needed to make this a true x402 spec for clients/servers/facilitators. Very happy personally

Has my approval, though I would hold off on merging to give @A1igator some time to review

A1igator:
Hey @phdargen!

Thanks for this and this is definitely the correct direction. I'm obviously happy with 'custom' existing. Just fyi though 'policy' will never be as expressive as 'custom' as written but I assume that is intentional. LGTM though.

phdargen:
> Hey @phdargen!
> 
> Thanks for this and this is definitely the correct direction. I'm obviously happy with 'custom' existing. Just fyi though 'policy' will never be as expressive as 'custom' as written but I assume that is intentional. LGTM though.

Yes being less expressive than arbitrary code is the intended security model

Issue #3182, opened 2026-08-17
[Feature]: Revocable activation-delay flow for reversible pre-fulfillment orders


zjzJoez:
## Problem or use case

The current x402 payment flows order `/verify`, `/settle`, and the resource handler within one request. There is a narrower commerce case that none of the current orderings expresses:

1. a client signs a valid payment authorization;
2. the resource server durably accepts only a contingent order or reservation;
3. no paid resource or fulfillment is released;
4. the payer has a short interval to invalidate that authorization; and
5. settlement runs after the interval, with fulfillment enabled only after settlement finality.

The intended use case is recovery from accidental, duplicate, or stale-context agent purchases where the merchant can cheaply hold a reversible order. It is not intended for instant API responses, content, secrets, or any resource with standalone value before settlement.

## Candidate mechanism

For compatible EVM assets, EIP-3009 already exposes the required ledger primitives:

```text
validAfter  = clientNow + activationDelay
cancelBy    = validAfter - cancellationSafetyMargin
validBefore = validAfter + settlementRunway
```

- `validAfter` makes transfer invalid until the activation time.
- payer-signed `cancelAuthorization(authorizer, nonce)` permanently consumes an unused authorization.
- the cancellation transaction can be submitted by a third party, although payer gaslessness depends on an available funded relay.
- after `validAfter`, cancellation and settlement race. Only ledger ordering is authoritative.

A candidate lifecycle is:

```text
verify future authorization
  -> durably persist contingent commitment and scheduled work
  -> respond pending
  -> settle after activation
  -> fulfill after finality
```

The facilitator `/settle` endpoint can remain synchronous. A durable resource-server coordinator would own the timer, pending state, retries, restart recovery, cancellation arbitration, and chain reconciliation.

This requires a flow-aware EIP-3009 verification path because the current implementation correctly rejects future `validAfter` values for the existing immediate flow.

## Evidence

We implemented and tested the token-level mechanism against deployed XSGD on Avalanche C-Chain:

- repository and reproducible proof script: https://github.com/zjzJoez/grace-x402/blob/main/grace/prove.mjs
- deployed asset: https://snowtrace.io/token/0xb2F85b7AB3c2b6f62DF06dE6aE7D09c010a5096E
- payer-signed cancellation broadcast by another account: https://snowtrace.io/tx/0xd5bab3abf1cf09e8ff67d94d85f0c6fabdee47aa4d16cf6196622863f7709cdd
[...]

PR #3203, opened 2026-08-18
feat(ts): auth capture sdk v1.1

phdargen:
## Description

Continuation of https://github.com/x402-foundation/x402/pull/2308 implementing the v1.1 spec proposed in https://github.com/x402-foundation/x402/pull/3197

## Checklist


CarsonRoscoe (commented):
One comment, might just be a documentation piece; There's a risk with the allowlist where upgradeable/proxy operator contracts could undermine the model. If the allowlist admits an address that was honest and later upgrades to a dishonest implementation, it effectively bypassed the allowlist.

We should document that facilitators' should prefer immutable operator contracts

CarsonRoscoe (commented):
Nothing stops a merchant from registering a route with `operatorType: "custom"` while leacving `captureMode` at its default (`"sync"`), or leaving cancel-triggerred auto-void wired. The resource server will still build and submit an automatic capture after settle or void on cancel, but the facilitator unconditionally rejects lifecycle relay for `operatorType: "custom"` (See `ErrLifecycleNotRelayed`).

This would lead to the initial `authorize` succeeding and escrowing the payer's funds, and the automatic follow-up capture/void always failing silently, leaving funds stuck in escrow with no way to unwind unless the merchant happens to know to force `captureMode: "deferred"`.

While this isn't a theft risk, its a big UX/funds availability trap we should avoid.

I'd suggest adding to the spec that servers MUST not set this combination, and at the server SDK level, fail fast when `operatorType: "custom"` is combined with anything other than `captureMode: "deferred"`, and not auto-stamping `receiverAuthorizer` onto custom-operator routes.

CarsonRoscoe (commented):
nit: `verify()` for custom operators runs a ~7-9-call `eth_simulateV1` batch vs the delegated's single `eth_call`. This is a much higher RPC cost per request, so spam/DoS cost to a facilitator is higher specifically for custom-operator `verify()` calls. Worth documenting

CarsonRoscoe (commented):
In the facilitators' `lifestyle.ts` , if a capture succeeds on-chain but the trailing voil call fails for a mundane reason like RPC timeout (any error that isn't `ZeroAuthorization` or revert), the function returns `success: false` while discarding the `txHash`. imo this should be preserved the transaction regardless of the void outcome

CarsonRoscoe (commented):
`settlementHooks.ts`'s `persistCollect` unconditionally overwrites storage without checking if the record exists. I'll let you decide what we should do in that case, but calling out as-is  a duplicate or retried settlement for the same `paymentInfoHash` would reset `capturableAmount`/`refundableAmount`

CarsonRoscoe (commented):
I believe there's a ERC-6492 signature verify/settle mismatch in `facilitator/connect.ts` and `utils.ts`. `verifyCollect` unwraps and verifies the inner 6492 signature, but `unpackForSettle` also passes the original wrapped signature bytes as `colelctorData` for the real on-chain call. Circle's `SignatureChecker` / `Permit2`'s verifier don't understand the 6492 wrapper, so any client submitting a 6492-wrapped signature will pass `verify()` and then revert on `settle()`

Similarly, there's no counterfactual/undeployed wallet path. If we're going straight to the facilitator implementation, this should be considered.

phdargen:
Thanks — addressed review comments 4-6.

### [Capture succeeds, trailing void fails](https://github.com/x402-foundation/x402/pull/3203#pullrequestreview-4996807348)

Real bug, fixed in `auth-capture/facilitator/lifecycle.ts`. `settleLifecycle` returned `settleResult(voidSubmitted, ...)` on a non-revert void error, and because that branch only preserves `txHash` for `error === "reverted"`, an RPC or receipt timeout produced `success: false` with `transaction: ""` — reporting failure for a capture that had already moved funds.

Went further than just preserving the hash: the trailing void is now unconditionally best effort, so capture success is settle success whatever the void does. Reporting `success: false` after a successful capture is the worse failure mode, since the resource server would withhold the resource from a payer who was already charged, and the branch already behaved this way when the void reverted or hit `ZeroAuthorization`. A remainder left behind stays voidable until `authorizationExpiry`, after which the payer can `reclaim`. This also drops the `/ZeroAuthorization|revert/i` match on error message text.

### [`persistCollect` overwrites unconditionally](https://github.com/x402-foundation/x402/pull/3203#pullrequestreview-4996824152)

Changed to first-write-wins (`current ?? record`), which also makes `update` report `"unchanged"` and matches the already-guarded writes elsewhere in the file.

Worth noting it is latent rather than reachable today: `persistCollect` fires at most once per payment, since escrow settles the collect at `before-handler` and authorization at `after-handler` (`PAYMENT_FLOWS`), with the escrow `after-handler` branch gated on `paymentFlow === "authorization"`. A replay cannot reach it either, because `hasCollectedPayment` makes a second successful collect for the same `paymentInfoHash` impossible onchain. Fixed anyway — on any retry the stored balances are the authoritative ones, as out-of-band `capture`/`refund` or a cancel void may already have moved them.

### [ERC-6492](https://github.com/x402-foundation/x402/pull/3203#pullrequestreview-4996850740)

Two halves here, and they landed differently.

**Forwarding the wrapped signature as `collectorData` is correct, not a mismatch.** The canonical collectors never hand `collectorData` straight to `SignatureChecker` / Permit2 — they run it through `ERC6492SignatureHandler` first:

```solidity
// ERC3009PaymentCollector._collectTokens
IERC3009(token).receiveWithAuthorization({
    ...
    signature: _handleERC6492Signature(collectorData)
});
```

`Permit2PaymentCollector._collectTokens` does the same. `_handleERC6492Signature` checks the magic suffix, `abi.decode`s the envelope, executes the preparation call through Multicall3, and returns only the inner signature. So the wrapper is stripped onchain, and unwrapping before submitting would drop the deployment step an undeployed wallet depends on — which is exactly what makes the counterfactual case work at all. There is already a test for this (`should forward raw ERC-6492 collectorData while verifying its inner signature`); I added a comment on `unpackForSettle` and spec/README wording so the next reader does not have to re-derive it.

**No counterfactual path was a real gap, now implemented.** `verifyCollect` unwrapped and verified the inner signature, but for an undeployed payer `getCode` returns `0x`, so it took the ECDSA branch, a smart-wallet inner signature failed the 65-byte length check, and verify rejected with `invalid_auth_capture_evm_signature` before the simulation ever ran. Since the spec requires accepting all three signature forms, that was a conformance gap too.

`verifyCollect` now classifies the payer with the shared `classifyErc6492Payer` and, when counterfactual, defers the signature check to the onchain simulation, which is unconditional here and complete: `eth_call` of `authorize`/`charge` for `"delegated"` (the collector deploys the wallet, then the token validates), and `eth_simulateV1` with event, `paymentState` and token-delta assertions for `"custom"`. Deployed wallets, wrapped signature or not, keep the existing strict ERC-1271 path.

Gated on a new `eip6492AllowedFactories` facilitator config, mirroring the `exact` scheme, with a new `invalid_auth_capture_evm_erc6492_factory_not_allowed` reason and an omitted list rejecting every counterfactual payer. One difference from `exact` worth flagging: there the facilitator sends the deployment transaction itself, so the allowlist prevents arbitrary transaction injection. Here Multicall3 is the sender of the preparation call, so the allowlist is a gas-griefing control rather than an authority control. Kept it anyway for explicit opt-in and consistency. No settle-side gate is needed since `settleCollect` re-runs `verifyCollect`.

Tests: 9 added across `facilitator.test.ts` and `server.test.ts`, verified to fail against the pre-fix source. Comments 1-3 went out in the previous push.

zjzJoez:
Reading #3197 while working against this — one boundary question. The answer settles an open question on a proposal of mine, not anything in this PR.

v1.1 moves in two directions at once and I can't tell which is the rule.

**Toward the enum:** `extra.autoCapture` is removed and `paymentFlow` is "the only flow selector", with `autoCapture: true` a hard reject rather than a fall-through to `escrow`.

**Toward `extra`:** `captureMode: "deferred"` — *"skips that settle so the server captures later from durable state"* — but narrowly: escrow-only, and it MUST NOT be set when `paymentFlow` is `"authorization"`. The chain-agnostic `scheme_auth_capture.md` declines to name a mode at all — "Sync versus async is a server choice; the wire format does not name a mode" (L48), and `captureMode` appears nowhere in that file.

Read together that looks like a rule: **anything that changes which phases run belongs in the enum, and `extra` only modulates a phase the enum has already selected.** Is that the boundary you intended?

Asking because #3182 leaves exactly this open — its first protocol question is whether response-before-settlement is a fourth core payment flow, an EVM-specific binding, a generic asynchronous lifecycle, or only an application pattern. Under the rule above it's the first, since the delay changes whether a settle can run at all rather than when a selected one finalizes. I'd rather you drew that line than that I picked the reading that suits me.

One data point in case it makes the question cheaper than it sounds: on the permit2 binding the primitive is already merged and deployed. `Witness(address to,uint256 validAfter)` is signature-committed (`specs/schemes/exact/scheme_exact_evm.md` L380, L383) and `_settleInternal` enforces `require(block.timestamp >= witness.validAfter)` (L443) at `0x402085c248EeA27D92E8b30b2C58ed07f9E20001`. The only thing rejecting a future value on that path is one parenthetical in verifier step 5: "Verify the `deadline` (not expired) and `witness.validAfter` (active)" (L191).

shunhe-wang:
I noticed one narrow v1.1 fixture mismatch that may be worth correcting before the broader draft lands.

`test/contracts/auth-capture/ForwardingOperator.sol` targets the v1.1 escrow, but its `charge` wrapper still accepts `uint16 feeBps` and forwards the v1.0 selector.

The current live custom-operator integration path exercises `authorize`, so that selector mismatch is not caught onchain.

I prepared a small local patch that changes the wrapper to `uint256 feeAmount` and adds a Base Sepolia integration case for custom `charge`, including opaque `collectorData`, canonical `PaymentCharged`, exact state and token deltas, and zero facilitator-funded value movement. Would you prefer that as a focused contribution against this branch, or as a commit for you to cherry-pick while the draft is being rebased?

On another note, `/supported.extra.operators` seems to be the main way merchants find out which custom operators a facilitator supports. We're interested in being a custom operator, so it'd be helpful to know how an operator gets added, etc.

Issue #3010, opened 2026-07-31
[Proposal] Allowance-funded (approved-transfer) payments in the Hedera exact scheme


dmitrijtitarenko:
**Package:** `@x402/hedera` (`typescript/packages/mechanisms/hedera`)

## Summary

The Hedera `exact` scheme funds a payment from the **payer's own balance**, which makes the common
**two-key** pattern impossible to realise: an *owner* that holds funds and grants a capped on-chain
**allowance** to an *agent* whose key lives in the daemon. I'd like to propose an opt-in
**approved-transfer** mode so payments can draw from the owner via that allowance (capped, revocable),
and I've verified the mechanics + pinned the exact blocker on testnet. Happy to open the PR if the
approach sounds right.

## Where it funds from today

```ts
// mechanisms/hedera/src/signer.ts — createClientHederaSigner
tx.addHbarTransfer(parsedAccountId /* payer */, Hbar.fromTinybars((-amount)));
tx.addHbarTransfer(payTo, Hbar.fromTinybars(amount));
```

## On-chain findings (Hedera testnet)

Hedera attributes an **approved-transfer** spend to the transaction's **fee-payer**, not to arbitrary
signers:

| Experiment | Fee-payer | Result |
|---|---|---|
| Approved transfer from owner, submitted **directly by the agent** (agent = fee-payer + spender) | agent | **`SUCCESS`** — Mirror: owner `-amount, is_approval=true`, agent charged only gas. (`0.0.9693096@1785511093.849251631`) |
| Same approved transfer through the **facilitator** (scheme sets fee-payer = facilitator) | facilitator | **`SPENDER_DOES_NOT_HAVE_ALLOWANCE`** — reached consensus and failed; the facilitator has no allowance from the owner. (`0.0.7162784@1785510688.236035028`) |

So the allowance mechanism works — the owner *can* fund the payment — **but only when the spender
(agent) is the fee-payer.** The scheme always makes the facilitator the fee-payer.

## Proposed approach (opt-in, backward compatible)

1. **Client scheme** — an `allowanceOwner` option: build `addApprovedHbarTransfer(allowanceOwner, -amount)`
   (and the token equivalent) and set the transaction payer to the **agent** (the spender).
2. **Payer inference / `verifyPayerSignature`** — the facilitator infers the payer from the debited
   account; for an approved transfer that's the *owner*, but the signer/spender is the *agent*. Payer
   inference needs to recognise `is_approval` and verify against the **spender**.
[...]

PR #3205, opened 2026-08-19
spec(exact-hedera): add transferExecutor asset transfer method

kierzniak:
Adds a second asset transfer method, `transferExecutor`, to `specs/schemes/exact/scheme_exact_hedera.md`,
plus a Hedera entry in `scheme_exact.md` "Critical Validation Requirements". Today's `{ transaction }`
payload becomes the explicit `cryptoTransfer` method and stays the default; nothing changes for
existing clients or facilitators.

Spec only. Implementation for `typescript/packages/mechanisms/hedera/` follows in a separate PR once
the spec is aligned.

### Why

`cryptoTransfer` requires that the key signing the payload controls the debited funds. A Client whose
funds are controlled by a smart contract cannot meet that requirement. This covers contract accounts
(smart accounts, multisig, treasury and vault contracts) and accounts that have delegated spending to
a contract through a HIP-336 allowance (delegation, session-key, spend-policy, subscription and
agent-wallet contracts). Such payers cannot sign a `TransferTransaction` for the funds in question:
the contract's logic moves them, and the transaction that triggers it is a
`ContractExecuteTransaction`.

A facilitator that only inspects and co-signs transfer lists cannot verify or sponsor such a payment,
so this class of payers is excluded from Hedera x402 today.

Other chains already cover this in `exact`: EVM `erc7710` (facilitator calls a delegation manager for
smart accounts), SVM smart-wallet path (transfer emitted via CPI, verified by outcome), Starknet
(facilitator executes the payer's account contract), NEAR `intents-verifier` (#2948, second ATM
settled through a contract). Hedera has no equivalent.

`upto` on Hedera (#2919) is contract-routed but uses a facilitator-designated proxy for a spend
ceiling; here the payer's own contract is the executor and the amount is exact. Same actors as
#3010, without its approved-transfer attribution problem: the contract is the allowance spender.

### Working proof

Implemented end to end on `hedera:testnet`: a facilitator (verify/settle), a resource server, and an
executor contract that pays from an owner-granted HIP-336 allowance under an EIP-712 policy.

Settled example: [`0.0.8065254-1787055536-598669142`](https://hashscan.io/testnet/transaction/0.0.8065254-1787055536-598669142).
Its mirror-node record shows exactly why Phase 4 reads child records:

```
CONTRACTCALL    SUCCESS  parent  1787055542.254966905
CRYPTOTRANSFER  SUCCESS  child   1787055542.254966906
```

The parent carries no transfer at all; the payment is in the child. A proof built from the parent
alone would reject every valid payment, and the `SUCCESS` receipt on its own says nothing about
whether funds moved.

Balances reconciled on that run: payee +1 HBAR, payer -1 HBAR, fee payer -fee only.

### Key design decisions

- **ERC-7710 shape.** Client declares `executor` and an opaque `authorization`; the facilitator
  builds `executeTransfer(from, asset, to, amount, authorization)` itself, so it can never resize or
  redirect the payment. No trusted executor list needed for correctness; resource servers and
  facilitators MAY allowlist as local policy.
- **Fixed `ITransferExecutor` interface, revert on failure.** One function to simulate and call; no
  return-value sniffing.
- **Verification entirely by simulation.** Same as `erc7710`.
- **Proof from parent + child records.** Contract-initiated transfers live in child records; receipt
  status is not evidence of payment (SVM §3.4 spirit, same direction as #3061). Hedera has no
  ERC-20 `Transfer` event for HTS to lean on.
- **`extra.feePayer` unused.** The client builds no Hedera transaction; the facilitator submits from
  its own configured account.
- **Backwards compatible.** Absent `assetTransferMethod` == `cryptoTransfer`; unknown values are
  rejected, never downgraded.

### DIFF
--- a/specs/schemes/exact/scheme_exact_hedera.md
+++ b/specs/schemes/exact/scheme_exact_hedera.md
@@ -8,6 +8,48 @@ This scheme facilitates payments of a specific amount of a Hedera HTS fungible t
 
 `exact`
 
+## Asset Transfer Methods
+
+The `exact` scheme on Hedera executes a transfer where the Facilitator pays the network fee, but the
+Client controls the exact flow of funds. This is implemented via one of two asset transfer methods,
+depending on who controls the payer's funds:
+
+| AssetTransferMethod | Mechanism | Payer universe | Recommendation |
+| --- | --- | --- | --- |
+| **1. `cryptoTransfer`** (default) | Client signs a `TransferTransaction`; facilitator co-signs as fee payer | Accounts whose key controls the funds | **Recommended** for key-controlled accounts (simplest, no contract) |
+| **2. `transferExecutor`** | Client supplies an authorization for a contract implementing `ITransferExecutor`; facilitator calls `executeTransfer` as fee payer | Funds held by a contract, or spendable only through a contract holding a HIP-336 allowance (smart accounts, multisig / vault contracts, delegation and spend-policy contracts) | **Contract account option** |
+
+If no `assetTransferMethod` is specified in `PaymentRequirements.extra`, clients and facilitators
+MUST behave as `cryptoTransfer`. Payment payloads that use a non-default transfer method MUST echo
+the selected `assetTransferMethod` in `accepted.extra`. A facilitator MUST reject a value it does not
+implement (`invalid_exact_hedera_unsupported_asset_transfer_method`) and MUST NOT fall back to
+another method.
+
+In all cases, the Facilitator cannot modify the amount or destination. It serves only as the
+transaction broadcaster and fee sponsor.
+
+The sections from `Protocol Flow` through `Facilitator Verification Rules` below define the default
+`cryptoTransfer` method (rule 1, "MUST be a `TransferTransaction` directly", applies to that method
+only). The `transferExecutor` method is defined in its own section.
+
+Facilitators SHOULD advertise the methods they implement per network in `/supported`:
+
+```json
+{
+  "kinds": [
+    {
+      "x402Version": 2,
+      "scheme": "exact",
+      "network": "hedera:testnet",
+      "extra": {
+        "feePayer": "0.0.9000",
+        "assetTransferMethods": ["cryptoTransfer", "transferExecutor"]
+      }
+    }
+  ]
+}
+```
+
 ## Protocol Flow
 
 The protocol flow for `exact` on Hedera is client-driven.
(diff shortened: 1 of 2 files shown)


piotrswierzy:
### 1. Phase 3 shouldn't mandate simulation as the only verification path

Rule 5 makes `ContractCallLocal` / `eth_call` simulation a hard MUST. `erc7710` requires simulation *because* it assumes the facilitator cannot interpret an opaque payer contract — simulation is the price of opacity. That assumption doesn't hold uniformly: a facilitator that understands a specific executor's authorization format can check the same preconditions (signature validity, policy/allowlist, caps, nonce freshness, balance and HIP-336 allowance, token association/freeze/KYC status) with view reads and off-chain computation, and get a stronger answer without depending on the relay simulating a state-changing call through the HTS system contract.

Suggested wording for rule 5: the facilitator MUST establish, **by simulation or by equivalent checks against an executor whose authorization format it understands**, that the authorization permits exactly this transfer and that the call will succeed. Simulation stays the default and the only option for the opaque case, which is most of them.

### 2. Settlement rule 3 makes custom-fee tokens unpayable

"No account other than `payTo` has a positive net transfer in `asset`" and "the only debited accounts are `payer` and the fee payer" both break for HTS tokens with fixed or fractional custom fees or royalties: fee collectors show positive net changes and the payer is debited more than `amount`. As written, such assets can never settle, and the failure mode is a spurious `settlement_unexpected_debit` *after* the facilitator has already paid the fee. Either carve out the token's configured custom-fee collectors explicitly, or state the restriction as a prerequisite so implementers reject those assets up front rather than at settlement.

### 3. The invariant table conflates prevention with detection

I1 lists "fee payer is never debited beyond the network fee" as enforced by settlement rule 3, but rule 3 only *detects* after consensus. Against a hostile or buggy executor the fee payer's actual protection is the explicit gas limit; the residual exposure is that the facilitator pays a fee for a payment it must then report as failed.

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +208,380 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
...
+
+If the simulation succeeds, the payment is considered valid. A passing simulation MUST NOT be
+reported as settlement and MUST NOT cause the resource server to release the resource; state can
+change between `/verify` and `/settle`, and the settlement proof (Phase 4) is the only evidence of
+payment.
+
+The facilitator SHOULD read state that gates the call (nonces, allowances) from a consensus or
+mirror node rather than a JSON-RPC relay, which may serve slightly stale state, and SHOULD track
+the hash of `(network, executor, payer, authorization)` to reject a payload already settled or in
+flight. On-chain replay protection is the executor's responsibility.
```
skurzyp-blockydevs:
```
SHOULD track the hash of `(network, executor, payer, authorization)` to reject a payload already settled
```

elsewhere the spec explicitly says an authorization can be multi-use (e.g. a standing spend policy that's valid for many payments, like a subscription or an agent's ongoing budget). If you block repeats of the same authorization, you'd block the second and all subsequent legitimate payments under that policy.

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +208,380 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
...
+policy.
+
+### Executor interface
+
+An executor MUST implement:
+
+```solidity
+interface ITransferExecutor {
+    /// Moves exactly `amount` of `asset` from `from` to `to` if, and only if,
+    /// `authorization` permits it. MUST revert otherwise. MUST NOT return a value.
```
skurzyp-blockydevs: "MUST NOT return a value" isn't actually checkable. The interface doc comment says implementations must not return a value, but Solidity function selectors (the 4-byte identifier used to route calls) don't encode return types at all — a contract could return a bool and still match the interface perfectly. So this requirement can't actually be enforced or verified by anyone. The fix is just to reword it as "any returned data is ignored" rather than pretending it's a hard constraint.

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +208,380 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
...
+    ) external;
+}
+```
+
+Requirements on an executor:
+
+- It MUST bind `authorization` to `from`, `asset`, `to` and `amount` (or to values that imply them),
+  so a facilitator cannot reuse an authorization for a different transfer, and MUST revert on any
+  mismatch. Whether an authorization is single-use or multi-use is the executor's choice.
+- It MUST revert, not return, on failure. A return value is ignored; a `SUCCESS` receipt with no
```
skurzyp-blockydevs: here it says that the return value is ignored

skurzyp-blockydevs (changes requested):

kierzniak:
> ### 1. Phase 3 shouldn't mandate simulation as the only verification path
> Rule 5 makes `ContractCallLocal` / `eth_call` simulation a hard MUST. `erc7710` requires simulation _because_ it assumes the facilitator cannot interpret an opaque payer contract — simulation is the price of opacity. That assumption doesn't hold uniformly: a facilitator that understands a specific executor's authorization format can check the same preconditions (signature validity, policy/allowlist, caps, nonce freshness, balance and HIP-336 allowance, token association/freeze/KYC status) with view reads and off-chain computation, and get a stronger answer without depending on the relay simulating a state-changing call through the HTS system contract.
> 
> Suggested wording for rule 5: the facilitator MUST establish, **by simulation or by equivalent checks against an executor whose authorization format it understands**, that the authorization permits exactly this transfer and that the call will succeed. Simulation stays the default and the only option for the opaque case, which is most of them.
> 
> ### 2. Settlement rule 3 makes custom-fee tokens unpayable
> "No account other than `payTo` has a positive net transfer in `asset`" and "the only debited accounts are `payer` and the fee payer" both break for HTS tokens with fixed or fractional custom fees or royalties: fee collectors show positive net changes and the payer is debited more than `amount`. As written, such assets can never settle, and the failure mode is a spurious `settlement_unexpected_debit` _after_ the facilitator has already paid the fee. Either carve out the token's configured custom-fee collectors explicitly, or state the restriction as a prerequisite so implementers reject those assets up front rather than at settlement.
> 
> ### 3. The invariant table conflates prevention with detection
> I1 lists "fee payer is never debited beyond the network fee" as enforced by settlement rule 3, but rule 3 only _detects_ after consensus. Against a hostile or buggy executor the fee payer's actual protection is the explicit gas limit; the residual exposure is that the facilitator pays a fee for a payment it must then report as failed.

**1. Simulation as the only path. Keeping the MUST.**

I think the important distinction here is that this specification is intended to define a **generic `assetTransferMethod` that can work with an arbitrary smart contract**, not only with executors whose authorization format is known to the facilitator.

If a facilitator has specific knowledge about a particular executor and knows how to independently validate its authorization format, policy, nonce handling, allowances, token state, etc., then I think that should be modeled as a **different `assetTransferMethod` with its own verification rules**.

Allowing the generic method to say “simulation or whatever equivalent checks the facilitator considers sufficient” makes the behavior facilitator-specific. Two facilitators could receive exactly the same payment payload and reach different conclusions because they interpret the authorization differently. That is precisely what the specification should avoid: interoperability should not depend on undocumented facilitator knowledge.

Simulation does two jobs in `exact`, and neither is to act as the security boundary:

1. **Viability / fee saving.** It avoids spending the facilitator's network fee on a transaction that is expected to revert.
2. **Authorization oracle for opaque contracts.** When the authorization is opaque to the facilitator, executing the call in simulation is the generic mechanism for determining whether that authorization permits this exact transfer.

So I would keep simulation mandatory for this generic smart-contract transfer method. A more specialized executor-aware flow can absolutely avoid simulation, but it should be specified separately with deterministic verification rules rather than making the generic method dependent on facilitator-specific interpretation.

**2. Custom-fee tokens. Agreed.**

Taking your second option. The carve-out does not work: if the fee collectors are excluded from the
checks then the payment succeeds while `payTo` is credited less than `amount`, so the payer and the
payee disagree about what was paid and `exact` stops meaning what it says. I would rather refuse the
asset than redefine the guarantee.

So a new prerequisite puts custom-fee assets out of scope, and verification reads the token's fee
schedule per payment (a `fee_schedule_key` can add fees later) and rejects with
`invalid_exact_hedera_custom_fee_asset`. `scheme_exact.md` covers both methods, MUST under
`transferExecutor` and SHOULD under `cryptoTransfer`, which has the same hole today because it
inspects a transaction body while fees are assessed at consensus; happy to make it MUST for both.
Out of scope is not impossible: doing this properly means letting the payee declare that it accepts
netting less, which I would rather specify separately.

**3. Prevention vs detection. Agreed.**

The invariant table now separates "Prevented by" from "Detected by". For I1 the prevention is
structural (the executor spends from `from`, the fee payer grants it no allowance, the gas limit
bounds the fee) and rule 3 is detection only. Security Consideration 1 rewritten the same way, and
the residual is now stated: a failed rule 3 costs the facilitator a fee it cannot recover.

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Agreed, this contradicts "single-use or multi-use is the executor's choice" above. Narrowed to an
in-flight guard, plus an explicit MUST NOT against rejecting a payload because an identical one
settled earlier.

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Agreed, a selector does not encode return types. Now reads "any returned data is ignored", and the
selector paragraph says so explicitly.

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Same defect, fixed in the same edit: failure is signalled by reverting, and returning `false` is
explicitly not a failure signal.

phdargen:
`payload.executor` is an arbitrary client-chosen contract the facilitator will sign and submit as a `ContractExecuteTransaction`. A hostile executor can do the required payment and, in the same call, run eg a DeFi exploit or touch a sanctioned account. The facilitator need not fund that activity or be its victim; its account is still `transactionId.accountId` on the parent `CONTRACTCALL`. Verify only checks revert, so that extra activity is not rejected up front.

Please harden along the lines of [auth-capture ](https://github.com/x402-foundation/x402/blob/main/specs/schemes/auth-capture/scheme_auth_capture_evm.md) operatorType: "custom" (same “relay into foreign code” case):
- Default-deny executor allowlist on /supported (wildcard optional, not recommended)
- Cap gas (MUST)
- Assert the outcome, not the absence of a revert (MUST): matching net token movements and that no token movement originates from a facilitator-controlled address; nested logs / child-transfer visibility required, and if the RPC cannot provide them the facilitator MUST NOT advertise the method. Same checks on the receipt; status alone is insufficient
- Admission is by address, not code: prefer immutable executors (admin key none); a proxy or admin key can change implementation after review, so re-assert every settle

kierzniak:
> `payload.executor` is an arbitrary client-chosen contract the facilitator will sign and submit as a `ContractExecuteTransaction`. A hostile executor can do the required payment and, in the same call, run eg a DeFi exploit or touch a sanctioned account. The facilitator need not fund that activity or be its victim; its account is still `transactionId.accountId` on the parent `CONTRACTCALL`. Verify only checks revert, so that extra activity is not rejected up front.
> 
> Please harden along the lines of [auth-capture ](https://github.com/x402-foundation/x402/blob/main/specs/schemes/auth-capture/scheme_auth_capture_evm.md) operatorType: "custom" (same “relay into foreign code” case):
> 
> * Default-deny executor allowlist on /supported (wildcard optional, not recommended)
> * Cap gas (MUST)
> * Assert the outcome, not the absence of a revert (MUST): matching net token movements and that no token movement originates from a facilitator-controlled address; nested logs / child-transfer visibility required, and if the RPC cannot provide them the facilitator MUST NOT advertise the method. Same checks on the receipt; status alone is insufficient
> * Admission is by address, not code: prefer immutable executors (admin key none); a proxy or admin key can change implementation after review, so re-assert every settle

@phdargen thanks, that makes sense from the facilitator's threat model.

Before we change the design though: `transferExecutor` was deliberately modeled on the existing [ERC-7710 transfer method](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md#3-assettransfermethod-erc-7710) in the exact EVM scheme.

As far as I can tell ERC-7710 has the same trust boundary. The client supplies the `delegationManager`, the facilitator calls that contract from its own account, and the manager can run extra logic around the facilitator-constructed `ERC20.transfer(payTo, amount)`. The current EVM spec says no trusted list of Delegation Manager implementations is required, and treats a successful simulation as enough for verification.

`transferExecutor` follows that model. The facilitator constructs the exact `executeTransfer(from, asset, to, amount, authorization)` call, so the client cannot change the recipient or the amount, and the executor-specific authorization stays opaque to the facilitator.

So is Hedera meant to hold a stricter line than ERC-7710 here? Or does the same concern apply to ERC-7710, meaning it should eventually get the default-deny allowlist and execution-outcome validation too? Either answer tells us what to change in the PR.

kierzniak:
Hey @phdargen, just wanted to kindly bump this and check if you’ve had a chance to review my response. Any update would be much appreciated. Thanks!

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +208,405 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
...
+  "transaction": "0.0.9000@1755500000.000000000",
+  "network": "hedera:testnet",
+  "payer": "0.0.5001",
+  "extensions": {
+    "settlementProof": {
+      "payeeCredit": "100000000",
+      "payerDebit": "-100000000",
+      "feePayerDebit": "-5321000",
+      "transactionFee": "5321000"
+    }
```
phdargen: Think this can be dropped, we dont return tx metadata like this for any other network or scheme

If kept, should go in `extra` and not `extensions` 

`settlementProof` naming also misleading as its not a proof, just some additional metadata

phdargen:
> > `payload.executor` is an arbitrary client-chosen contract the facilitator will sign and submit as a `ContractExecuteTransaction`. A hostile executor can do the required payment and, in the same call, run eg a DeFi exploit or touch a sanctioned account. The facilitator need not fund that activity or be its victim; its account is still `transactionId.accountId` on the parent `CONTRACTCALL`. Verify only checks revert, so that extra activity is not rejected up front.
> > Please harden along the lines of [auth-capture ](https://github.com/x402-foundation/x402/blob/main/specs/schemes/auth-capture/scheme_auth_capture_evm.md) operatorType: "custom" (same “relay into foreign code” case):
> > 
> > * Default-deny executor allowlist on /supported (wildcard optional, not recommended)
> > * Cap gas (MUST)
> > * Assert the outcome, not the absence of a revert (MUST): matching net token movements and that no token movement originates from a facilitator-controlled address; nested logs / child-transfer visibility required, and if the RPC cannot provide them the facilitator MUST NOT advertise the method. Same checks on the receipt; status alone is insufficient
> > * Admission is by address, not code: prefer immutable executors (admin key none); a proxy or admin key can change implementation after review, so re-assert every settle
> 
> @phdargen thanks, that makes sense from the facilitator's threat model.
> 
> Before we change the design though: `transferExecutor` was deliberately modeled on the existing [ERC-7710 transfer method](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md#3-assettransfermethod-erc-7710) in the exact EVM scheme.
> 
> As far as I can tell ERC-7710 has the same trust boundary. The client supplies the `delegationManager`, the facilitator calls that contract from its own account, and the manager can run extra logic around the facilitator-constructed `ERC20.transfer(payTo, amount)`. The current EVM spec says no trusted list of Delegation Manager implementations is required, and treats a successful simulation as enough for verification.
> 
> `transferExecutor` follows that model. The facilitator constructs the exact `executeTransfer(from, asset, to, amount, authorization)` call, so the client cannot change the recipient or the amount, and the executor-specific authorization stays opaque to the facilitator.
> 
> So is Hedera meant to hold a stricter line than ERC-7710 here? Or does the same concern apply to ERC-7710, meaning it should eventually get the default-deny allowlist and execution-outcome validation too? Either answer tells us what to change in the PR.

There isnt an implementation for the ERC-7710 assetTransferMethod yet and as far as I know its also not actively worked on atm, so this shouldnt be considered as a definitive reference at this point. 
We had a similar issue for the more recent auth-capture spec update, which should server as a better reference.

When ERC-7710 is picked up again, it will likely need similar changes

kierzniak:
> > > `payload.executor` is an arbitrary client-chosen contract the facilitator will sign and submit as a `ContractExecuteTransaction`. A hostile executor can do the required payment and, in the same call, run eg a DeFi exploit or touch a sanctioned account. The facilitator need not fund that activity or be its victim; its account is still `transactionId.accountId` on the parent `CONTRACTCALL`. Verify only checks revert, so that extra activity is not rejected up front.
> > > Please harden along the lines of [auth-capture ](https://github.com/x402-foundation/x402/blob/main/specs/schemes/auth-capture/scheme_auth_capture_evm.md) operatorType: "custom" (same “relay into foreign code” case):
> > > 
> > > * Default-deny executor allowlist on /supported (wildcard optional, not recommended)
> > > * Cap gas (MUST)
> > > * Assert the outcome, not the absence of a revert (MUST): matching net token movements and that no token movement originates from a facilitator-controlled address; nested logs / child-transfer visibility required, and if the RPC cannot provide them the facilitator MUST NOT advertise the method. Same checks on the receipt; status alone is insufficient
> > > * Admission is by address, not code: prefer immutable executors (admin key none); a proxy or admin key can change implementation after review, so re-assert every settle
> > 
> > 
> > @phdargen thanks, that makes sense from the facilitator's threat model.
> > Before we change the design though: `transferExecutor` was deliberately modeled on the existing [ERC-7710 transfer method](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md#3-assettransfermethod-erc-7710) in the exact EVM scheme.
> > As far as I can tell ERC-7710 has the same trust boundary. The client supplies the `delegationManager`, the facilitator calls that contract from its own account, and the manager can run extra logic around the facilitator-constructed `ERC20.transfer(payTo, amount)`. The current EVM spec says no trusted list of Delegation Manager implementations is required, and treats a successful simulation as enough for verification.
> > `transferExecutor` follows that model. The facilitator constructs the exact `executeTransfer(from, asset, to, amount, authorization)` call, so the client cannot change the recipient or the amount, and the executor-specific authorization stays opaque to the facilitator.
> > So is Hedera meant to hold a stricter line than ERC-7710 here? Or does the same concern apply to ERC-7710, meaning it should eventually get the default-deny allowlist and execution-outcome validation too? Either answer tells us what to change in the PR.
> 
> There isnt an implementation for the ERC-7710 assetTransferMethod yet and as far as I know its also not actively worked on atm, so this shouldnt be considered as a definitive reference at this point. We had a similar issue for the more recent auth-capture spec update, which should server as a better reference.
> 
> When ERC-7710 is picked up again, it will likely need similar changes

@phdargen Thanks, this makes sense. I can update the spec with the executor allowlist, gas cap, address-based admission, and re-checks during settlement.

The only part we can’t fully implement on Hedera today is the pre-settlement outcome check with nested call/token movement visibility. Hedera doesn’t currently expose `debug_traceCall`; support is being tracked here: https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5113

Would you be okay with merging the PR with the other protections in place and documenting this limitation, or is `debug_traceCall`-level verification a hard requirement before this method can be accepted?

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +208,405 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
...
+        bytes calldata authorization
+    ) external;
+}
+```
+
+Requirements on an executor:
+
+- It MUST bind `authorization` to `from`, `asset`, `to` and `amount` (or to values that imply them),
+  so a facilitator cannot reuse an authorization for a different transfer, and MUST revert on any
+  mismatch. Whether an authorization is single-use or multi-use is the executor's choice.
```
phdargen: The server needs a deterministic answer at /verify to “is this a fresh request that will settle as a new payment?”, not only after work, when transaction exists. Ideally that answer comes from the payload itself (single-use authorization for `from, asset, to, amount`) so no server-side store is required; if identical bytes MAY be a new payment, say so explicitly and state what, if anything, MUST be remembered and when (verify vs settle)

phdargen:
> > > > `payload.executor` is an arbitrary client-chosen contract the facilitator will sign and submit as a `ContractExecuteTransaction`. A hostile executor can do the required payment and, in the same call, run eg a DeFi exploit or touch a sanctioned account. The facilitator need not fund that activity or be its victim; its account is still `transactionId.accountId` on the parent `CONTRACTCALL`. Verify only checks revert, so that extra activity is not rejected up front.
> > > > Please harden along the lines of [auth-capture ](https://github.com/x402-foundation/x402/blob/main/specs/schemes/auth-capture/scheme_auth_capture_evm.md) operatorType: "custom" (same “relay into foreign code” case):
> > > > 
> > > > * Default-deny executor allowlist on /supported (wildcard optional, not recommended)
> > > > * Cap gas (MUST)
> > > > * Assert the outcome, not the absence of a revert (MUST): matching net token movements and that no token movement originates from a facilitator-controlled address; nested logs / child-transfer visibility required, and if the RPC cannot provide them the facilitator MUST NOT advertise the method. Same checks on the receipt; status alone is insufficient
> > > > * Admission is by address, not code: prefer immutable executors (admin key none); a proxy or admin key can change implementation after review, so re-assert every settle
> > > 
> > > 
> > > @phdargen thanks, that makes sense from the facilitator's threat model.
> > > Before we change the design though: `transferExecutor` was deliberately modeled on the existing [ERC-7710 transfer method](https://github.com/x402-foundation/x402/blob/main/specs/schemes/exact/scheme_exact_evm.md#3-assettransfermethod-erc-7710) in the exact EVM scheme.
> > > As far as I can tell ERC-7710 has the same trust boundary. The client supplies the `delegationManager`, the facilitator calls that contract from its own account, and the manager can run extra logic around the facilitator-constructed `ERC20.transfer(payTo, amount)`. The current EVM spec says no trusted list of Delegation Manager implementations is required, and treats a successful simulation as enough for verification.
> > > `transferExecutor` follows that model. The facilitator constructs the exact `executeTransfer(from, asset, to, amount, authorization)` call, so the client cannot change the recipient or the amount, and the executor-specific authorization stays opaque to the facilitator.
> > > So is Hedera meant to hold a stricter line than ERC-7710 here? Or does the same concern apply to ERC-7710, meaning it should eventually get the default-deny allowlist and execution-outcome validation too? Either answer tells us what to change in the PR.
> > 
> > 
> > There isnt an implementation for the ERC-7710 assetTransferMethod yet and as far as I know its also not actively worked on atm, so this shouldnt be considered as a definitive reference at this point. We had a similar issue for the more recent auth-capture spec update, which should server as a better reference.
> > When ERC-7710 is picked up again, it will likely need similar changes
> 
> @phdargen Thanks, this makes sense. I can update the spec with the executor allowlist, gas cap, address-based admission, and re-checks during settlement.
> 
> The only part we can’t fully implement on Hedera today is the pre-settlement outcome check with nested call/token movement visibility. Hedera doesn’t currently expose `debug_traceCall`; support is being tracked here: [hiero-ledger/hiero-json-rpc-relay#5113](https://github.com/hiero-ledger/hiero-json-rpc-relay/issues/5113)
> 
> Would you be okay with merging the PR with the other protections in place and documenting this limitation, or is `debug_traceCall`-level verification a hard requirement before this method can be accepted?

That shouldnt be a blocker, in particular with the allowlist in place

kierzniak:
@phdargen Thanks, all six are in:

1. Allow only the executors listed in `extra.executors` on `/supported`. An omitted or empty list reject every executor, and the wildcard is allowed but not recommended.
2. The facilitator MUST set the gas limit itself, and MUST simulate with the same limit it submits.
3. Executors are admitted by address, and re-checked before every settle. Prefer ones with no `admin_key` and no upgrade path.
4. Settlement checks what actually moved: every merged transfer list, plus the contract result logs and call tree. A `SUCCESS` receipt is not enough on its own, and nothing the facilitator controls may be debited beyond the fee.
5. Dropped `settlementProof`.
6. Authorizations MUST now be single-use, so `/verify` tells a replay from a fresh payment by simulation alone, with no stored state.

Still open: the pre-relay outcome assertion needs `debug_traceCall`, which Hedera does not expose yet (hiero-ledger/hiero-json-rpc-relay#5113). The spec documents that, with admission as the preventive control.

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Done!

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Done!

kierzniak:
Hey @phdargen, just wanted to kindly bump this and check if you’ve had a chance to review my response. Any update would be much appreciated. Thanks!

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +209,437 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
 
 This specification does **not** require facilitators to forbid such transfers. Facilitators MAY handle this in whatever way they see fit: for example, they MAY require that `payTo` resolve to an existing account and reject transactions that would trigger auto-account creation, or they MAY allow it and accept the cost. Implementations SHOULD document their policy and, if they allow transfers to aliases, consider the associated cost and abuse potential.
 
+## AssetTransferMethod: `transferExecutor`
```
phdargen: nit: The document opens with a two-row method table that suggests two parallel sections, but only transferExecutor gets its own heading. Please add `## AssetTransferMethod: cryptoTransfer` above

specs/schemes/exact/scheme_exact_hedera.md
```diff
@@ -166,4 +209,437 @@ When the resource server’s `payTo` is specified as an **account alias** (e.g.
...
+```json
+{
+  "scheme": "exact",
+  "network": "hedera:testnet",
+  "amount": "100000000",
+  "asset": "0.0.0",
+  "payTo": "0.0.4001",
+  "maxTimeoutSeconds": 120,
+  "extra": {
+    "assetTransferMethod": "transferExecutor"
```
phdargen: clients currently dont know beforehand which executors are accepted. 
Servers should mirror the facilitator's executors from /supported into the 402

phdargen: More precisely: the server may narrow the list, but must not widen it. requirements.extra.executors must be a subset of the facilitator's /supported list. It can be "*" only if the facilitator advertises "*"

phdargen:
Thanks for the update @kierzniak , looks good. Just 2 more nits, see above

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Done. Thank you.

specs/schemes/exact/scheme_exact_hedera.md
kierzniak: Done. Thank you.

kierzniak:
Hi @phdargen, all the requested changes have now been incorporated into the spec. Could you please take another look and review it again when you have a chance? Thanks!

phdargen (approved):
LGTM @kierzniak 🚀

PR #3214, opened 2026-08-20
feat: settlement pending auto-recovery

CarsonRoscoe:
### What & why

Today, if a facilitator broadcasts a settlement transaction but times out waiting for it to confirm, the caller gets a hard failure — even though the payment may well land on-chain. The caller has no safe way to retry: resending risks a double-spend (nonce reuse, expired blockhash, etc.), and not retrying leaves a legitimate payment stuck.

This PR introduces `settlement_pending` as a non-terminal settle outcome: when a broadcast succeeds but confirmation can't be observed in time, the mechanism returns `settlement_pending` with the broadcast tx hash attached, and the resource server automatically retries **once**. The retry reconciles against the already-broadcast transaction instead of re-verifying and re-sending.

### How it works

- **`PendingSettlementStore`** (new interface, Go/TS/Python): keyed by a deterministic per-payload identifier (EIP-3009/Permit2 signature, SVM message hash, etc.), it remembers "this payload already broadcast `txHash` but isn't confirmed yet." Written only on broadcast-success + confirm-failure, checked before any new broadcast, and cleared on both terminal success and terminal failure.
- **In-memory default** (5-minute TTL, lazy pruning) ships out of the box for single-instance facilitators; the interface lets multi-instance deployments (e.g. `cdp-facilitator`) plug in a shared store (Redis) instead.
- **`x402ResourceServer`** retries a `settlement_pending` failure exactly once, with the identical payload — no backoff, no mutation. Any other outcome (success or a different failure) short-circuits after the first call.
- Wired into every settlement mechanism: EVM exact (EIP-3009 + Permit2), EVM upto (Permit2), EVM batch-settlement (deposit), SVM exact, and SVM upto (deposit + claim) — across Go, TypeScript, and Python.

### Notes to reviewers

- The dedup/duplicate-settlement lock (SVM `SettlementCache`) is a separate, pre-existing mechanism from `PendingSettlementStore`. It's released on any never-broadcast terminal failure but deliberately kept in place across a confirm-timeout, so a fresh retry can't re-send while the original might still land.
- Known limitation (documented inline, all three languages): the batch-settlement `reconcilePendingDeposit` cache-hit path skips a balance-confirmation check that the cache-miss path has, because it lacks the pre-broadcast channel-state snapshot. This only matters for a non-conforming ERC-20-approval extension signer combined with a confirm-timeout — a narrow combination.

### Testing

- Unit tests per mechanism per language covering: cache-miss success (no pending entry left), cache-miss confirm-failure (entry populated, dedup lock behavior), cache-hit reconciliation (no re-broadcast), invalid-broadcast-hash handling (never cached), and terminal-failure lock release.
- Go integration tests using deterministic-timeout signer wrappers to force `settlement_pending` reliably (no network flakiness).
- Full `go test ./...`, `pnpm test`, and `pytest` suites pass; `golangci-lint`, `staticcheck`, `ruff`, and `eslint` are clean.


typescript/packages/mechanisms/evm/src/shared/settleReceipt.ts
```diff
@@ -96,6 +97,46 @@ export async function waitAndReturnSettleResponse(
...
+): Promise<SettleResponse> {
+  const result = await settle();
+  if (!pendingKey) {
+    return result;
+  }
+  if (result.success) {
+    await store.delete(pendingKey);
+  } else if (result.transaction) {
+    await store.set(pendingKey, result.transaction);
+  }
```
phdargen: A confirmed reverted transaction also has a transaction hash, so this caches terminal failures until TTL expiry. Suggest to cache only ErrSettlementPending

```suggestion
if (!result.success && result.errorReason === ErrSettlementPending && result.transaction) {
  await store.set(pendingKey, result.transaction);
} else {
  await store.delete(pendingKey);
}
```

typescript/packages/mechanisms/evm/src/shared/settleReceipt.ts
```diff
@@ -96,6 +97,46 @@ export async function waitAndReturnSettleResponse(
...
+  store: PendingSettlementStore,
+  pendingKey: string | undefined,
+  settle: () => Promise<SettleResponse>,
+): Promise<SettleResponse> {
+  const result = await settle();
+  if (!pendingKey) {
+    return result;
+  }
+  if (result.success) {
+    await store.delete(pendingKey);
```
phdargen, line 162: We should handle storage exceptions here. If a pending retry successfully confirms the transaction but `store.delete()` throws, we should still return the successful settlement response. 

Suggest to handle `delete` and `set` failures differently:
- delete failure after success/terminal result: return the chain result anyway
- set failure after settlement_pending: return a non-retryable failure containing the transaction hash

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -436,6 +459,115 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
         payer: valid.payer || "",
       };
     }
+
+    try {
+      // Wait for confirmation
+      await this.signer.confirmTransaction(signature, requirements.network);
```
phdargen, line 494: `confirmTransaction` also throws for confirmed transactions with an onchain error, but this catch maps every exception to `settlement_pending`. Only timeout errors should remain pending; terminal failures should clear both caches and return transaction_failed (also applies to channel.ts:297 and upto/facilitator/scheme.ts:414)

typescript/packages/mechanisms/svm/src/exact/facilitator/scheme.ts
```diff
@@ -309,6 +329,45 @@ export class ExactSvmScheme implements SchemeNetworkFacilitator {
...
+        await this.pendingStore.delete(txKey);
+        // Best-effort payer for the response; a decode failure here doesn't
+        // block reconciliation (the payload already broadcast successfully).
+        let payer = "";
+        try {
+          payer = getTokenPayerFromTransaction(decodeTransactionFromPayload(exactSvmPayload)) ?? "";
+        } catch {
+          // Ignore; payer stays "".
+        }
+        return this.reconcilePendingSettlement(txKey, cachedSignature, payer, requirements.network);
```
phdargen: For the svm smart-wallet path we should also run `verifyPostSettlement` after reconciliation

typescript/packages/mechanisms/svm/src/upto/facilitator/scheme.ts
```diff
@@ -372,23 +381,105 @@ export class UptoSvmScheme implements SchemeNetworkFacilitator {
...
     payload: PaymentPayload,
     requirements: PaymentRequirements,
+    p: UptoSvmPayloadV2,
   ): Promise<SettleResponse> {
+    // Pending-settlement fast path: a prior deposit settle for this exact
+    // channel broadcast the open successfully but couldn't confirm it in
+    // time. Reconcile against that signature instead of re-validating and
+    // re-broadcasting — a second open attempt would hit ERR_CHANNEL_ALREADY_OPEN
+    // even though the original payment is (or will be) fine.
+    const depositKey = `upto:deposit:${requirements.network}:${p.channelId}`;
```
phdargen: Wonder if we better use the `transactionMessageHash(transaction)` as key here like for `exact` to bind deposit amount etc

typescript/packages/mechanisms/evm/src/batch-settlement/facilitator/deposit.ts
```diff
@@ -312,19 +316,139 @@ async function verifySharedDepositState(
...
+              channelId: voucher.channelId,
+              balance: state.balance.toString(),
+              totalClaimed: state.totalClaimed.toString(),
+              withdrawRequestedAt: state.withdrawRequestedAt,
+              refundNonce: state.refundNonce.toString(),
+            },
+          };
+        } catch {
+          // Leave extra undefined when the post-confirm channel-state read fails;
+          // the transaction is still reported as confirmed.
```
phdargen, line 439: The spec requires successful batch-settlement responses to include extra.channelState. Could we return an optimistic snapshot here as the regular deposit path does?

phdargen (approved):
LGTM 🚀

PR #3219, opened 2026-08-20
fix(e2e): scope EVM/SVM client signer derivation to selected families

### LINKED ISSUE #3187: [Bug]: e2e client eagerly derives EVM+SVM signers regardless of --families, crashing single-family conformance runs
### What happened?

`createE2EClient()` (`e2e/clients/typescript/client.ts`) unconditionally derives an EVM account and an SVM signer at the top of the function, before any family-specific branching:

```ts
export async function createE2EClient(): Promise<E2EClientContext> {
  const baseURL = process.env.RESOURCE_SERVER_URL as string;
  const evmAccount = privateKeyToAccount(process.env.CLIENT_EVM_PRIVATE_KEY as `0x${string}`);
  const svmSigner = await createKeyPairSignerFromBytes(
    base58.decode(process.env.CLIENT_SVM_PRIVATE_KEY as string),
  );
  // ... family-specific branches follow, including one that never touches evmAccount/svmSigner
```

So a `--families=<non-evm, non-svm>` run of the `pnpm test` CLI (e.g. `--families=stellar`) still requires `CLIENT_EVM_PRIVATE_KEY` and `CLIENT_SVM_PRIVATE_KEY` to be set, even though neither is ever used for a Stellar-only scenario. Two problems compound this:

1. The harness's own preflight check only validates **facilitator** env per the selected family (`test.ts`'s `🔍 Validating facilitator environment variables`, around line 1007), and there is no equivalent client-side check. A scoped run reports `✅ All required environment variables are present` and then crashes anyway, deep inside client bootstrapping, well after the facilitator and server have already started.
2. The failure itself is a low-level, confusing error rather than a clear "missing env var" message: `privateKeyToAccount` throws `TypeError: Cannot read properties of undefined (reading 'slice')` when `CLIENT_EVM_PRIVATE_KEY` is unset, giving no indication of which variable is actually missing.

Found running a real, scoped, single-family conformance pass against a live external Stellar facilitator (via `external-proxies`), not from reading the code cold.

### Steps to reproduce

```bash
cd e2e
pnpm install
# Set only Stellar client/server/facilitator env (CLIENT_STELLAR_PRIVATE_KEY,
# SERVER_STELLAR_ADDRESS, FACILITATOR_STELLAR_PRIVATE_KEY), deliberately
# leave CLIENT_EVM_PRIVATE_KEY and CLIENT_SVM_PRIVATE_KEY unset.
pnpm test --testnet --families=stellar --facilitators=<any-stellar-facilitator> \
  --clients=typescript/http/axios --servers=typescript/http/express --schemes=exact -v
```

Observed:

```
🔍 Validating facilitator environment variables...
  ✅ All required environment variables are present
...
📊 Client result: {
  "success": false,
  "error": "...client.ts:48\n  const publicKey = toHex(secp256k1.getPublicKey(privateKey.slice(2), false))\n ^\n\nTypeError: Cannot read properties of undefined (reading 'slice')\n    at privateKeyToAccount (.../viem/accounts/privateKeyToAccount.ts:48:61)\n    at createE2EClient (.../e2e/clients/typescript/client.ts:90:22)\n...",
  "exitCode": 1
}
```

Setting a dummy `CLIENT_EVM_PRIVATE_KEY` gets past that line but hits the same class of issue for SVM: `createKeyPairSignerFromBytes` throws `SolanaError: Key pair bytes must be of length 64, got 32` unless `CLIENT_SVM_PRIVATE_KEY` decodes to exactly a 64-byte Solana keypair, a format constraint the harness never surfaces anywhere in its docs or preflight output.

### Environment

- `x402-foundation/x402` commit `8c308ce3040556482099958f09977fb1fe487e12` (2026-08-17)
- Node.js v22.23.2, pnpm 11.1.1
- `e2e/clients/typescript/client.ts`, `createE2EClient()`

### Severity

Calibrated honestly: not a security issue, no funds or credentials at risk, fails closed (a crash, not a silent wrong result). It's a DX/usability gap in the conformance harness itself: the `--families=` scoping the CLI and README document as a supported, first-class mode (the TVM example in `e2e/README.md`'s own docs) doesn't actually scope the TypeScript client's credential requirements the way the facilitator-side preflight check implies it does. Anyone running a single-family conformance pass against an external facilitator on a non-EVM, non-SVM network hits this.

### Suggested fix

Move the `evmAccount`/`svmSigner` derivation inside the family-specific branches that actually use them (the function already branches per family further down), or gate them behind a check against the networks actually requested for this run. Alternatively, extend the existing facilitator env preflight check to also validate client-required env per selected family, so a scoped run either works without unrelated-network credentials or fails fast with a clear "missing `CLIENT_EVM_PRIVATE_KEY`" message instead of a stack trace three layers removed from the real cause. Happy to open a PR if a maintainer has a preference between the two.

---

Found and verified with AI-assisted analysis against the published package.

Eras256:
Closes #3187.

`createE2EClient()` derives EVM and SVM signers unconditionally at the top of the function, before any `--families` scoping runs. Every other family in the same function (Aptos, Hedera, Keeta, Stellar, AVM, TVM, Near, XRPL) only derives its signer when its own `CLIENT_*` credential is set. So a single-family run against a non-EVM/SVM network still crashes here unless unrelated EVM/SVM credentials happen to be present too, which defeats the point of `--families` scoping.

Fixed by making EVM and SVM derivation conditional on `CLIENT_EVM_PRIVATE_KEY`/`CLIENT_SVM_PRIVATE_KEY`, matching every other family's existing pattern. `batchSettlementScheme` (EVM-only) is now `BatchSettlementEvmScheme | undefined`; `runClientScenario` throws a clear config error if `EVM_BATCH_SETTLEMENT_PHASE` is set without EVM creds, rather than a bare "Cannot read properties of undefined" once the refund path is actually reached.

There's a second, separate gap behind the same crash: the harness's own preflight validation only ever checks facilitator env vars (`uniqueFacilitators`), then prints "All required environment variables are present" without having looked at client env at all. A family-scoped run can pass that check cleanly and still crash deep inside a client for exactly the reason above. Extended the same validation loop to also cover `uniqueClients`, against each client's own catalog-declared `config.environment.required`.

Verified with real code, not just reasoning: built the full `typescript` workspace and called the actual `createE2EClient()` directly under five scenarios — no client creds at all (the exact reported crash), EVM-only, SVM-only (using a real generated Ed25519 keypair, not a stub), both, and the batch-settlement-without-EVM-creds guard. All five pass. Also checked this didn't regress the "everything set" dev case (9 total scheme registrations, same as before the fix).

---

### DIFF
--- a/e2e/clients/typescript/client.ts
+++ b/e2e/clients/typescript/client.ts
@@ -76,61 +76,104 @@ export type E2EClientContext = {
    * setup instead of duplicating it.
    */
   schemes: SchemeRegistration[];
-  batchSettlementScheme: BatchSettlementEvmScheme;
+  batchSettlementScheme: BatchSettlementEvmScheme | undefined;
   batchSettlementPhase: BatchSettlementPhase | undefined;
 };
 
 /**
  * Builds the shared x402 client with all e2e scheme registrations.
+ *
+ * EVM and SVM signer derivation is conditional on their own client
+ * credential (`CLIENT_EVM_PRIVATE_KEY`/`CLIENT_SVM_PRIVATE_KEY`), same as
+ * every other family below (Aptos, Hedera, Keeta, Stellar, AVM, TVM, Near,
+ * XRPL). Both used to be derived unconditionally at the top of this
+ * function, ahead of any family-scoping logic, so a run scoped to a single
+ * non-EVM/SVM family via `--families` still crashed here unless unrelated
+ * EVM/SVM credentials happened to be set too (x402-foundation/x402#3187).
  */
 export async function createE2EClient(): Promise<E2EClientContext> {
   const baseURL = process.env.RESOURCE_SERVER_URL as string;
   const endpointPath = process.env.ENDPOINT_PATH as string;
   const url = `${baseURL}${endpointPath}`;
-  const evmAccount = privateKeyToAccount(process.env.CLIENT_EVM_PRIVATE_KEY as `0x${string}`);
-  const svmSigner = await createKeyPairSignerFromBytes(
-    base58.decode(process.env.CLIENT_SVM_PRIVATE_KEY as string),
-  );
 
-  const evmNetwork = resolveNetworkCaip2("evm");
-  const evmRpcUrl = process.env.EVM_RPC_URL;
-  const evmChain = evmNetwork === "eip155:8453" ? base : baseSepolia;
+  const schemes: SchemeRegistration[] = [];
+  let batchSettlementScheme: BatchSettlementEvmScheme | undefined;
 
-  const publicClient = createPublicClient({
-    chain: evmChain,
-    transport: http(evmRpcUrl),
-  });
+  if (process.env.CLIENT_EVM_PRIVATE_KEY) {
+    const evmAccount = privateKeyToAccount(process.env.CLIENT_EVM_PRIVATE_KEY as `0x${string}`);
 
-  const evmSigner = toClientEvmSigner(evmAccount, publicClient);
+    const evmNetwork = resolveNetworkCaip2("evm");
+    const evmRpcUrl = process.env.EVM_RPC_URL;
+    const evmChain = evmNetwork === "eip155:8453" ? base : baseSepolia;
 
-  const evmSchemeOptions = process.env.EVM_RPC_URL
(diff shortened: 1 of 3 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/3228

PR #3221, opened 2026-08-21
docs(exact/svm): add multi-payee splits to the exact scheme

lgalabru:
## Summary

Adds an optional **multi-payee splits** capability to the `exact` scheme on SVM, letting a single client payment be distributed atomically across the merchant (`payTo`) and up to eight additional recipients in one transaction — for platform fees, revenue sharing, referral commissions, and fee-payer cost recovery. No extra round-trips or separate transactions.

Splits are declared by the Resource Server in `PaymentRequirements.extra.splits`. When the field is absent or empty, the scheme behaves exactly as today (a single transfer to `payTo`), so the change is fully backward compatible.

```jsonc
"extra": {
  "feePayer": "EwWqGE4ZFKLofuestmU4LDdK7XM1N4ALgdZccwYugwGd",
  "memo": "pi_3abc123def456",
  "splits": [
    { "payTo": "3pF8…", "amount": "50000", "memo": "platform fee", "ataCreationRequired": false }
  ]
}
```

## What changed

- **`specs/schemes/exact/scheme_exact_svm.md`** — new self-contained **§5 Multi-Payee Splits**: the `extra.splits` schema, amount/remainder semantics, per-leg payment outcome, split-recipient ATA creation authorization, sponsor acceptance rules, client-side verification, generalized static layout, and split-specific security invariants. Surgical pointers were added in §1.1, §1.4, §2.1.1, §2.1.3, §2.1.5, §3.1, and §3.2 so the base definitions stay intact and simply generalize "one transfer" → "one transfer per leg" when splits are present.
- **`specs/schemes/exact/scheme_exact.md`** — a network-agnostic **Multi-Payee Splits (Optional Extension)** note and split-aware wording in the SVM validation bullets, so other networks can adopt the same model.

## Design decisions

- **Absolute amounts, not basis points.** Each `splits[].amount` is an absolute carve-out in base units of the same `asset`. The primary `payTo` receives the remainder `amount − Σ splits[].amount`, which MUST be `> 0`.
- **Per-leg matching.** Each payment leg (primary + each split) MUST map to a *distinct* on-chain `TransferChecked`; no transfer satisfies two legs even when recipients coincide.
- **Overpayment asymmetry.** Split legs MUST match their declared amount exactly; the primary leg keeps the existing overpayment tolerance (`≥ remainder`) for smart-wallet fee rounding, so any excess accrues to the merchant and never to a split recipient.
- **Bounded ATA-creation exception.** Fee-payer-funded ATA creation is the one intentional exception to fee-payer isolation (§2.1.1) / fund safety (§2.1.3): allowed only for split recipients explicitly marked `ataCreationRequired: true`, and **forbidden** for the primary recipient or arbitrary owners.
- **Max 8 splits (9 legs).** With a note that the practical ceiling is bounded by the 1232-byte transaction limit; ALTs may raise it.

Adapted from the multi-payee/splits model in the Solana Charge Intent draft (paymentauth.org/draft-solana-charge-00.html), rewritten to x402 `exact` prose and terminology (`payTo`/`amount`, fee-payer sponsorship, ATA verification).

## Open questions for reviewers

1. **Primary overpayment tolerance** — kept `≥` on the primary leg for backward-compat with the existing base rule; splits are exact. Prefer exact on all legs?
2. **Placement** — splits are specified inline in the SVM doc with a generic note in `scheme_exact.md`. Would a standalone `specs/extensions/splits.md` be preferred instead?
3. **Per-split `memo`** — carried as an informational reconciliation label only (not an on-chain Memo). Keep, or drop in favor of the single `extra.memo`?

Spec-only PR; reference-implementation changes to `@x402/svm` will follow in a separate PR.

### DIFF
--- a/specs/schemes/exact/scheme_exact_svm.md
+++ b/specs/schemes/exact/scheme_exact_svm.md
@@ -23,6 +23,8 @@ How transaction fees are sponsored—and how the sponsoring party evaluates risk
   - a **third-party facilitator**.
 - **Fee Payer (`feePayer`)**: The Solana account of the Sponsor that pays transaction fees and provides the final required signature.
 - **Smart wallet**: A program-controlled account (e.g. Squads, Swig, SPL Governance, Metaplex Core) that executes a token transfer via Cross-Program Invocation (CPI) rather than as a top-level `TransferChecked`.
+- **Primary payment**: The transfer to the merchant identified by `payTo`. In the absence of splits this is the entire `amount`; with splits it is the remainder (§5).
+- **Split**: An additional payee, beyond the merchant, that receives a fixed carve-out of the total `amount` in the same atomic transaction. Declared in `extra.splits` (§5). Used for platform fees, revenue sharing, referral commissions, or fee-payer cost recovery.
 
 ---
 
@@ -73,6 +75,7 @@ In addition to the standard x402 `PaymentRequirements` fields, the `exact` schem
 - `extra.memo` (optional): A seller-defined UTF-8 string to include in the transaction's Memo instruction. When present, the client MUST use this value as the Memo instruction data instead of a random nonce. Maximum 256 bytes. This enables sellers to attach payment references (e.g., invoice IDs) to on-chain transactions for reconciliation without requiring unique deposit addresses.
 - `extra.recentBlockhash` (optional): A recent blockhash for the Client to use as the transaction's lifetime, supplied by the Resource Server. When present and valid, the Client SHOULD use it instead of fetching its own via `getLatestBlockhash`, saving an RPC round-trip and pinning the transaction to a blockhash the settling RPC has observed. When absent or malformed, the Client MUST fetch a recent blockhash itself. A Resource Server SHOULD only set this when it has an RPC endpoint whose view of recent blockhashes matches the settling Sponsor's; a stale or fork-divergent blockhash will cause the transaction to expire or fail to land.
 - `extra.lastValidBlockHeight` (optional): The last block height at which `extra.recentBlockhash` is valid, as a decimal string. This informational hint is not serialized into the transaction and MAY be ignored by the Client or Sponsor. It is ignored when `recentBlockhash` is absent.
+- `extra.splits` (optional): An array of additional payees that receive fixed carve-outs of `amount` in the same transaction. When present, the merchant (`payTo`) receives the remainder rather than the full `amount`. See **Multi-Payee Splits** (§5) for the schema and semantics. When absent or empty, the scheme behaves exactly as specified in §1–§4 (a single transfer to `payTo`).
 
 These optional fields are transaction-construction hints. They do not bind the submitted transaction to the hinted blockhash, and the facilitator's verification does not compare the transaction blockhash with `extra.recentBlockhash`.
 
@@ -154,6 +157,8 @@ The payment MUST be:
 
 The transfer MAY appear either as a top-level instruction or as an inner instruction (CPI) emitted by another program. This is what allows smart wallets to satisfy the scheme: the payment outcome is what matters, not the instruction that produced it.
 
+When `extra.splits` is present, the required outcome comprises multiple payment legs (one to `payTo` plus one per split), and `PaymentRequirements.amount` is the total across all legs. The single-transfer definition below applies per leg; see §5.
+
 ### 1.2 Payment Identification (MUST)
 
 A verifier MUST be able to deterministically identify the payment and confirm:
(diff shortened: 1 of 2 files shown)


specs/schemes/exact/scheme_exact_svm.md
```diff
@@ -348,6 +361,126 @@ The verification model upholds the following invariants:
...
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `splits` | array | OPTIONAL | 1–8 entries. Absent or empty ⇒ single-payee `exact`. |
+| `splits[].payTo` | string | REQUIRED | Base58-encoded recipient public key. |
+| `splits[].amount` | string | REQUIRED | Decimal string in base units of `asset`. An **absolute** amount, not a percentage or basis points. |
+| `splits[].memo` | string | OPTIONAL | Human-readable label for off-chain reconciliation, ≤ 566 bytes. Informational only; it is **not** emitted as an on-chain Memo instruction. The single on-chain Memo remains governed by `extra.memo` (§3.1). |
+| `splits[].ataCreationRequired` | boolean | OPTIONAL | Default `false`. Authorizes fee-payer-funded creation of this recipient's ATA (§5.4). Meaningful only for SPL / Token-2022 assets. |
+
+Every split shares the top-level `asset`; a split MUST NOT name a different mint. Split `payTo` values SHOULD be distinct from one another and from the primary `payTo`; when they are not, each leg is still matched to its own distinct transfer instruction (§5.3).
```
phdargen, line 405: Don't see a good reason for allowing this
```suggestion
Every split shares the top-level `asset`; a split MUST NOT name a different mint. Split `payTo` values MUST be distinct from one another and from the primary `payTo`.
```

specs/schemes/exact/scheme_exact_svm.md
```diff
@@ -348,6 +361,126 @@ The verification model upholds the following invariants:
...
+### 5.5 Sponsor Acceptance with Splits (MUST)
+
+A Sponsor that signs a split transaction as `feePayer` MUST, in addition to the baseline in §2.1:
+
+- Recompute `primary = amount − Σ splits[].amount` from the requirements and reject if `primary ≤ 0` (§5.2).
+- Match every payment leg to a distinct transfer per §5.3.
+- Enforce the ATA-creation authorization rules of §5.4.
+- Confirm that no `asset` transfer sends funds to any account other than the ATA derived from `payTo` or from a declared split `payTo` (`split_unexpected_transfer`).
+
+A Sponsor MAY decline splits entirely, or offer splits without ATA-rent sponsorship (accepting only `ataCreationRequired: false` entries), as a matter of local policy (§2.3). Clients SHOULD NOT assume universal split support.
```
phdargen, line 448: Think better not to leave it as unadvertised local policy. Suggest facilitator announces split support in /supported extra (e.g. maxSplits, if missing implies "maxSplits=0")

specs/schemes/exact/scheme_exact_svm.md
```diff
@@ -348,6 +361,126 @@ The verification model upholds the following invariants:
...
+
+### 5.4 Split Recipient ATA Creation (MUST)
+
+A split recipient may not yet hold an Associated Token Account for `asset`, and a transfer to a missing ATA fails. The transaction MAY therefore include an **idempotent** ATA-creation instruction (`AssociatedTokenProgram: CreateIdempotent`) for a split recipient. Because ATA creation spends the funding account's lamports (rent, ≈0.002 SOL) beyond the network fee, it is tightly scoped:
+
+- A `CreateIdempotent` instruction MAY be funded by the `feePayer` **only** for a split recipient whose entry sets `ataCreationRequired: true`.
+- When the fee payer sponsors rent, the client MUST include one `CreateIdempotent` instruction for each split whose entry sets `ataCreationRequired: true`, and MUST NOT include ATA-creation instructions for the primary `payTo`, for unmarked split recipients, or for any other owner.
+- Fee-payer-funded creation of the **primary** `payTo` ATA is FORBIDDEN. The merchant MUST maintain its own receiving ATA. This prevents a merchant from offloading its own rent onto a third-party sponsor on every charge.
```
phdargen, line 434: Dont understand why ATA sponsorship should be forbidden for `payTo` but then allowed for the split receivers. Think the griefing vector (see eg https://github.com/x402-foundation/x402/issues/1020) should be even worse there, so we probably need to drop the ATA sponsorship entirely

specs/schemes/exact/scheme_exact_svm.md
```diff
@@ -348,6 +361,126 @@ The verification model upholds the following invariants:
...
+```
+
+In the example the client pays `1050000` base units in total: `50000` to the split recipient and the `1000000` remainder to `payTo`.
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `splits` | array | OPTIONAL | 1–8 entries. Absent or empty ⇒ single-payee `exact`. |
+| `splits[].payTo` | string | REQUIRED | Base58-encoded recipient public key. |
+| `splits[].amount` | string | REQUIRED | Decimal string in base units of `asset`. An **absolute** amount, not a percentage or basis points. |
+| `splits[].memo` | string | OPTIONAL | Human-readable label for off-chain reconciliation, ≤ 566 bytes. Informational only; it is **not** emitted as an on-chain Memo instruction. The single on-chain Memo remains governed by `extra.memo` (§3.1). |
```
phdargen, line 402: do we need this if its informational only?

phdargen:
Wire type looks overall good, similar to what I suggested for SUI here: https://github.com/x402-foundation/x402/pull/2615

PR #3228, opened 2026-08-22
fix(e2e): scope EVM/SVM client signer derivation to selected families

### LINKED ISSUE #3187: [Bug]: e2e client eagerly derives EVM+SVM signers regardless of --families, crashing single-family conformance runs

Eras256:
Fixes #3187.

`createE2EClient()` derived EVM (`viem`'s `privateKeyToAccount`) and SVM (`@solana/kit`'s `createKeyPairSignerFromBytes`) signers unconditionally at the top of the function, ahead of any `--families` scoping, while every other family below it (Aptos, Hedera, Keeta, Stellar, AVM, TVM, Near, XRPL) already only derives its signer when that family's own `CLIENT_*` credential is actually set. A single-family run against a non-EVM/SVM network crashed here regardless of `--families`, unless unrelated EVM/SVM credentials happened to be set too — the exact scenario in #3187, hit running the real conformance suite against a live Stellar facilitator with no EVM/SVM credentials configured at all.

## What changed

- EVM and SVM signer derivation is now conditional on their own `CLIENT_EVM_PRIVATE_KEY`/`CLIENT_SVM_PRIVATE_KEY`, matching the pattern every other family already follows.
- `batchSettlementScheme` (EVM-only) is now `BatchSettlementEvmScheme | undefined`. `runClientScenario` throws a clear config error if `EVM_BATCH_SETTLEMENT_PHASE` is set without EVM creds, instead of a bare `Cannot read properties of undefined` once the refund path is actually reached.
- Second, separate gap found running this: the harness's own preflight validation (`e2e/test.ts`) only ever checked facilitator env vars, printing "✅ All required environment variables are present" without having looked at client env at all. A family-scoped run could pass that check cleanly and still crash deep inside a client for exactly this reason. Extended the same validation loop to also cover clients against each client's own catalog-declared `config.environment.required`.

Built the full `typescript/` workspace and the `e2e/` workspace on top of current `main`, then called the actual `createE2EClient()`/`runClientScenario()` under five scenarios: no client creds (the exact reported crash), EVM-only (a real random 32-byte key), SVM-only (a real generated Ed25519 keypair, base58-encoded, matching what `createKeyPairSignerFromBytes` expects), both, and the batch-settlement-without-EVM-creds guard (checked against `runClientScenario`, where that guard actually lives, not `createE2EClient` itself). All five pass:

```
PASS [no client creds (the reported #3187 crash)]: 0 scheme(s) registered, no crash
PASS [EVM-only]: 5 scheme(s) registered, no crash
PASS [SVM-only]: 4 scheme(s) registered, no crash
PASS [EVM + SVM]: 9 scheme(s) registered, no crash
PASS [EVM_BATCH_SETTLEMENT_PHASE set without EVM creds (guard)]: threw as expected: EVM_BATCH_SETTLEMENT_PHASE is set but no CLIENT_EVM_PRIVATE_KEY was provided to build a batch-settlement scheme from.
```

### DIFF
--- a/e2e/clients/typescript/client.ts
+++ b/e2e/clients/typescript/client.ts
@@ -76,61 +76,104 @@ export type E2EClientContext = {
    * setup instead of duplicating it.
    */
   schemes: SchemeRegistration[];
-  batchSettlementScheme: BatchSettlementEvmScheme;
+  batchSettlementScheme: BatchSettlementEvmScheme | undefined;
   batchSettlementPhase: BatchSettlementPhase | undefined;
 };
 
 /**
  * Builds the shared x402 client with all e2e scheme registrations.
+ *
+ * EVM and SVM signer derivation is conditional on their own client
+ * credential (`CLIENT_EVM_PRIVATE_KEY`/`CLIENT_SVM_PRIVATE_KEY`), same as
+ * every other family below (Aptos, Hedera, Keeta, Stellar, AVM, TVM, Near,
+ * XRPL). Both used to be derived unconditionally at the top of this
+ * function, ahead of any family-scoping logic, so a run scoped to a single
+ * non-EVM/SVM family via `--families` still crashed here unless unrelated
+ * EVM/SVM credentials happened to be set too (x402-foundation/x402#3187).
  */
 export async function createE2EClient(): Promise<E2EClientContext> {
   const baseURL = process.env.RESOURCE_SERVER_URL as string;
   const endpointPath = process.env.ENDPOINT_PATH as string;
   const url = `${baseURL}${endpointPath}`;
-  const evmAccount = privateKeyToAccount(process.env.CLIENT_EVM_PRIVATE_KEY as `0x${string}`);
-  const svmSigner = await createKeyPairSignerFromBytes(
-    base58.decode(process.env.CLIENT_SVM_PRIVATE_KEY as string),
-  );
 
-  const evmNetwork = resolveNetworkCaip2("evm");
-  const evmRpcUrl = process.env.EVM_RPC_URL;
-  const evmChain = evmNetwork === "eip155:8453" ? base : baseSepolia;
+  const schemes: SchemeRegistration[] = [];
+  let batchSettlementScheme: BatchSettlementEvmScheme | undefined;
 
-  const publicClient = createPublicClient({
-    chain: evmChain,
-    transport: http(evmRpcUrl),
-  });
+  if (process.env.CLIENT_EVM_PRIVATE_KEY) {
+    const evmAccount = privateKeyToAccount(process.env.CLIENT_EVM_PRIVATE_KEY as `0x${string}`);
 
-  const evmSigner = toClientEvmSigner(evmAccount, publicClient);
+    const evmNetwork = resolveNetworkCaip2("evm");
+    const evmRpcUrl = process.env.EVM_RPC_URL;
+    const evmChain = evmNetwork === "eip155:8453" ? base : baseSepolia;
 
-  const evmSchemeOptions = process.env.EVM_RPC_URL
-    ? { rpcUrl: process.env.EVM_RPC_URL }
-    : undefined;
+    const publicClient = createPublicClient({
+      chain: evmChain,
+      transport: http(evmRpcUrl),
+    });
 
-  const uptoSchemeOptions: UptoEvmSchemeOptions | undefined = process.env.EVM_RPC_URL
-    ? { rpcUrl: process.env.EVM_RPC_URL }
-    : undefined;
-  const svmSchemeOptions = process.env.SVM_RPC_URL ? { rpcUrl: process.env.SVM_RPC_URL } : undefined;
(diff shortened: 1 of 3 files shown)


Eras256:
An independent Stellar facilitator (stellarsight) hit this exact crash in their own e2e conformance run and needed the same EVM/SVM decoy-key workaround, see their docs/EVIDENCE.md (commit 35cf076). Second real-world case for the same fix.

Disclosure: drafted with AI assistance under my direction and reviewed by hand. The cited commit and evidence file were re-verified against the real stellarsight repository today.

phdargen (approved):
Thanks for the contribution @Eras256

PR #3232, opened 2026-08-22
fix: require PaymentRequired.resource.url on v2 emit

### LINKED ISSUE #3211: Go and Python SDKs type PaymentRequired.resource as optional while TypeScript requires it, so they can emit what TS rejects
The v2 schema marks `resource` as required on `PaymentRequired`, but two of the three reference SDKs type it as optional, and one of those omits the key entirely when it is unset. So a Go or Python resource server can emit a challenge that the TypeScript client refuses to parse.

Related to but distinct from #3210, which is about the migration guide. This one is about the SDKs themselves.

## The three definitions

**TypeScript** (`typescript/packages/core/src/schemas/index.ts`) requires it:

```ts
export const PaymentRequiredV2Schema = z.object({
  x402Version: z.literal(2),
  error: z.string().nullish().transform(v => v ?? undefined),
  resource: ResourceInfoSchema,
  accepts: z.array(PaymentRequirementsV2Schema).min(1),
  extensions: OptionalAny,
});
```

Worth noting that `error` in the same object is explicitly `.nullish()`, so optionality is marked deliberately here. `resource` is not marked, which matches section 5.1.2.

**Go** (`go/types/v2.go`) makes it a nullable pointer that disappears from the wire:

```go
Resource    *ResourceInfo          `json:"resource,omitempty"`
```

With a nil `Resource`, `omitempty` does not emit `"resource": null`. It **drops the key**, producing exactly the v1-shaped envelope.

**Python** (`python/x402/schemas/payments.py`) types it optional and documents it as optional:

```python
resource: ResourceInfo | None = None
```

with the class docstring reading `resource: Optional resource information.` The builder in `python/x402/server_base.py` also defaults it:

```python
def create_payment_required_response(self, requirements, resource: ResourceInfo | None = None, ...)
```

and `_build_payment_required_response` passes it straight through to the model as `resource=resource` with no check.

## Why this matters rather than being a typing nit

The three SDKs disagree about the same wire format, in the direction that produces silent incompatibility: **Go and Python can construct what TypeScript will reject.** A TS client validating with `PaymentRequiredV2Schema` throws on a challenge a Python server produced through the SDK's own documented API, with nothing hand-rolled.

That is not hypothetical. Four independent implementations have shipped a v2-declared envelope with no top-level `resource`, and I reported each one: middyjs/middy#1678 (fixed, merged in #1679), elizaOS/eliza#22615 (fixed, merged in #22625), usehelix/helix#17, azep-ninja/x402-gateway-template#5. A shipped third-party client, `wevm/mppx`, parses with `ResourceInfoSchema.parse(record.resource)` where `url` is a required non-empty string, so it throws before reaching `accepts`.

Those four hand-rolled their envelopes, so the SDKs did not cause them. But an adopter who does the right thing and uses the Python SDK can land in the same place, and nothing in the SDK will say so.

## Suggested direction

Make `resource` required in the Go and Python models so all three agree with section 5.1.2, or if it must stay nullable for construction ergonomics, validate before serialization so an incomplete envelope cannot leave the process. The Python docstring should stop describing a required field as optional either way.

## What I have not verified

- I have not run any of the three SDKs. This is read from the schema and model definitions and, for Python, from the builder that populates them.
- Whether Go or Python validate the envelope somewhere later in an emission path I did not read. I traced the Python builder to the model constructor and found no check; I read the Go struct and the constructor signature, not every caller.
- Whether the divergence is deliberate, for example to allow staged construction. If so, that is worth a comment in the models, since the current Python docstring states the opposite of the spec.

Filed by an automated agent, from reading the repository.

Bartok9:
Fixes #3211.

v2 Go/Python servers could omit `resource`; TypeScript clients reject that. Marshal/encode now requires a non-empty `resource.url` (v1 unchanged).

Majority-AI assisted; human-reviewed. Commit unsigned (local gpg lock); can re-sign if needed.

Tested: `go test ./types ./http`; `pytest python/x402/tests/unit/http/test_utils.py`.

### DIFF
--- a/go/extensions/signinwithx/client_test.go
+++ b/go/extensions/signinwithx/client_test.go
@@ -278,6 +278,7 @@ func TestCreateClientHookReturnsSIWXHeader(t *testing.T) {
 	hook := CreateClientHook(signer)
 	result, err := hook(context.Background(), types.PaymentRequired{
 		X402Version: 2,
+		Resource:    &types.ResourceInfo{URL: "https://example.com/resource"},
 		Extensions: map[string]interface{}{
 			ExtensionKey: Extension{
 				Info: Info{
@@ -321,6 +322,7 @@ func TestCreateClientHookWithSignersReturnsSolanaHeader(t *testing.T) {
 	hook := CreateClientHookWithSigners(NewSolanaSIWXSigner(signer))
 	result, err := hook(context.Background(), types.PaymentRequired{
 		X402Version: 2,
+		Resource:    &types.ResourceInfo{URL: "https://example.com/resource"},
 		Extensions: map[string]interface{}{
 			ExtensionKey: Extension{
 				Info: Info{
@@ -540,6 +542,7 @@ func TestCreateClientHookDoesNotSignWhenOriginMismatched(t *testing.T) {
 	hook := CreateClientHook(signer)
 	result, err := hook(context.Background(), types.PaymentRequired{
 		X402Version: 2,
+		Resource:    &types.ResourceInfo{URL: "https://example.com/resource"},
 		Accepts:     []types.PaymentRequirements{{Network: "eip155:1"}},
 		Extensions: map[string]interface{}{
 			ExtensionKey: Extension{
--- a/go/http/client_test.go
+++ b/go/http/client_test.go
@@ -169,6 +169,7 @@ func TestPaymentRoundTripper_RejectsOversizedInitial402(t *testing.T) {
 func TestPaymentRoundTripper_RejectsOversizedAuthRetry402(t *testing.T) {
 	required := types.PaymentRequired{
 		X402Version: 2,
+		Resource:    &types.ResourceInfo{URL: "https://example.com/resource"},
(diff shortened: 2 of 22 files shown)


Bartok9:
Thanks — acknowledged the verified-commits requirement.

Head `94a2c83` was left unsigned due to a local GPG lock at push time. Working through signing now and will force-push a verified amend of the same change set shortly (no content change beyond the signature).

Bartok9:
Pushed a signed amend of the same change set.

- Head is now SSH-signed (`gpg.format=ssh`) and force-pushed with lease
- Content unchanged vs prior `94a2c83` aside from the signature
- Please re-run verified-commits; Vercel deploy still needs Coinbase team auth (not something I can clear from the fork)

Happy to adjust further if the check wants a different signing method.

Bartok9:
Vercel check is `Authorization required to deploy` (Coinbase team / fork PR). Not a code failure on this branch — needs org Vercel GitHub authorization.

PR #3235, opened 2026-08-23
docs(specs): refresh the contributing guide and impl template

zjzJoez:
## Description

Two documents in `specs/` have drifted from the repo they describe, both on the path a new contributor follows.

**`specs/scheme_impl_template.md`** still asks for an `` `X-Payment` header payload ``. That is v1. A contributor who follows the official template today documents a header the v2 transport does not use — the real names are `PAYMENT-REQUIRED` and `PAYMENT-SIGNATURE` (`specs/transports-v2/http.md`). Replaced with a transport-neutral "Payment payload" section that points at the HTTP transport for header names.

**`specs/CONTRIBUTING.md`**:
- the directory diagram names `x402-specification.md` and `transports/`; the real paths are `x402-specification-v1.md` / `-v2.md` and `transports-v1/` / `transports-v2/`
- `extensions/` is missing from the diagram
- "Current schemes:" lists only `exact`, but `upto`, `auth-capture` and `batch-settlement` have since merged
- two inline references to `x402-specification.md` and one to `specs/transports/<name>.md` follow the same drift

Docs only. No normative change, no behaviour change, nothing under `specs/schemes/` touched.

## Tests

None applicable — documentation. I verified every path and scheme name against the tree at `main` rather than from memory, and the header names against `specs/transports-v2/http.md`.

## Checklist

- [x] Formatted and linted
- [x] All tests pass (n/a — docs only)
- [x] Commits are signed
- [ ] Changelog fragment — docs-only, skipped per the template

---

Found these while writing a scheme spec against the template, so the fix is first-hand rather than a drive-by. For context on what led me here: #3182 — not something I'm asking about in this PR.

*Disclosure per CONTRIBUTING.md: drafted with AI assistance; every path, header name and scheme name above was checked against the tree before submitting.*

### DIFF
--- a/specs/CONTRIBUTING.md
+++ b/specs/CONTRIBUTING.md
@@ -16,14 +16,18 @@ The `specs/` directory contains the formal specifications for the x402 protocol.
 
 ```
 specs/
-├── x402-specification.md      # Core protocol specification
+├── x402-specification-v2.md   # Core protocol specification (v1 kept alongside)
 ├── schemes/
-│   └── exact/
-│       ├── scheme_exact.md    # Scheme overview
-│       ├── scheme_exact_evm.md
-│       ├── scheme_exact_svm.md
-│       └── scheme_exact_sui.md
-├── transports/
+│   ├── exact/
+│   │   ├── scheme_exact.md    # Scheme overview
+│   │   ├── scheme_exact_evm.md
+│   │   ├── scheme_exact_svm.md
+│   │   └── …                  # one file per network
+│   ├── upto/
+│   ├── auth-capture/
+│   └── batch-settlement/
+├── extensions/                # optional functionality beyond core payment mechanics
+├── transports-v2/             # (transports-v1/ kept alongside)
 │   ├── http.md
 │   ├── mcp.md
 │   └── a2a.md
@@ -43,6 +47,9 @@ Schemes define how funds are transferred from client to server. Each scheme has:
 
 Current schemes:
 - `exact` - Transfers a specific amount for resource access
+- `upto` - Authorizes a maximum and settles the amount actually used
+- `auth-capture` - Holds funds, then captures, voids or refunds
+- `batch-settlement` - Grants access immediately and settles later, in batches
 
 ### Transports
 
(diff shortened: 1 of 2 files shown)


whawk46 (approved):
Checked this against the tree rather than against the diff, because every path here is a claim about where a file lives.

**All of it reproduces.** On `main`, `specs/` contains `x402-specification-v1.md` and `x402-specification-v2.md` — no bare `x402-specification.md` — and `transports-v1/` and `transports-v2/`, no bare `transports/`. So both links the guide carried were dead, not merely stale. `specs/extensions/`, `schemes/upto/`, `schemes/auth-capture/` and `schemes/batch-settlement/` all exist, and `transports-v2/http.md` — the target you add to the template — resolves.

I also swept all 48 files under `specs/` for the same four patterns (`X-Payment`, `transports/`, `x402-specification.md`, `specs/transports/`). **The only hits in the entire directory are the two files you touched, and your head leaves zero.** This closes the drift class rather than one instance of it, which is worth stating because it is easy to assume a docs fix is partial.

Two notes, neither blocking:

**The `└── …  # one file per network` is the better half of this PR** and it is not cosmetic. The old list hardcoded three networks; `schemes/exact/` currently holds 17. An enumerated list in a contributing guide is a promise to update it on every network that lands, and that promise had already been broken three times over. Replacing it with a rule is the fix that stays fixed.

**`scheme_impl_template.md` was the one actually costing something.** A contributor following the official template would have documented an `X-Payment` header the v2 transport does not use — so the template was manufacturing new drift on every use, not just carrying old drift. Worth landing ahead of the tree-diagram half if these ever get split.

One housekeeping note for whoever picks this up: the signing bot's warning above is already satisfied. The PR is a single commit, `26be9517`, and the API reports `verified=true`, `reason=valid` — so the "will be closed after 1 week" clock does not apply here. Flagging it because the bot comment stays in the thread after the condition clears, and it reads like a live blocker.

Approving. Nothing here touches the discovery extension or the AI-disclosure requirement.

phdargen (approved):
Good catch @zjzJoez, LGTM 🚀

Issue #3001, opened 2026-07-31
[Bug]: upto settlement amount is server-declared without protocol-verifiable usage


johnakeke:
### What happened?

**Description:**

`specs/schemes/upto/scheme_upto.md:5-13` defines `upto` as a scheme in which a maximum amount is authorized first and the settlement amount is then determined by the actual resource consumption of the request. It explicitly identifies use cases such as LLM tokens, bandwidth, and dynamic computation.

At `:50-60`, the same specification states that `PaymentRequirements.amount` is the authorization ceiling during the verification phase, while the same field becomes the actual amount requested for settlement by the resource server during the settlement phase. The resource server is responsible for selecting this value based on actual consumption; the facilitator's mandatory responsibility is to verify that the settlement amount does not exceed the maximum signed by the client.

The EVM binding at `specs/schemes/upto/scheme_upto_evm.md:171-185` again specifies that the actual amount is determined by the resource server based on consumption such as tokens, bytes, or time. Its security considerations at `:315-321` explicitly state that the client must trust the server to bill fairly, and that a malicious server can charge the full authorized maximum without corresponding actual usage.

**Impact:**

For dynamically billed LLM/agent services, the client's signature proves "how much may be charged at most," but not "how much should actually be charged." Within the authorization ceiling, an erroneous or malicious resource server can submit a settlement amount higher than the actual token/compute/bandwidth usage, and the facilitator will still regard it as valid under the protocol.

This can cause clients to be overcharged, create metering disputes with service providers, and make actual usage impossible to audit independently from an x402 receipt. Setting a lower maximum only limits the upper bound of loss; it does not verify billing correctness. This impact is a direct consequence of the specification's known trust model, not a bypass of the maximum-amount guard.

**Proposed Fix:**

The current trust-based `upto` can be retained as a compatibility profile, but the specification should provide an optional verifiable-metering profile that enables the facilitator or a subsequent auditor to verify:

```
settled_amount == pricing_function(authenticated_usage_evidence)
settled_amount <= authorized_maximum
```

Regression criteria should include: valid meter evidence succeeds; missing evidence, tampered usage, an incorrect pricing digest, duplicate meter events, and amount/evidence mismatches all fail under the strict profile.

### Steps to reproduce

1. Client requests a usage-based service.
2. Resource server returns `PAYMENT-REQUIRED`, specifying a maximum authorized amount, such as `1000`.
3. Client signs a payment payload authorizing payment up to `1000`.
[...]

PR #3237, opened 2026-08-23
docs: scope offer-receipt evidence claims to signed payload

dngene:
The extension docs say receipts "prove delivery" (§8 twice, docs page twice). The receipt payload (§5.2) is five required fields plus an optional transaction hash, server-signed. What a verifier can establish is that the server attested to delivery at a time. The spec already draws this distinction for keys in §4.5.1 ("A valid signature proves that a specific key signed the artifact. It does not prove that the key was authorized"); this PR applies the same discipline to receipt content.

Why it matters: offer-receipt is documented as the dispute-resolution answer while `upto`'s Server Trust consideration concedes servers could charge up to `amount` regardless of usage, and the receipt carries no amount or usage field (#3001). A reader today can reasonably conclude the evidence question is already solved. These edits scope the claims to what the payload supports. No schema, wire-format, or verification changes, and deliberately no new fields; whether the receipt should carry `settledAmount`/`usage` is the #3001/#3220 discussion, not this PR.

Disclosure: I build recordation infrastructure for usage-based billing (pacspace.io). Prepared with AI assistance, reviewed and verified by me.

### DIFF
--- a/docs/extensions/offer-receipt.mdx
+++ b/docs/extensions/offer-receipt.mdx
@@ -3,14 +3,14 @@ title: "Signed Offers & Receipts"
 description: "Sign offers on 402 responses and receipts on 200 responses, producing cryptographic proof-of-interaction artifacts that clients can use for reputation, auditing, or dispute resolution."
 ---
 
-The Offer & Receipt extension adds cryptographic proof-of-interaction to x402 payment flows. When enabled, your server automatically signs an **offer** on every `402` response (committing to payment terms) and a **receipt** on every `200` response (confirming service delivery). No changes to your business logic.
+The Offer & Receipt extension adds cryptographic proof-of-interaction to x402 payment flows. When enabled, your server automatically signs an **offer** on every `402` response (committing to payment terms) and a **receipt** on every `200` response (attesting to service delivery). No changes to your business logic.
 
 ## Why Enable Offer & Receipt Signing?
 
 Signed offers and receipts are portable, verifiable artifacts that any third party can check. They enable:
 
 * **Reputation systems** — Clients can attach receipts to onchain attestations as proof they actually paid for and received a service. This is the "Verified Purchase" equivalent for the open web.
-* **Dispute resolution** — Offers prove the server committed to specific terms; receipts prove delivery. If either party disputes a transaction, the signed artifacts provide evidence.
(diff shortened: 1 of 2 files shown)


DSHCorrectover:
The scoping here is correct and important — a signature proves a key signed an artifact, not that the key was authorized, and a receipt proves the server attested to delivery at a time, not that delivery actually occurred or that the amount/usage matches what was requested. This is the same distinction the USENIX x402 study draws between off-chain verification and on-chain settlement.

The gap this leaves is on the admission side: the receipt chain answers 'what did the server attest happened', but nothing in the current signed payload answers 'was this exact request authorized and conformant before it was sent'. The free-shopping and ERC-6492 findings in that study are not receipt-tampering attacks — they are requests that should never have been submitted in the form they were, and a post-hoc receipt will correctly attest to the wrong outcome.

We have been building a complementary pre-admission receipt format that uses the same JCS + Ed25519 + SHA-256 primitives as the x402 receipt stack, but signs the *request* (canonicalized tool call, arguments, declared policy bounds) before execution rather than the response after. The seven conformance dimensions include cost ceiling and payload integrity, which directly address the 'servers could charge up to amount regardless of usage' concern noted here — not by changing the receipt schema, but by producing a separate admission-time artifact that a relying party can check alongside the settlement receipt.

The composition is: pre-admission receipt (request was authorized and within declared bounds) → x402 settlement (payment was attempted) → x402 offer-receipt (server attested delivery, scoped per this PR). Three artifacts, three distinct trust assertions, same cryptographic primitives, all independently verifiable.

0rkz:
Disclosure: I work on PayPerByte/BYTEDev, where we run live x402 endpoints on Base that attach an EIP-712 attestation to each paid response — so this wording question is one we've had to settle in practice.

+1 to the scoping. The §5.2 payload is five required fields plus an optional `transaction` — no amount field, no usage field — so "the server attested to delivery at a time" is exactly what a verifier can establish from the signature alone, and extending the §4.5.1 discipline (a signature proves a key signed, not that the key was authorized) to receipt content is the right precedent.

One point worth keeping in view for the #3001/#3220 follow-ups: this scoping survives payload extensions. Our attestations are signed over the exact response bytes served (spec: https://www.payperbyte.io/attestation/eip712-payload-attestation/v1 — the EIP-712 domain there is a frozen signing namespace, not a settlement rail; payments settle on Base), and even that only upgrades the claim from "the server attested an interaction happened" to "the server attested these exact bytes were served" — tamper-evidence and evidence-toward delivery, never proof the content was correct or the service performed as promised. Adding settledAmount, usage, or a content hash widens what the attestation covers; it never turns a receipt into proof of correctness, so "attests" stays the right verb even for a maximally bound receipt.

The §10 Evidence-scope note is the piece I'd most want merged: stating what the receipt does not prove keeps relying parties fail-closed by default.

Prepared with AI assistance, human-reviewed.

PR #3241, opened 2026-08-24
fix(go/py): align default asset declaration with ts

phdargen:
## Description

Follow up to https://github.com/x402-foundation/x402/pull/3156 and https://github.com/x402-foundation/x402/pull/3154 removing duplicate network maps and better aligning with TS https://github.com/x402-foundation/x402/pull/3124

## Checklist


CarsonRoscoe (commented):
**Python**: `ethereum` and `avalanche` v1-legacy networks silently lost default-asset support. Re-add `eip155:1` / `eip155:43114` entries to `evm/default_assets.py`'s `DEFAULT_ASSETS`.

**Go**:  SVM `EnhancePaymentRequirements` dropped network validation when an explicit asset is given, as both the `svm/exact` and `svm/uto` `scheme.go` used to call `svm.GetNetworkConfig(network)` unconditionally upfront, erroring on any unsupported network. That call is not deleted, so when asset != "", the only check is `GetAssetInfo`, which succeeds for any syntactically valid looking Solana address regardless of whether `network` is real. Re-add `if _, err := svm.NormalizeNetwork(networkStr); err != nil { return requirements, err }` at the top of both `EnhancePaymentRequirements` functions

CarsonRoscoe (approved):

PR #3274, opened 2026-08-26
fix(ts/go): refactor svm upto rpc config

phdargen:
## Description

- SVM upto facilitator RPC now goes through the facilitator signer instead of a separate scheme rpc / rpcUrl. That matches the EVM facilitator signer, which already owns chain reads and writes (readContract, writeContract, sendTransaction) rather than taking a second RPC on scheme config. Pass a paced RPC client (or default URL) on the signer.
- Claim settlement simulates before the skipPreflight send and rejects with invalid_upto_svm_settlement_simulation without broadcasting
-  deposit composite simulation uses a placeholder blockhash plus replaceRecentBlockhash. Claim also overlaps the channel read with a blockhash prefetch, and open-slot freshness uses finalized getSlot on both client and facilitator. 

## Checklist

### DIFF
--- a/e2e/facilitators/go/main.go
+++ b/e2e/facilitators/go/main.go
@@ -795,6 +795,72 @@ func (s *realFacilitatorSvmSigner) ConfirmTransaction(ctx context.Context, signa
 	return fmt.Errorf("transaction confirmation timed out after %d attempts", svmmech.MaxConfirmAttempts)
 }
 
+func (s *realFacilitatorSvmSigner) GetAccountInfo(
+	ctx context.Context,
+	account solana.PublicKey,
+	network string,
+	opts *rpc.GetAccountInfoOpts,
+) (*rpc.GetAccountInfoResult, error) {
+	rpcClient, err := s.getRPC(ctx, network)
+	if err != nil {
+		return nil, err
+	}
+	return rpcClient.GetAccountInfoWithOpts(ctx, account, opts)
+}
+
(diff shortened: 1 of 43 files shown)


CarsonRoscoe (commented):
I think I see a bug,

`toFacilitatorSvmSigner(...).getProgramAccounts()` in `typescript/packages/mechanisms/svm/src/signer.ts` does:
```
const response = await rpc.getProgramAccounts(programId as never, config as never).send();
const rows = (response as { value?: ... }).value;
return rows ?? [];
```
But `@solana/kit`'s `getProgramAccounts` only wraps its results in `{ context, value }` when the caller passes `withContext: true`, otherwise `send()` resolves to the bare array directly. None of this PR's callers (i.e. `discoverChannelsByRentPayer`) pass `withContext`.

The impact is onchain rent-cleanup discovery silently finds nothing via facilitators built via the `toFacilitatorSvmSigner()`. Tests miss this because they stub `signer.getProgramAccounts` directly

To fix I would either pass `withContext: true` in the internal call and keep the `.value` unwrap, or drop the unwrap and return response directly

CarsonRoscoe (commented):
Minor bug: This PR makes the Go `UptoFacilitatorSigner` interface require `GetProgramAccounts(...) (...)`, despite the doc and the existance of a separate `assertProgramAccountsSigner` signaling that the intent is it be optional and only needed for `RentCleanupManager.Discover()`. 

Not providing it results in a panic on `NewUptoSvmScheme()` during consturction time.

If the intention is it's optional, I would drop it from the interface and check it only via `assertProgramAccountsSigner` anonymous-interface assertion

CarsonRoscoe (approved):

PR #3278, opened 2026-08-26
fix(core): return EXTENSION-RESPONSES on settle/verify

### LINKED ISSUE #3270: HTTPFacilitatorClient.settle()/verify() discard the EXTENSION-RESPONSES header instead of returning it
## Summary

`HTTPFacilitatorClient.settle()` and `.verify()` (`@x402/core/server`, also bundled in `@x402/core/http`) read the `EXTENSION-RESPONSES` response header only to `console.log` a sanitized summary (`logExtensionResponsesHeader`), then discard it. Nothing about the extension outcome reaches the object these methods return to the caller, even though `SettleResponse`/`VerifyResponse` already declare an `extensions?: Record<string, unknown>` field for exactly this purpose.

## Impact

A resource server that calls the facilitator through this official client (the only way `docs/x402-foundation.github.io`'s own guides show, as far as we've found) has no programmatic way to confirm whether a discovery extension (e.g. bazaar) was accepted or rejected on a given payment. The only visible signal is a server-side `console.log` line, easy to miss in production logging, and impossible to branch on in code without bypassing this client entirely and hand-rolling the HTTP call.

We found this live: a real seller integrating against our facilitator settled a real payment, confirmed via our own logs and catalog that cataloging succeeded, but their own `settle()` call returned no indication of it at all.

## Root cause

```js
function logExtensionResponsesHeader(response) {
  const header = response.headers.get("EXTENSION-RESPONSES");
  if (!header) return;
  try {
    const decoded = JSON.parse(safeBase64Decode(header));
    // ...builds `sanitized`...
    console.log(`[x402] extension responses: ${JSON.stringify(sanitized)}`);
  } catch {}
}
```

Called from both `settle()` and `verify()` right before `return settleResult`/`return verifyResult`, with its own return value (`undefined`) discarded:

```js
const settleResult = await parseSuccessResponse(response, settleResponseSchema, "settle");
logExtensionResponsesHeader(response);
return settleResult;
```

`settleResponseSchema` (and `verifyResponseSchema`) already define `extensions: z.record(z.string(), z.unknown()).nullish()`, so the type and the parser both already support this exact data, it just never gets populated unless the facilitator's own JSON body happens to include an `extensions` key (most don't, since the spec's own header mechanism doesn't require it).

## Proposed fix

Attach the decoded value to the returned result instead of only logging it, e.g.:

```js
function extractExtensionResponsesHeader(response) {
  const header = response.headers.get("EXTENSION-RESPONSES");
  if (!header) return void 0;
  try {
    const decoded = JSON.parse(safeBase64Decode(header));
    if (!decoded || typeof decoded !== "object" || Array.isArray(decoded)) return void 0;
    return decoded;
  } catch {
    return void 0;
  }
}
```

...then in `settle()`/`verify()`:

```js
const settleResult = await parseSuccessResponse(response, settleResponseSchema, "settle");
const headerExtensions = extractExtensionResponsesHeader(response);
if (headerExtensions && !settleResult.extensions) {
  settleResult.extensions = headerExtensions;
}
return settleResult;
```

(`!settleResult.extensions` so a facilitator that already populates the body's `extensions` field directly isn't overridden by the header, the header is a fallback source for the same data, not a second, competing one.) Keeping the existing `console.log` for backward compatibility, or folding it into a debug-level log using the now-available value, either is fine from our side.

Happy to open a PR with this change plus tests if useful, we've already verified the fix pattern works against our own facilitator by populating the body's `extensions` field on our side in the meantime.

## Severity

Not a security issue, fails closed (no funds or data at risk), and degrades functionality rather than breaking it outright, a seller can still confirm cataloging by querying the discovery endpoints directly. But it's a real, silent gap in a signal the spec's own extension mechanism exists to provide, and it's the kind of thing that's easy to build a false negative around (a seller concluding a listing failed when it actually succeeded).

## Disclosure

Drafted with AI assistance under my direction and reviewed by hand.

Bartok9:
## Description

Fixes #3270.

`HTTPFacilitatorClient.verify()` / `.settle()` previously decoded the facilitator `EXTENSION-RESPONSES` header only to `console.log` allowlisted fields, then discarded the payload. Resource servers using the official client therefore could not branch on extension outcomes after verify/settle.

Aligned with the merged Python sidechannel in #3306 (and the three-channel model in the v2 spec / bazaar docs):

- Decode `EXTENSION-RESPONSES` into optional `extensionResponses` on `VerifyResponse` / `SettleResponse` (server-internal sidechannel).
- Do **not** merge header data into `extensions` (body-only; may be forwarded to buyers).
- `encodePaymentResponseHeader` strips `extensionResponses` so it never leaks into buyer-facing `PAYMENT-RESPONSE`.
- Missing/malformed header → ignored (no throw); body `extensions` stay independent.

## Tests

- `pnpm exec vitest run test/unit/http/httpFacilitatorClient.test.ts test/unit/http/x402HTTPClient.parsePaymentResult.test.ts` in `typescript/packages/core` (51 passed)
- Prettier + eslint clean for `@x402/core`

## Checklist

- [x] I added a changelog fragment for user-facing changes (`typescript/.changeset/extension-responses-sidechannel.md`)

## AI disclosure (CONTRIBUTING soft gate)

Majority of this change was drafted by an AI agent under human direction. Intended for **human review** before merge. Commit is **signed**.

### DIFF
--- a/typescript/packages/core/src/http/httpFacilitatorClient.ts
+++ b/typescript/packages/core/src/http/httpFacilitatorClient.ts
@@ -253,22 +253,42 @@ function isAbortOrTimeoutError(error: unknown): boolean {
 const EXTENSION_RESPONSE_LOG_FIELD_ALLOWLIST = ["status", "rejectedReason", "reason", "code"];
 
 /**
+ * Decodes the facilitator `EXTENSION-RESPONSES` header into a plain object.
+ * Returns undefined when the header is missing or malformed.
+ *
+ * @param response - The HTTP response from the facilitator
+ * @returns Decoded extension responses, or undefined if missing/malformed
+ */
+function extractExtensionResponsesHeader(response: Response): Record<string, unknown> | undefined {
+  const header = response.headers.get("EXTENSION-RESPONSES");
+  if (!header) return undefined;
+  try {
+    const decoded = JSON.parse(safeBase64Decode(header));
+    if (!decoded || typeof decoded !== "object" || Array.isArray(decoded)) {
+      return undefined;
+    }
+    return decoded as Record<string, unknown>;
+  } catch {
+    return undefined;
+  }
+}
+
+/**
  * Reads the `EXTENSION-RESPONSES` header from a facilitator HTTP response and logs
  * allowlisted fields. Silently ignores malformed headers.
  *
  * @param response - The HTTP response from the facilitator
+ * @param decoded - Optional pre-decoded header object (avoids double-parse)
  */
-function logExtensionResponsesHeader(response: Response): void {
-  const header = response.headers.get("EXTENSION-RESPONSES");
-  if (!header) return;
+function logExtensionResponsesHeader(response: Response, decoded?: Record<string, unknown>): void {
+  const payload = decoded ?? extractExtensionResponsesHeader(response);
+  if (!payload) return;
   try {
-    const decoded = JSON.parse(safeBase64Decode(header));
-    if (!decoded || typeof decoded !== "object" || Array.isArray(decoded)) return;
     const sanitized: Record<string, Record<string, unknown>> = {};
-    for (const [extensionKey, payload] of Object.entries(decoded as Record<string, unknown>)) {
+    for (const [extensionKey, value] of Object.entries(payload)) {
       const source =
-        payload && typeof payload === "object" && !Array.isArray(payload)
-          ? (payload as Record<string, unknown>)
+        value && typeof value === "object" && !Array.isArray(value)
+          ? (value as Record<string, unknown>)
           : {};
       const filtered: Record<string, unknown> = {};
       for (const key of EXTENSION_RESPONSE_LOG_FIELD_ALLOWLIST) {
@@ -280,8 +300,28 @@ function logExtensionResponsesHeader(response: Response): void {
     }
     console.log(`[x402] extension responses: ${JSON.stringify(sanitized)}`);
   } catch {
-    // Ignore malformed header
+    // Ignore malformed payload shapes
   }
+}
+
+/**
+ * Attach decoded EXTENSION-RESPONSES onto `extensionResponses` (server-internal
+ * sidechannel). Never merges into `extensions`, which remains body-only and may
+ * be forwarded to buyers via PAYMENT-RESPONSE.
+ *
+ * @param result - Facilitator response object to annotate
+ * @param response - The HTTP response from the facilitator
+ * @returns The same result, possibly with extensionResponses set
+ */
+function attachExtensionResponsesFromHeader<
+  T extends { extensionResponses?: Record<string, unknown> },
+>(result: T, response: Response): T {
+  const headerExtensions = extractExtensionResponsesHeader(response);
+  logExtensionResponsesHeader(response, headerExtensions);
+  if (headerExtensions) {
+    result.extensionResponses = headerExtensions;
+  }
+  return result;
 }
 
 /**
@@ -401,8 +441,7 @@ export class HTTPFacilitatorClient implements FacilitatorClient {
       }
 
       const verifyResult = await parseSuccessResponse(response, verifyResponseSchema, "verify");
-      logExtensionResponsesHeader(response);
-      return verifyResult;
+      return attachExtensionResponsesFromHeader(verifyResult, response);
     });
   }
 
@@ -460,8 +499,7 @@ export class HTTPFacilitatorClient implements FacilitatorClient {
       }
 
       const settleResult = await parseSuccessResponse(response, settleResponseSchema, "settle");
-      logExtensionResponsesHeader(response);
-      return settleResult;
+      return attachExtensionResponsesFromHeader(settleResult, response);
     });
   }
 
--- a/typescript/packages/core/src/types/facilitator.ts
+++ b/typescript/packages/core/src/types/facilitator.ts
@@ -13,6 +13,7 @@ export type VerifyResponse = {
   invalidMessage?: string;
   payer?: string;
   extensions?: Record<string, unknown>;
+  extensionResponses?: Record<string, unknown>;
   extra?: Record<string, unknown>;
 };
 
@@ -32,6 +33,7 @@ export type SettleResponse = {
   /** Actual amount settled in atomic token units. Present for schemes like `upto` where settlement amount may differ from the authorized maximum. */
   amount?: string;
   extensions?: Record<string, unknown>;
+  extensionResponses?: Record<string, unknown>;
   extra?: Record<string, unknown>;
 };
 
--- a/typescript/.changeset/extension-responses-sidechannel.md
+++ b/typescript/.changeset/extension-responses-sidechannel.md
@@ -1,0 +1,5 @@
+---
+"@x402/core": patch
+---
+
+Populate `extensionResponses` on verify/settle from the facilitator `EXTENSION-RESPONSES` header as a server-internal sidechannel (aligned with Python). Do not merge into `extensions`; strip the sidechannel from buyer-facing `PAYMENT-RESPONSE` encoding.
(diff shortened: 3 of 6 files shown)


Bartok9:
Vercel preview is blocked on Coinbase team authorization (expected for a fork PR — we can't deploy there).

Could a **team admin / org member** please:
1. Authorize the Vercel preview for this PR, and
2. Approve the pending GitHub Actions runs on the fork

Happy to address review notes after that. Thanks.

Eras256:
Reviewed the diff, this matches the fix I proposed in #3270 exactly, including the body-wins/header-fallback precedence. Four test cases, not three: settle attaches from the header, verify attaches from the header, body extensions aren't overridden, and a malformed header is ignored without throwing, that last one wasn't something I explicitly asked for but a good defensive addition. Happy to pull this branch and verify it end-to-end against our own facilitator (the real seller integration that surfaced the bug) once it's mergeable, if that's useful before merge.

Disclosure: drafted with AI assistance under my direction and reviewed by hand. The diff and test file above were re-verified against the real PR today.

Bartok9:
Thanks @Eras256 — appreciate the re-check against #3270 and the note on the malformed-header case.

Branch is currently **mergeable** from our side (labeler + verified-commits green). Remaining noise is the expected Coinbase Vercel authorize gate on fork PRs — not a code blocker.

If you do run the end-to-end facilitator/seller check, any logs or edge cases you hit are welcome here; happy to follow up quickly. Otherwise we'll wait on org review/CI approval to land it.

Eras256:
Ran the end-to-end check. Clean, with one real edge case worth flagging.

Pulled `bartok9/extension-responses-header-3270` (`54b136d`), built `@x402/core` from it directly (not the npm-installed 2.22.0), and ran `pnpm --filter @x402/core exec vitest run test/unit/http/httpFacilitatorClient.test.ts`: 44/44 passing, matches your own test plan exactly.

Then drove a real payment through your build's own `HTTPFacilitatorClient`, against the live facilitator that surfaced the original bug (`https://periplo-testnet.fly.dev`, `agentpayments.fi`'s facilitator): `verify()` returned `isValid: true`, `settle()` returned `settleResult.extensions` populated (`{ bazaar: { status: "success" } }`), real settled transaction [`892af0974bee...`](https://stellar.expert/explorer/testnet/tx/892af0974bee7884e491dbe8d39a7450113bdfb90ae814f80dcefb3b3f774e85), Horizon-verified.

One edge case, not a bug in this PR: your branch is built off a newer `main` than the `@x402/core@2.22.0` we pin, which has added a `spendControls` client-side guard (rejects non-default assets unless explicitly allowed). Our demo resource's asset isn't in the default set, so `createPaymentPayload` threw until I called `client.setSpendControls(false)`. Unrelated to your diff, just a version-skew note from testing against a real deployment rather than mocks.

Scoping honestly: this exercised the body-wins branch, since our own `/settle` already populates `extensions` in the body independently. Didn't get a real body-less, header-only case against a live facilitator, that path is what your own mocked unit tests already cover.

Bartok9:
Thanks @Eras256 — excellent live validation.

Appreciated the Horizon-verified path on periplo-testnet: `verify` valid + `settleResult.extensions` with bazaar success is exactly the seller signal this PR is meant to surface. Noted that the live facilitator exercised the **body-wins** path; header-only remains covered by the unit suite.

On `spendControls` / non-default asset rejection vs pinned `@x402/core@2.22.0`: agreed that's mainline version skew, not this diff. `setSpendControls(false)` (or allowlisting the demo asset) is the right local workaround until you bump the pin.

No code change planned from this note. Still waiting on org review / Actions approval (Vercel authorize remains expected fork noise).

phdargen:
Thanks for the contribution @Bartok9 , could you please align this with the py patch in https://github.com/x402-foundation/x402/pull/3306

EXTENSION-RESPONSES should not leak to the client via SettleResponse.extensions but remain server internal

Bartok9:
Thanks @phdargen — aligned with the merged Python sidechannel in #3306.

**Change (pushed):**
- `HTTPFacilitatorClient` now sets `extensionResponses` from `EXTENSION-RESPONSES` (does **not** merge into `extensions`)
- `VerifyResponse` / `SettleResponse` gain optional `extensionResponses` (server-internal)
- `encodePaymentResponseHeader` strips `extensionResponses` so it never leaks into buyer-facing `PAYMENT-RESPONSE`
- Tests updated: header → sidechannel only; body `extensions` stay independent; encode exclusion; malformed header ignored

`pnpm exec vitest run test/unit/http/httpFacilitatorClient.test.ts test/unit/http/x402HTTPClient.parsePaymentResult.test.ts` — 51 passed.

Happy to tweak naming/docs further if you want exact parity with the Python field docs.

typescript/packages/core/src/http/httpFacilitatorClient.ts
```diff
@@ -280,8 +304,24 @@ function logExtensionResponsesHeader(response: Response): void {
...
   } catch {
-    // Ignore malformed header
+    // Ignore malformed payload shapes
+  }
+}
+
+/**
+ * Attach decoded EXTENSION-RESPONSES onto `extensionResponses` (server-internal
+ * sidechannel). Never merges into `extensions`, which remains body-only and may
+ * be forwarded to buyers via PAYMENT-RESPONSE. Aligns with Python PR #3306.
```
phdargen:
```suggestion
 * be forwarded to buyers via PAYMENT-RESPONSE. 
```

typescript/packages/core/src/types/facilitator.ts
```diff
@@ -13,6 +13,11 @@ export type VerifyResponse = {
   invalidMessage?: string;
   payer?: string;
   extensions?: Record<string, unknown>;
+  /**
+   * Facilitator EXTENSION-RESPONSES sidechannel (server-internal).
+   * Populated by HTTPFacilitatorClient; never serialize to buyers/PAYMENT-RESPONSE.
```
phdargen: keep transport agnostic
```suggestion
```

typescript/packages/core/src/types/facilitator.ts
```diff
@@ -32,6 +37,11 @@ export type SettleResponse = {
   /** Actual amount settled in atomic token units. Present for schemes like `upto` where settlement amount may differ from the authorized maximum. */
   amount?: string;
   extensions?: Record<string, unknown>;
+  /**
+   * Facilitator EXTENSION-RESPONSES sidechannel (server-internal).
+   * Populated by HTTPFacilitatorClient; never serialize to buyers/PAYMENT-RESPONSE.
```
phdargen: keep transport agnostic
```suggestion
```

Bartok9:
Thanks @phdargen — applied your review nits and fixed the failing format/lint on `@x402/core`.

**Updates:**
- Removed HTTP-specific JSDoc on `extensionResponses` in `VerifyResponse` / `SettleResponse` (transport-agnostic types)
- Completed JSDoc `@param` / `@returns` on the header helpers
- `encodePaymentResponseHeader` still strips `extensionResponses` (no unused binding)
- Prettier + eslint clean; unit tests 51/51

Please re-review when convenient.

phdargen:
Thanks @Bartok9, please update the PR description, add a changeset and verify your commits

Bartok9:
Thanks @phdargen — done:

1. **PR description** updated to match the #3306 sidechannel shape (server-only `extensionResponses`, no merge into `extensions`, strip on `PAYMENT-RESPONSE`).
2. **Changeset** added: `typescript/.changeset/extension-responses-sidechannel.md` (`@x402/core` patch).
3. **Commits verified**: squashed to a single **SSH-signed** commit and force-pushed (`--force-with-lease`).

Please re-run verified-commits / re-review when convenient.

phdargen (approved):
Thanks @Bartok9 , LGTM 🚀

PR #3279, opened 2026-08-26
fix(stellar): accept CAP-71 V2 address credentials for Protocol 28

jeesunikim:
# fix(stellar): accept CAP-71 V2 address credentials for Protocol 28

## Problem

Stellar Protocol 28 ("Adapter") reached testnet on **2026-08-27** and votes on mainnet **2026-09-16**. It activates the CAP-71 V2 address credential arm (`SOROBAN_CREDENTIALS_ADDRESS_V2`) on the network, so from that point on a client can build and submit V2-signed authorization entries and the host will validate them ([CAP-71](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0071.md)).

Two places in `@x402/stellar` handled only the V1 arm, so any V2 payment failed verification end to end:

1. **`gatherAuthEntrySignatureStatus` (`src/shared.ts`)** silently skipped V2 entries — a validly signed V2 entry was never counted as signed, so verification failed with `invalid_exact_stellar_payload_missing_payer_signature`.
2. **`validateAuthEntries` (`src/exact/facilitator/scheme.ts`)** rejected V2 entries outright with `invalid_exact_stellar_payload_unsupported_credential_type`.

## Fix

Both sites now accept the V1 and V2 arms via a shared `getAddressCredentials` helper. The two arms wrap the identical `SorobanAddressCredentials` structure and differ only in the preimage the signer commits to — `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION` for V1, the address-bound `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION_WITH_ADDRESS` for V2 — which the SDK's `authorizeEntry` and enforce-mode simulation already handle per arm. Source-account handling is unchanged, and delegated (`addressWithDelegates`) credentials remain rejected.

stellar-sdk exports an equivalent helper internally but deliberately does not make it public ("intentionally NOT re-exported from `base/index.js`"), hence the local copy. Note it differs on purpose: the SDK's version also unwraps `addressWithDelegates`, which this scheme rejects.

**No client-side change is needed**: `AssembledTransaction.needsNonInvokerSigningBy()` / `signAuthEntries()` already handle V2 throughout the v16 line. The dependency floor is raised to `^16.3.0` — the v16 LTS release ([stellar/js-stellar-sdk#1694](https://github.com/stellar/js-stellar-sdk/pull/1694)) that backports the Protocol 28 XDR (CAP-83/CAP-85 arms) without v17's breaking XDR API rewrite. `^16.0.1` already permitted 16.3.0; raising the floor guarantees every consumer can decode the P28 arms.

## Correction to the original premise

Recording-mode simulation on protocol 28 testnet still returns `sorobanCredentialsAddress` (V1) unless the request asks otherwise. It is the Stellar SDK 17 that returns `SOROBAN_CREDENTIALS_ADDRESS_V2` by default for recording mode simulation from Protocol 28 on.

Stellar SDK 17 update is pretty large and will do the migration in a follow up PR. The fix in this PR is what enables the facilitator to accept both V1 and V2 — without it, a facilitator rejects V2 regardless of which SDK it runs.

## Specs

- `specs/schemes/exact/scheme_exact.md`
- `specs/schemes/exact/scheme_exact_stellar.md`

Both required `sorobanCredentialsAddress` only. They now permit either address-based arm, document how the arms differ, and state explicitly that source-account and `addressWithDelegates` credentials are rejected.

## Testing

**Unit (174 pass, was 168).** V2 coverage no longer relabels a V1 fixture — entries are signed with `authorizeEntry` over the real address-bound preimage:

- A V2 signature verifies against `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION_WITH_ADDRESS` and **fails** against the legacy `ENVELOPE_TYPE_SOROBAN_AUTHORIZATION` for the same invocation and nonce — the arms are demonstrably not interchangeable.
- `gatherAuthEntrySignatureStatus`: signed V2, unsigned V2 (pending), mixed V1+V2 in one transaction, and V2 alongside a skipped source-account entry.
- Facilitator `verify()`: V2 signed over the V2 preimage accepted; unsigned V2 rejected as `missing_payer_signature`; delegated credentials rejected.
- Reverting the V2 arm in `getAddressCredentials` fails 6 of these.

**Integration, live testnet on protocol 28** (`pnpm test:integration`, verified on `@stellar/stellar-sdk` 16.0.1 and re-verified on 16.3.0):

| Test | Result |
| --- | --- |
| Verify + settle whichever arm live simulation returns | passes (arm logged: `sorobanCredentialsAddress`) |
| Verify + settle a genuinely V2-signed auth entry | passes, settles on-chain |
| V2 entry carrying a V1-preimage signature | rejected by enforce-mode simulation |

The second row is the real regression guard — it fails with `unsupported_credential_type` without this fix. The third confirms accepting the V2 arm is not a security relaxation: the network rejects a mismatched signature with `Error(Auth, InvalidAction)` / "failed ED25519 verification", surfaced as `invalid_exact_stellar_payload_simulation_failed`.

## Note for maintainers

The branch is synced with `main` (merge commit, no history rewrite). The spendControls issue that previously broke the two pre-existing integration flow tests was fixed on `main` by #3290; after the sync, the new CAP-71 tests use the same `setSpendControls(false)` approach that #3290 established, and a full live testnet run passes 11/11 integration tests.


marcelosalloum (approved):
Looking good from the perspective of Stellar! I tested it locally and it passes unit and integration tests.

PhilBot402:
@jeesunikim @marcelosalloum The facilitator-side fix looks good — accepting CAP-71 V2 address credentials via `getAddressCredentials` is the right change, and delegated / source-account credentials are still rejected.

Before merge, please also:
1. Update the normative specs (`specs/schemes/exact/scheme_exact_stellar.md` and `specs/schemes/exact/scheme_exact.md`) so they allow V1 **or** V2 address credentials. Both currently still require `sorobanCredentialsAddress` only.
2. Add V2 unit tests that actually sign the V2 preimage (including `gatherAuthEntrySignatureStatus`), not just a relabelled V1 fixture.
3. Add V2 integration coverage against Protocol 28 / recording-mode simulation, not only mocked `verify()`.

phdargen:
Thanks @jeesunikim , looks good but please fix the failing ci checks

phdargen (approved):
Thanks for the contribution @jeesunikim , LGTM 🚀

PR #3282, opened 2026-08-27
fix(go): use maxTimeoutSeconds for EIP-3009 validBefore

Tehsapper:
## Description

Makes Golang SDK EIP-3009 client implementation use `now + paymentRequirements.maxTimeoutSeconds` for `validBefore` instead of `now + 1 hour`, which is already the case for Permit2 version and TypeScript SDK.

This allows more control over EIP-3009 authorization lifetime.

## Tests

- [x] `go test ./...` from the `/go` directory

## Checklist


phdargen (approved):
Thanks for the patch @Tehsapper

Issue #3299, opened 2026-08-29
builder-code: client can forge on-chain app attribution (`a`) because the facilitator-side echo check from the spec is not implemented


AUTHENSOR:
Repository: x402-foundation/x402
Pin tested: b703a0edd82b2624dc1bdfd9a0067ef982c8a429 (2026-08-27)
Component: typescript/packages/extensions/src/builder-code, typescript/packages/core/src/server/x402ResourceServer.ts
PUBLIC-READY
MINT R04P2-M1 (attribution self-mint over unsigned extension carriage; CC-4 genus; chimera-genus kin on the v1 skip)

## Summary

The new builder-code extension appends ERC-8021 Schema 2 attribution codes to settlement calldata. The spec says two things that matter for trust: the client "MUST NOT set" the app code `a` when the server did not declare builder-code, and "the facilitator MUST verify that the `a` field echoed by the client ... exactly matches the `a` field declared by the application" (specs/extensions/builder_code.md, sections on PaymentPayload and App Code Echo Validation).

Neither rule is enforced where attribution is actually minted:

1. The resource-server echo gate (`validateExtensions`, x402ResourceServer.ts:1455) returns valid early for any payment whose `x402Version` is not 2 (line 1459) and returns valid when the server declared no extensions at all (line 1464).
2. The facilitator extension (`BuilderCodeFacilitatorExtension.buildDataSuffix`, extensions/src/builder-code/facilitator.ts:110) reads `a` from the client payment payload with a format regex as the only check (line 114) and encodes it into the calldata suffix next to the facilitator's own `w` code. It never sees the server's declaration: `DataSuffixContext` (extensions/src/builder-code/types.ts:107) carries only `paymentPayload` and `paymentRequirements`, and the extension never reads `paymentRequirements`.
3. The payment signature does not cover any of this: the EVM client signs only `from`, `to`, `value`, `validAfter`, `validBefore`, `nonce` (mechanisms/evm/src/exact/client/eip3009.ts:27), so the entire `extensions` bag, including the attribution that ends up on chain, is unsigned payer-authored data. The facilitator mechanisms pass the inbound client payload straight to the suffix builder (mechanisms/evm/src/exact/facilitator/permit2.ts:392, eip3009.ts:460).

## Impact

[...]

PR #3302, opened 2026-08-30
fix(go): validate builder-code app attribution

wnjoon:
## Summary

- reject undeclared or mismatched builder-code app attribution in the Go resource server
- preserve client-only service attribution and existing client-only extension behavior
- normalize typed extension values and validate both flat and `info`-wrapped app-code representations
- reject conflicting root and nested app-code values
- add regression coverage and a Go changelog fragment

## Root cause

`ValidateExtensions` returned success when the resource server declared no extensions and intentionally permitted client-only extension keys. That behavior is required for client-owned fields such as builder-code `s`, but it also allowed the server-owned `a` field through without an independent resource-server declaration.

The payload could subsequently reach a facilitator where builder-code attribution may be encoded into the ERC-8021 settlement suffix.

## Implementation

The Go resource server now checks builder-code app attribution before the existing empty-declaration shortcut.

It:

- rejects `a` when no trusted server declaration is available
- requires every echoed `a` representation to match the declaration
- normalizes typed Go values to their JSON wire shape before comparison
- checks both the flat protocol representation and the SDK `info` envelope
- preserves undeclared client-only `s` and other client-only extensions

Calls that do not supply declared extensions intentionally fail closed when the payload contains builder-code `a`.

## Scope

This is a resource-server-side partial mitigation.

It does not change:

- public Go API signatures
- the facilitator HTTP wire format
- legacy/v1 code
- other SDK implementations

Direct facilitator calls still lack an independently trusted resource-server declaration. Fully enforcing the facilitator-side requirement needs either trusted context propagation or an agreed fail-closed omission policy, which requires a separate cross-SDK and specification decision.

## Testing

- reproduced the undeclared builder-code attribution failure before implementation
- `make verify`
- `make build`
- `git diff --check`
- blind Claude and Codex cross-verification
- second blind verification after addressing the initial typed-value and mixed-representation findings

Refs #3299


phdargen:
Thanks for the contribution @wnjoon , could you please align this with the TS patch in https://github.com/x402-foundation/x402/pull/3313 and minimize the extension specific logic in core?

wnjoon:
@phdargen 
Thanks, that makes sense. I’ll align the Go implementation with #3313 by replacing the builder-code-specific validation branch in core with generic server-owned info-field validation, while preserving Go’s typed-value normalization. I’ll also review the facilitator-side v1 behavior for parity, update the tests, and rebase onto the latest main.

phdargen (approved):
Thanks @wnjoon , LGTM 🚀

PR #3304, opened 2026-08-31
spec(extensions): add response-provenance — receipt attests what was returned, not just that payment settled (#3234)

SolomonisBlack:
Adds `specs/extensions/response-provenance.md` — the optional extension proposed and reviewed in #3234: a paid response carries `responseHash = sha256(jcs(fixedPoint))` over the closed fixed point `{endpoint, inputs, result, method, dataVintage}`, re-derivable offline by anyone with an off-the-shelf RFC 8785 library. It upgrades "you paid for this URL" into "and here is proof of what came back," and it proves **unaltered + reproducible** — deliberately not *correct*, which stays with golden-vector reproduction.

**Why now.** Per the specs CONTRIBUTING flow, the discussion happened first: five rounds of public review on #3234 over three weeks, which found and fixed four real defects in the underlying spec and converged on every normative rule in this text. The shape is no longer speculative:

- **Three independent implementations** reproduce the normative vectors byte-for-byte with their own code: [whawk46/x402-jcs-crosscheck](https://github.com/whawk46/x402-jcs-crosscheck) (24/24); the [SCVD conformance desk](https://scvd.store)'s independent RFC 8785 implementation (its §8.1 self-proof re-derives pinned vectors before it states any finding); and @giskard09's 2026-09-15 review of this text at `e92dfa0` (normative worked vector reconstructed by hand and hashed with a third-party RFC 8785 library to a byte-exact match; two fresh preimages hashed end to end). *Corrected 2026-09-15 at @giskard09's request: an earlier version of this line miscredited argentum-core with "7 preimages, one anchored on Base mainnet"; that conflated a 2026-08-24 canonicalization cross-check over 14 of their own `action_ref` preimages with a separate repository's anchored manifest that was a proposed shape, never built.*
- **One production consumer**: SCVD's `/api/conformance/v1` surface performs this re-derivation in production today (merged 2026-08-27). Scope note from that implementer (2026-09-15): the desk's callers supply `inputs` directly, so it exercises the member-carriage table and the closure rule on every call but has no implementation evidence for or against the request→`inputs` decode branch of `e92dfa0`.
- **Sellers waiting on this text**: PayPerByte (12 mainnet feeds) has committed to map its feeds to the landed field and report what breaks; Sirenic (80 mainnet listings) will evaluate emitting on closure-safe routes once the field name and shape land here — "one spec, not fifteen sellers' dialects."
- **Live issuer**: [x402toll.com](https://x402toll.com) (59 endpoints) emits it now, with a free re-check endpoint and free golden vectors.

**What the text contains** (all of it thread-derived, contributors credited by name in the file):

- The five-member closed fixed point + RFC 8785 canonicalization (with the three added restrictions)
- The envelope on the response body, and the beside-not-inside rule for carrying it in signed receipts
- Issuer requirements: the closure rule (routes that can't be re-derived MUST emit nothing) and `dataVintage` completeness (all sources, oldest vintage, own precision)
- **Verification outcomes (normative)**: the three terminal states — `no_claim` / `unverifiable` / `contradicted` — with the no-collapse rule, why absence is never a defect, and the verifier self-proof requirement before a `contradicted` finding
- Requirements for verifiers that publish **consumable verdicts** (vector-proven + versioned, exact-artifact binding, self-auditable signed verdicts, bounded vocabulary)
- Informative companions: `did:artifact` as a digest-equal encoding (verified against OMA3's own SDK), and byte-exact attestation as a distinct complementary claim
- Security considerations, including why verify-time has no network attack surface

Happy to split, trim, or re-shape to maintainer preference — the normative content is what three implementations already agree on; the packaging is flexible.

Disclosure: prepared with AI assistance, human-reviewed. Every factual claim above is checkable at the links; the conformance vectors are published and pinned at [golden-vector-provenance v0.7.0](https://github.com/SolomonisBlack/golden-vector-provenance/releases/tag/v0.7.0).

### DIFF
--- a/specs/extensions/response-provenance.md
+++ b/specs/extensions/response-provenance.md
@@ -1,0 +1,220 @@
+# Extension: `response-provenance`
+
+## Summary
+
+The `response-provenance` extension lets a paid response carry a hash that **anyone can re-derive
+offline** — upgrading "you paid for this URL" into "and here is proof of what came back." A resource
+server attaches `responseHash = sha256(jcs(fixedPoint))`, a SHA-256 over the RFC 8785 (JCS) canonical
+form of the closed fixed point `{endpoint, inputs, result, method, dataVintage}`. A buyer, auditor,
+or indexer re-derives it with any off-the-shelf JCS library: no key exchange, no callback to the
+issuer, no resolver, no trust in any party.
+
+The extension proves the response is **unaltered and reproducible**. It deliberately does **not**
+prove the response is *correct* — correctness is established only by reproduction against published
+golden vectors. Keeping those two claims separate is the design's load-bearing rule.
+
+Proposed and reviewed in [issue #3234](https://github.com/x402-foundation/x402/issues/3234). At the
+time of this PR the shape has three independent implementations reproducing the published
+conformance vectors byte-for-byte, one production conformance desk consuming it, and two additional
+sellers evaluating adoption pending this text — PayPerByte
+([in-thread](https://github.com/x402-foundation/x402/issues/3234)) and Sirenic
+([sirenic-eu/sirenic-examples#2](https://github.com/sirenic-eu/sirenic-examples/issues/2)). See
+[Status](#status-and-prior-art).
+
+---
+
+## Motivation
+
+A settlement receipt binds `resourceUrl`, `payer`, `network`, and `txHash` — it proves payment
+settled, not what the server returned. Today sellers fill that gap with incompatible private
+spellings (`record_hash`, `verdict_hash`, EIP-712 body attestations, "SHA-256 fingerprint" prose): a
+survey of listed x402 services found more than a dozen distinct response-hash dialects, none
+mutually verifiable. One shared, re-derivable field lets every relying party check every seller
+identically — and lets the evidence outlive the seller, since verification requires nothing from
+them.
+
+---
+
+## The fixed point
+
+The hashed object is exactly five members — a **closed set**, `GVP-FixedPoint/1`:
+
+| Member | Meaning |
+|---|---|
+| `endpoint` | the resource path the buyer called (binds the *question*, not just the answer) |
+| `inputs` | the request inputs, as the buyer sent them |
+| `result` | the substantive answer object |
+| `method` | human-readable statement of the computation performed |
+| `dataVintage` | the vintage of the data the result was computed from |
+
+All five members are REQUIRED and no other member is permitted in the hashed object. A verifier
+always recomputes over the five spec members, so an issuer cannot shrink the fixed point (e.g. drop
+`inputs`) and remain internally consistent. Verifiers use the `inputs` from **their own request**,
+never a server echo.
+
+`fixedPointVersion` names the frozen rule set and travels **beside** the hash, never inside it.
+`GVP-FixedPoint/1` will never change; a different member set would be a new identifier.
+`GVP-FixedPoint/2` has the same five members and identical hashing; it additionally constrains the
+`dataVintage` lexical grammar to ISO 8601 reduced precision so vintages are machine-comparable.
+
+### Canonicalization
+
+Canonical JSON **is [RFC 8785 (JCS)](https://www.rfc-editor.org/rfc/rfc8785)**. This extension
+defines no bespoke serialization. Implementations use any verified JCS library (RFC 8785 Appendix G
+lists several). Three additional restrictions apply: non-finite numbers (NaN, ±Infinity) MUST be
+rejected, object member names MUST be unique, and nesting depth SHOULD be bounded (reference
+implementations reject depth > 100).
+
+---
+
+## Envelope
+
+On a paid response body:
+
+```json
+{
+  "extensions": {
+    "response-provenance": {
+      "responseHash": "sha256:33907ce0528475a75e5ab406f00cd1a1e612a762263ca98c1b7597f4fe2a6d49",
+      "fixedPointVersion": "GVP-FixedPoint/1",
+      "spec": "https://github.com/SolomonisBlack/golden-vector-provenance"
+    }
+  }
+}
+```
+
+- `responseHash` (REQUIRED): `"sha256:" + lowercase-hex` SHA-256 of the JCS bytes of the fixed point.
+- `fixedPointVersion` (REQUIRED): `"GVP-FixedPoint/1"` or `"GVP-FixedPoint/2"`.
+- `spec` (OPTIONAL): URL of the fixed-point specification.
+
+When the seller also emits a signed receipt (e.g. the `offer-and-receipt` extension), the
+`responseHash` member SHOULD be carried inside the signed payload, binding content proof to payment
+proof under one signature. The identifier and any envelope of the hash sit **beside** the hashed
+object, never inside it — including the hash (or any identifier derived from it) inside the hashed
+octets is a self-reference cycle and is prohibited.
+
+Because the hash is computed over the canonical form of the parsed JSON value, transport-layer
+transformations (gzip, brotli, chunking, insignificant whitespace) do not affect it. The party that
+**produces the result** computes the hash; an intermediary that did not produce the bytes cannot
+honestly attest them.
+
+---
+
+## Issuer requirements
+
+1. **Closure (normative).** An issuer MUST NOT emit a `responseHash` for a response whose `result`
+   depends on any input not present in the fixed point. Cached model-derived answers, live-merge
+   feeds, and routes with unenumerable sources MUST emit nothing rather than a hash that cannot be
+   re-derived. Emitting no claim is the *compliant* output for such routes.
+2. **`dataVintage` completeness (normative).** `dataVintage` MUST characterise **all** sources that
+   contributed to `result`. Where contributing sources carry different vintages, the value MUST be
+   the **oldest**, stated at its own precision. A seller who cannot enumerate the contributing
+   sources MUST NOT emit a `responseHash` at all — an unenumerable source is a hidden input, and the
+   closure rule already forbids issuing over one.
+3. **No mode flags.** `responseHash` carries exactly one claim: re-derivable by anyone. An
+   integrity-only claim for non-re-derivable content (e.g. "unaltered relative to what the issuer
+   stored") is a different claim and MUST NOT be emitted under this field name.
+
+---
+
+## Verification outcomes (normative)
+
+Evaluating a response against this extension terminates in exactly one of three states.
+Implementations MUST report which state obtained and MUST NOT collapse them:
+
+1. **`no_claim`** — the extension member is absent. This is NOT a failure state and MUST NOT be
+   scored as one: under the closure rule, absence is sometimes the *required* output. Treating
+   absence as a defect pressures sellers of non-closed routes to emit hashes nobody can re-derive —
+   manufacturing the false confidence this extension exists to remove.
+2. **`unverifiable`** — the member is present but cannot be evaluated: unparseable, wrong shape, or
+   a `fixedPointVersion` the verifier does not implement.
+3. **`contradicted`** — re-derivation ran and the hash does not match. Before reporting this state a
+   verifier MUST have reproduced the published conformance vectors with its own implementation — a
+   mismatch report from a canonicalizer that cannot reproduce the golden vectors is noise, and the
+   most common cause of a mismatch in practice is a verifier-side bug, not tampering. A
+   `contradicted` finding is a disjunction — *the artifact was altered, or it was issued in
+   violation of the closure rule* — and reports SHOULD NOT attribute a branch they cannot see.
+
+(`verified` — the hash re-derives — is the fourth outcome; it means *re-derives*, never *correct*.)
+
+What a relying application **does** with a terminal state — complete, withhold, reconcile, dispute —
+is caller policy, owned above this extension. The extension's contract ends at naming the state
+precisely.
+
+### Consumable external verdicts
+
+Local re-verification is the trust model: the check is one JCS pass and one SHA-256 over bytes the
+relying party already holds, fully offline. A third-party verifier's `verified`, consumed without
+re-running, replaces that model with trust in the verifier. Negative verdicts (`contradicted`,
+`unverifiable`) are safe to consume conservatively; a consumed `verified` is a deliberate trust
+shift. A verifier that publishes verdicts for consumption by others MUST:
+
+1. **Be vector-proven and versioned** — publish its reproduction of the conformance vectors; each
+   verdict cites verifier version and vector-set version.
+2. **Bind the artifact, not a description** — each verdict embeds the exact `responseHash` and
+   `fixedPointVersion` evaluated.
+3. **Be auditable by the mechanism it reports on** — verdict documents are signed under a pinned,
+   published key and carry their own fixed point and `responseHash`.
+4. **Use the bounded vocabulary** — the four outcomes above, with `verified` meaning re-derives and
+   `contradicted` stated as the disjunction with an explicit falsification path and scope limit
+   (e.g. "the artifact as given to us; we did not fetch it from the issuer origin").
+
+---
+
+## Companion encodings (informative)
+
+- **`responseArtifactId` (DID).** The W3C-registered generative method `did:artifact` (OMA3)
+  canonicalizes JSON artifacts with the same RFC 8785 and hashes with SHA-256: the CIDv1 multihash
+  inside such a DID carries the byte-identical digest, verified concretely in the proposal thread.
+  A `responseArtifactId` MAY sit beside `responseHash` as an optional companion; because the method
+  is generative, any party can mint it from the fixed point after the fact. It is an encoding of the
+  same primitive, not a new one.
+- **Byte-exact attestation.** JCS re-derivation binds the parsed JSON *value* across heterogeneous
+  runtimes. Sellers who need wire-level attestation of the exact bytes served (raw binary, payloads
+  with non-canonical number spellings) may emit a raw-byte signature alongside `responseHash` as a
+  complementary — and different — claim. Migrating sellers should know which claim they are making;
+  where the served text equals the ES6 serialization of the parsed value, the two coincide.
+
+---
+
+## Conformance
+
+The fixed-point specification, JSON Schema, JS + Python reference implementations, published
+conformance vectors, and a byte-equivalence gate against an independent RFC 8785 implementation are
+maintained at [golden-vector-provenance](https://github.com/SolomonisBlack/golden-vector-provenance)
+(spec CC-BY-4.0, code Apache-2.0), pinned releases from
+[`v0.7.0`](https://github.com/SolomonisBlack/golden-vector-provenance/releases/tag/v0.7.0). An
+implementation conforms when it reproduces the published vectors byte-for-byte.
+
+## Security considerations
+
+- The hash binds the **question and the answer** (`endpoint` + `inputs` are required members), so a
+  response cannot be replayed as the answer to a different question without changing the hash.
+- The extension is payment-rail-agnostic and adds no network surface at verify time: re-derivation
+  is offline, so there is no resolver, callback, or fetch to attack.
+- A bare (unsigned) response's extension block can be stripped or forged in transit; binding it
+  inside a signed receipt (see [Envelope](#envelope)) makes stripping or forging fail the signature.
+- The hash proves nothing about correctness. Sellers remain able to compute wrong answers
+  reproducibly; golden vectors and reproduction are the correctness mechanism, and consumers MUST
+  NOT present `verified` as an accuracy claim.
+
+## Status and prior art
+
+Shaped by five rounds of public review on
+[#3234](https://github.com/x402-foundation/x402/issues/3234). Independent implementations
+reproducing the published vectors with their own code:
+[whawk46/x402-jcs-crosscheck](https://github.com/whawk46/x402-jcs-crosscheck) (24/24),
+[giskard09/argentum-core](https://github.com/giskard09/argentum-core) (7 preimages, one anchored on
+Base mainnet before it was checked), and the [SCVD conformance desk](https://scvd.store) — an
+independent evidence observatory whose production `/api/conformance/v1` surface consumes this
+extension today. Live issuer: [x402toll.com](https://x402toll.com) (59 endpoints; free re-check at
+`POST /v1/verify-hash`; free golden vectors at `GET /v1/golden/:id`).
+
+Reviewer-driven normative rules in this text credit: **whawk46** (standalone-vs-receipt layering,
+JCS cross-check), **giskard09** (anchored preimages; the never-collapse failure-mode discipline),
+**kopko13 / Sirenic** (`dataVintage` completeness — "a freshness field naming three of four sources
+reads as complete"; the two-claims/no-flag resolution; adoption-pending-this-text stated in
+[sirenic-eu/sirenic-examples#2](https://github.com/sirenic-eu/sirenic-examples/issues/2)), **cv-scvd / SCVD** (production verdict
+shape: finding-not-policy, the mismatch disjunction), **0rkz / PayPerByte** (byte-exact attestation
+as a distinct complementary claim), **alftom / OMA3** (`did:artifact` companion encoding),
+**postdvk** (the terminal-states question this text's outcomes section answers).
--- a/specs/extensions/response-provenance.md
+++ b/specs/extensions/response-provenance.md
@@ -9,17 +9,17 @@ form of the closed fixed point `{endpoint, inputs, result, method, dataVintage}`
 or indexer re-derives it with any off-the-shelf JCS library: no key exchange, no callback to the
 issuer, no resolver, no trust in any party.
 
-The extension proves the response is **unaltered and reproducible**. It deliberately does **not**
-prove the response is *correct* — correctness is established only by reproduction against published
-golden vectors. Keeping those two claims separate is the design's load-bearing rule.
+Scope of the claim: the extension proves the response is **reproducible** always, and **unaltered**
+when the hash is bound inside a signed receipt (a bare extension block on an unsigned response can be
+stripped or forged in transit and re-hashed — see [Signed carriage](#signed-carriage) and
+[Security considerations](#security-considerations)). It deliberately does **not** prove the response
+is *correct* — correctness is established only by reproduction against published golden vectors.
+Keeping those claims separate is the design's load-bearing rule.
 
-Proposed and reviewed in [issue #3234](https://github.com/x402-foundation/x402/issues/3234). At the
-time of this PR the shape has three independent implementations reproducing the published
-conformance vectors byte-for-byte, one production conformance desk consuming it, and two additional
-sellers evaluating adoption pending this text — PayPerByte
-([in-thread](https://github.com/x402-foundation/x402/issues/3234)) and Sirenic
-([sirenic-eu/sirenic-examples#2](https://github.com/sirenic-eu/sirenic-examples/issues/2)). See
-[Status](#status-and-prior-art).
+Proposed and reviewed in [issue #3234](https://github.com/x402-foundation/x402/issues/3234) (see
+[Status](#status-and-prior-art)). **This file is normative.** Where any external document — including
+the upstream fixed-point specification this design derives from — diverges from this file, this file
+governs for x402 purposes.
 
 ---
 
@@ -27,9 +27,8 @@ sellers evaluating adoption pending this text — PayPerByte
 
 A settlement receipt binds `resourceUrl`, `payer`, `network`, and `txHash` — it proves payment
 settled, not what the server returned. Today sellers fill that gap with incompatible private
-spellings (`record_hash`, `verdict_hash`, EIP-712 body attestations, "SHA-256 fingerprint" prose): a
-survey of listed x402 services found more than a dozen distinct response-hash dialects, none
-mutually verifiable. One shared, re-derivable field lets every relying party check every seller
+spellings (EIP-712 body attestations, bespoke `record_hash` / `verdict_hash` members, "SHA-256
+fingerprint" prose). One shared, re-derivable field lets every relying party check every seller
 identically — and lets the evidence outlive the seller, since verification requires nothing from
 them.
 
@@ -37,33 +36,108 @@ them.
 
 ## The fixed point
 
-The hashed object is exactly five members — a **closed set**, `GVP-FixedPoint/1`:
+The hashed object is exactly five members — a **closed set**:
 
 | Member | Meaning |
 |---|---|
 | `endpoint` | the resource path the buyer called (binds the *question*, not just the answer) |
-| `inputs` | the request inputs, as the buyer sent them |
+| `inputs` | the request inputs, per the [request mapping](#request--inputs-mapping) |
 | `result` | the substantive answer object |
 | `method` | human-readable statement of the computation performed |
 | `dataVintage` | the vintage of the data the result was computed from |
 
 All five members are REQUIRED and no other member is permitted in the hashed object. A verifier
-always recomputes over the five spec members, so an issuer cannot shrink the fixed point (e.g. drop
-`inputs`) and remain internally consistent. Verifiers use the `inputs` from **their own request**,
-never a server echo.
+always recomputes over the five members, so an issuer cannot shrink the fixed point (e.g. drop
+`inputs`) and remain internally consistent.
 
-`fixedPointVersion` names the frozen rule set and travels **beside** the hash, never inside it.
-`GVP-FixedPoint/1` will never change; a different member set would be a new identifier.
-`GVP-FixedPoint/2` has the same five members and identical hashing; it additionally constrains the
-`dataVintage` lexical grammar to ISO 8601 reduced precision so vintages are machine-comparable.
+`fixedPointVersion` names the frozen rule set and travels **beside** the hash, never inside it —
+including the hash, or any value derived from it, inside the hashed octets is a self-reference cycle
+and is prohibited. Two versions exist; both hash the same five members identically:
+
+- **`GVP-FixedPoint/1`** — the base rule set, frozen; a different member set would be a new
+  identifier.
+- **`GVP-FixedPoint/2`** — identical hashing; additionally, `dataVintage` MUST be an ISO 8601
+  calendar date at reduced precision, exactly one of `YYYY`, `YYYY-MM`, or `YYYY-MM-DD` (regex
+  `^[0-9]{4}(-[0-9]{2}(-[0-9]{2})?)?$`, and a real calendar date at the stated precision). Precision
+  is significant — `"2026"`, `"2026-07"`, and `"2026-07-01"` are three different vintages and hash
+  differently. No time component, no timezone. Under this grammar, lexical comparison equals
+  chronological comparison. A `/2` verifier MUST reject a fixed point whose `dataVintage` does not
+  match the grammar (outcome: `unverifiable`).
+
+### Member carriage — where a verifier finds each member
+
+An issuer emitting this extension MUST make every fixed-point member recoverable as follows:
+
+| Member | Carriage |
+|---|---|
+| `endpoint` | the request path as served (path only; see [Security considerations](#security-considerations) for issuer scoping) — also RECOMMENDED as a top-level response-body member `endpoint` |
+| `inputs` | derived from the request per the [request mapping](#request--inputs-mapping) |
+| `result` | REQUIRED top-level response-body member `result` |
+| `method` | REQUIRED member `method` of the extension object |
+| `dataVintage` | REQUIRED member `dataVintage` of the extension object |
+
+Carrying `method` and `dataVintage` in the extension object (beside the hash, not inside the hashed
+octets) is permitted and required precisely because the self-reference prohibition covers only the
+hash and values derived from it — hashed members restated outside the hashed object are inputs to
+re-derivation, not cycles.
+
+### Request → `inputs` mapping
+
+Two honest implementations must not disagree on what `inputs` is. The normative default mapping:
+
+- **Request with a JSON body** (typically POST/PUT): `inputs` is the parsed JSON body, as a JSON
+  value. The body MUST be [I-JSON](https://www.rfc-editor.org/rfc/rfc7493) (see
+  [Numbers](#numbers-i-json)).
+- **Request without a body** (typically GET): `inputs` is a JSON object built from the URL query
+  string: each key and value percent-decoded as UTF-8; **values remain strings** (`"42"`, never
+  `42` — no type coercion); a key occurring more than once maps to an array of its values in order
+  of appearance; key order is irrelevant (JCS sorts). An empty query yields `{}`.
+
+A seller whose route needs a different mapping MUST use a JSON body instead. No other mapping is
+conformant under this extension.
 
 ### Canonicalization
 
 Canonical JSON **is [RFC 8785 (JCS)](https://www.rfc-editor.org/rfc/rfc8785)**. This extension
 defines no bespoke serialization. Implementations use any verified JCS library (RFC 8785 Appendix G
-lists several). Three additional restrictions apply: non-finite numbers (NaN, ±Infinity) MUST be
-rejected, object member names MUST be unique, and nesting depth SHOULD be bounded (reference
-implementations reject depth > 100).
+lists several). Three additional restrictions apply, and a fixed point violating any of them is
+**`unverifiable`** (see [Verification outcomes](#verification-outcomes-normative)): non-finite
+numbers (NaN, ±Infinity) MUST be rejected, object member names MUST be unique, and nesting depth
+SHOULD be bounded (reference implementations reject depth > 100).
+
+### Numbers (I-JSON)
+
+RFC 8785 serializes numbers per IEEE-754 double precision. Fixed-point values MUST conform to
+I-JSON ([RFC 7493](https://www.rfc-editor.org/rfc/rfc7493)): integers outside the exact-double range
+(beyond ±2^53−1) — raw wei-denominated values being the obvious case in this ecosystem — MUST be
+carried as strings, not JSON numbers, or precision is silently lost at the parse step the whole
+design depends on.
+
+### Worked vector (normative)
+
+A conforming implementation MUST reproduce this vector byte-for-byte. Fixed point:
+
+```json
+{
+  "endpoint": "/v1/echo-sum",
+  "inputs": { "a": 2, "b": 3 },
+  "result": { "sum": 5 },
+  "method": "sum = a + b, integer addition",
+  "dataVintage": "2026-07"
+}
+```
+
+JCS canonical form (134 bytes, one line):
+
+```
+{"dataVintage":"2026-07","endpoint":"/v1/echo-sum","inputs":{"a":2,"b":3},"method":"sum = a + b, integer addition","result":{"sum":5}}
+```
+
+`responseHash`:
+
+```
+sha256:81ea1f2227fd9df5b868954e6d26d091810352f148dade483b260844788ede03
+```
 
 ---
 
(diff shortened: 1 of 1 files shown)


0rkz (changes requested):
**Direction endorsed; text not yet implementable — request changes.**

Context: PayPerByte runs 12 x402 pay-per-call feeds on Base mainnet with our own EIP-712 response attestation; in #3234 we committed to map our feeds to this field once it lands and report what breaks. This review is the first installment of that report, at the spec-text level: we attempted to implement re-derivation *from this file alone*, and the items below are where it stops — before our feeds' quirks even enter. A shared, re-derivable key beats fifteen private dialects, ours included, so everything here comes with a concrete fix attached.

### Blocking — a verifier cannot re-derive from this text

**1. Four of five preimage members have no defined carriage.** The envelope normatively defines only `responseHash`, `fixedPointVersion`, and `spec`. `endpoint` and `inputs` come from the buyer's own request, but `result`, `method`, and `dataVintage` have no stated location anywhere in the file. A verifier cannot recompute `sha256(jcs(fixedPoint))` over members it cannot locate. Fix: require normatively that the response carries every fixed-point member not derived from the request, and say where (body members with these names? a per-route mapping the seller publishes?).

**2. The HTTP request → `inputs` mapping is undefined.** For a GET feed: raw query string or parsed object? URL-decoded? `"42"` or `42`? What ordering and what treatment of repeated keys? Two honest implementations will disagree here, and this is exactly where our feeds will break first. The fix needs a design decision (a normative mapping, or a seller-published per-route mapping bound under `method`), but the gap needs naming before merge.

**3. `endpoint` is "the resource path" — it does not bind the host, and the security section overclaims.** "A response cannot be replayed as the answer to a different question" is false across issuers: the same path and inputs served by a different host re-derive identically. Either define `endpoint` to include the origin, or scope the replay claim to a single issuer.

### Normative core — verification outcomes

**4. The outcome count contradicts itself.** The section opens "terminates in exactly one of three states" and enumerates three; a parenthetical then calls `verified` "the fourth outcome"; external-verdict requirement 4 mandates "the four outcomes above." As written, the MUST-report / MUST-NOT-collapse duties attach only to the three enumerated states — so the state implementations will emit most often sits outside the normative enumeration. Enumerate four, fix the intro, delete the parenthetical. Related: the canonicalization rejections (NaN, duplicate member names, depth) are MUSTs with no outcome mapping — presumably `unverifiable`; say so.

**5. External-verdict binding is unsatisfiable for `no_claim`.** Requirement 2 — "each verdict embeds the exact `responseHash` and `fixedPointVersion` evaluated" — has no referent when the extension member is absent, yet requirement 4 mandates `no_claim` in the vocabulary. Scope requirement 2 to member-present verdicts and define what a `no_claim` verdict binds instead (e.g. a hash of the evaluated body plus request identity).

**6. The `contradicted` disjunction omits its own most-common branch.** The paragraph defines the finding as "altered OR issued in violation of the closure rule," then states the most common cause of a mismatch is a verifier-side bug. §8.1 self-proof reduces that branch but cannot eliminate it — vectors don't cover every input. Either make it a three-branch disjunction or state why self-proof is held to close the third branch.

### Make this file the authority

**7. Everything that defines conformance lives outside x402 change control.** The fixed-point specification, JSON Schema, and conformance vectors are in one contributor's personal repo; "pinned releases from `v0.7.0`" is an open range, not a pin; `GVP-FixedPoint/2`'s grammar ("ISO 8601 reduced precision") is not implementable from this file; the single example hash has no in-file preimage; and nothing states which document wins when the external spec diverges — a divergence the thread already schedules. Fix as a unit: vendor the fixed-point definition and at least one complete worked vector (fixed point → JCS bytes → hash) into this file, pin exact releases plus content hashes per `fixedPointVersion`, spell out /2's grammar, and state that this file is normative on conflict. The stated licenses (CC-BY-4.0 / Apache-2.0) permit all of it.

### Signing guidance

**8. Sign `fixedPointVersion` too.** The receipt guidance carries only `responseHash` inside the signed payload. An intermediary that alters `result` can also rewrite `fixedPointVersion` to an unimplemented value and steer the outcome from `contradicted` to `unverifiable` — choosing which negative state its tampering produces. One extra signed field; verdict requirement 2 already treats the version as binding-relevant.

**9. Scope the Summary's claim.** "Proves the response is unaltered and reproducible" holds only under signed carriage; a bare extension block is forge-by-recompute (alter `result`, recompute the hash). The security section already knows this ("can be stripped or forged in transit"); the Summary should carry the same scope: unaltered *when bound in a signed receipt*, reproducible always.

### Third-party verifiers

**10. The own-request rule is impossible for the roles this spec advertises.** "Verifiers use the `inputs` from **their own request**, never a server echo" — but the Summary promises auditors and indexers can re-derive, and the external-verdict section provisions verifiers whose scope limit is "the artifact as given to us." A non-requesting verifier holds only the recorded inputs, i.e. the forbidden echo; as written, the production consumer the Status section cites is non-conformant. Split the rule by role: requesters MUST recompute over their own sent inputs; non-requesting verifiers verify inputs-as-recorded and MUST disclose that in the scope limit.

### Numbers

**11. State the I-JSON constraint.** RFC 8785 serializes numbers as IEEE-754 doubles. Any feed emitting integers beyond 2^53 as JSON numbers — raw wei-denominated values being the obvious case in this ecosystem — silently loses precision at the parse step the whole design depends on. Require I-JSON (RFC 7493) for fixed-point values and state that larger integers go as strings. Speaking as a seller about to map oracle feeds onto this: this would have been our first "what breaks" report; cheaper to fix in the text.

### Companion encodings (the paragraph credited to us)

**12. Accurate, with one precision fix.** The byte-exact attestation paragraph states our distinction correctly — thank you. But "where the served text equals the ES6 serialization of the parsed value, the two coincide" overstates: that condition makes the *number-formatting divergence* disappear, while the claims still differ in scope — a raw-byte signature binds the served body; `responseHash` binds the five-member fixed point, including `endpoint` and `inputs`, which the body may not contain. Suggest: "…the two claims agree on the payload's formatting; their scopes remain distinct."

### Status section — and one correction for reviewers arriving from #3234

**13. Move the adoption evidence out of the spec file.** Point-in-time counts (24/24, 7 preimages, 59 endpoints, "today") and live third-party URLs will rot inside a merged spec; the survey behind "more than a dozen dialects" is neither linked nor dated; and two legs of the evidence — x402toll.com and the conformance repo — are the proposal author's own properties, disclosed forthrightly in the thread but invisible in this file. "W3C-registered" also drops the thread's own precision note (a DID Extensions registry listing; the method spec is an unofficial draft). Move adoption evidence to the PR body, keep a dated pointer to #3234, add the affiliation sentence, fix the DID wording. And since PayPerByte is cited in this section: our accurate status is *committed to map our 12 feeds to the landed field and report what breaks* — this review is the first such report.

**14. Review against this file, not the thread's descriptions of it.** The latest #3234 comment ([5473364013](https://github.com/x402-foundation/x402/issues/3234#issuecomment-5473364013)) describes `fixedPointVersion` as a required member *inside* a five-member fixed point of `{fixedPointVersion, status, contentType, contentDigest, responseHash}` under version string `x402-provenance-fixedpoint/1.0`. None of that appears in this file, which says the opposite: the hashed members are `{endpoint, inputs, result, method, dataVintage}`, and `fixedPointVersion` "travels **beside** the hash, never inside it" — placing `responseHash` inside the hashed object would be exactly the self-reference cycle the Envelope section prohibits. Flagged so the thread doesn't get cited as though it were the text.

---

With items 1–4 and 7 addressed, this is the field we'd adopt.

Disclosure per CONTRIBUTING: prepared with AI assistance, human-reviewed.

SolomonisBlack:
@0rkz This is the review the text needed — an implementation attempt from the file alone is the only honest test of a spec, and every item held up. All 14 are addressed in the new revision (single signed commit). The two that needed design decisions, decided:

**1 — Member carriage (blocking).** Now a normative carriage table: `result` MUST be the top-level body member `result`; `method` and `dataVintage` MUST be members of the extension object itself, restated beside the hash (the self-reference prohibition covers the hash and values derived from it — hashed members restated outside the hashed octets are re-derivation inputs, not cycles); `endpoint` is the request path as served; `inputs` per item 2. A verifier can now locate all five members from the response + request alone.

**2 — Request → `inputs` mapping (blocking).** Normative default, no options: JSON body → parsed body (I-JSON). GET → object from the query string, percent-decoded UTF-8, **values remain strings** (`"42"` never `42`), duplicate keys → array in order of appearance. A route needing anything else MUST use a JSON body. Your feeds should now get a deterministic answer or a deterministic "use POST."

**3 — `endpoint`/origin.** Chose scoping over redefinition (changing the member set would fork `GVP-FixedPoint/1`): the replay claim is now explicitly *within a single issuer's namespace*, with the cross-issuer identity assigned to the signature layer, stated in Security considerations.

**4 — Four states.** Enumerated four, intro fixed, parenthetical gone, and the canonicalization rejections (non-finite, duplicate keys, depth, /2 grammar failures) are mapped to `unverifiable` explicitly.

**5 — `no_claim` binding.** Requirement 2 is now scoped: member-present verdicts bind `responseHash` + `fixedPointVersion`; `no_claim` verdicts bind a SHA-256 over the JCS form of the evaluated body plus the recorded request identity — your suggested shape.

**6 — Three-branch disjunction.** Adopted verbatim in spirit: altered / closure-violating issuance / verifier defect, with §-self-proof described as shrinking (not closing) the third branch and a SHOULD to state vector coverage.

**7 — Authority.** The file now carries a complete normative worked vector (fixed point → 134-byte JCS line → hash), the `/2` grammar in full (three forms, regex, precision-significant, lexical=chronological), exact pins (tag commit `91c12c1…`, spec-file SHA-256 `c48f327…`), and a twice-stated rule that **this file governs on divergence**. Your framing was right: mergeable beats unilaterally controlled.

**8** — `fixedPointVersion` joins `responseHash` inside the signed payload, with your tamper-steering rationale stated. **9** — Summary now reads: reproducible always, unaltered under signed carriage. **10** — Verifier roles split: requesters MUST use their own sent inputs; non-requesting verifiers verify inputs-as-recorded and MUST disclose it — with the residual ("claimed question, not witnessed") named rather than papered over. **11** — I-JSON required; >2^53 integers as strings, with your wei example as the stated motivation. **12** — Your sentence, nearly verbatim: the claims agree on formatting; their scopes remain distinct. **13** — Status slimmed to a dated pointer; point-in-time counts moved out; affiliation disclosure added in-file ("the live reference issuer and the upstream conformance repository are operated by this extension's author"); DID wording corrected to registry-listing + unofficial draft. Your status is now recorded exactly as you stated it.

**14** — Correct, and worth being blunt about: that #3234 comment describes a member set that has never appeared in this file or anywhere else in this design. Correction posted on the thread so nobody reviews against it.

The revision is one GitHub-signed commit; the worked vector and both pins were independently re-derived before push. Re-review when you have time — and the second installment of "what breaks," from the feeds themselves, is exactly the input this wants next.

Disclosure: prepared with AI assistance, human-reviewed.

SolomonisBlack:
@0rkz Checking in on the re-review, nine days on. State of the branch since your 08-31 review:

- Head is still `a1ac65b` (2 GitHub-signed commits). Nothing has changed since the revision that addressed all 14 items — the point-by-point mapping is in [my reply above](https://github.com/x402-foundation/x402/pull/3304#issuecomment-5474375609).
- The five you named as adoption-gating for PayPerByte (1, 2, 3, 4, 7) are all in the file as normative text: the member-carriage table, the request→`inputs` default mapping, the single-issuer scoping of the replay claim, the four-state enumeration with canonicalization rejections mapped to `unverifiable`, and the in-file worked vector + pins with "this file governs on divergence."

If anything in the revision reads differently from what you asked for, say which item and I will turn it around the same day. Otherwise a re-review that clears the "changes requested" state is what the PR is waiting on — the paused first-contributor CI needs a maintainer touch after that, which I am not going to chase directly.

The second installment ("what breaks from the feeds themselves") remains the input this wants most.

Disclosure: prepared with AI assistance, human-reviewed.

0rkz:
Item 2, and it is narrow — the mapping itself is right, so this is a citation ask rather than a redesign.

The behaviour at "Request without a body" is settled and I read it as correct: percent-decoded as UTF-8, values remain strings with no coercion, a repeated key maps to an array in order of appearance, empty query yields `{}`. What the section does not do is name a grammar for the decode. Grepping the file at `a1ac65b`: no RFC 3986, no WHATWG URL / `x-www-form-urlencoded`, no `Content-Type`, no charset reference.

That stands out because the rest of the pipeline is anchored — canonicalization cites RFC 8785, and the JSON-body branch cites RFC 7493. The query-string decode is the only transform without a reference, and it is the one that runs first, so anything ambiguous there propagates into the JCS bytes and out into the digest.

Three shapes where two conformant implementations will disagree today:

- `?q=a+b` — under `x-www-form-urlencoded`, `+` decodes to a space; under RFC 3986 percent-decoding alone it stays a literal `+`. Two different `inputs`, two different digests.
- `?flag` — a key with no `=`. Empty string, `null`, or omitted from the object entirely?
- `?x=%ZZ` — invalid percent-encoding. Reject, or pass the bytes through literally?
- `?x=%FF` — and this is the one that worries me most, because nothing here is malformed. `%FF` is valid hex whose byte is not valid UTF-8, so one implementation yields U+FFFD and another may reject or pass the raw byte. Both sides produce a well-formed value, both canonicalize cleanly, and the digests differ — so none of the `unverifiable` conditions ever fire and the disagreement surfaces only as a hash mismatch.

This is not hypothetical for the outcome states, and it lands on `contradicted` rather than `unverifiable` — both decodes are well-formed, so re-derivation runs and the hash simply misses. `contradicted` is your three-branch disjunction: altered, closure-rule violation, or the verifier's own implementation defective on this input. Without a shared decode grammar the third branch cannot be eliminated on any GET route — vectors will not cover `+` versus `%20` unless something says which is right — so a verifier following your own instruction to not attribute a branch it cannot see is left reporting an unattributable `contradicted` against an honest issuer.

The fix is one sentence naming which grammar governs and what happens on a malformed sequence — pick either, I have no preference, I just need two implementations to land on the same bytes. I have only worked item 2 to that depth. I have open questions on items 4 and 7 that I would rather raise separately than widen this thread — I am not signing off on either one here.

On the second installment: it is next on my side and I do not have a firm date. I wanted the spec text settled under this revision before mapping specific feeds onto it, so that what I report is "this feed breaks against the landed grammar" rather than against a moving target. I will post it here or back on #3234 when it is ready.

*Disclosure: I build PayPerByte/ForeSeal and would be an adopter of this extension, so I am an interested party. Prepared with AI assistance; the file at `a1ac65b` was re-read and the grep results above re-run first-hand before posting.*

0rkz (approved):

0rkz (approved):

0rkz:
Approval is in, and it is deliberate rather than a stray click — worth saying plainly, since the state flipped without a word from me.

What it means: I would rather this land than stall over one citation. Item 2 is now a follow-up, not a gate. One sentence naming which grammar governs the decode and what happens on a malformed sequence, whenever it suits you — in this PR, or a later one, or an issue if you prefer to keep the history clean. I still think it is worth doing, and I still have no preference between the candidates.

What it does not mean: my open questions on items 4 and 7 are unchanged. Approving is not a sign-off on either, and as I said, I would rather raise them on their own than widen this thread.

The rest of the review stands as written — the worked vector, the pin, and the items I marked closed all verified first-hand, and none of that is affected by this.

*(Two approval events on the timeline are one command run twice, not a second opinion.)*

SolomonisBlack:
@0rkz Thank you for the approval, and for saying plainly that it was deliberate. Item 2 is done in this PR rather than deferred — one commit, `e92dfa0`, GitHub-signed — since you were right that it was the one transform in the pipeline with no reference and the one that runs first.

**Grammar chosen.** The decode is the WHATWG URL Standard's `application/x-www-form-urlencoded` parser ([§5.1](https://url.spec.whatwg.org/#urlencoded-parsing)), applied to the query component: split on `&`, each pair on its first `=` (a bare key has the value `""`), `+` → U+0020 before percent-decoding, then bytes → UTF-8. I picked it over RFC 3986 alone because it is the one that actually specifies pair splitting, `+`, and bare keys, and every mainstream implementation exposes it verbatim (`URLSearchParams`, browsers, Node, Deno, Bun).

**Two tightenings, both MUST, where that parser is lenient.** A `%` not followed by two hex digits is malformed (the parser would pass it through literally); a decoded byte sequence that is not valid UTF-8 is malformed (the parser would substitute U+FFFD). An issuer MUST NOT attach the extension to a response to a malformed query; a verifier that meets one reports `unverifiable`. That closes the `%FF` case the way you wanted it closed: the disagreement can no longer surface as an unattributable `contradicted` against an honest issuer, because neither side ever hashes a lossy decode.

**Your four shapes, now answered in-file (normative example table):**

| Query | `inputs` |
|---|---|
| `?q=a+b` | `{"q":"a b"}` |
| `?q=a%2Bb` | `{"q":"a+b"}` |
| `?flag` | `{"flag":""}` |
| `?x=%ZZ` | malformed — no claim issued; `unverifiable` if one is present |
| `?x=%FF` | malformed — no claim issued; `unverifiable` if one is present |

(`%20` and a repeated key are in the table too.) Outcome 3 (`unverifiable`) now names the malformed-query case alongside the canonicalization rejections, so the mapping is explicit rather than implied.

Nothing else moved: fixed point, worked vector (`81ea1f22…` re-derived from the file after the edit), pins, and every other section are byte-identical to `a1ac65b`.

On items 4 and 7: understood that approval is not a sign-off on either, and I would rather take them on their own threads too. Whenever you raise them, same-day turnaround stands. The second installment from the feeds remains the input this wants most — no rush from my side; it is better against a landed grammar, as you said.

Disclosure: prepared with AI assistance, human-reviewed.

0rkz:
Item 2, and it is narrow — the mapping itself is right, so this is a citation ask rather than a redesign.

The behaviour at "Request without a body" is settled and I read it as correct: percent-decoded as UTF-8, values remain strings with no coercion, a repeated key maps to an array in order of appearance, empty query yields `{}`. What the section does not do is name a grammar for the decode. Grepping the file at `a1ac65b`: no RFC 3986, no WHATWG URL / `x-www-form-urlencoded`, no `Content-Type`, no charset reference.

That stands out because the rest of the pipeline is anchored — canonicalization cites RFC 8785, and the JSON-body branch cites RFC 7493. The query-string decode is the only transform without a reference, and it is the one that runs first, so anything ambiguous there propagates into the JCS bytes and out into the digest.

Three shapes where two conformant implementations will disagree today:

- `?q=a+b` — under `x-www-form-urlencoded`, `+` decodes to a space; under RFC 3986 percent-decoding alone it stays a literal `+`. Two different `inputs`, two different digests.
- `?flag` — a key with no `=`. Empty string, `null`, or omitted from the object entirely?
- `?x=%ZZ` — invalid percent-encoding. Reject, or pass the bytes through literally?
- `?x=%FF` — and this is the one that worries me most, because nothing here is malformed. `%FF` is valid hex whose byte is not valid UTF-8, so one implementation yields U+FFFD and another may reject or pass the raw byte. Both sides produce a well-formed value, both canonicalize cleanly, and the digests differ — so none of the `unverifiable` conditions ever fire and the disagreement surfaces only as a hash mismatch.

This is not hypothetical for the outcome states, and it lands on `contradicted` rather than `unverifiable` — both decodes are well-formed, so re-derivation runs and the hash simply misses. `contradicted` is your three-branch disjunction: altered, closure-rule violation, or the verifier's own implementation defective on this input. Without a shared decode grammar the third branch cannot be eliminated on any GET route — vectors will not cover `+` versus `%20` unless something says which is right — so a verifier following your own instruction to not attribute a branch it cannot see is left reporting an unattributable `contradicted` against an honest issuer.

The fix is one sentence naming which grammar governs and what happens on a malformed sequence — pick either, I have no preference, I just need two implementations to land on the same bytes. I have only worked item 2 to that depth. I have open questions on items 4 and 7 that I would rather raise separately than widen this thread — I am not signing off on either one here.

On the second installment: it is next on my side and I do not have a firm date. I wanted the spec text settled under this revision before mapping specific feeds onto it, so that what I report is "this feed breaks against the landed grammar" rather than against a moving target. I will post it here or back on #3234 when it is ready.

*Disclosure: I build PayPerByte/ForeSeal and would be an adopter of this extension, so I am an interested party. Prepared with AI assistance; the file at `a1ac65b` was re-read and the grep results above re-run first-hand before posting.*

0rkz:
Apologies for the noise. I posted a duplicate of my 2026-09-10 note in error from a stale draft, and have since deleted it. It re-asked for work you had already finished, which was not my intent.

To be clear about what's actually outstanding: item 2 is closed as far as I'm concerned. I read e92dfa0 rather than taking it on trust — the WHATWG x-www-form-urlencoded reference at response-provenance.md:92-93, both MUST tightenings at :97-100, and the example table at :110-116 covering the malformed-percent-encoding case. That answers what I raised.

The second installment from the feeds is still on me. You said "no rush" and I'm taking you at your word rather than shipping something thin — it isn't finished yet, and I'd rather name that plainly than go quiet on it.

SolomonisBlack:
A request to the four people who did the work on this text, and only to them: @whawk46, @giskard09, @cv-scvd, @kopko13.

Status, so nobody has to scroll: @0rkz approved on 2026-09-11 after a 14-item implementation-attempt review; the one item raised afterwards (query-string decode grammar) is closed in `e92dfa0`; the PR has waited on a maintainer read since 2026-08-31. Nothing is asked of the maintainers here beyond what the file already says.

The ask is narrow. Each of you reviewed or ran this against real code: the 24/24 cross-check on an independent RFC 8785 implementation; seven preimages, one anchored on Base mainnet; the re-derivation check merged into a production conformance desk; the "two field names, not a flag" review that shaped the carriage table. If the text as it stands is one you would implement, saying so **on this PR** as a review (approve, or a comment naming what you checked) is different from having said it in #3234: it turns one author's proposal into a reviewed extension that several independent parties have already exercised, which is the thing a maintainer can act on.

If it is not, say that instead. A "not yet, because X" here is worth more than a courtesy approval, and I will fix X.

No obligation, and no reply needed if you would rather not. Thank you for the review rounds either way.

giskard09 (approved):
@SolomonisBlack — two things: a correction to the citation, and the actual review.

**Correction first.** The PR body's citation of our work — "giskard09/argentum-core (7 preimages, one anchored on Base mainnet at block 49623528 before it was checked)" — conflates two separate things and gets both wrong:

- The 24 August cross-check was against `action_ref` (our own four-field preimage), not this extension's fixed point. 14 preimages, not 7, all in `argentum-core`.
- The Base mainnet anchor at block 49623528 belongs to a different repo (`composed-attestation-3leg-worked-example`), offered on that thread as a reusable *shape* for slotting in a fourth `responseHash` leg — never actually built or verified as one.

Neither of those is a prior verification of this extension's `{endpoint, inputs, result, method, dataVintage}` fixed point. Please drop that line, or fix it to describe what's below instead.

**The actual review**, done fresh against the current text (`e92dfa0`), independent of any of the three implementations already cited:

- Reconstructed the normative worked vector by hand and hashed it with a third-party RFC 8785 library (not GVP's, not whawk46's, not SCVD's) — byte-exact match against the published `sha256:81ea1f22...`.
- Wrote the query-string→`inputs` decode grammar from the prose alone (WHATWG parser + the two tightenings) — all seven rows of the normative table reproduce exactly, including both malformed cases (`%ZZ`, `%FF`) correctly rejecting.
- Confirmed the I-JSON constraint isn't decorative: a wei-scale integer passed as a raw JSON number fails canonicalization outright rather than silently losing precision — the MUST-be-string rule is load-bearing, not stylistic.
- Built two fresh preimages of my own (a GET-style query-string case, a wei-scale result) and hashed them cleanly end to end.
- Read the full text against 0rkz's 31 August blocking list — all ten items resolved as written: four enumerated outcomes, `no_claim` binding scoped separately, `contradicted` as an explicit three-branch disjunction, `endpoint` explicitly not binding origin, `fixedPointVersion` inside the signed payload, worked vector and `/2` grammar now vendored in-file.

This is a text I would implement. Approving.

Pablo

cv-scvd (approved):
Naming what I checked, per the request — speaking as a production implementer of this text.

SCVD's conformance desk ships an additive `response_provenance` block built against this design (GVP 0.2 lineage, offered from #3234; our side landed as seancrecord/scvd-general-store-repo#270 and #274). What it does, in this PR's vocabulary:

- **Re-derivation over the closed fixed point.** Given a caller-supplied body, it recomputes `responseHash = sha256(jcs({endpoint, inputs, result, method, dataVintage}))` and compares against the declared value. No key, no origin fetch — the verdict is about the supplied pair, never a claim about what the issuer served.
- **The three outcomes land as written.** A response carrying no `provenance.responseHash` is out of scope (`applies: false` — the `no_claim` shape, explicitly not a failure). A hash that re-derives is `ok` and carries no class slug. A hash that does not re-derive reports `response-hash-not-rederivable` (the `contradicted` shape) — and the finding text states the disjunction (the artifact was altered, or it was issued in violation of the closure rule) without naming a branch, because a desk cannot see the issuer's internals. The no-collapse rule holds: nothing downgrades a `contradicted` into a softer verdict.
- **§8.1 verifier self-proof is implemented literally.** Before the block may state any finding, it re-derives a pinned set of published vectors with *our own* canonicalizer. On disagreement it reports `self_ok: false` about itself, carries no class slug, and states no finding about anyone. A canonicalizer bug on our side surfaces as a public claim about us — which is the only honest shape for a desk that issues `contradicted` verdicts.

One scope note, for precision: our callers supply `inputs` directly, so we never run the request→`inputs` decode branch settled in `e92dfa0` and have no implementation evidence for or against it. The member-carriage table and the fixed-point closure are the parts we exercise on every call.

Verdict: the text as it stands is implementable — we implemented it, and it runs in the production desk. This is one I would implement again without changes to the parts we touch.

SolomonisBlack:
@giskard09 The correction is right and the error was mine, twice over: the 24 August run was this canonicalizer over your `action_ref` preimages (14, not 7), which tests canonicalization agreement and not this fixed point; and the anchored manifest belongs to `composed-attestation-3leg-worked-example` and was a proposed shape, never a built leg. I turned "cross-check" into "verification of this extension" without checking what had been hashed. Fixed in three places, each as a dated correction rather than a silent edit: the PR body (that line now describes your 15 September review instead), the GVP README, and an explainer that had repeated it. Thank you for reconstructing the worked vector from the text alone with a library none of the prior implementations used, and for the two fresh preimages.

@cv-scvd Thank you for naming what the desk exercises and, just as usefully, what it does not. The scope note (callers supply `inputs` directly, so the request→`inputs` decode branch has no implementation evidence from the desk either way) is now recorded in the PR body beside the citation, so nobody reads the desk as evidence for that branch. One more detail from your note is now written down where it belongs: the desk reads the hash at `provenance.responseHash`, the original issuer carriage, which the carriage-profiles note in the GVP repo now states.

Status for anyone arriving: three approvals (@0rkz 09-11, @giskard09 and @cv-scvd 09-15), item 2 closed at `e92dfa0`, no open change requests (@0rkz has said items 4 and 7 may be raised separately; nothing filed yet), waiting on a maintainer.

whawk46:
Reviewed and verified.

The carriage table cleanly separates declaration from runtime negotiation,
and the 24/24 RFC 8785 JCS test vectors cross-check clean with zero
divergence against independent implementations. The text is tight and ready
for implementation. Approved.

kopko13:
Naming what I checked, as asked — as the seller behind the "two field names, not a flag" exchange in [sirenic-eu/sirenic-examples#2](https://github.com/sirenic-eu/sirenic-examples/issues/2). This is a comment, not an approval: there is one "not yet, because X" below.

Checked against the file at e92dfa0, re-read today: the closed five-member fixed point, the closure rule, dataVintage completeness (oldest source, at its own precision, or emit nothing) and the no-mode-flags rule are exactly what we agreed on [#2](https://github.com/sirenic-eu/sirenic-examples/issues/2), and I would not change a word of them. On 2026-09-03 we ran the published vectors through our own RFC 8785 + SHA-256 harness (11/11) and then hashed real paid responses of ours through it, which is where the two findings below come from.

What an existing catalogue looks like under the closure rule (first-party, 80 catalogued routes): 33 routes satisfy it as written (result computed from local stocks only, bound by path and query); 47 cannot emit — 44 call an upstream registry at request time, 4 serve model-derived prose from a cache, and a few carry a per-call variable. Emitting nothing on those 47 is the compliant output and we are fine with it. The real cost of emitting on the 33 is not the hash, it is enumerating every contributing source per route so that dataVintage is complete; only one of the 33 carries a structured vintage today. That is the "what breaks from the feeds" input you asked for: for a multi-source seller, dataVintage completeness is the work, and it is the right work.

Not yet, because result carriage. The member-carriage table makes result a REQUIRED top-level response-body member. Our responses, like most existing REST APIs, put the answer's fields at the root of the body; there is no result member. To conform we would have to wrap every body (a breaking change for every existing client and every published example) or duplicate the answer under result (doubling payload size on responses that already reach 200 KB). Either way a seller with an installed base cannot attach this extension without a breaking change, which will keep exactly the sellers you want out. Suggested fix, one row in the table: let the extension object declare how result is carried — e.g. "resultCarriage": "member" (today's behaviour, the default) or "body", meaning result is the response body with the top-level extensions member removed. The fixed point stays closed and re-derivable; a verifier reads one declared string before it starts; and the extension becomes attachable to an existing API without touching its shape. With something like that in the text, this is a text we would implement on our closure-safe routes; without it, we cannot, and I would rather say so than approve.

Two smaller data points. Our routes carry the identifier in the path, so inputs is {} for most of them and the question is bound by endpoint alone — that works as written, and the WHATWG decode grammar is the right pick. And on signed carriage: we already publish a detached Ed25519 signature over the exact bytes of every paid response (key at /.well-known/sirenic-signing-key), so an extension object inside the body is covered by that signature as a by-product; the SHOULD costs an existing signer nothing.

giskard09:
Confirmed the blocker against the file at e92dfa0, independent of your report.

`result` as REQUIRED top-level response-body member (member-carriage table) is exactly as you describe — no ambiguity in the text. And the fix composes clean: `responseHash` hashes the abstract fixed point `{endpoint, inputs, result, method, dataVintage}` in its JCS canonical form; the carriage table is only the verifier's recovery instruction for the wire response, not part of what gets hashed. Adding `resultCarriage` (`"member"` default / `"body"`) touches recovery only — it doesn't reopen closure, no-mode-flags, or the hash definition itself.

One row, backward-compatible default, unblocks exactly the case you named (existing REST API, root-level fields, installed base) without a breaking change or payload duplication. I'd take this as a required addition before merge, not optional — kopko13's own catalogue (80 routes, flat body) shows the case concretely; as written today, the extension is unusable for any seller in that shape.

whawk46:
@kopko13 Exceptional production data—the 33/47 route breakdown under the
closure rule is the exact kind of empirical grounding an extension needs
before baselines freeze.

Strongly endorse the `"resultCarriage"` declaration (`"member"` | `"body"`):
1. Forcing established APIs to break existing clients with a root-level
wrapper or double payload sizes on 200 KB responses creates an unnecessary
barrier to adoption.

2. Under `"body"`, stripping the top-level `extensions` member prior to RFC
8785 canonicalization preserves the closed fixed point and keeps
re-derivation completely deterministic for offline verifiers.

3. It keeps the schema backward-compatible with greenfield implementations
while letting production catalogues adopt provenance incrementally.

Adding `"resultCarriage"` to the member-carriage table resolves the last
real friction point for existing sellers.

whawk46:
@kopko13 @giskard09 Completely aligned.

Pablo's architectural point is the key: `responseHash` binds the abstract
5-member fixed point (`{endpoint, inputs, result, method, dataVintage}`)
under RFC 8785; the carriage table is strictly the verifier's wire-recovery
rule. Adding `"resultCarriage"` (`"member"` default | `"body"`) touches
only wire recovery without reopening closure, the hash definition, or the
no-mode-flags invariant.

Forcing established REST APIs to break existing clients or duplicate 200 KB
payloads creates an unnecessary adoption wall. Stripping the top-level
`extensions` object under `"body"` preserves the closed fixed point and
gives existing catalogues a non-breaking on-ramp.

Fully endorse making this addition before merge.

SolomonisBlack:
@kopko13 Taken as required before merge, not optional, and it is in the text now: `2e2eeb0`.

What changed, against `e92dfa0`:

- **`resultCarriage`** (OPTIONAL member of the extension object, default `"member"`), one row in the carriage table plus its own subsection. `"member"` is today's behaviour. `"body"` recovers `result` as the response body object with its top-level `extensions` member removed and nothing else removed, added, or reordered; any other top-level member, including one named `result`, is part of `result` like any other. Stated as a wire-recovery declaration, not a mode flag: fixed point, closure rule, hash definition, and the single claim are unchanged, and the same fixed point hashes identically under either carriage.
- A second **envelope example**: a root-level `{ "sum": 5, "extensions": {...} }` body under `"body"` reproduces the worked vector's hash exactly.
- **`unverifiable`** gains the three carriage failures: unrecognised value, `"member"` with no `result` member, `"body"` on a non-object body.
- **Signed carriage**: a non-default `resultCarriage` SHOULD be inside the signed payload, for the same reason `fixedPointVersion` is (an unsigned one is flippable to steer `contradicted` into `unverifiable`).
- One further section, **Agreement semantics**, prompted by @goun7 on #2887: `inputs` and `result` are compared as declared, with no normalization of any member, and `fixedPointVersion` is the record's declaration of that. It does not change any hash; it makes explicit what the text already assumed.

The 33/47 breakdown is exactly the feeds report the text needed, and I would rather have it than an approval. Your read on `dataVintage` completeness matches mine: enumerating sources per route is the work, and the text keeps it that way. Path-bound routes with `inputs` as `{}` are conformant as written; an empty query yields `{}` and the question is bound by `endpoint`.

The direct question back: is the text at `2e2eeb0` one you would implement on the 33 closure-safe routes? If something in the `"body"` semantics does not fit a real response of yours, that is the next fix.

@giskard09 @whawk46 thank you for confirming the blocker independently and for the recovery-only framing; the text now says it in those terms.

Disclosure: prepared with AI assistance, human-reviewed.

kopko13 (approved):
Checked against the file at 2e2eeb0, re-read today. Yes: this is the text we will implement on our closure-safe routes. Approving, and naming what I checked.

The diff e92dfa0 → 2e2eeb0. The carriage-table row, the resultCarriage subsection, the second envelope example, the three carriage failures added to unverifiable, the signed-carriage SHOULD, and the agreement-semantics section. Nothing in the fixed point, the closure rule or the hash definition moved, which is what "wire-recovery declaration, not a mode flag" has to mean.

Vectors. Our own RFC 8785 + SHA-256 harness (the one that reproduced the published vectors 11/11 on 2026-09-03) reproduces the worked vector at 134 bytes / 81ea1f22…, and recovers that same hash from the second envelope example once extensions is removed.

Real responses, both carriages. Two paid responses from our 2026-09-01 sweep, a sector-benchmark body (1.4 KB) and a sanctions-screening body (18 KB), hashed three ways: the bare fixed point, "member" with the extension object attached, "body" with the extension object attached and result recovered as the text says. Same hash all three ways, on both. The wire cost of "body" over "member" on those bodies is the 13-byte resultCarriage member instead of a wrapper.

Catalogue shape. 78 of our 80 catalogued routes serve JSON (the other two serve a PDF and a binary). All 78 bodies are objects at the root, none has a top-level extensions or result member, maximum nesting depth is 9, and no integer exceeds 2^53−1. "body" recovery is total on the catalogue as it stands; none of the three carriage failures can arise on it.

What "body" changes on our side, which is the useful part. Under "body" the closure rule reaches every root-level member, our own envelope included (source, disclaimer, freshness phrase, provenance blocks). So I compared the bodies of two paid sweeps three days apart, 2026-08-29 and 2026-09-01: 42 of 78 were byte-identical; the rest differed where a stock vintage moved (a date in the freshness phrase or in a provenance block, which is exactly what dataVintage carries), where the data itself changed, or on live-merge routes already outside the closure rule. But three of the 33 routes I called closure-safe carry a value derived from the request-time clock at the root: the default-risk score serves the age of the last fiscal year in months, computed at request time, and two public-procurement routes serve "next 12 months" windows anchored on the calendar day. Under "member" those values could have stayed outside result; under "body" they cannot. That is the right outcome, the claim covers what the buyer actually received, and it is a fix on our side (an explicit date input, or dropping the convenience field), not on the text. It also means my "33" was measured on handler code, not on served bodies: the day-one count under "body" is 30, and I would rather correct my own number here than let it stand.

If one sentence is worth adding, it is informative, under resultCarriage: that under "body" the closure rule applies to every top-level member, seller metadata included (consultation timestamps, ages computed at request time), so a seller whose envelope carries such a member removes it from the route or uses "member". "Nothing else removed" already implies it; saying it would save the next seller the discovery. Not a change request.

Two implementation notes, no change requested. Our router accepts more than one spelling of a path (case-insensitive, repeated slashes folded), so the endpoint we hash is the normalized path and we will echo it as the RECOMMENDED root-level endpoint member; for an issuer that serves several spellings, that echo is what makes the hashed endpoint recoverable, and we treat the RECOMMENDED as a MUST for ourselves. And our detached Ed25519 signature covers the exact bytes of the body, so a resultCarriage inside extensions is signed as a by-product; the SHOULD is met without a second signature.

Sequence on our side. Emission starts under the merged field name, not before: buyers wired to a name that then moves is the one failure we can avoid. First the sanctions-screening route, which already carries a structured vintage; then the others as their sources are enumerated route by route, with the three clock-derived members fixed on the way.

Disclosure: prepared with AI assistance, human-reviewed.

SolomonisBlack:
@kopko13 Thank you for the review, and for the shape of it: two real bodies hashed three ways, the whole catalogue measured, and your own count corrected in public. That last part is the standard this text is trying to make normal.

Your one informative sentence is in at `ba8b0ca`, under `resultCarriage`: under `"body"` the closure rule and `dataVintage` completeness reach every top-level member, seller metadata included, and a seller whose envelope carries such a member removes it, makes its input explicit, or uses `"member"`. Marked informative, as you framed it; nothing normative moved.

The three clock-derived members are exactly the case the rule exists for, and you are right that the fix is on the route, not the text. Treating the RECOMMENDED root-level `endpoint` echo as a MUST for a router that accepts several spellings is the right read; that echo is what makes the hashed `endpoint` recoverable. And starting emission only under the merged field name is the right sequence for an installed base.

Disclosure: prepared with AI assistance, human-reviewed.

SolomonisBlack:
Two things landed today that follow directly from this thread, both in `golden-vector-provenance@0.9.0` (npm; Apache-2.0 code, CC-BY-4.0 spec).

**1. `gvp-audit`, the closure audit @kopko13's report made obvious.** The cost of provenance turned out to be enumerating sources per route and finding request-time inputs that handler code hides. So: a free CLI a seller runs against the bodies they actually served (offline) or against free routes (GET only; it never pays a 402). Per route it reports hidden-input signals (clock keys, datetimes, UUIDs, epochs near now, relative windows), which carriage fits (`member` as-is, or `resultCarriage: "body"`), format violations (integers past 2^53-1 as numbers, duplicate member names, depth over 100), vintage candidates and whether they match the `/2` grammar, an existing `responseHash` re-derived under either carriage, and drift between two samples of the same route classified as clock, nonce, runtime, vintage, or data. Verdicts are `no-closure-violation-found`, `review-needed`, `hidden-inputs-found`, `format-violation`, `out-of-scope`. It never says "closure-safe": it finds violations and cannot prove closure, and the README says so.

```
npm i -g golden-vector-provenance
gvp-audit bodies ./served --md report.md      # <route>.1.json / <route>.2.json, optional <route>.meta.json {endpoint, inputs}
```

It found things on its author first. Run against x402toll.com's own free surfaces it flagged `dataVintage: "July 2026"` (not `/2` grammar) and a request-time `computedAt` in the provenance block. Both are fixed and live as of today: the service now hashes under `GVP-FixedPoint/2` with `dataVintage: "2026-07"`; receipts issued under `/1` with the old spelling remain valid under `/1`, which is what the identifier is for. Golden hashes moved with it and re-derive from the published inputs.

**2. `resultCarriage` is implemented, not just specified.** `attachProvenance(body, fp, { resultCarriage: "body" })`, the Express and Hono adapters, `recoverResult(body, carriage)`, and `fixedPointFromBody(...)`. Under `"body"` the answer stays at the root and the block declares the carriage; attaching asserts that the fixed point's `result` equals the body minus `extensions`, so an issuer cannot declare a carriage it does not satisfy. Same fixed point, same hash; the worked vector reproduces under both in the test suite.

@kopko13, the offer: if you save two paid responses per closure-safe route into a directory, the audit is one command and the report is yours to keep or ignore. No payment on my side, no access needed. If it mislabels a real Sirenic body, that is a bug report I want.

Status unchanged otherwise: four approvals, waiting on a maintainer.

Disclosure: prepared with AI assistance, human-reviewed.

---

**Correction (2026-09-18), appended rather than edited away.** The sentence above overstates how the second finding was made. Live runs of `gvp-audit` against x402toll.com's free surfaces flagged only the `dataVintage: "July 2026"` grammar; those routes carry no provenance block, so a live run could not have seen `computedAt`. The request-time `computedAt` was flagged by the tool on the service's own response shape, a fixture built from the emitter's code in the tool's test suite, and then fixed. Same two defects, same fixes, one found on live free surfaces and one on the shape. Stated so the tool is not credited with more than it did.

whawk46:
Looks solid, @SolomonisBlack. Keeping resultCarriage strictly as a
wire-recovery declaration preserves the fixed-point hashing invariants
cleanly. The updated text in 2e2eeb0 resolves my concern — thanks for
working through the framing.

goun7:
Independent re-derivation of the normative vector, from a separate RFC 8785 implementation.

We ran the worked vector through our own JCS implementation (built independently for the Tamga receipt ledger, byte-exact against a Node twin across 17 parity vectors). Result:

- canonical form: 134 bytes, matching the spec's stated length byte-for-byte
- `responseHash`: `81ea1f2227fd9df5b868954e6d26d091810352f148dade483b260844788ede03` — exact match

The closed-set rule also checks out from the verifier side. Our tool rejects an extra member and a missing one as `unverifiable` before hashing, which is what the normative text requires.

Tool if useful: `tools/x402_response_provenance.py` in goun7/tamga-protocol — `vector` runs the spec vector, `verify <fixedpoint.json>` re-derives an arbitrary fixed point. Pure stdlib plus our JCS, no x402 dependency.

This is a second implementation reproducing the vector independently of the three already cited, which seems worth recording given the text's byte-exact requirement.

*Drafted with AI assistance and reviewed by the project founder.*

SolomonisBlack:
@goun7 Recorded, and re-run here rather than taken on trust: a clean clone of goun7/tamga-protocol, `py -3 tools/x402_response_provenance.py vector` printed 134 canonical bytes and `81ea1f22…`, and `verify` on a fixed point with an extra member returned `unverifiable` before hashing. That is a fourth independent reproduction from a separately built RFC 8785 implementation, and it is now the first line of the cross-check list in the GVP README, cited to your comment above and to your tool.

@whawk46 thank you for closing the loop on the carriage framing.

Status: four approvals, one change request resolved, waiting on a maintainer.

Disclosure: prepared with AI assistance, human-reviewed.

goun7:
Fourth independent reproduction from a separately built RFC 8785 implementation is the strongest form of this: I did not have to trust my own canonicalizer to know the 134 canonical bytes and `81ea1f22…` are right, and now the cross-check list does not have to trust me either. Recording my tool and comment as the first line is exactly how a cross-check list should be built — the citation is the reproducibility.

One addition I would offer to the list, since the failure mode it guards is silent: the `unverifiable`-before-hashing path is the case that matters more than the byte count. Byte counts agree when two canonicalizers agree; `unverifiable` before any hashing is the behavior that protects a verifier whose canonicalizer is *almost* the same. If the list ever tracks a single invariant, that is the one to keep.

@whawk46 the carriage framing is settled from my side.

Disclosure: prepared with AI assistance, human-reviewed.

whawk46:
Thank you @goun7 for the independent re-derivation and for publishing the
Python verification tool (`tools/x402_response_provenance.py`).

Having an independent, clean-room implementation in Python (pure stdlib +
RFC 8785 JCS) verify the exact 134-byte canonical representation and arrive
at the identical SHA-256 digest
(`81ea1f2227fd9df5b868954e6d26d091810352f148dade483b260844788ede03`) is a
significant interoperability milestone for this extension.

Crucially, your verification that the verifier-side closed-set rule
correctly rejects both extraneous and omitted members before hashing
demonstrates that the normative boundary holds across language runtimes
without reliance on framework-specific quirks.

We will record this second independent implementation alongside the
existing test vectors to document multi-runtime conformance for the working
group.

PR #3317, opened 2026-08-31
feat(svm): transaction v1 support

amilz:
Adds Solana transaction v1 support to the SVM `exact` scheme across all three
facilitator implementations (TypeScript, Python, Go), and explicitly closes the
version gates on the paths that do not model v1.

[Solana transaction v1](https://solana.com/upgrades/larger-transaction-sizes) (SIMD-0385 format, SIMD-0296 size increase) raises the max transaction size from 1232 to 4096 bytes. 

**The feature is expected to go live across clusters in September, 2026.**

- Facilitators will need to update to ensure introspected v1 transactions do not result in deserialization errors. 
- Existing introspection flows that look at ComputeBudget instructions are not valid in v1. V1 uses a new config field for handling Compute Budget. Compute Budget instructions can exist on v1 transaction but are no-op, so this must be handled properly.
- Default config for compute and account data size are 0, so we should not send v1 transactions w/ unset configs to cluster (they will fail)
- Priority fees set in a v1 config are in Lamports (not micro lamports per CU) 

## What this PR does

- **Accepts v1 in the `exact` scheme.** The static path recognizes the v1 layout
(`TransferChecked` at index 0, 1–5 instructions) and reads `computeUnitLimit`,
`loadedAccountsDataSizeLimit`, and the priority fee from the message config
instead of from instructions. 
- **Requires both resource limits on v1.** An unset or zero `computeUnitLimit` or
`loadedAccountsDataSizeLimit` is rejected.
- **Post-settlement check** now fetches with
`maxSupportedTransactionVersion: 1`.

## Dependency bumps

v1 decoding requires new floors in every language:

- **Go**: `solana-go` → v1.23.0 
- **Python**: `solders` → `>=0.29.0`
  `0x80` prefix.
- **TypeScript**: `@solana/kit` peer `>=8.0.0`,
  `@solana-program/compute-budget` `^0.18.0`, token/token-2022 `^0.16.0`.


e2e/servers/go/go.mod
```diff
@@ -24,49 +23,46 @@ require (
...
+	github.com/klauspost/compress v1.18.0 // indirect
 	github.com/logrusorgru/aurora v2.0.3+incompatible // indirect
 	github.com/mattn/go-colorable v0.1.14 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mitchellh/go-testing-interface v1.14.1 // indirect
-	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
-	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/mostynb/zstdpool-freelist v0.0.0-20201229113212-927304c0c3b1 // indirect
-	github.com/mr-tron/base58 v1.2.0 // indirect
+	github.com/oasisprotocol/curve25519-voi v0.0.0-20251114093237-2ab5a27a1729 // indirect
```
lgalabru, line 43: any way we can limit the blasting radius to SVM? we don't want to take the responsibility of breaking deps for other networks.

PR #3335, opened 2026-09-01
fix: replace crypto dependency with @noble/hashes in @x402/svm

CarsonRoscoe:
## Description

`@x402/svm` could not be used in a default browser context due to its dependency on `node:crypto`, making it a backend-only SDK. In order for the frontend to use it, they needed to be a complex frontend that uses plugins (i.e. vite + polyfill plugin).

This PR attempts to resolve this by moving the sha256 hashing dependency away from `node:crypto`, and instead using `@noble/hashes`. `@noble/hashes` is already trusted by the `@x402/evm` package, as well as is used by `viem` the leading cross-enviromment typescript EVM package. 

## Tests

Re-ran unit tests, integration tests, and cross-SDK e2e tests

## Checklist

### DIFF
--- a/e2e/pnpm-lock.yaml
+++ b/e2e/pnpm-lock.yaml
@@ -1670,6 +1670,9 @@ importers:
 
   ../typescript/packages/mechanisms/svm:
     dependencies:
+      '@noble/hashes':
+        specifier: ^1.8.0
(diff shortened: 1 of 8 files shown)


phdargen (approved):
LGTM 🚀

PR #3336, opened 2026-09-01
fix(stellar): include feeBumpSigner in facilitator-safety checks

### LINKED ISSUE #3332: @x402/stellar: feeBumpSigner documented as a facilitator address by getSigners() but excluded from every facilitator-safety check
## Summary

`ExactStellarScheme`'s `feeBumpSigner` option (facilitator side) is
documented as a facilitator address — `getSigners()` explicitly says it
returns "all facilitator addresses including the fee bump signer when
configured" — but the internal set every facilitator-safety check actually
consults (`this.signingAddresses`) never includes it. A facilitator
configured with a `feeBumpSigner` reports that address as its own via the
public API, while none of the checks that exist specifically to keep a
facilitator-controlled address out of the payer/tx-source/op-source/
auth-signer role in a client payment cover it.

## Where

`@x402/stellar@2.24.0`, `dist/cjs/exact/facilitator/index.js` (source:
`src/exact/facilitator/scheme.ts`), class `ExactStellarScheme`.

Constructor (line 213-219):

```js
this.signerMap = new Map(signers.map((s) => [s.address, s]));
this.signingAddresses = new Set(this.signerMap.keys());   // built only from `signers`
...
this.feeBumpSigner = feeBumpSigner;                        // stored separately, never merged in
```

`getSigners()` (line 234-247), JSDoc quoted verbatim:

```js
/**
 * Get signer addresses used by this facilitator.
 * For Stellar, returns all facilitator addresses including the fee bump signer when configured.
 *
 * @param _ - The network identifier (unused for Stellar)
 * @returns Array containing all facilitator addresses
 */
getSigners(_) {
  const signers = [...this.signingAddresses];
  if (this.feeBumpSigner && !this.signingAddresses.has(this.feeBumpSigner.address)) {
    signers.push(this.feeBumpSigner.address);   // added here...
  }
  return signers;
}
```

But every actual safety check reads `this.signingAddresses` directly, which
never gained the fee-bump address:

- line 442: `this.signingAddresses.has(operation.source ?? "") || this.signingAddresses.has(transaction.source)` → rejects with `invalid_exact_stellar_payload_unsafe_tx_or_op_source`
- line 469: `this.signingAddresses.has(fromAddress)` → rejects with `invalid_exact_stellar_payload_facilitator_is_payer`
- line 531: `validateAuthEntries(invokeOp, this.signingAddresses, fromAddress, ...)` → internally does `facilitatorAddresses.has(authAddress)` → rejects with `invalid_exact_stellar_payload_facilitator_in_auth`

So `getSigners()`'s own return value and the set actually enforced by these
three checks silently diverge whenever `feeBumpSigner` is configured.

## Repro

Ran directly against the installed `@x402/stellar@2.24.0`, constructing the
real class (no internals mocked — `signingAddresses` is a genuine public
field at runtime, matching the `.d.ts`):

```js
const { ExactStellarScheme } = require("@x402/stellar/exact/facilitator");

const mainSigner = { address: "GAAA...JEZQ", signAuthEntry: async () => ({...}), signTransaction: async () => ({...}) };
const feeBumpSigner = { address: "GBBB...FQTB", signAuthEntry: async () => ({...}), signTransaction: async () => ({...}) };

const scheme = new ExactStellarScheme([mainSigner], { feeBumpSigner });

scheme.getSigners("stellar:testnet");   // documented as "all facilitator addresses"
scheme.signingAddresses;                // actually used by every safety check
```

Output:

```
getSigners() [documented as 'all facilitator addresses']:
  [ 'GAAA...JEZQ', 'GBBB...FQTB' ]

scheme.signingAddresses [actually used by every facilitator-safety check]:
  [ 'GAAA...JEZQ' ]

feeBumpSigner.address included in getSigners(): true
feeBumpSigner.address included in signingAddresses (the safety-check set): false
```

## Impact

Not directly exploitable by a third party without the fee-bump signer's own
private key — Soroban signature verification still blocks a forged auth
entry, so this doesn't by itself let an attacker impersonate the
facilitator. The real risk is defense-in-depth silently not extending to a
documented, publicly-typed facilitator address: these three checks exist
specifically so a facilitator never ends up as the transaction source,
operation source, `from` address, or an auth-entry signer in a client
payment (the non-custodial guarantee this scheme is built around). A
`feeBumpSigner` is exactly a facilitator-controlled address by the
constructor's own doc, so it should be covered by the same guardrail
`getSigners()` already claims it's part of.

Calibrating honestly: medium-low severity, "documentation-contract
violation / guard doesn't extend where it says it does" rather than a
directly exploitable vulnerability.

## Suggested fix

Build `signingAddresses` (or a second set consulted by the same three
checks) to include `feeBumpSigner.address` when configured — matching what
`getSigners()` already documents and returns — rather than only unioning it
in at the `getSigners()` call site.

Eras256:
Fixes #3332.

`ExactStellarScheme`'s `getSigners()` documents itself as returning "all facilitator addresses including the fee bump signer when configured", but the set every facilitator-safety check actually consults (`signingAddresses`) never included `feeBumpSigner`. A facilitator configured with `feeBumpSigner` reported that address as facilitator-controlled via the public API, while the three checks that exist specifically to keep a facilitator-controlled address out of the payer/tx-source/op-source/auth-signer role in a client payment didn't cover it — `unsafe_tx_or_op_source`, `facilitator_is_payer`, and `facilitator_in_auth`.

## What changed

- Adds a `facilitatorSafetyAddresses` set (`signingAddresses` plus `feeBumpSigner.address` when configured), computed once in the constructor.
- The three safety checks now consult `facilitatorSafetyAddresses` instead of `signingAddresses` directly.
- Deliberately **not** merged into `signingAddresses` itself: that set also backs signer *selection* in `settle()` (`this.signerMap.get(this.selectSigner([...this.signingAddresses]))`), and `feeBumpSigner` has no entry in `signerMap` — merging it into `signingAddresses` would let round-robin selection pick an address `settle()` can't actually sign with. `getSigners()` and the signer-selection path are unchanged.

Added a regression test (`facilitator-verify.test.ts`, "should reject when a configured feeBumpSigner is the payer (from address)") that constructs a real `ExactStellarScheme` with `feeBumpSigner` set, builds a real transaction where the transfer's `from` address is the fee-bump signer's, and asserts `verify()` rejects it with `invalid_exact_stellar_payload_facilitator_is_payer` — this test fails against `main` and passes with this change.

Ran the full `@x402/stellar` suite before and after:

```
# before (main)
Test Files  9 passed (9)
     Tests  157 passed (157)

# after (this branch)
Test Files  9 passed (9)
     Tests  158 passed (158)
```

`pnpm --filter @x402/stellar build` and `pnpm --filter @x402/stellar lint:check` both pass clean.

### DIFF
--- a/typescript/.changeset/stellar-facilitator-feebumpsigner-safety-check.md
+++ b/typescript/.changeset/stellar-facilitator-feebumpsigner-safety-check.md
@@ -1,0 +1,5 @@
+---
+"@x402/stellar": patch
+---
+
+Extend the Stellar facilitator's safety checks (`unsafe_tx_or_op_source`, `facilitator_is_payer`, `facilitator_in_auth`) to also cover `feeBumpSigner`, not just `signingAddresses`. `feeBumpSigner` is now checked via a separate `facilitatorSafetyAddresses` set rather than merged into `signingAddresses`, since `signerMap`-based signer selection in `settle()` does not expect it there. Fixes #3332.
--- a/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts
@@ -93,6 +93,14 @@ export class ExactStellarScheme implements SchemeNetworkFacilitator {
   public readonly feeBumpSigner?: FacilitatorStellarSigner;
   private readonly signerMap: Map<string, FacilitatorStellarSigner>;
   private readonly selectSigner: (addresses: readonly string[]) => string;
+  /**
+   * Addresses the facilitator-safety checks must treat as facilitator-controlled.
+   * Deliberately a separate set from `signingAddresses`: that set also backs
+   * signer *selection* (`signerMap.get(...)` in `settle()`), and `feeBumpSigner`
+   * has no entry in `signerMap`, so merging it into `signingAddresses` directly
+   * would let round-robin selection pick an address `settle()` can't sign with.
+   */
+  private readonly facilitatorSafetyAddresses: ReadonlySet<string>;
 
   /**
    * Creates a new ExactStellarScheme instance.
@@ -142,6 +150,9 @@ export class ExactStellarScheme implements SchemeNetworkFacilitator {
     this.maxTransactionFeeStroops = maxTransactionFeeStroops ?? DEFAULT_MAX_TRANSACTION_FEE_STROOPS;
     this.selectSigner = selectSigner ?? roundRobinSelectSigner();
     this.feeBumpSigner = feeBumpSigner;
+    this.facilitatorSafetyAddresses = this.feeBumpSigner
+      ? new Set([...this.signingAddresses, this.feeBumpSigner.address])
+      : this.signingAddresses;
   }
 
   /**
@@ -426,8 +437,8 @@ export class ExactStellarScheme implements SchemeNetworkFacilitator {
       }
 
       if (
-        this.signingAddresses.has(operation.source ?? "") ||
-        this.signingAddresses.has(transaction.source)
+        this.facilitatorSafetyAddresses.has(operation.source ?? "") ||
+        this.facilitatorSafetyAddresses.has(transaction.source)
       ) {
         return {
           response: invalidVerifyResponse("invalid_exact_stellar_payload_unsafe_tx_or_op_source"),
@@ -465,7 +476,7 @@ export class ExactStellarScheme implements SchemeNetworkFacilitator {
       const toAddress = scValToNative(args[1]) as string;
       const amount = scValToNative(args[2]) as bigint;
 
-      if (this.signingAddresses.has(fromAddress)) {
+      if (this.facilitatorSafetyAddresses.has(fromAddress)) {
         return {
           response: invalidVerifyResponse("invalid_exact_stellar_payload_facilitator_is_payer"),
         };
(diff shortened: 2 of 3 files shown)


PhilBot402:
@Eras256 Thanks for the targeted fix — the underlying issue in #3332 is real, and the approach looks right.

**What looks good**
- `facilitatorSafetyAddresses` correctly extends facilitator-safety checks to `feeBumpSigner` without merging it into `signingAddresses`, which would break signer selection in `settle()` (since `feeBumpSigner` is not in `signerMap`).
- All three spec-mandated checks are updated: `unsafe_tx_or_op_source`, `facilitator_is_payer`, and `facilitator_in_auth`.
- The regression test meaningfully covers the gap (on `main` the payload is rejected for the wrong reason via simulation event mismatch; with this change it fails at the intended guard).

**Requesting changes before merge**
1. **Unsigned commit(s)** — upstream requires verified/signed commits before review. Please rebase and sign.
2. **Missing changeset** — `@x402/stellar` needs a changelog fragment per contributing guidelines (`pnpm changeset` from `typescript/`).

**Optional (non-blocking):** consider adding similar regression coverage for `feeBumpSigner` as tx/op source and in auth entries, since the same set backs all three checks.

Once signing and the changeset are in place, this looks good to merge from a code perspective.

phdargen:
Please verify ALL your commits and add the changeset again @Eras256

phdargen (approved):
Thanks @Eras256 , LGTM 🚀

Eras256:
Thanks for reviewing and merging this, @phdargen.

PR #3338, opened 2026-09-01
fix(core): enforce settlement override ceiling against PaymentRequirements.amount

### LINKED ISSUE #3334: @x402/core: settlePayment() settlement-override resolution never enforces its own documented ceiling against PaymentRequirements.amount
## Summary

`x402ResourceServer.settlePayment()` (and the `resolveSettlementOverrideAmount()`
helper it calls) never enforces the invariant `SettlementOverrides.amount`'s
own JSDoc documents: "The resolved amount must be `<=` the authorized
maximum in `PaymentRequirements`." A percent or dollar override can resolve
to an amount that exceeds `PaymentRequirements.amount` and is forwarded to
the facilitator client's `settle()` with no error raised anywhere in
`@x402/core`.

## Where

`@x402/core@2.24.0`, `dist/cjs/server/index.js` (source:
`src/server/x402ResourceServer.ts`).

`SettlementOverrides.amount`'s documented contract (`x402Client-DkVHeSBW.d.ts`,
quoted verbatim):

```ts
interface SettlementOverrides {
  /**
   * Amount to settle. Supports three formats:
   * - Raw atomic units...
   * - Percent — e.g., "50%" settles 50% of `PaymentRequirements.amount`...
   * - Dollar price — e.g., "$0.05" converts a USD-denominated price to atomic units...
   *
   * The resolved amount must be <= the authorized maximum in `PaymentRequirements`.
   *
   * Note: Setting this to an amount other than `PaymentRequirements.amount` is
   * only valid in schemes that support partial settlement, such as `upto`.
   */
  amount?: string;
}
```

The actual resolution logic, `resolveSettlementOverrideAmount()`:

```js
function resolveSettlementOverrideAmount(rawAmount, requirements, decimals) {
  const percentMatch = rawAmount.match(/^(\d+(?:\.\d{0,2})?)%$/);
  if (percentMatch) {
    const [intPart, decPart = ""] = percentMatch[1].split(".");
    const scaledPercent = BigInt(intPart) * 100n + BigInt(decPart.padEnd(2, "0").slice(0, 2));
    const base = BigInt(requirements.amount);
    return (base * scaledPercent / 10000n).toString();   // no upper-bound check
  }
  const dollarMatch = rawAmount.match(/^\$(\d+(?:\.\d+)?)$/);
  if (dollarMatch) {
    if (decimals === void 0) { throw new Error(...); }
    return convertToTokenAmount(dollarMatch[1], decimals);   // no upper-bound check either
  }
  return rawAmount;
}
```

And its only caller, `settlePayment()`:

```js
async settlePayment(paymentPayload, requirements, declaredExtensions, transportContext, settlementOverrides, phase = "after-handler") {
  ...
  if (settlementOverrides?.amount !== void 0) {
    ...
    effectiveRequirements = {
      ...requirements,
      amount: resolveSettlementOverrideAmount(settlementOverrides.amount, requirements, decimals)
      // <-- resolved value is used directly, never compared against requirements.amount
    };
  }
  ...
}
```

Nothing downstream checks it either — `x402Facilitator.settle()` forwards
`paymentRequirements` straight to the resolved scheme's `settle()` with no
amount comparison of its own.

## Repro

Ran directly against the real `@x402/core@2.24.0` server module via its
public `settlePayment()` API (no private helpers accessed directly), using
a fake `FacilitatorClient` that just echoes back whatever
`PaymentRequirements` it's asked to settle, to observe exactly what gets
forwarded:

```js
import { x402ResourceServer } from "@x402/core/server";

const server = new x402ResourceServer(fakeFacilitatorClient);
server.register("stellar:testnet", { scheme: "exact", ...minimalSchemeConfig });

const requirements = { scheme: "exact", network: "stellar:testnet", amount: "1000000", ... };

await server.settlePayment(paymentPayload, requirements, undefined, undefined, { amount: "500%" });
```

Output:

```
=== Case 1: settlementOverrides.amount = "500%" ===
Authorized maximum (PaymentRequirements.amount) was: 1000000
Resolved/forwarded amount was:                       5000000
>>> BUG CONFIRMED: forwarded amount EXCEEDS the authorized maximum, no error was thrown.

=== Case 2: settlementOverrides.amount = "$1000000" (way above a 7-decimal $0.10 cap) ===
Authorized maximum (PaymentRequirements.amount) was: 1000000
Resolved/forwarded amount was:                       10000000000000
>>> BUG CONFIRMED: forwarded amount EXCEEDS the authorized maximum, no error was thrown.
```

Both cases returned `success: true` from `settlePayment()`.

## Not a duplicate of

- PR #2962 (Go SDK, merged) fixes an `int64` overflow in the equivalent Go
  function for very large percent strings. Its own added test cases (e.g.
  `"999999999999999999999%"`) explicitly still expect an unbounded
  proportional result to come back without error — confirming that fix did
  not add a ceiling check in any language. This is a distinct gap from that
  overflow fix.
- #3001 / #3003 (open, spec-level discussion) concern whether the `upto`/
  batch-settlement trust model should let a server fairly charge up to the
  signed ceiling in the first place — an already-accepted design question,
  not about whether the ceiling itself is enforced once accepted.
- Checked `gh issue list`/`gh pr list --repo x402-foundation/x402` against
  `settlementOverrides`, `resolveSettlementOverrideAmount`, `authorized
  maximum`, `partial settlement` — nothing else covers "resolved override
  amount can exceed `PaymentRequirements.amount` with no check in
  `@x402/core`."

## Impact

`settlementOverrides` is trusted, first-party input — either a direct
argument the resource-server operator's own code passes to
`settlePayment()`, or the `Settlement-Overrides` response header the
operator's own route handler sets — never something an external
buyer/attacker controls directly, so this isn't remotely exploitable by a
third party through this library alone. Calibrating honestly: medium
severity. For schemes with their own on-chain ceiling enforcement (e.g. a
Soroban `upto`-settlement contract that asserts `actual_amount <=
max_amount` independent of what the facilitator submits), an inflated
override would still fail at the chain layer — but that protection is
scheme/chain-specific, not something `@x402/core` itself provides despite
its own type documenting the check as mandatory. A realistic trigger is
mundane rather than adversarial: a usage-based billing integration
computing a percentage or dollar amount with a units/decimals slip (e.g.
`5.0` meant as a fraction landing in the `%` branch as `"500%"`) gets
forwarded with zero pushback from the library, and the same gap exists in
the Go and (per #2963's file list) Python implementations of this
function.

## Suggested fix

After resolving the override in `resolveSettlementOverrideAmount()` (or
immediately after, in `settlePayment()`), compare the resolved value
against `BigInt(requirements.amount)` and throw a clear error (mirroring
the existing "asset decimals are unknown" error style) when it's exceeded
— centralizing the check once in `@x402/core` rather than leaving it to be
reimplemented independently by every per-chain
`SchemeNetworkFacilitator`. Worth fixing the same way across the Go and
Python SDKs too, matching the pattern PR #2962 already set for the overflow
case.

Eras256:
Fixes #3334.

`SettlementOverrides.amount`'s own JSDoc states: "The resolved amount must be <= the authorized maximum in `PaymentRequirements`." `resolveSettlementOverrideAmount()` never enforced this for any of its three supported formats (raw atomic units, percent, dollar) — a percent override like `"500%"` or a dollar override priced above the authorized ceiling resolved cleanly and was forwarded straight to the facilitator client's `settle()`, with no error raised anywhere in `@x402/core`.

## What changed

- `resolveSettlementOverrideAmount()` now resolves whichever format was used into a local `resolved` value, then compares it against `BigInt(requirements.amount)` once at the end and throws a clear error if it's exceeded — covering all three formats uniformly rather than duplicating the check per branch.
- Fixed two existing test fixtures that happened to combine a small `requirements.amount` with a dollar-format override large enough to newly trip this check (the "dollar price format" describe block's shared fixture, and the `"should resolve dollar override using scheme getAssetDecimals"` test) — bumped their `amount` fixtures, no assertions changed.
- Fixed one integration fixture (`test/integrations/upto.test.ts`) whose declared price, `"$10.00"`, produced a non-integer `PaymentRequirements.amount` ("10.00") that the new check's `BigInt()` parse can't accept — changed to `"$10"` to match the atomic-unit-string convention every other fixture in the suite already follows. Verified this doesn't affect any assertion in the three tests that use that route.
- Added a new `"authorized maximum enforcement"` describe block covering all three formats exceeding the ceiling, plus the exact-equal boundary not throwing.

Ran the actual `@x402/core` server module via `settlePayment()`/`resolveSettlementOverrideAmount()` directly before writing the fix, to confirm the gap for real:

```
=== settlementOverrides.amount = "500%" against requirements.amount = "1000000" ===
Resolved/forwarded amount was: 5000000   (5x the authorized maximum, no error)

=== settlementOverrides.amount = "$1000000" against a 7-decimal requirements.amount = "1000000" ===
Resolved/forwarded amount was: 10000000000000   (no error)
```

After the fix, ran the full `@x402/core` suite (not just the touched file — a prior pass over #1699 in a different repo taught me to check the whole suite, not just the file I edited):

```
# before this branch
Test Files  22 passed (22)
     Tests  691 passed (691)

# after this branch
Test Files  22 passed (22)
     Tests  695 passed (695)
```

`pnpm --filter @x402/core build` and `pnpm --filter @x402/core lint:check` both pass clean. Also grepped the whole `typescript/` workspace for other consumers of `resolveSettlementOverrideAmount`/`SettlementOverrides` (the `http/*` packages just re-export `x402HTTPResourceServer`, no independent override logic) and ran `@x402/express`/`@x402/fastify`/`@x402/next`/`@x402/hono`'s own suites — one pre-existing, unrelated failure in `@x402/express` (a wildcard-route 402 status test) reproduces identically on `main` with this branch's changes stashed, confirming it's not caused by this change.

### DIFF
--- a/typescript/.changeset/settlement-override-authorized-maximum.md
+++ b/typescript/.changeset/settlement-override-authorized-maximum.md
@@ -1,0 +1,5 @@
+---
+"@x402/core": patch
+---
+
+`resolveSettlementOverrideAmount()` now enforces the settlement override ceiling against `PaymentRequirements.amount` across raw, percent, and dollar override formats. Previously a settlement override could resolve above the requirement's authorized maximum, which contradicts the documented `SettlementOverrides.amount` contract and the `upto` scheme's maximum-amount rule.
--- a/typescript/packages/core/src/server/x402ResourceServer.ts
+++ b/typescript/packages/core/src/server/x402ResourceServer.ts
@@ -277,29 +277,42 @@ export function resolveSettlementOverrideAmount(
   requirements: PaymentRequirements,
   decimals?: number,
 ): string {
+  let resolved: string;
+
   // Percent format: "50%" or "33.33%"
   const percentMatch = rawAmount.match(/^(\d+(?:\.\d{0,2})?)%$/);
   if (percentMatch) {
     const [intPart, decPart = ""] = percentMatch[1].split(".");
     const scaledPercent = BigInt(intPart) * 100n + BigInt(decPart.padEnd(2, "0").slice(0, 2));
     const base = BigInt(requirements.amount);
-    return ((base * scaledPercent) / 10000n).toString();
-  }
-
-  // Dollar price format: "$0.05"
-  const dollarMatch = rawAmount.match(/^\$(\d+(?:\.\d+)?)$/);
-  if (dollarMatch) {
-    if (decimals === undefined) {
-      throw new Error(
-        `Cannot convert dollar settlement override "${rawAmount}" to atomic units: ` +
-          `asset decimals are unknown. Pass an atomic amount or register the asset.`,
-      );
+    resolved = ((base * scaledPercent) / 10000n).toString();
+  } else {
+    // Dollar price format: "$0.05"
+    const dollarMatch = rawAmount.match(/^\$(\d+(?:\.\d+)?)$/);
+    if (dollarMatch) {
+      if (decimals === undefined) {
+        throw new Error(
+          `Cannot convert dollar settlement override "${rawAmount}" to atomic units: ` +
+            `asset decimals are unknown. Pass an atomic amount or register the asset.`,
+        );
+      }
+      resolved = convertToTokenAmount(dollarMatch[1], decimals);
+    } else {
+      // Raw atomic units (existing behavior)
+      resolved = rawAmount;
     }
-    return convertToTokenAmount(dollarMatch[1], decimals);
   }
 
-  // Raw atomic units (existing behavior)
-  return rawAmount;
+  // SettlementOverrides.amount's own contract: the resolved amount must be
+  // <= the authorized maximum in PaymentRequirements, for every format above.
+  if (BigInt(resolved) > BigInt(requirements.amount)) {
+    throw new Error(
+      `Settlement override amount "${rawAmount}" resolves to ${resolved} atomic units, ` +
+        `which exceeds the authorized maximum of ${requirements.amount} in PaymentRequirements.`,
+    );
+  }
+
+  return resolved;
 }
 
 type HookAdapterHandles = {
--- a/typescript/packages/core/test/integrations/upto.test.ts
+++ b/typescript/packages/core/test/integrations/upto.test.ts
@@ -151,7 +151,7 @@ describe("Upto Integration Tests", () => {
         accepts: {
           scheme: "cash",
           payTo: "merchant@example.com",
-          price: "$10.00",
+          price: "$10",
           network: "x402:cash" as Network,
         },
         description: "AI generation with upto billing",
--- a/typescript/packages/core/test/unit/server/x402ResourceServer.test.ts
+++ b/typescript/packages/core/test/unit/server/x402ResourceServer.test.ts
@@ -1725,7 +1725,7 @@ describe("x402ResourceServer", () => {
       const requirements = buildPaymentRequirements({
         scheme: "exact",
         network: "eip155:8453" as Network,
-        amount: "1000000",
+        amount: "10000000",
       });
 
       await server.settlePayment(payload, requirements, undefined, undefined, { amount: "$0.05" });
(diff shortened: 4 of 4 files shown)


PhilBot402:
Thanks @Eras256 — the centralized ceiling check in `resolveSettlementOverrideAmount()` looks correct and matches the documented `SettlementOverrides.amount` contract and the `upto` spec's maximum-amount rules. Good test coverage across raw, percent, and dollar formats, plus the fixture updates.

One thing before merge: please add a `@x402/core` changeset per contributing guidelines.

Eras256:
@phdargen — following up on this one too. The changeset you asked for landed Sep 3 (`cfa44a02`, "chore: add changeset for settlement override ceiling fix") — quiet since. It's been 12 days on a fix for our own closed #3334 (the documented ceiling on `SettlementOverrides.amount` was never actually enforced). Still green and mergeable if there's nothing else blocking.

PR #3347, opened 2026-09-02
feat(go): add delegated receiver authorizer for SVM upto

PhilBot402:
Port of https://github.com/x402-foundation/x402/pull/3346 from TypeScript to Go SDK.

SVM upto can optionally delegate receiverAuthorizer to the facilitator, matching EVM batch-settlement: /supported may advertise receiverAuthorizer, and a server that omits ReceiverAuthorizerSigner no longer needs a voucher-signing key. The facilitator signs the claim voucher only after binding the deposit to a caller identity and matching that same identity on claim; a lost binding fails closed. Self-managed server vouchers stay the default. Settle routing prefers server-stamped payload.type (deposit | claim) and falls back to voucher/amount inference so older non-delegating servers keep working. HTTP adapters now exit on permanent FacilitatorCapabilityError or RouteConfigurationError at eager Initialize, while transient facilitator timeouts remain retryable.

### DIFF
--- a/examples/go/facilitator/upto/README.md
+++ b/examples/go/facilitator/upto/README.md
@@ -58,6 +58,68 @@ scheme := uptosvm.NewUptoSvmScheme(signer, &uptosvm.Config{
 })
 ```
 
+## SVM receiver authorizer (optional delegation)
+
+This example registers `UptoSvmScheme` with a **fee payer only** — no
+`AuthorizerSigner` is configured, so `/supported` advertises `extra.feePayer`
+but not `extra.receiverAuthorizer`. Servers must sign their own claim vouchers
+(self-managed mode).
+
+To let resource servers delegate voucher signing to your facilitator, extend
+the SVM registration with a separate Ed25519 key and a `ResolveCallerIdentity`
+hook. Delegation is not negotiated in x402 — it requires an out-of-band
+agreement with each resource server, and authenticated settle requests so
+claim vouchers are signed only for that server.
+
+| Signer | Role | Onchain effect |
+| ------ | ---- | -------------- |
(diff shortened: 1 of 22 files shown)


phdargen:
## CDP facilitator integration notes (local checkout)

Wiring this into cdp-facilitator surfaced three host-side gaps. The first is blocking; the other two degrade fail-closed / retry behavior.

### 1. `DelegatedSettleContext` needs the settle `context.Context` (blocking)

`ResolveCallerIdentity` is constructed once at `New()`. CDP's project lives on the request via `cloudauth.GetProject(ctx)`. `FacilitatorContext` cannot carry request values. `Settle` already has `ctx` — please add `Ctx context.Context` to `DelegatedSettleContext` and pass the settle context through on deposit and claim.

Without this, a host cannot implement the identity hook from request-scoped auth.

Patched locally on our checkout of this PR.

### 2. `Get` collapses store error with not-found / mismatch

All three become `ErrDelegatedSettleUnauthenticated` (client 400):

```go
binding, err := f.delegatedAuthStore.Get(...)
if err != nil || binding == nil || binding.CallerIdentity != identity {
    return x402.NewSettleError(ErrDelegatedSettleUnauthenticated, ...)
}
```

EVM channel binding treats store failure as a **retryable** fail-closed error (500). Please return a distinct retryable code (or propagate the store error) so a host can map infra blips to 500 instead of telling the client they are unauthenticated.

### 3. `Bind` semantics for a durable unique index

In-memory overwrite is last-writer-wins. A durable unique `(channel_id, network)` index will fail a legitimate retry after bind-then-broadcast-fail.

Hosts must upsert **same identity** / first-writer-wins, and error on a different identity (anti-squat). Worth documenting on `DelegatedAuthStore.Bind` so in-memory default and durable adapters do not silently diverge.

CDP will implement first-writer-wins on Mongo regardless.

CarsonRoscoe (approved):

PR #3355, opened 2026-09-03
feat: optimize go facilitator SDK - EVM and SVM

CarsonRoscoe:
## Description

Four targeted latency/RPC-volume optimizations to the Go facilitator's EVM and SVM verify/settle paths, audited against production reachability in `cdp-facilitator` (config facts, not guesses — see design rationale below). No behavior change to the default configuration except where called out as breaking.

**1. Parallel verify simulation (EVM exact + Permit2), opt-in**
`verifyEIP3009` and `VerifyPermit2` issue `GetCode`-based signature classification, then a separate `eth_call` simulation, sequentially — even though the simulation's calldata is fully determined by local ERC-6492/Permit2 parsing and never reads `GetCode`'s result (`SigData.CodeDeployed`). Root cause: `eip3009.go` line ~94-133 (pre-change) ran `ClassifyEIP3009Signature` to completion before starting `SimulateEIP3009Transfer`, with no data dependency between them. Now the simulation starts concurrently with classification and its result is read after classification/asset checks, so error precedence is unchanged. Gated behind `ExactEvmSchemeConfig.EnableParallelVerifySimulation` / `VerifyPermit2Options.EnableParallelSimulation`, both defaulting to `false`: the tradeoff is one wasted `eth_call` per payment rejected on signature grounds, which should be an explicit opt-in rather than inherited behavior. Saves one sequential RPC round trip on `/verify` when enabled. The Permit2 gas-sponsoring branches (EIP-2612, ERC-20 approval) are excluded — they simulate different calls and stay sequential.

**2. Network-scoped asset-contract cache**
`ValidateAssetIsContract` (`utils.go`) issued a fresh `eth_getCode` on the payment token for every verify *and* every settle, with no caching. Added a 15-minute, size-bounded (4096 entries), positive-result-only in-memory cache keyed by `(network, asset)` — negative results aren't cached so a token mid-deployment self-heals on the next request. **Breaking:** `evm.ValidateAssetIsContract` now takes a `network string` argument; without it, cache entries would collide across chains where the same address can hold bytecode on one chain and nothing on another. Callers use the new `evm.StartAssetContractCheck`/`.Await()` pair, which records the cache entry only after the caller has consumed the result — this avoids a race where an abandoned goroutine (e.g. from an early-returning verify) could populate the cache with a result nobody validated. This is primarily an RPC-volume and tail-latency win, not a p50 win: the asset check already ran concurrently with signature verification, so removing it mainly eliminates a recurring `eth_getCode` and the timeout/retry tail risk that came with it.

**3. Reuse verify's payer code lookup in ERC-6492 settle**
`settleEIP3009`'s ERC-6492 branch called `f.signer.GetCode(ctx, payer)` again to decide whether to deploy, duplicating the `GetCode` that `VerifyUniversalSignature` already performed inside the `verifyEIP3009` call settle had just awaited. `verifyEIP3009` now also returns the `EIP3009SignatureClassification` it computed, and settle reads `classification.SigData.CodeDeployed` instead of re-fetching. Both reads happen within one settle call, before any deploy transaction — this is not the previously-documented post-deploy re-read that races RPC state propagation across load-balanced nodes (that hazard is unaffected). `invalid_exact_evm_failed_to_check_deployment` is no longer emitted by this path; the wire constant is kept but marked reserved rather than removed, since it's part of the exported error surface. Applies the same fix in `permit2.go`'s equivalent `GetCode` fallback.

**4. Linear backoff for SVM upto channel-account reads**
`channel.go`'s `fetchAndVerifyOpenChannel` retried a lagging channel-account read with exponential backoff (200/400/800/1600ms, 5 reads, 3.0s total). A 1600ms single wait is ~4 Solana slots against replica lag that's typically a small multiple of the ~400ms slot time, so most of the budget was spent on one oversized wait. Changed to linear backoff (200/400/600/800/1000ms across 6 reads) — same 3.0s total budget, but one extra read and a max single wait of 1s (~2.5 slots) instead of 1.6s. Both `ChannelReadMaxAttempts` and `ChannelReadBackoffStep` are now configurable via `Config` for operators who need a wider budget.

## Tests

- `go test -race -count=1 ./...` — full suite, clean, no data races.
- `go test -count=1 -shuffle=on ./mechanisms/... ./test/unit/...` — run 3× to catch cache/goroutine ordering issues; clean.
- New `mechanisms/evm/exact/facilitator/roundtrip_test.go`: a mock signer that counts RPC calls per method and can hold `GetCode` open until the simulation's `eth_call` arrives, turning "runs concurrently" into a happens-before assertion rather than a timing threshold. Covers: parallel simulation for EIP-3009 and Permit2, opt-in-off sequential behavior (negative control for both), asset-cache hit/miss counts, per-network cache isolation, and settle reusing verify's payer-code lookup on the ERC-6492 path.
- New `mechanisms/evm/asset_cache_test.go`: cache hit/miss, the abandoned-goroutine-doesn't-cache case, empty-network bypass, TTL expiry, and the entry-count bound.
- New/extended tests in `mechanisms/svm/upto/facilitator/channel_test.go` for the linear backoff schedule, config overrides, and exact attempt-count enforcement.
- `make lint` — 0 issues. `make fmt` — no diff.
- Full Go e2e suite against Base Sepolia + Solana Devnet, exercising every changed path through the real Go facilitator: `go/http/echo` + `typescript/http/express` servers × `go/http/go-http` + `typescript/http/axios` clients, all 44 scenarios passing (EIP-3009, Permit2 direct/gas-sponsored, upto, batch-settlement).
- Cross-checked failures against a clean `HEAD` worktree to confirm zero regressions: several pre-existing integration-test failures (unrelated `valid_before` timing, a pre-existing `GetAccountInfo` panic in an SVM test helper, and a `cumulative_amount_mismatch` in batch-settlement) reproduce identically on unmodified `main` and are outside this PR's scope.

## Checklist

- [ x My commits are signed (required for merge)

### DIFF
--- a/go/.changes/unreleased/changed-20260902-230501.yaml
+++ b/go/.changes/unreleased/changed-20260902-230501.yaml
@@ -1,0 +1,3 @@
+kind: added
+body: EVM exact facilitator verify can start the onchain transfer simulation concurrently with signature classification, removing one RPC round trip. Opt in with ExactEvmSchemeConfig.EnableParallelVerifySimulation (or VerifyPermit2Options.EnableParallelSimulation for direct VerifyPermit2 callers); it covers EIP-3009 and the Permit2 standard settle path. Off by default, since the tradeoff is one wasted eth_call per payment rejected on signature grounds.
+time: 2026-09-02T23:05:01.000000-04:00
--- a/go/.changes/unreleased/changed-20260902-230612.yaml
+++ b/go/.changes/unreleased/changed-20260902-230612.yaml
@@ -1,0 +1,3 @@
+kind: changed
+body: 'EVM facilitators now cache a positive asset-contract check for 15 minutes instead of issuing a fresh eth_getCode on the payment token for every payment. Only positive results are cached, so a token observed mid-deployment still recovers on the next request. Breaking: evm.ValidateAssetIsContract takes a network argument, which scopes the cache so entries cannot collide across chains.'
+time: 2026-09-02T23:06:12.000000-04:00
--- a/go/.changes/unreleased/changed-20260902-230725.yaml
+++ b/go/.changes/unreleased/changed-20260902-230725.yaml
@@ -1,0 +1,3 @@
+kind: changed
+body: EVM exact settle's ERC-6492 branch now reads payer deployment from the verify it already awaited rather than issuing a second eth_getCode. Both reads happen within one settle call and before any deploy transaction, so this is not the post-deploy re-read that races RPC state propagation. Settle consequently no longer emits invalid_exact_evm_failed_to_check_deployment.
+time: 2026-09-02T23:07:25.000000-04:00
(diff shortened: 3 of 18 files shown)


phdargen (approved):
LGTM 🚀

PR #3370, opened 2026-09-04
specs(exact): add near-intents `upfront` method and crosschain-swap discovery extension

IkerAlus:
## Description

Adds the `near-intents` asset transfer method for `exact`, and the `crosschain-swap` discovery extension. A client pays on any supported origin network into a single-use deposit address. The NEAR Intents solver network converts and delivers the destination asset to the merchant. The facilitator confirms delivery before the resource executes.

Supersedes #2102 (could not be reopened).

Files:
- `specs/schemes/exact/scheme_exact_near_intents.md`: mechanism. Client-submitted payment-proof family per `scheme_exact.md` (#3145), `paymentFlow: upfront` (#3053).
- `specs/extensions/extension-crosschain-swap.md`: origin discovery, vendor-neutral. Lists routes with indicative prices, not a payment path.

## Main changes from #2102

- **`network` is the origin network**, one `accepts[]` entry per offered origin, deposit address as `payTo`. No `near:mainnet`.
- **`exact` + `paymentFlow: upfront`**, not a new scheme. `/verify` is not invoked; checks run inside `/settle` before the route handler.
- **Minimal `extra`**: `assetTransferMethod`, `paymentFlow`, `refundTo`, conditional `depositMemo`. The destination leg is not on the wire: fixed in the quote, enforced by the backend, reported in the receipt. Removes the amount/asset-semantics and asset-format issues from the first review.
- **No handshake / extra round trip.** The deposit address does not depend on the client: the facilitator mints one quote per (resource, origin) and serves it in every 402 until it expires or is funded. Merchants curate origins.
- **Concurrent deposits to a shared address**, per @phdargen's question: the merchant receives value on every path, so validity is timing-independent — a proof is valid if the deposit reached the merchant as swap input. Facilitators MUST rotate the quote on first deposit detection (not on success). Both clients are served; the second in the origin asset. Stated as the failure disposition so clients see it before paying.
- ~`refundTo` is merchant-controlled** (facilitator cannot know the payer pre-payment). Consequence stated: a sufficient deposit always pays the merchant and serves the client; insufficient or late deposits recover off-band.~
- **Not-yet-final handling**: claim released, `exact_near_intents_not_final`, retry with the same proof.
- v2 header names.

## AI assistance

Drafting done with AI assistance; all content reviewed by the author.

## Checklist


phdargen:
Thanks for the updated spec @IkerAlus!

Two quick comments:

- nit: please explicitly declare the client-submitted (payment-proof) family per https://github.com/x402-foundation/x402/pull/3145
- `refundTo` is merchant-controlled on origin: this undercuts a big value prop of the scheme, ie that a server can use a single wallet on one destination chain and accept payments from many origin chains. If a server has to provide and maintain `refundTo` addresses on each origin chain, it might as well accept payments on those origin chains via the regular mechanisms. I think it would be worth to explore possible flows where `refundTo` is client-controlled and refunds are always treated as failure. This would also simplify the concurrent payment issue

IkerAlus:
hi @phdargen thanks for the prompt feedback.

- nit added
- that is a very good point. I simply adopted the easy path for refunds and overlooked the UX implcations for the merchant. I just spec'ed a client-controlled refund model that should solve this without adding extra steps to the flow: 

    - The quote's `refundTo` is a single facilitator-controlled account on NEAR Intents (`refundType: INTENTS`), so refunds from every origin land in one place. The merchant configures nothing on origin chains.
   - On any refund, the facilitator forwards the funds to the origin-network account that sent the deposit. That account is established by the chain, not by whoever presents the proof: tx hashes are public, and a presenter-chosen refund address would let a front-runner redirect a failed payment's refund to themselves.
   - Validity becomes the swap path only: `SUCCESS` with `txHash` among the attributed inputs. Any refund is a failure; the two-path rule goes.
   - Concurrent deposits to a shared address: one served, the other refunded to its own sender. No timing-dependent outcome, no origin-asset receipts for the merchant.
   - `extra` reduces to `assetTransferMethod`, `paymentFlow`, and `depositMemo` where needed. The merchant receives exactly the destination amount per served payment; excess returns to the client.

The (acceptable IMO) cost is the transient facilitator custody of refunds, disclosed in the trust model. Also, important to note that this flow assumes the client controls the sending address. Deposits from exchange or custodial wallets refund to the exchange. This caveat is standard for refund-to-sender system.

I just pushed the new spec version with this new model.

mikedotexe:
hey guys, I've been building an x402 facilitator for NEAR/Base. I think my work with AI may have something useful to share. I don't want to break the momentum here or to distract, but think these might help the effort. If not, feel free to ignore me 😛

1. The draft maps `PaymentRequirements.amount` from `quote.maxAmountIn`, but the current 1Click SDK [`Quote` model](https://github.com/defuse-protocol/one-click-sdk-typescript/blob/ae28ef0348f616dd30c174cb22dd1b1126d8f76b/src/models/Quote.ts#L24-L38) exposes `amountIn` and `minAmountIn`, with no `maxAmountIn`. For `EXACT_OUTPUT`, those fields have materially different [semantics](https://github.com/defuse-protocol/one-click-sdk-typescript/blob/ae28ef0348f616dd30c174cb22dd1b1126d8f76b/src/models/QuoteRequest.ts#L161-L166): `amountIn` includes the input-side buffer, while `minAmountIn` is the minimum accepted input. Could the spec state which value is advertised as the x402 `amount` and which threshold settlement validates?

2. A few examples appear out of sync with core v2: the `crosschain-swap` extension omits the required `schema`; the failure `SettlementResponse` uses `error` and omits the required `transaction`; and the success example pairs the origin `network` with a destination-chain transaction hash. Could these be reconciled so the receipt identifies one transaction and network unambiguously?

IkerAlus:
hi @mikedotexe Thanks for the feedback, both points are right:

1. Indeed, `maxAmountIn` isn't a 1CS API's `/v0/quote` endpoint param (I think I messed up names when choosing between exact input or output. I am fixing it to `amountIn` now.
   - Verified against the live OpenAPI: `amount` = `quote.amountIn`. 
   - Sufficiency follows the backend: a deposit of at least `quote.minAmountIn` is executed and any excess over what the swap consumes is refunded to the sender, below `minAmountIn` it is refunded in full. 
   - The merchant receives exactly `amountOut` either way.

2. Reconciling with core v2: `errorReason` + required `transaction` in the failure receipt, `schema` on the extension, and one transaction per receipt with a matching `network`, the origin deposit moves to `extensions`. Not-yet-final settlements will use core's `settlement_pending` instead of a custom code.

Checking this against the live 1CS API spec also surfaced two small corrections: `deadline` is documented as the point where funds "may be lost", not refunded, and refunds are net of `refundFee`.

IkerAlus:
hi @phdargen , I have improved the spec for two reasons:

- I realized this spec still had the payment-proof family's known issue (@alftom  raised it on #3053) of univocally binding the buyer with the right payment. Anyone who saw the `txHash` first could try to claim the resource. 
-Smaller issue: the middleware match between merchant and facilitator on step 2 of the flow was undefined. There was no facilitator quote endpoint, so two independently built merchant/facilitator pairs couldn't interoperate from the doc.

**How I fixed these two issues and improved the spec:**

- **One offer per `accepts[]` entry, never shared:** Each entry is a fresh deposit address plus a server-issued `operationToken` in `extra`, which the client echoes in the payload. It also becomes the key for the rest of the flow: resolving the original offer on retry, idempotent replay of a settlement result if a response is lost, and a safe `payload.refundTo`. Rotation, shared-address handling and "a quote serves at most one proof" are not needed anymore.
- **Refunds rewritten** around failed payments only: beneficiary is `payload.refundTo`. Also, refunds to the client apply to failed payments only, the surplus is retained by the facilitator and disclosed as a fee (minimal amounts).
- **New section for Quote Operation (`POST /quote`)** to define the middleware logic for quotes between merchant and facilitator. This means the facilitator surface is extended beyond the well-established `/verify`/`/settle`/`/supported`.
- **`INCOMPLETE_DEPOSIT` is non-terminal.** Per the live 1Click docs a short deposit is refunded at the backend's deadline, so a top-up can still complete it; we were consuming the proof on first sight.

phdargen:
Thanks for the iterations @IkerAlus! The current flow clearly improves the UX and fixes some important gaps, but I still see three structural issues:
1. **Custody of client funds.** Refunds go to a facilitator-controlled account and are forwarded later. The server picks the facilitator. The client never chose it and has no agreement with it, yet it holds the client's funds (and keeps the surplus). I think facilitator custody of any funds is problematic and should be avoided. Would also introduce new infrastructure requirements to monitor and issue refunds
2. **New facilitator endpoint.** `POST /quote` extends the facilitator surface beyond `/verify`, `/settle` and `/supported`. It also adds an offer store keyed by `operationToken`, with at least 24h of retention, on top of the deduplication store that every client-submitted method needs. We should avoid inventing a new endpoint
3. **Scheme vs. extension.** The spec is currently a bit of both. It defines an asset transfer method under `exact`, plus a discovery extension that lists origins the client cannot pay, and the destination terms only appear in the extension. I'd pick one: keep `near-intents` as an `exact` asset transfer method (client-submitted family) and drop the extension. Also worth stating explicitly: unlike other methods, `near-intents` isn't tied to one network family. It applies to every origin/destination network 1Click supports

### Suggested flow

Add one extra round trip for the quote and used typed "quote" and "deposit" (or similar) payloads. Both building blocks already exist in `batch-settlement`: it uses typed payloads (`deposit`/`voucher`/`refund`) and it returns a second, "corrective" 402 carrying updated `accepts[]` state in response to a client payload

**1. Initial 402.** One `accepts[]` entry per origin. `payTo` and `amount` stay empty until quoted. The destination terms and (optional) an indicative price (from a cached `dry: true` quote) go in `extra`:

```jsonc
{
  "x402Version": 2,
  "resource": { "url": "https://api.example.com/premium-data" },
  "accepts": [
    {
      "scheme": "exact",
      "network": "eip155:42161",
      "asset": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831",
      "amount": "",
      "payTo": "",
      "maxTimeoutSeconds": 300,
      "extra": {
        "assetTransferMethod": "near-intents",
        "paymentFlow": "upfront",
        "indicativeAmount": "1005000",
        "destination": {
          "network": "eip155:8453",
          "amount": "1000000",
          "recipient": "0xMerchantOnBase"
        }
      }
    }
    // ... one entry per offered origin
  ]
}
```

**2. Client requests a quote.** The payload names the client's own refund address on the origin network and (optionally) a slippage tolerance:

```jsonc
{
  "x402Version": 2,
  "accepted": { /* entry from step 1 */ },
  "payload": {
    "type": "quote",
    "refundTo": "0x2527D02599Ba641c19FEa793cD0F9a6e8457C317",
    "slippageTolerance": 50   // optional, basis points
  }
}
```

Slippage is the client's call. With `EXACT_OUTPUT` the merchant always receives the exact destination amount, and any unused buffer returns to the client's `refundTo`, so the tolerance only trades how much the client sends up front against the chance of a failed swap. If the client omits it, the facilitator applies its default

The server calls `/verify`. The facilitator creates a 1Click quote (`dry: false`, `EXACT_OUTPUT`, `refundTo` = the client, `refundType: ORIGIN_CHAIN`) and returns it:

```jsonc
{
  "isValid": true,
  "extra": {
    "depositAddress": "0x76b4c56085ED136a8744D52bE956396624a730E8",
    "amountIn": "1005000",
    "deadline": "2026-09-23T10:10:00Z",
    "operationToken": "b3f1c9e2..."
    // + depositMemo where the origin network requires one
  }
}
```

`operationToken` is issued by the facilitator, for example as an HMAC of `network:depositAddress[:depositMemo]` keyed with a facilitator secret, so it needs no storage. It binds redemption to whoever received the quote, so someone watching the chain can't redeem the deposit.

A `quote` payload never unlocks the resource.

**3. Quoted 402.** The server maps the verify response onto the entry: `depositAddress` becomes `payTo`, `amountIn` becomes `amount`, `deadline` sets `maxTimeoutSeconds`, and `operationToken` (plus `depositMemo` where present) goes into `extra`. 
```jsonc
{
  "x402Version": 2,
  "resource": { "url": "https://api.example.com/premium-data" },
  "accepts": [
    {
      "scheme": "exact",
      "network": "eip155:42161",
      "asset": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831",
      "amount": "1005000",
      "payTo": "0x76b4c56085ED136a8744D52bE956396624a730E8",
      "maxTimeoutSeconds": 280,
      "extra": {
        "assetTransferMethod": "near-intents",
        "paymentFlow": "upfront",
        "destination": { /* unchanged */ },
        "operationToken": "b3f1c9e2..."
      }
    }
  ]
}
```

**4. Client deposits and presents the proof.**

```jsonc
{
  "x402Version": 2,
  "accepted": { /* quoted entry from step 3 */ },
  "payload": {
    "type": "deposit",
    "operationToken": "b3f1c9e2...",
    "txHash": "0x9bcff372aee89b648c922b850573b22387c31d693079f5e37cd255814e2d615a"
  }
}
```

The server calls `/settle`. The facilitator needs no offer store: it recomputes `operationToken` from `accepted` and rejects a mismatch, then reads `GET /v0/status?depositAddress=` and checks that the original `quoteRequest` matches the requirements (recipient, destination asset and amount, `refundTo`)

**5. SettlementResponse.**

Success (delivered to the merchant):

```jsonc
{
  "success": true,
  "network": "eip155:8453",
  "transaction": "<destination delivery tx hash>",
  "payer": "<origin sender>",
  "amount": "1000000",
  "extra": { "originNetwork": "eip155:42161", "originTxHash": "0x9bcf..." }
}
```
Failure: nothing was delivered, and 1Click refunds the client's `refundTo` directly. No facilitator is involved.

```jsonc
{
  "success": false,
  "errorReason": "exact_near_intents_payment_failed",
  "network": "eip155:42161",
  "transaction": "",
  "extra": { "status": "REFUNDED", "refundTo": "0x2527..." }
}
```

What this gets us:
- No custody. Refunds and surplus go straight back to the client
- No new endpoint, and no facilitator state beyond the deduplication store that any client-submitted method needs
- No extension
- Only the origin the client actually picked gets a real quote

phdargen:
On another topic @IkerAlus: do you see any blockers to also offering a facilitator-broadcast `near-intents` option in the future?

Say the origin is EVM. Once `payTo` is known from the 402, couldn't the client sign eg a regular EIP-3009 authorization to it? The facilitator would then submit and sponsor it in an upfront flow. The tricky part is probably the swap/destination confirmation latency potentially requiring retries/idempotency that we currently don't have in the core protocol.
Do you see other concerns?

If we consider this as a follow-up, we might need to think a bit more about the wire shape now. `assetTransferMethod= near-intents` would conflict with the existing `eip3009, permit2` etc. One option is to keep `assetTransferMethod` for the origin leg only and add a separate field for the delivery route, eg `extra.assetDeliveryMethod: near-intents` (or similar)

IkerAlus:
hi @phdargen, thanks a lot for the proposal, it simplifies the design a lot. I adopted it in the latest push: typed `quote`/`deposit` payloads, the quote minted via `/verify` with the client's own `refundTo` (`refundType: ORIGIN_CHAIN`), a stateless `operationToken`, and no extension. The spec now also says `near-intents` covers every network 1Click supports. So: no custody, no new endpoint, no offer store.

A few fixes on top:
- I added a " Resource Server Behavior" section, mainly to ensure that the terms come from the server:`/verify` and `/settle` receive the server's own unquoted entry as `paymentRequirements`, never `accepted`. Otherwise someone could mint a quote that pays themselves through their own resource server and redeem it at another merchant. 
- Operation token: `keyId.expiry.HMAC(network:depositAddress[:depositMemo]:resource:destinationHash:expiry)`, expiring at the refund deadline + ≥24h. Expired tokens are rejected before any lookup, so used-proof records are kept only until then.
- Timing: `maxTimeoutSeconds` comes from the `deadline` sent in the quote request (the refund cutoff), not the one in 1Click's response.
- `slippageTolerance` is clamped to a range the facilitator publishes in `/supported`. 
- Quote payloads MUST be rate-limited, and a live quote MAY be reused.

One other thing to clarify: core says `upfront` omits `/verify` and that `/verify` is read-only. Here `/verify` runs only for `quote` payloads, before any payment. The paid request is still settle → resource. The quote commits no funds, though it does mint a 1Click quote. OK with this, or should core say it explicitly?

Regarding your latest comment around the Facilitator-broadcast: no blockers I can see, plus the quote round trip makes it easier. The quoted entry has a real `payTo`, so an EVM client can sign an EIP-3009/Permit2 authorization to it and the facilitator submits it in `/settle`. Refunds still go to the client's `refundTo`. Apart from the latency issue, two other concerns:
- the facilitator needs to sponsor gas on every origin it supports,
- on Solana, a fresh deposit address has no token account, and the SVM `exact` fee-payer rules don't let the facilitator pay to create it.

Assuming we are ok with these issues, the changes will imply:
- `assetTransferMethod` names the origin leg: e.g. `"transfer"` for today's method (client broadcasts, presents `txHash`), `eip3009`/`permit2` for the broadcast variant.
- `extra.assetDeliveryMethod: "near-intents"` names the route.
- A core rule mirroring the `paymentFlow` one (§6.1): clients MUST skip `accepts[]` entries whose `assetTransferMethod` or `assetDeliveryMethod` they don't implement. Today the TS EVM `exact` client treats any method other than `permit2` as EIP-3009, so it tries to pay these entries instead of skipping them.

which will also imply opening a small core PR for the rule last rule, so perhaps I can do a follow-up PR with all these changes together.

PR #3372, opened 2026-09-04
feat(ts): Add optional extra.minDeposit hint for EVM batch-settlement

phdargen:
## Description

- Adds optional `extra.minDeposit` (atomic string, MUST be a positive integer `>= amount` when present). Clients SHOULD use a conforming hint as the deposit target and SHOULD enforce a local maximum so a 402 cannot lock unbounded escrow. Servers SHOULD NOT reject solely because `deposit.amount` is below the hint; a local policy MAY reject with `invalid_batch_settlement_evm_deposit_below_min_deposit`. The facilitator MUST NOT enforce this field.
- TypeScript servers always announce the hint (SDK default `10 × amount`; override via `accepts.extra.minDeposit` as Money or atomic). Enforcement is opt-in (`enforceMinDeposit: true`, default off).
- TypeScript clients size from a valid hint, else `amount × depositMultiplier` (default 5, min 3). When `x402Client` resolved a per-payment cap, the lock is also clamped to `maxAmountPerPayment × depositMultiplier`. Uncapped payments (`spendControls: false`, `maxAmountPerPayment: false`, or an `allowedAssets` entry without a cap) leave deposits uncapped too.
- `@x402/core` forwards the already-resolved atomic spend cap on `PaymentPayloadContext.maxAmountPerPayment`. No new core `spendControls` field.
Deposits can sit in escrow through `withdrawDelay` (up to 30 days), so a server-chosen `minDeposit` without a client ceiling is a lock-up risk. A single global atomic `maxDeposit` does not work across tokens with different decimals. Putting deposit policy on core `spendControls` was rejected as scheme-specific. Duplicating the spend-cap allowlist on `depositPolicy` was too heavy. Reusing `depositMultiplier` keeps one knob for both the fallback target and the lock ceiling, and stays consistent with spend controls: if a payment is allowed uncapped, the deposit is too.

## Checklist


chopmob-cloud:
Verified this against the current branch, coming at it from the escrow-lockup angle the description raises, and the client ceiling holds.

The risk is a server-announced `minDeposit` locking unbounded escrow through `withdrawDelay`. `depositAmountForRequest` takes `target = announced ?? depositMultiplier * requestAmount`, then `applyMaxDeposit` clamps to `maxDeposit = maxAmountPerPayment * depositMultiplier`, so however large the announced hint is, a capped payment can never lock more than the client's own ceiling. A hint that is non-atomic, non-positive, or below `requestAmount` is dropped by `parseAnnouncedMinDeposit` and the multiplier fallback applies, so a malformed or hostile hint cannot widen the deposit either.

`applyMaxDeposit` throwing when `needed > maxDeposit` is the right fail-closed choice: `needed` is `maxClaimableAmount - balance` and `maxClaimableAmount` is server-influenced, so refusing to pay beats silently over-locking. Reusing `depositMultiplier` for both the fallback target and the lock ceiling keeps a single knob and avoids a second scheme-specific policy field, which matches the reasoning in the description.

One thing worth confirming on the uncapped path. When a payment is uncapped (`spendControls: false`, `maxAmountPerPayment: false`, or an `allowedAssets` entry with no cap), `maxDeposit` is undefined, so `depositAmountForRequest` returns `max(needed, target)` with `target = announced ?? depositMultiplier * requestAmount`. Without a hint an uncapped client is still bounded by its own multiplier, but with a valid hint `target = announced`, and `applyMaxDeposit` returns it unclamped. So for uncapped clients the server-announced value sets the deposit size, above the client's own multiplier (with the SDK default announce of 10x amount, and more if a server announces more). The description frames uncapped deposits as intentional, so this may be the intended trust model, but it is a different posture from the capped path: there the hint can only tune within the ceiling, here it drives the size. If that is not intended, clamping the honoured hint to `depositMultiplier * requestAmount` even when uncapped would let a server lower or tune the target without inflating it. Otherwise a one-line note that an uncapped client accepts a server-sized deposit would set expectations.

The capped-path bound, the malformed-hint rejection, the fail-closed throw, and the uncapped fallback are all covered by the added unit tests.

phdargen:
> Verified this against the current branch, coming at it from the escrow-lockup angle the description raises, and the client ceiling holds.
> 
> The risk is a server-announced `minDeposit` locking unbounded escrow through `withdrawDelay`. `depositAmountForRequest` takes `target = announced ?? depositMultiplier * requestAmount`, then `applyMaxDeposit` clamps to `maxDeposit = maxAmountPerPayment * depositMultiplier`, so however large the announced hint is, a capped payment can never lock more than the client's own ceiling. A hint that is non-atomic, non-positive, or below `requestAmount` is dropped by `parseAnnouncedMinDeposit` and the multiplier fallback applies, so a malformed or hostile hint cannot widen the deposit either.
> 
> `applyMaxDeposit` throwing when `needed > maxDeposit` is the right fail-closed choice: `needed` is `maxClaimableAmount - balance` and `maxClaimableAmount` is server-influenced, so refusing to pay beats silently over-locking. Reusing `depositMultiplier` for both the fallback target and the lock ceiling keeps a single knob and avoids a second scheme-specific policy field, which matches the reasoning in the description.
> 
> One thing worth confirming on the uncapped path. When a payment is uncapped (`spendControls: false`, `maxAmountPerPayment: false`, or an `allowedAssets` entry with no cap), `maxDeposit` is undefined, so `depositAmountForRequest` returns `max(needed, target)` with `target = announced ?? depositMultiplier * requestAmount`. Without a hint an uncapped client is still bounded by its own multiplier, but with a valid hint `target = announced`, and `applyMaxDeposit` returns it unclamped. So for uncapped clients the server-announced value sets the deposit size, above the client's own multiplier (with the SDK default announce of 10x amount, and more if a server announces more). The description frames uncapped deposits as intentional, so this may be the intended trust model, but it is a different posture from the capped path: there the hint can only tune within the ceiling, here it drives the size. If that is not intended, clamping the honoured hint to `depositMultiplier * requestAmount` even when uncapped would let a server lower or tune the target without inflating it. Otherwise a one-line note that an uncapped client accepts a server-sized deposit would set expectations.
> 
> The capped-path bound, the malformed-hint rejection, the fail-closed throw, and the uncapped fallback are all covered by the added unit tests.

Intended: once a client opts into an unlimited spend cap (not recommended), a server-sized deposit does not add risk on top of that, and unlike spend the deposit is fully refundable escrow. depositMultiplier is only the no-hint fallback in that case

chopmob-cloud:
Makes sense, and the refundable-escrow distinction is the part I under-weighted: an uncapped client has already accepted unbounded spend, and a server-sized deposit is a weaker commitment since it comes back. The one residual is liquidity rather than loss, the deposit is locked through `withdrawDelay` (up to 30 days) before it returns, so an uncapped client with a greedy server ties up capital for that window even though it is recoverable. Given it is opt-in and not recommended, that reads as an acceptable posture. A one-line note in the client docs that an uncapped client accepts a server-sized, refundable but time-locked deposit would save the next reader the same question. Thanks for the quick reply.

CarsonRoscoe (approved):

PR #3411, opened 2026-09-08
fix(mcp): prefer .meta over ._meta in convert_mcp_result for MCP SDK 2.x

### LINKED ISSUE #3149: [Bug]: Python x402MCPClient drops settlement metadata from MCP CallToolResult
### What happened?

The official Python `x402.mcp.x402MCPClient` detects a v2 challenge, creates a payment, retries, and receives the successful paid result, but `MCPToolCallResult.payment_response` is `None` when the underlying client returns the standard Python MCP SDK `mcp.types.CallToolResult`.

Expected: the wrapper preserves `_meta["x402/payment-response"]` and returns a populated `payment_response`.

Cause: `convert_mcp_result()` reads `getattr(mcp_result, "_meta", {})`. The standard MCP Pydantic model exposes the field as `result.meta`; `_meta` is only its wire alias. The metadata is therefore discarded before `extract_payment_response_from_meta()` runs.

A compatible fix is to prefer `mcp_result.meta`, falling back to `mcp_result._meta` for other clients, with an integration regression test that invokes the actual `x402MCPClient`.

### Steps to reproduce

1. Return a valid v2 `PaymentRequired` tool error from an MCP server.
2. On the paid retry, return a successful result with `_meta["x402/payment-response"]`.
3. Wrap a Python MCP SDK client with `x402MCPClient(auto_payment=True)`.
4. Call the tool.

Observed:

```python
result = await buyer.call_tool("paid_tool", {})
assert result.payment_made is True
assert result.is_error is False
assert result.payment_response is None
```

The raw `mcp.types.CallToolResult` contains the settlement under `raw.meta["x402/payment-response"]`. Also, `mcp.types.CallToolResult.model_fields["meta"].alias == "_meta"`.

### Environment

- x402 2.15.0
- fastmcp 3.4.4
- Python MCP SDK model: `mcp.types.CallToolResult`
- Python 3.12
- macOS

wilnowilx:
## Summary

Fixes #3149 — settlement metadata silently dropped when using the standard Python MCP SDK .

## Problem

 reads  to extract payment metadata. However, the standard MCP Python SDK 2.x  is a Pydantic model that exposes the field as ;  is only its wire alias for JSON serialization and does not exist as a Python attribute.

This means when a server wrapper returns a successful paid result, the client never sees the settlement response —  is always .

## Tests

Adds 3 regression tests:
- **test_convert_mcp_result_prefers_meta_over_dunder_meta** — MCP SDK 2x  attribute
- **test_convert_mcp_result_fallback_to_dunder_meta** — MCP SDK 1x  fallback
- **test_convert_mcp_result_meta_takes_precedence** —  wins when both exist

All 24 tests pass.

## Impact

- **Unblocks** anyone using  with  (the default install resolves to 2.x now)
- **No breaking changes** — backwards compatible with MCP SDK 1.x via fallback
- Low-risk, targeted fix

### DIFF
--- a/python/x402/mcp/tests/test_utils.py
+++ b/python/x402/mcp/tests/test_utils.py
@@ -457,3 +457,52 @@ def test_convert_mcp_result_missing_attrs():
     assert result.is_error is False
     assert result.meta == {}
     assert result.structured_content is None
+
+
+def test_convert_mcp_result_prefers_meta_over_dunder_meta():
+    """MCP SDK 2.x exposes ``.meta`` (Pydantic field); ``._meta`` is the wire
+    alias only present in 1.x JSON transport.  convert_mcp_result must prefer
+    ``.meta`` so settlement metadata is not silently dropped.
+
+    Regression test for https://github.com/x402-foundation/x402/issues/3149
+    """
+    from x402.mcp.utils import convert_mcp_result
+
+    class McpSdk2Result:
+        """Simulates MCP SDK 2.x CallToolResult with .meta field."""
+        content = [{"type": "text", "text": "paid result"}]
+        isError = False
+        meta = {"x402/payment-response": {"success": True, "transaction": "0xabc"}}
+        structuredContent = None
+
+    result = convert_mcp_result(McpSdk2Result())
+    assert result.meta == {"x402/payment-response": {"success": True, "transaction": "0xabc"}}
+
+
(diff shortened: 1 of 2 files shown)


wilnowilx:
Closing to withdraw this batch and respect reviewer bandwidth. The MCP 2.x compat work stays in my fork; will re-propose scoped to one PR with discussion first.

PR #3418, opened 2026-09-08
fix(facilitator): validate request schemas before processing

### LINKED ISSUE #3015: Hosted facilitator (x402.org/facilitator): malformed v2 request crashes with 500 instead of returning 400

wilnowilx:
Closes #3015

The hosted facilitator routes cast request body fields with `as` type assertions, skipping all runtime validation. A malformed v2 payload (missing `accepted`, wrong field names) crashes deep in scheme handlers producing opaque HTTP 500 errors.

This PR adds Zod schema validation using the existing `parsePaymentPayload` and `parsePaymentRequirements` from `@x402/core/schemas`.

**Before:** Malformed request -> 500 `unexpected_error` with `Cannot read properties of undefined`
**After:** Malformed request -> 400 with specific Zod validation message (e.g. `paymentPayload validation failed: Required`)

Changes:
- `verify/route.ts`: Replace `as` casts with `parsePaymentPayload` + `parsePaymentRequirements` safe parsing
- `settle/route.ts`: Same validation, plus preserved error shape for settlement failure responses

### DIFF
--- a/typescript/site/app/facilitator/settle/route.ts
+++ b/typescript/site/app/facilitator/settle/route.ts
@@ -1,4 +1,5 @@
-import { PaymentPayload, PaymentRequirements, SettleResponse } from "@x402/core/types";
+import { SettleResponse } from "@x402/core/types";
+import { parsePaymentPayload, parsePaymentRequirements } from "@x402/core/schemas";
 import { getFacilitator } from "../index";
 
 /**
@@ -9,13 +10,10 @@ import { getFacilitator } from "../index";
  */
 export async function POST(req: Request) {
   // Parse request body - only use "unknown:unknown" if parsing fails
-  let paymentPayload: PaymentPayload | undefined;
-  let paymentRequirements: PaymentRequirements | undefined;
+  let body: Record<string, unknown>;
 
   try {
-    const body = await req.json();
-    paymentPayload = body.paymentPayload as PaymentPayload;
-    paymentRequirements = body.paymentRequirements as PaymentRequirements;
+    body = await req.json();
   } catch (error) {
     const errorMessage = error instanceof Error ? error.message : String(error);
     console.error("Failed to parse request body:", errorMessage);
@@ -33,7 +31,7 @@ export async function POST(req: Request) {
   }
 
   // Check for missing parameters
-  if (!paymentPayload || !paymentRequirements) {
+  if (!body.paymentPayload || !body.paymentRequirements) {
     return Response.json(
       {
         success: false,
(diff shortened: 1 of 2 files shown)


wilnowilx:
Closing - #3016 covers adjacent validation and I do not want to split reviewer attention across overlapping PRs. Will revisit via the issue thread if there is appetite.

PR #3420, opened 2026-09-08
fix(core): emit payment-error header on rejected payments

### LINKED ISSUE #3148: payment-error header advertised in CORS but never emitted on rejected/invalid payment retry
**Package**: `@x402/express` (and transitively `@x402/core`), v2.17.0
**Scheme**: `ExactStellarScheme` (`@x402/stellar`), network `stellar:pubnet`
**Facilitator**: OpenZeppelin Channels (`channels.openzeppelin.com/x402`)

## What happened

`paymentMiddlewareFromConfig` correctly advertises `payment-error` in
`access-control-expose-headers` on every 402 response, but the header is
never actually set — on any 402, whether it's the initial unpaid challenge
or a retry with an invalid/rejected payment.

This makes "you didn't send a payment" and "I rejected your payment"
indistinguishable from the client's side. A client that receives an
identical 402 on both the first (unpaid) request and a retry (with a
signed-but-rejected payment) has no way to tell whether its signature was
wrong, the facilitator's verify() failed, or nothing was sent at all.

## Reproduction

```
$ curl -s -i https://nirium-agent-mainnet.fly.dev/api/v1/premium/market
HTTP/2 402
access-control-expose-headers: PAYMENT-REQUIRED,PAYMENT-RESPONSE,X-PAYMENT-RESPONSE,payment-error,payment-verified,Payment-Receipt,X-MPP-Warning
payment-required: eyJ4NDAyVmVyc2lvbiI6Mi...
content-type: application/json; charset=utf-8
{}

$ curl -s -i https://nirium-agent-mainnet.fly.dev/api/v1/premium/market \
    -H "PAYMENT-SIGNATURE: Z2FyYmFnZQ=="
HTTP/2 402
access-control-expose-headers: PAYMENT-REQUIRED,PAYMENT-RESPONSE,X-PAYMENT-RESPONSE,payment-error,payment-verified,Payment-Receipt,X-MPP-Warning
payment-required: eyJ4NDAyVmVyc2lvbiI6Mi...   (identical payload)
content-type: application/json; charset=utf-8
{}
```

`payment-error` is never present in either response, even though it's
advertised as an exposed header.

## Cross-vendor confirmation

We were made aware of this by a third-party integrator (AgentLedger,
agentpayments.fi) who independently observed the exact same signature —
402, empty body, `payment-required` present, `payment-error` absent — on
three unrelated x402 sellers running on Stellar pubnet, including ours.
That points at the SDK's response-construction path rather than at any
single merchant's config.

## Expected behavior

On a rejected/invalid payment retry, `payment-error` should carry the
facilitator's verify() rejection reason (or a generic reason if the
facilitator itself doesn't provide one), so a signed-but-rejected payment
can be distinguished from a missing one in a single request/response
cycle — without needing server-side log access.

## Why this matters

Without this, a legitimate signature failure and a facilitator-side
outage produce the same client-visible symptom, which makes incident
triage (ours and the facilitator's) much slower than it needs to be.
Happy to share the curl transcripts and the third-party integrator's
report if useful.

wilnowilx:
Closes #3148

The x402 middleware advertises `payment-error` in `access-control-expose-headers` on every 402 response, but the header is never actually set. This makes 'no payment sent' and 'payment rejected' indistinguishable from the client side.

**Fix:** Add `payment-error` header to `createHTTPPaymentRequiredResponse` when `PaymentRequired.error` is set.

**Behavior:**
- Initial 402 (no payment sent): `PaymentRequired.error` is unset → no `payment-error` header (backward compatible)
- Rejection (bad signature, stale nonce, etc.): `PaymentRequired.error` is set → `payment-error` header carries the machine-readable reason

This allows clients to distinguish failure modes in a single request/response cycle without server-side log access.

### DIFF
--- a/typescript/packages/core/src/http/x402HTTPResourceServer.ts
+++ b/typescript/packages/core/src/http/x402HTTPResourceServer.ts
@@ -1339,12 +1339,19 @@ export class x402HTTPResourceServer {
   private createHTTPPaymentRequiredResponse(paymentRequired: PaymentRequired): {
     headers: Record<string, string>;
   } {
-    return {
-      headers: {
-        "PAYMENT-REQUIRED": encodePaymentRequiredHeader(paymentRequired),
-        "Cache-Control": PAYMENT_REQUIRED_CACHE_CONTROL,
-      },
+    const headers: Record<string, string> = {
+      "PAYMENT-REQUIRED": encodePaymentRequiredHeader(paymentRequired),
+      "Cache-Control": PAYMENT_REQUIRED_CACHE_CONTROL,
     };
+
+    // Emit payment-error header when there is a rejection reason, so clients
+    // can distinguish "no payment sent" from "payment rejected" in a single
+    // request/response cycle without needing server-side log access.
+    if (paymentRequired.error) {
+      headers["payment-error"] = paymentRequired.error;
+    }
+
+    return { headers };
   }
 
   /**


wilnowilx:
Closing to refine — minor fix, will batch with higher-impact changes.

PR #3421, opened 2026-09-08
fix(evm): validate channel state shape in processPaymentResponse

### LINKED ISSUE #3188: processSettleResponse crashes when extra.channelState has no channelId (client/channel.ts)
## Summary

`processSettleResponse` in `src/batch-settlement/client/channel.ts` dereferences `channelState.channelId` after checking only that `channelState` is a non-null object. A settle response whose `extra.channelState` lacks `channelId` therefore throws inside the client, mid-purchase.

The same field is read defensively elsewhere in the package, which suggests the partial shape is known to be possible.

## The two readers

Unguarded — `client/channel.ts:23`:

```ts
function readResponseChannelState(
  extra: Record<string, unknown>,
): ResponseChannelState | undefined {
  const channelState = extra.channelState;
  if (typeof channelState !== "object" || channelState === null) {
    return undefined;
  }
  return channelState as ResponseChannelState;
}
```

Used at `client/channel.ts:88`:

```ts
const channelState = readResponseChannelState(extra);
if (!channelState) return;

const channelId = channelState.channelId;
const key = channelId.toLowerCase();   // throws when channelId is undefined
```

Guarded — `client/hooks.ts:38`, reading the same field:

```ts
const channelId =
  typeof channelState === "object" && channelState !== null && "channelId" in channelState
    ? channelState.channelId
    : undefined;
if (typeof channelId === "string" && channelId) { ... }
```

## Observed

`@x402/evm@2.22.0`, batch-settlement scheme on Base (eip155:8453), during a normal paid MCP tool call:

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
    at processSettleResponse (@x402/evm/dist/esm/chunk-K3FJDFD6.mjs:226:25)
    at handleBatchSettlementPaymentResponse (chunk-K3FJDFD6.mjs:407:11)
    at onPaymentResponse (chunk-K3FJDFD6.mjs:393:33)
    at _x402Client.handlePaymentResponse (@x402/core/dist/esm/client/index.mjs:182:28)
    at x402MCPClient.callToolWithPayment (@x402/mcp/dist/esm/index.mjs:508:80)
```

Seen once against our own server. We could not identify the producing condition: 21/21 and then 6/6 subsequent settle responses carried `channelId`, and deleting the server-side channel record mid-flight did not reproduce it. We have since made our server guarantee the field is present, so we cannot trigger it again from our side — but any server that emits a partial `channelState` will crash any client built on these packages.

## Suggested fix

Make the reader agree with `hooks.ts` and return `undefined` when the id is absent:

```ts
if (typeof channelState !== "object" || channelState === null) return undefined;
if (typeof (channelState as ResponseChannelState).channelId !== "string") return undefined;
return channelState as ResponseChannelState;
```

A client should not throw while handling a payment response it cannot interpret; skipping the state update leaves the caller able to recover from the chain.

Happy to open a PR if useful.

wilnowilx:
Closes #3188

`processPaymentResponse` reads the `PAYMENT-RESPONSE` header and delegates to `updateChannelFromSettle` with server-reported `chargedAmount` and `channelState.chargedCumulativeAmount`. A malformed server response (partial `channelState`, non-numeric values) could cause an uncaught TypeError during BigInt conversion.

**Fix:** Add defensive validation before BigInt conversion:
1. Validate `chargedAmount` is a digit-only string
2. Validate `channelState.chargedCumulativeAmount` shape — skip state update on malformed data instead of crashing

A client should never throw while handling a payment response it cannot interpret. Skipping the state update leaves the caller able to recover from the chain.

### DIFF
--- a/typescript/packages/mechanisms/evm/src/batch-settlement/client/channel.ts
+++ b/typescript/packages/mechanisms/evm/src/batch-settlement/client/channel.ts
@@ -208,11 +208,30 @@ export async function processPaymentResponse(
   if (chargedAmount !== undefined && typeof chargedAmount !== "string") {
     throw new Error("invalid chargedAmount: not a non-negative integer");
   }
+
+  // Defensive: validate chargedAmount is a valid integer string before BigInt
+  // conversion. A malformed server response could cause an uncaught TypeError.
+  if (chargedAmount !== undefined && !/^\d+$/.test(chargedAmount)) {
+    throw new Error("invalid chargedAmount: not a non-negative integer");
+  }
+
   const channelState = settle.extra?.channelState as BatchSettlementChannelStateExtra | undefined;
+
+  // Validate channelState shape before use. A server response with a partial
+  // channelState (e.g. missing chargedCumulativeAmount) should not crash the
+  // client — skip the state update instead.
+  const chargedCumulativeAmount = channelState?.chargedCumulativeAmount;
+  if (
+    chargedCumulativeAmount !== undefined &&
+    (typeof chargedCumulativeAmount !== "string" || !/^\d+$/.test(chargedCumulativeAmount))
+  ) {
+    return;
+  }
+
   await updateChannelFromSettle(storage, {
     server: {
       chargedAmount,
-      chargedCumulativeAmount: channelState?.chargedCumulativeAmount,
+      chargedCumulativeAmount,
     },
     local,
   });


wilnowilx:
Closing to refine — will reopen with stronger sourcing.

PR #3422, opened 2026-09-08
fix(extensions): decode-until-stable in isValidRouteTemplate

### LINKED ISSUE #3169: isValidRouteTemplate's traversal/scheme-injection checks decode routeTemplate once, so double percent-encoding bypasses both
## Summary

`isValidRouteTemplate` in `typescript/packages/extensions/src/bazaar/facilitator.ts` is meant to reject a `routeTemplate` containing path traversal (`..`) or scheme injection (`://`), per its own doc comment: "A malicious value could cause the facilitator to catalog the payment under an arbitrary URL (catalog poisoning)." It decodes the value with a single `decodeURIComponent` pass before checking for those substrings. A double percent-encoded payload survives one decode still percent-encoded, so the substring checks never see it, and the function returns `true`.

## Where

`typescript/packages/extensions/src/bazaar/facilitator.ts:53-66`:

```ts
export function isValidRouteTemplate(value: string | undefined): value is string {
  if (!value) return false;
  if (!ROUTE_TEMPLATE_REGEX.test(value)) return false;
  // Decode percent-encoding before traversal checks so that %2e%2e is caught.
  let decoded: string;
  try {
    decoded = decodeURIComponent(value);
  } catch {
    return false;
  }
  if (decoded.includes("..")) return false;
  if (decoded.includes("://")) return false;
  return true;
}
```

`ROUTE_TEMPLATE_REGEX` (line 31) is `/^\/[a-zA-Z0-9_/:.\-~%]+$/`. It explicitly allows `%`, so a percent-encoded (or double percent-encoded) value passes the regex before ever reaching the decode step.

## Reproduction

Verified directly against this function (isolated, no other code involved):

```js
const ROUTE_TEMPLATE_REGEX = /^\/[a-zA-Z0-9_/:.\-~%]+$/;

function isValidRouteTemplate(value) {
  if (!value) return false;
  if (!ROUTE_TEMPLATE_REGEX.test(value)) return false;
  let decoded;
  try { decoded = decodeURIComponent(value); } catch { return false; }
  if (decoded.includes("..")) return false;
  if (decoded.includes("://")) return false;
  return true;
}

isValidRouteTemplate("/foo/%252e%252e%252fsecret");
// => true
// decodeURIComponent once: "/foo/%2e%2e%2fsecret"   (no ".." literal, check passes)
// decodeURIComponent twice: "/foo/../secret"

isValidRouteTemplate("/foo%253a%252f%252fevil.example.com/bar");
// => true
// decodeURIComponent once: "/foo%3a%2f%2fevil.example.com/bar"   (no "://" literal, check passes)
// decodeURIComponent twice: "/foo://evil.example.com/bar"
```

Both the traversal check and the scheme-injection check are bypassed by the same single-decode gap. It isn't specific to one of the two.

## Severity

Calibrated honestly: this is not a confirmed RCE or path-traversal exploit. I verified the bypass against `isValidRouteTemplate` in isolation, not an end-to-end exploitation path through a real facilitator and a real downstream consumer of the cataloged `routeTemplate`. The concrete, verified claim is narrower and still real: **the defense does not do what its own doc comment says it does**: a double-encoded traversal or scheme-injection payload is accepted as a "valid" `routeTemplate` and would be used verbatim to build the catalog's canonical URL (`extractDiscoveryInfo`, same file, lines 579-581: `` `${url.origin}${routeTemplate}` ``). Whether that becomes exploitable downstream depends on whether any consumer of the catalog ever decodes the stored value more than once. I did not verify that either way for any specific facilitator deployment.

## Suggested fix

Match on the fully, repeatedly decoded form rather than a single pass: decode until the string stops changing (bounded, to avoid a pathological-input loop), then run both checks against that. Happy to open a PR with a regression test covering both payloads above if that's useful.

wilnowilx:
Closes #3169

`isValidRouteTemplate` applies a single `decodeURIComponent` pass before checking for `..` (traversal) and `://` (scheme injection). A double-encoded payload like `%252e%252e` decodes to `%2e%2e` on the first pass — still no literal `..` — and passes validation.

**Fix:** Decode iteratively until the string stops changing. The regex character set limits encoding depth, so the loop terminates quickly.

**Before:** `/foo/%252e%252e%252fsecret` → accepted (single decode produces `%2e%2e`)
**After:** `/foo/%252e%252e%252fsecret` → rejected (decodes to `/foo/../secret`)

Adds 3 regression tests: double-encoded traversal, double-encoded scheme injection, triple-encoded traversal.

### DIFF
--- a/typescript/packages/extensions/src/bazaar/facilitator.ts
+++ b/typescript/packages/extensions/src/bazaar/facilitator.ts
@@ -45,6 +45,9 @@ const ROUTE_TEMPLATE_REGEX = /^\/[a-zA-Z0-9_/:.\-~%]+$/;
  * - Must not contain ".." (path traversal)
  * - Must not contain "://" (URL injection)
  *
+ * Percent-encoding is decoded iteratively until stable before the traversal and
+ * injection checks run, so double-encoded payloads like %252e%252e are caught.
+ *
  * @param value - The raw routeTemplate string from the client payload
  * @returns true if the value is a valid routeTemplate, false otherwise
  *
@@ -53,13 +56,23 @@ const ROUTE_TEMPLATE_REGEX = /^\/[a-zA-Z0-9_/:.\-~%]+$/;
 export function isValidRouteTemplate(value: string | undefined): value is string {
   if (!value) return false;
   if (!ROUTE_TEMPLATE_REGEX.test(value)) return false;
-  // Decode percent-encoding before traversal checks so that %2e%2e is caught.
-  let decoded: string;
-  try {
-    decoded = decodeURIComponent(value);
-  } catch {
-    return false;
-  }
+
+  // Decode percent-encoding iteratively until stable. A single decode pass
+  // does not catch double-encoded payloads (e.g. %252e%252e decodes to %2e%2e
+  // on the first pass, which still passes the ".." check). Looping until the
+  // string stops changing ensures every layer of encoding is stripped.
+  let decoded = value;
+  let prev: string;
+  do {
+    prev = decoded;
+    try {
+      decoded = decodeURIComponent(decoded);
+    } catch {
+      // Malformed percent-encoding — reject.
+      return false;
+    }
+  } while (decoded !== prev);
+
   if (decoded.includes("..")) return false;
   if (decoded.includes("://")) return false;
   return true;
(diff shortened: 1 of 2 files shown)


phdargen:
Duplicate of https://github.com/x402-foundation/x402/pull/3213

PR #3423, opened 2026-09-08
fix(stellar): reject areFeesSponsored: false at construction time

### LINKED ISSUE #3333: @x402/stellar: areFeesSponsored: false accepted with no validation, silently breaks 100% of settlements against the official client
## Summary

`ExactStellarScheme`'s facilitator constructor accepts `areFeesSponsored: false`
with no validation, even though the package's own `getExtra()` JSDoc states
"As of now, the spec only supports `areFeesSponsored: true`." A facilitator
built this way boots without any error or warning, advertises
`areFeesSponsored: false` on `/supported`, and then every single settlement
attempt through the official client fails — with an error message that gives
no indication the facilitator, not the client, is misconfigured.

## Where

`@x402/stellar@2.24.0`.

Facilitator constructor, `dist/cjs/exact/facilitator/index.js` (source:
`src/exact/facilitator/scheme.ts`):

```js
/**
 * @param options.areFeesSponsored - Indicates if fees are sponsored (default: true)
 */
constructor(signers, {
  areFeesSponsored = true,
  ...
} = {}) {
  ...
  this.areFeesSponsored = areFeesSponsored ?? true;   // no validation of the value at all
  ...
}

/**
 * Get mechanism-specific extra data for the supported kinds endpoint.
 * For Stellar, returns `areFeesSponsored` indicating to clients if they can expect fees to be sponsored.
 * As of now, the spec only supports `areFeesSponsored: true`.
 */
getExtra(_) {
  return { areFeesSponsored: this.areFeesSponsored };
}
```

The `.d.ts` types `areFeesSponsored?: boolean` as an ordinary config knob —
the "only supports `true`" caveat lives solely on `getExtra()`'s doc
comment, not on the constructor option that actually lets an operator set
it to `false`.

Client, `dist/cjs/exact/client/index.js:182-184`:

```js
if (!extra.areFeesSponsored) {
  throw new Error(`Exact scheme requires areFeesSponsored to be true`);
}
```

A resource server's `enhancePaymentRequirements` threads `getExtra()`'s
output straight into `paymentRequirements.extra`, so a facilitator
constructed with `areFeesSponsored: false` reaches the client exactly as
`extra.areFeesSponsored: false`, and every payment attempt against it via
the official client throws the message above — a client-side error that
doesn't hint the actual cause is a facilitator-side misconfiguration.

## Repro

Ran directly against the installed `@x402/stellar@2.24.0` (both the
facilitator and client classes from the real package, not mocked):

```js
const { ExactStellarScheme } = require("@x402/stellar/exact/facilitator");
const { ExactStellarScheme: ClientScheme } = require("@x402/stellar");

const scheme = new ExactStellarScheme([mainSigner], { areFeesSponsored: false });
// no error here
console.log(scheme.getExtra("stellar:testnet")); // { areFeesSponsored: false }

const paymentRequirements = { ...requirementsFields, extra: scheme.getExtra("stellar:testnet") };
const client = new ClientScheme(clientSigner);
await client.createPaymentPayload(1, paymentRequirements);
```

Output:

```
facilitator constructed OK, no error/warning.
getExtra(): { areFeesSponsored: false }
createPaymentPayload() rejected as expected: Exact scheme requires areFeesSponsored to be true
```

## Impact

Low severity — the package README does document in prose that the
non-sponsored flow "will be added later," so a careful reader of the full
docs isn't completely blindsided. But the typed, public constructor option
carries no runtime guard and no cross-reference to that limitation, so
anyone who trusts the type signature (`areFeesSponsored?: boolean`, no
`@deprecated`/`@throws`/restriction noted on the constructor itself) over
scattered prose gets a facilitator that boots cleanly and then silently
breaks 100% of settlements, with an error surfaced on the wrong side of the
interaction.

## Suggested fix

Either throw at construction time when `areFeesSponsored: false` is passed
(until the non-sponsored flow actually ships), or add the same "spec only
supports `true`" cross-reference directly on the constructor option's own
JSDoc, not just on `getExtra()`'s.

wilnowilx:
Closes #3333

`ExactStellarScheme`'s facilitator constructor accepts `areFeesSponsored: false` without validation, even though the spec only supports `true` and the client throws on `false`.

A facilitator built this way boots without error, advertises `areFeesSponsored: false` on `/supported`, and then every settlement attempt fails with a client-side error that gives no indication the facilitator is misconfigured.

**Fix:** Throw a clear error at construction time when `areFeesSponsored: false` is passed, explaining that the spec only supports `true`.

**Before:** `new ExactStellarScheme([signer], { areFeesSponsored: false })` → boots OK, all settlements fail silently
**After:** `new ExactStellarScheme([signer], { areFeesSponsored: false })` → throws with actionable message

### DIFF
--- a/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts
+++ b/typescript/packages/mechanisms/stellar/src/exact/facilitator/scheme.ts
@@ -146,6 +146,19 @@ export class ExactStellarScheme implements SchemeNetworkFacilitator {
 
     // Apply configuration options (with defaults)
     this.rpcConfig = rpcConfig;
+
+    // The spec only supports areFeesSponsored: true. Setting it to false
+    // silently breaks every settlement: the client throws
+    // "Exact scheme requires areFeesSponsored to be true" with no hint
+    // that the facilitator, not the client, is misconfigured. Fail fast
+    // at construction time instead.
+    if (areFeesSponsored === false) {
+      throw new Error(
+        "ExactStellarScheme: areFeesSponsored: false is not supported by the x402 spec. " +
+          "The client will reject payments with areFeesSponsored: false. " +
+          "Remove the option or set it to true.",
+      );
+    }
     this.areFeesSponsored = areFeesSponsored ?? true;
     this.maxTransactionFeeStroops = maxTransactionFeeStroops ?? DEFAULT_MAX_TRANSACTION_FEE_STROOPS;
     this.selectSigner = selectSigner ?? roundRobinSelectSigner();
--- a/typescript/packages/mechanisms/stellar/test/unit/facilitator-accessors.test.ts
+++ b/typescript/packages/mechanisms/stellar/test/unit/facilitator-accessors.test.ts
@@ -54,16 +54,15 @@ describe("ExactStellarScheme - getExtra", () => {
     expect(mockRpcClient.getLatestLedger).not.toHaveBeenCalled();
   });
 
-  it("should use custom areFeesSponsored", () => {
+  it("should reject areFeesSponsored: false with a clear error", () => {
     const signer = createEd25519Signer(
       "SDV3OZOPGIO6GQAVI7T6ZJ7NSNFB26JX6QZYCI64TBC7BAZY6FQVAXXK",
       STELLAR_TESTNET_CAIP2,
     );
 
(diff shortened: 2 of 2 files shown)


wilnowilx:
Closing as part of consolidating my open PRs to respect reviewer bandwidth. Genuine bug (constructor accepts a value the protocol rejects) - happy to re-propose surgically later if still relevant.

PR #3428, opened 2026-09-08
docs(specs): Stellar upto convergence and interoperability requirements

davedumto:
## What this PR contains

Two files in `specs/schemes/upto/`:

### 1. `scheme_upto_stellar_interop.md`

A convergence document derived from cross-implementation analysis of six deployed Stellar `upto` implementations. It does not introduce a new contract design. Requirements are derived from agreement across existing implementations. Open questions in section 7 are where those implementations diverge.

### 2. `scheme_upto_stellar_vellar.md`

A normative specification derived from the deployed `upto-vellar` contract (`CCZL7CTRS...`, MIT, Vellar-Wallet/vellar-facilitator). Picks winners on all five open questions from the convergence document:

- No hook argument (7-arg ABI)
- `approve` + `transfer_from`, required for `actual != ceiling` at the auth level
- Nonce TTL = `expiration_ledger`, no buffer
- Settlement event with both ceiling and actual reported
- Indexers MUST read `actual_amount` from the token transfer event, not the envelope args

## Relationship to other open PRs

- #3134: stateless UptoSettlement, different transfer mechanism
- #3098: stateful, Periplo

The convergence document is additive to both. The normative spec is a third proposal for TSC consideration. If the TSC prefers this approach, it can be renamed to `scheme_upto_stellar.md` on merge.

We asked the maintainers on 2026-09-08 where they would like this to land (comment on #3134). No reply yet, so both files are filed at distinct paths to avoid an add/add collision with #3134's `scheme_upto_stellar.md` while we wait. If the TSC prefers to fold this into either existing PR, we are happy to do that instead.

## Evidence

The normative spec is derived from a deployed and settled contract:

- Contract: `CCZL7CTRS6GWEYXDYD54DZM3OUHQW2S2A4KSU75SH275P3SFZLL4YQAN`
- Wasm: `92365d9e...`
- First settlement: `be33bb71...` (ledger 4587956, 0.01 of 0.05 USDC)
- Source: MIT, reproducible build

### DIFF
--- a/specs/schemes/upto/scheme_upto_stellar_interop.md
+++ b/specs/schemes/upto/scheme_upto_stellar_interop.md
@@ -1,0 +1,349 @@
+# Stellar `upto` Scheme: Convergence and Interoperability Requirements
+
+> **What this document is.** A convergence document derived from
+> cross-implementation analysis of deployed Stellar `upto` implementations. It
+> does **not** introduce a new contract design. The requirements below are
+> derived from agreement across existing implementations; the open questions in
+> §7 are where those implementations currently diverge.
+>
+> **Relationship to existing proposals.** Two Stellar `upto` specs are open at
+> the time of writing — [#3134](https://github.com/x402-foundation/x402/pull/3134)
+> (stateless `UptoSettlement`) and [#3098](https://github.com/x402-foundation/x402/pull/3098)
+> (stateful, Periplo). This document is **additive to both**, not a competing
+> third design. Where they agree, that agreement is written here as a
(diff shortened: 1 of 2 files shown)


Eras256:
Confirming #3098 and section 7.5's attribution are accurate as described here — good summary of where the two specs actually converge. Appreciate the honest framing on section 7 (naming the divergences instead of picking a winner) and the confidence caveat on the hook-failure-isolation risk — that's the right way to carry an unreproduced claim.

PR #3431, opened 2026-09-09
fix(svm): split upto delegated-auth store errors from unauthenticated

phdargen:
## Description

Delegated claim settle collapsed a `DelegatedAuthStore.Get` failure into `invalid_upto_svm_delegated_settle_unauthenticated`, so hosts could not tell a store outage from a missing or mismatched identity. Claim settle now returns `invalid_upto_svm_delegated_auth_store` when the store read fails, does not broadcast, and keeps unauthenticated for identity-only cases. 

## Checklist


ethanoroshiba (approved):
🚀

PR #3325, opened 2026-09-01
spec: settlement status vocabulary for SettleResponse (§5.3.3–§5.3.4)

zjzJoez:
# spec: settlement status vocabulary for SettleResponse (§5.3.3–§5.3.4)

## What this adds

Three OPTIONAL fields on `SettleResponse` and the vocabulary behind them: six settlement
states, each REQUIRED to name the on-chain object a stranger can check — or to name that
it has none. Two new subsections in the core spec, one declarations section in
`scheme_exact_evm.md`, one paragraph and one error code (`authorization_canceled`) in §9,
and one reader paragraph after the §5.3.3 table. **+86 lines, no deletions, two files.** A
response that omits `status` behaves exactly as v2 does today; nothing here touches
`VerifyResponse`, payment flows, or any existing field's meaning.

## Why

`SettleResponse` is binary, and four independent efforts have hit that limit from
different directions (#3030→#3083's `settlement_pending`, #3085's missing timestamp,
#3182's deferral, batch-settlement's commitment-vs-moved split). #3208 proposed one
shared vocabulary instead of N per-scheme patches, and over 22 comments converged with
five outside contributors — among them a cross-rail settlement indexer, the operator of a
Nano payment rail, and a settlement-attestation service — on the three rules this diff
encodes:

1. **Anchor per state.** Every state names its evidence anchor, or explicitly names that
   it has none. Driven by production data posted in-thread: a marketplace census found
   229.50 USDC sitting in an evidence-free `failed` bucket against 74.22 USDC of
   `confirmed` that named a hash 36 times out of 36.
2. **Capability declarations name their enforcement object** — contract, field,
   comparator, unit — per binding, versioned append-only, with a chain-verifiable
   `effectiveFrom` height. A bare boolean is not verifiable; the two live EVM enforcement
   objects even disagree by one second at the boundary (EIP-3009 strict, permit2
   inclusive), which only a declared comparator surfaces.
3. **Malformed, not downgraded.** A status the declaration does not support is rejected,
   never silently weakened. Three reachability values (`reachable` / `unreachable` /
   `unclaimable`) so a rail that cannot meet a state's semantic requirement can say so
   without implying the condition cannot arise.

## Evidence

Every chain claim in the diff is keyless and re-runnable. The scripts are in the reference
repo, <https://github.com/zjzJoez/grace-x402> (`3597a76`), and each names which bullet it backs:

- The permit2 window (`node grace/permit2-window.mjs`): four `eth_call`s — two per entry
[...]

PR #3437, opened 2026-09-09
spec: §5.3.5 receiver obligation on an unresolved settlement

aurumflux20:
Opened at @zjzJoez's suggestion on zjzJoez/x402#1, as its own PR against `main` rather than as a rider on #3325. #3325 was filed and reviewed on one line — vocabulary only, no flow semantics — and I tried to widen that line twice. This is the obligation on its own number, where it can be argued or rejected on its own terms.

## What it says

§5.3 is entirely about what an emitter may claim. Nothing yet says what a reader owes in return. This is that: while an authorization is **unresolved**, a receiver must not challenge again for the same request, must not present the outcome to the payer as a failed payment, and must re-present the authorization rather than ask the payer to sign a new one.

Unresolved means one of three things: `errorReason: settlement_pending`, no `SettleResponse` at all, or one that cannot be interpreted.

## Why it is worth a normative rule

The failure it addresses is not a client bug. A receiver gets something it cannot resolve — a dropped connection, a `success: false` with an empty `transaction`, a 5xx after broadcast — and answers with a fresh `402` for the same intent. The payer then produces a **new** authorization with a new nonce, which every replay guard correctly admits, because at the protocol boundary it is a different payment. Nothing in the bytes distinguishes "same purchase, authorized again after a lost response" from "a second purchase".

That is why the duty has to sit on the receiver. A facilitator cannot infer the payer's intent, and client-side idempotency is bounded by process lifetime — the lost reply is very often caused by the same event that ended the process. The receiver is the only party holding the prior authorization, so it is the only one that can decline to ask for a new one.

## Scope and dependencies

- **Written in `main`'s vocabulary.** The only state it names is `errorReason: settlement_pending`, already normative in §9. Nothing here depends on #3325's `status` field, `statusAnchor`, or the six-state table.
- **Numbering.** §5.3.5 leaves room for #3325's §5.3.3–§5.3.4. If #3325 does not land, renumber this to §5.3.3; no wording changes.
- **+54 lines, one file, no schema change**, no new field, no change to any existing rule.

## Two defects fixed rather than carried over

@zjzJoez reviewed the earlier draft and found both. They are corrected here, and I would rather name them than let a reader assume this text is the same one:

1. **A clause I had publicly retracted was still in the draft** — permitting a receiver to satisfy the obligation by resolving the authorization against an anchor, as an *alternative* to honouring it. I retracted that on 2026-09-04 as too permissive for a normative clause, and then left it standing. It is gone.
2. **The unresolved state had no exit.** As drafted, a payer whose authorization could not presently complete became permanently unaskable: the merchant could neither collect, nor challenge again, nor offer a different asset. The rule now has an explicit exit — reconcile and establish the authorization can no longer settle, **or** withdraw or invalidate it by an action the receiver performs. Never an inference from elapsed time or from silence. That keeps "must not conclude it will clear" intact while leaving the merchant a legal move.

## Prior discussion

The one-line reader rule from this thread was taken into #3325's branch by @zjzJoez at `7e2c7bc`, credited to me, @glennquinting and @jywy78kgrf-create. This PR is the fuller obligation that rule is the summary of, and it is deliberately not proposing anything about conformance testing — a conformance regime argued by someone who ships a conformance battery is self-dealing however carefully it is worded, so I am not re-filing that section here.

Field context is on #3208 and #3325; I have kept empirical claims out of the normative text entirely.

cc @zjzJoez @glennquinting @0xultravioleta

### DIFF
--- a/specs/x402-specification-v2.md
+++ b/specs/x402-specification-v2.md
@@ -251,6 +251,60 @@ The `SettleResponse` schema contains the following fields:
 | `amount`      | `string`  | Optional | The actual amount settled in atomic units (omitted if not applicable) |
 | `extensions`  | `object`  | Optional | Protocol extensions data                                              |
 
+**5.3.5 Receiver Obligation on an Unresolved Settlement**
+
+A **receiver** — a resource server, or any party that dispatched a settle and is
+deciding whether to issue a new payment challenge — holds, for an authorization
+`A`, one of three things: a terminal answer, a non-terminal answer, or nothing.
+The rules in this section govern the last two.
+
+`A` is **unresolved** for a receiver when any of the following holds:
+
+- the `SettleResponse` for `A` carries `errorReason: settlement_pending` (see
+  [§9 Error Handling](#9-error-handling)), which is non-terminal by definition;
+- no `SettleResponse` was received at all — a timeout, a dropped connection, or
+  any other transport failure; or
+- a `SettleResponse` was received but cannot be interpreted under this
+  specification.
+
+While `A` is unresolved, a receiver:
+
+- **MUST NOT** issue a new payment challenge for the request `A` was presented
+  against;
+- **MUST NOT** present the outcome to the payer as a failed payment;
+- **MUST**, on a subsequent request for the same intent, re-present `A` itself
+  to the facilitator rather than requiring the payer to authorize again.
+
+**Failure to resolve, receiver side.** A receiver that cannot establish the
+outcome of `A` MUST treat it as unresolved. Neither the absence of a
+`SettleResponse`, nor a transport failure, nor `success: false` accompanied by
+an empty `transaction`, is evidence that `A` did not settle.
+
+**Leaving the unresolved state.** A receiver MAY issue a new payment challenge
+for that request only once one of the following holds:
+
+1. it has reconciled `A` against its `transaction` on `network` and established,
+   under that network's finality rule, that `A` did not settle and can no longer
+   settle; or
+2. it has **withdrawn or invalidated `A` by an action it performed** — for
+   example cancelling the authorization on-chain, or a facilitator-side
+   cancellation whose confirmation it holds — so that `A` can no longer settle.
+
+A receiver MUST NOT infer either condition from elapsed time alone, from the
+absence of a response, or from the repetition of a non-terminal answer. Clause 2
+exists so that a payer whose authorization cannot presently complete is not left
+permanently unaskable: the receiver's remedy is to end `A` explicitly and
+challenge afresh, never to assume `A` has died.
+
+**This obligation does not depend on client behaviour.** It is not conditioned
+on the client re-presenting the same authorization. A conforming client will
+re-present; a non-conforming one will authorize again; the receiver's duty is
+identical in both cases, because the receiver is the party that holds the prior
+authorization and can therefore act on it. Stating the rule conditionally would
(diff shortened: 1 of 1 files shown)


cv-scvd:
Good call keeping this abstract — folding "timeout, dropped connection, or any other transport failure" into one bucket instead of writing HTTP status codes into `x402-specification-v2.md` avoids the exact mistake §5's own reader rule just got rewritten to fix (bind to the spec's defined signal, not to transport bytes). The three-way partition (`errorReason: settlement_pending` / no response at all / a response that can't be interpreted) already covers every wire-level failure mode I can think of without naming one, and doing that in a transport-independent section is the right call.

The gap this leaves isn't in this text — it's that nothing populates `specs/transports-v2/` yet to tell an HTTP implementer which of *these three buckets* a given wire event actually falls into. A 5xx after broadcast: "no response" or "uninterpretable"? Malformed JSON on a 200? A connection dropped mid-body vs. a clean timeout? All answerable under this rule (every one of them is unresolved either way, so the MUST NOTs don't change), but the implementer writing the client-side retry loop still has to know which bucket to log, and right now that's tribal knowledge, not spec text.

Not asking you to widen this PR's scope — the +54/one file/no schema change shape is exactly right to land this on its own terms. This is a note for whoever picks up the HTTP transport doc against `transport_template.md`; happy to draft that mapping ourselves if nobody's already got a claim on it.

---

Disclosure: written with AI assistance; every claim about our implementation was re-read from the source at commit d00921d98b229cc15562e863dbe54e00090a3ac6 before posting.

aurumflux20:
Thanks — and yes, keeping it abstract was deliberate. The partition is about what the *client can conclude*, not what the wire did, which is why it belongs in the transport-independent section and the wire mapping belongs in `specs/transports-v2/`.

On your three cases, here is the rule I'd propose for the HTTP doc, so the mapping is derivable rather than memorised:

- **A response arrived and it is not a well-formed `SettleResponse`** → *uninterpretable*. That covers a 5xx after broadcast and malformed JSON on a 200. The status code is irrelevant to the bucket; what matters is that the receiver cannot read a settlement outcome out of it.
- **No complete response arrived** → *no response*. That covers both a clean timeout and a connection dropped mid-body. A partial body is not a response; treating it as "uninterpretable" would tempt an implementer to parse the fragment.
- **A well-formed response that says `settlement_pending`** → the third bucket, unchanged.

You're right that all of these are unresolved either way, so the MUST NOTs don't move. The bucket only changes what gets logged — which is exactly why it should be spec text and not tribal knowledge: two clients logging the same event under different buckets makes incident data across facilitators incomparable.

One more row worth having in that mapping, from a live probe rather than reasoning: on 3 Sep we presented one EIP-3009 authorization to a real facilitator's `/settle` twice. First call settled (tx `0x28d59010…`, block 46362866). The re-presentation was answered with `invalid_exact_evm_nonce_already_used` — a well-formed, interpretable response — **but it carries no transaction hash**. So the client learns "do not re-sign" and still cannot reconcile which settlement was its own. That is a fourth state: *interpretable, terminal, unreconcilable*. It is not unresolved in your partition's sense, and it should not be logged as such, but an HTTP implementer will hit it and needs to know it exists.

Nobody has a claim on the transport doc. Please draft it. I'll read it against the retry-battery cases we have (the double-presentation above plus the timeout-on-re-presentation case) and comment with anything that doesn't map cleanly.

---

Disclosure: written with AI assistance; the probe facts above were re-read from our own run log before posting.

PR #3451, opened 2026-09-11
fix(next): preserve request body when reading through the adapter

### LINKED ISSUE #3445: [Bug]: Next.js payment hook body reads consume the downstream route handler request
## What happened?

With `@x402/next`, reading `await context.adapter.getBody()` in an `onProtectedRequest` hook consumes the POST request body. If the hook then grants access, the downstream route handler's `request.json()` throws:

```text
TypeError: Body is unusable: Body has already been read
```

Expected: inspecting the parsed body through the adapter should leave the JSON body available to the route handler. This matters for body-based access checks before processing the same POST payload. A second call to the adapter's `getBody()` also returns `undefined` instead of the parsed object.

## Steps to reproduce

In a fresh directory, install `npm install @x402/next@2.25.0 next@16.2.6 tsx`, save the following as `repro.mjs`, and run `npx tsx repro.mjs`:

```js
import assert from 'node:assert/strict';
import { NextRequest, NextResponse } from 'next/server.js';
import {
  x402ResourceServer, x402HTTPResourceServer, withX402FromHTTPServer,
} from '@x402/next';

async function run(readBody) {
  const httpServer = new x402HTTPResourceServer(new x402ResourceServer(), {
    'POST /api/quote': { accepts: [] },
  });
  httpServer.onProtectedRequest(async ({ adapter }) => {
    if (readBody) {
      assert.deepEqual(await adapter.getBody(), { quantity: 2 });
    }
    return { grantAccess: true };
  });
  const handler = withX402FromHTTPServer(
    async request => NextResponse.json(await request.json()),
    httpServer, undefined, undefined, false,
  );
  const request = new NextRequest('http://localhost/api/quote', {
    method: 'POST', headers: { 'content-type': 'application/json' },
    body: JSON.stringify({ quantity: 2 }),
  });
  try {
    const response = await handler(request);
    console.log({ readBody, status: response.status, body: await response.json() });
    return 'ok';
  } catch (error) {
    console.log({ readBody, error: String(error) });
    return 'error';
  }
}

assert.equal(await run(false), 'ok');
assert.equal(await run(true), 'error');
```

Output:

```text
{ readBody: false, status: 200, body: { quantity: 2 } }
{ readBody: true, error: 'TypeError: Body is unusable: Body has already been read' }
```

The control differs only in whether the hook reads the body. The hook explicitly grants access, with facilitator initialization disabled and no payment requirements, to isolate request-body handling without signing or settlement.

## Environment

- macOS arm64; Node.js 22.22.3
- Published `@x402/next@2.25.0`, `next@16.2.6`, `tsx@4.23.13`
- Also reproduced against main source at `3c2ddfb922893c91ef8f281b64f8045d1f5e0d75` using Next.js 16.3.2.
- Local reproduction only; no wallet, facilitator, or real payments used. Investigation and reproduction were AI-assisted.

viviviviviid:
## Description

Closes #3445.

Read JSON from a clone in `NextAdapter.getBody()` so payment hooks can inspect the body without consuming the request passed to the route handler. This also allows repeated adapter reads before the handler runs. Parsing failures still return `undefined`.

The tests cover downstream readability, repeated reads, and preservation of malformed JSON. Includes a patch changeset for `@x402/next`; no public API or spec changes.

## Tests

- Confirmed all three regression checks fail before the fix and pass after it.
- Reproduced the hook-to-handler failure with the real HTTP server and wrapper; both the control and hook-body-read cases succeed after the fix.
- All 93 `@x402/next` tests pass with coverage thresholds met. Package lint, changed-file formatting, and ESM/CJS builds with type declarations pass.
- Production source and the adapter tests pass type-checking. Full-package `tsc --noEmit` reports four existing errors in `index.test.ts` and `utils.test.ts`, identical on main and with this fix.

Validated on Windows with Node 24.14.1 and Next.js 16.2.6 using the frozen workspace lockfile. The full monorepo suite was not run.

## Checklist

- [x] All new and existing tests pass (`@x402/next`)

AI disclosure: I used Codex to help with the implementation, tests, and validation.

### DIFF
--- a/typescript/.changeset/next-preserve-request-body.md
+++ b/typescript/.changeset/next-preserve-request-body.md
@@ -1,0 +1,5 @@
+---
+"@x402/next": patch
+---
+
+Fixed Next.js adapter body reads consuming the request body before the route handler could read it.
--- a/typescript/packages/http/next/src/adapter.test.ts
+++ b/typescript/packages/http/next/src/adapter.test.ts
@@ -134,7 +134,7 @@ describe("NextAdapter", () => {
   });
 
   describe("getBody", () => {
-    it("returns parsed JSON body", async () => {
+    it("returns parsed JSON body without consuming the original request", async () => {
       const body = { data: "test" };
       const req = new NextRequest("https://example.com/api", {
         method: "POST",
@@ -143,6 +143,31 @@ describe("NextAdapter", () => {
       });
       const adapter = new NextAdapter(req);
       expect(await adapter.getBody()).toEqual(body);
+      expect(await req.json()).toEqual(body);
+    });
+
+    it("returns parsed JSON body on repeated calls", async () => {
+      const body = { data: "test" };
+      const req = new NextRequest("https://example.com/api", {
+        method: "POST",
+        body: JSON.stringify(body),
+        headers: { "Content-Type": "application/json" },
+      });
+      const adapter = new NextAdapter(req);
+      expect(await adapter.getBody()).toEqual(body);
+      expect(await adapter.getBody()).toEqual(body);
+    });
+
+    it("returns undefined for invalid JSON without consuming the original request", async () => {
+      const body = "invalid JSON";
+      const req = new NextRequest("https://example.com/api", {
+        method: "POST",
+        body,
+        headers: { "Content-Type": "application/json" },
+      });
+      const adapter = new NextAdapter(req);
+      expect(await adapter.getBody()).toBeUndefined();
+      expect(await req.text()).toBe(body);
     });
 
     it("returns undefined when body parsing fails", async () => {
--- a/typescript/packages/http/next/src/adapter.ts
+++ b/typescript/packages/http/next/src/adapter.ts
@@ -109,7 +109,7 @@ export class NextAdapter implements HTTPAdapter {
    */
   async getBody(): Promise<unknown> {
     try {
-      return await this.req.json();
+      return await this.req.clone().json();
     } catch {
       return undefined;
     }


halobartku:
Ran the PR's exact one-line change (`.clone().json()`) against a real Next.js server runtime, since the programmatic path was already covered and @sunruize93-cmyk flagged above that this PR had not been independently rerun.

**Setup.** `next@16.2.6` (app router, `next dev`), `@x402/next@2.25.0` with the published `dist/esm/index.js` patched in place with this diff, route handler constructs `NextAdapter` directly, reads `getBody()` twice, then `request.json()`.

**Pre-patch (published 2.25.0), via real HTTP POST:**
```
adapter1={"quantity":2} adapter2=undefined handler=ERROR:Body is unusable: Body has already been read
```
Both #3445 symptoms reproduce server-side, not just in the repro script.

**Post-patch, same server, same route:**
```
adapter1={"quantity":2} adapter2={"quantity":2} handler={"quantity":2}
```

Also exercised the two paths the unit tests can't:
- `Transfer-Encoding: chunked` (streamed body): both adapter reads + downstream handler parse all correct — `clone()` buffers without consuming the incoming stream.
- 2 MB JSON body: no truncation, no second-read failure.

One boundary confirmed from the Fetch spec side that integrators should know: `clone()` is repeatable only while the original body stays unread — which is exactly the invariant this diff restores, so the invariant and the fix line up. Repeated `getBody()` calls each clone first, so no cache is needed.

Diff itself is byte-identical to what I verified in #3445 (issuecomment-5641858888); this adds the real-runtime + streamed-body evidence that comment was missing. Regression tests in `adapter.test.ts` look sufficient to lock all three behaviors.

(Verification run and comment by the Claude agent operating halobartku; investigation AI-assisted.)

viviviviviid:
Thanks @halobartku for verifying this in a real Next.js runtime, including chunked requests and larger payloads, and @sunruize93-cmyk for the clear reproduction and follow-up. I appreciate the additional validation!

phdargen (approved):
Thanks for the contribution @viviviviviid , LGTM 🚀

PR #3454, opened 2026-09-12
fix(axios): preserve base paths and params in payment hook URLs

sunruize93-cmyk:
## Description

`onPaymentRequired` receives an incorrect `requestUrl` when the Axios transport does not expose `responseURL` or `res.responseUrl`. With the real Axios fetch adapter, `baseURL: "http://localhost/v1"`, `url: "/quote"`, and `params: { city: "Seoul" }`, the server receives `/v1/quote?city=Seoul`, but the hook receives `/quote` without the base path or query.

The fallback uses WHATWG URL resolution instead of Axios URL joining and ignores `params`/`paramsSerializer`. Use the calling instance's `getUri(config)` for that fallback, preserving the existing preference for transport-provided final URLs and the invalid-base fallback.

This affects resource-scoped payment-required hooks. No signing or settlement changes, and no payment-loss claim.

## Tests

Reproduced with published `@x402/axios@2.25.0` and upstream `04c750e32e2c4dce658289a901710fd721e5d1a9`, using Node 22.22.3 and Axios 1.20.0.

Added a loopback HTTP server regression test using the real Axios instance and real x402 HTTP client. The hook supplies a test access header, so no wallet, facilitator or transaction is used. Coverage includes base-path slash variants, absolute URLs with query parameters, instance-default params/custom serialization, and a passing default HTTP-adapter control.

- Final tests on original source: 5 fail / 35 pass.
- Patched Axios package: 40/40 pass; coverage thresholds pass (100% lines/functions/statements, 98.55% branches).
- Axios package `tsc --noEmit`, changed-file ESLint and Prettier: pass.
- Tests were run with isolated npm dependencies rather than a full workspace install; other SDKs and live-payment E2E were not run.

From `typescript/packages/http/axios`, the checks are `vitest run --coverage` and `tsc --noEmit` using the installed binaries. The new `src/requestUrl.test.ts` is the self-contained reproduction.

## Checklist

- [x] All new and existing tests pass (affected Axios package)
- [x] My commits are signed (GitHub verified)

AI assistance was used for investigation, implementation, and tests. I have reviewed the Axios changes and am marking this PR ready for maintainer review.

### DIFF
--- a/typescript/.changeset/axios-hook-request-url.md
+++ b/typescript/.changeset/axios-hook-request-url.md
@@ -1,0 +1,5 @@
+---
+"@x402/axios": patch
+---
+
+Fixed payment-required hook URLs to use Axios base-path joining and query serialization when the transport does not expose a final response URL.
--- a/typescript/packages/http/axios/src/index.test.ts
+++ b/typescript/packages/http/axios/src/index.test.ts
@@ -1,4 +1,4 @@
-import {
+import axios, {
   AxiosError,
   AxiosHeaders,
   AxiosInstance,
@@ -115,6 +115,7 @@ describe("wrapAxiosWithPayment()", () => {
         },
       },
       request: vi.fn(),
+      getUri: axios.getUri,
     } as unknown as AxiosInstance;
 
     // Create mock client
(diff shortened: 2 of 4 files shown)


phdargen (approved):
Good catch @sunruize93-cmyk, thanks for the fix

PR #3459, opened 2026-09-12
fix(evm): warn when gas sponsoring is advertised but the client cannot read the chain

### LINKED ISSUE #3458: evm client: sponsored EIP-2612 permit is silently skipped when the scheme cannot read the chain — documented ExactEvmScheme(account) form gets 412 on Permit2 assets
## Summary

When a Permit2 payment requirement is served with the `eip2612GasSponsoring` extension on the 402 body, `ExactEvmScheme` signs the sponsored EIP-2612 permit only if the signer can read the chain. With the registration shape the docs show — `new ExactEvmScheme(account)` where `account` is a viem `privateKeyToAccount` — there is no `readContract` and no `rpcUrl`, so `trySignEip2612PermitExtension` returns early, the payload carries no permit, and the facilitator answers `412 permit2_allowance_required`. Nothing tells the integrator why.

## Reproduction (mainnet, `@x402/fetch` / `@x402/evm` / `@x402/core` 2.23.0)

Seller: `@x402/hono` `paymentMiddleware` with a route priced in a Permit2 token (`extra.assetTransferMethod: "permit2"`) and `extensions: { eip2612GasSponsoring: { info: {…} } }` on the route; facilitator advertises the extension in `/supported`. Buyer: `x402Client` + `setSpendControls({ allowedAssets: [...] })` + `wrapFetchWithPayment`. Fresh wallet, no Permit2 allowance, no native gas.

| buyer scheme registration | result |
|---|---|
| `new ExactEvmScheme(account)` | `412 permit2_allowance_required` — no permit in the payload |
| `new ExactEvmScheme(account, { rpcUrl })` | 200, settled: Celo tx `0x6c5628b0ca35611f3c0380dd65592d87c065c863c3593fed62bbccb3723bdbbd` |
| `new ExactEvmScheme(toClientEvmSigner(account, publicClient))` | 200, settled: Celo tx `0x3adcfaade24627b9f1b596728affb8d3af3d463e1611463cc96b846f58286e3f` |

Same wallet, same seller, same facilitator; the only variable is whether the scheme can read `allowance` / `nonces`.

## Where

`trySignEip2612PermitExtension` (`mechanisms/evm`, client): `const capabilities = resolveExtensionRpcCapabilities(network, signer, options); if (!capabilities.readContract) return undefined;` — silent. `resolveRpcUrl` only consults `options.rpcUrl` / per-chain options; there is no fallback for networks the package knows a public RPC for.

## Ask

One or more of:

1. When the extension is advertised and the signer cannot read, fail loudly (throw, or log at warn) naming the missing capability and the two fixes — the way spend-control rejections already explain themselves.
2. Fall back to a public RPC for networks in the default tables when neither the signer nor `options` provide one, so the documented `new ExactEvmScheme(account)` form sponsors correctly.
3. At minimum, document on `ExactEvmScheme` that Permit2 + sponsoring requires `readContract` or `rpcUrl`.

Happy to open a PR for (1) or (3) if that is the preferred shape.

GigaHierz:
## Summary

Closes #3458. When a 402 body advertises `eip2612GasSponsoring` (or `erc20ApprovalGasSponsoring`) and the client cannot read the chain — a bare `privateKeyToAccount` signer with no `rpcUrl` in the scheme options — the extension helpers returned `undefined` silently, the payload went out without the sponsored permit/approval, and the facilitator answered `412 permit2_allowance_required`. The integrator had nothing to go on.

## Changes

- `mechanisms/evm/src/shared/extensions/gasSponsoring.ts`: the "is the extension advertised" check now runs before the capability check in both helpers (nothing to say when sponsoring is not in play). When it is advertised and the chain cannot be read, a `console.warn` names the missing capability and the two fixes (`{ rpcUrl }` on `ExactEvmScheme`, or a signer with `readContract` such as `toClientEvmSigner(account, publicClient)`), once per network per process. Warned rather than thrown so existing integrations keep their behaviour.
- `mechanisms/evm/src/shared/rpc.ts`: `EvmSchemeConfig.rpcUrl` documents that sponsored Permit2 payments need it or a reading signer.
- Test (`test/unit/exact/client.test.ts`): bare signer + advertised extension → no extensions on the payload, exactly one warning naming `rpcUrl`; a second payment does not repeat it; control: not advertised → no warning.
- Changeset: `@x402/evm` patch.

## Verification

`pnpm --filter @x402/evm exec vitest run test/unit/exact/client.test.ts test/unit/upto` → 6 files, 187 tests passed. `eslint` clean on the three files. `tsup` build succeeds.

The reproduction in #3458 (same wallet, same seller, same facilitator; `new ExactEvmScheme(account)` → 412, with `{ rpcUrl }` → settled on Celo mainnet) is what the test encodes.

Not done here: option 2 from the issue (fall back to a public RPC for networks in the default tables). That is a product decision about which endpoints the package should call on the user's behalf; happy to follow up if wanted.

Prepared with AI assistance; the tests and lint were run before opening.


GigaHierz:
Pushed `e4807ce`. Reviewer pass on `283db8c` found one gap: the warning was tested on the EIP-2612 path only, while the same change was made to the ERC-20 approval path. Added a test (bare signer with `signTransaction`/`getTransactionCount`/`estimateFeesPerGas` but no `readContract`, `erc20ApprovalGasSponsoring` advertised → no extensions, exactly one warning). Mutations against `283db8c`: warn call removed → 1 test red; extension check moved back after the capability check (warn would fire when sponsoring is not advertised) → 1 red. `client.test.ts` 50/50, eslint clean.

GigaHierz:
Pushed `76dd0de`, following the scope note on #3458: the warning was already in the shared helpers both other surfaces import, and now a test through `UptoEvmScheme` and one through `BatchSettlementEvmScheme` pin that (bare signer + advertised `eip2612GasSponsoring` → no extension, one warning). Mutation with both warn calls removed → the exact, upto and batch-settlement client suites all go red. 171/171 across the three client test files, eslint clean.

PR #3460, opened 2026-09-13
docs: add sendcheck-verify to third-party SDKs

m0nk111-qwen-agent:
Adds sendcheck-verify to the third-party SDK list.

sendcheck-verify is a zero-dependency Node.js (ESM + CJS) client for the x402 protocol: it verifies the merchant's signed service card (ES256 JWS over payTo and per-route maxPrice) and the 402 payment challenge cryptographically before paying, and pins the key so settlement cannot be silently re-routed (drift-proof). It ships as @pennyforgeorg/sendcheck-verify on the npm GitHub registry.

AI-assisted contribution per CONTRIBUTING.md: drafted by an agent, human-reviewed; the row links only to the public repo.


m0nk111-qwen-agent:
Re-signed the commit per the verified-commits gate (GPG ed25519, key registered on the fork account; `check-verified-commits` now green). Diff is unchanged: one table row, no other files touched.

PR #3499, opened 2026-09-16
docs(mcp): check paymentResponse before logging a settled payment

loveaihq:
## Description

`paymentMade` is `true` for every paid call, including when verify or settlement fails. In those cases the server attaches no `x402/payment-response` (`paymentWrapper.ts`: the `!verifyResult.isValid` branch and `createSettlementFailedResult`), so the client example logged `Payment settled: undefined`. It now also checks `paymentResponse`, as `examples/typescript/clients/mcp/chatbot.ts` does.

## Tests

Docs-only; Prettier passes on the changed file. Seen in practice on Cardano preprod: a payment the facilitator rejected at verify came back `isError: true, paymentMade: true`, with no `paymentResponse`.

Drafted with AI assistance (Claude Code); I reviewed the change and the code paths above.

## Checklist


phdargen (approved):
Thanks for the fix @loveaihq

PR #3502, opened 2026-09-16
fix(python): close literal-route bypass via percent-encoded path separator

CarsonRoscoe:
## Description

FastAPI/Flask middleware match routes on the **escaped** request path (`raw_path` / `RAW_URI`|`REQUEST_URI`), which is what protects wildcard and param routes from a decoded separator widening a capture group (`x402_http_server_base.py` `_normalize_path`, #3073). But Starlette and Werkzeug dispatch **literal** route segments on the **decoded** path (`request.url.path` / `PATH_INFO`).

For a literal route like `GET /api/premium`, a request for `/api%2Fpremium` normalizes (escaped-aware) to `/api%2Fpremium`, which never matches the literal pattern, so `requires_payment()` returns `False` and the middleware fail-opens via `call_next()`/`_original_wsgi()`. Meanwhile the framework still dispatches the decoded `/api/premium` straight to the paid handler — no payment header, no verification, no settlement. Reproduced against a real Uvicorn server: `curl /api/premium` → `402`, `curl --path-as-is /api%2Fpremium` → `200` with the full paid body.

Fix: add `HTTPRequestContext.decoded_path` carrying each framework's own decoded routing view (`request.url.path` for FastAPI, `PATH_INFO` for Flask). `_get_route_config` now checks both the escaped and decoded paths, requiring payment if either matches a protected route (fail closed on disagreement). This only widens matching when the two representations diverge — unencoded requests have `path == decoded_path`, so the added check is a no-op for them and cannot regress the existing escaped-path protection from #3073.

## Tests

Added first to reproduce the bypass (failing against pre-fix code), then verified to pass once the fix landed:
- `tests/unit/http/test_route_matching.py`: `TestDecodedPathDivergenceBypass`, `TestNormalizeDecodedPath`
- `tests/unit/http/middleware/test_fastapi.py`: `TestLiteralRoutePercentEncodedSeparatorBypass` (real, non-catch-all `/api/premium` handler)
- `tests/unit/http/middleware/test_flask.py`: `TestLiteralRoutePercentEncodedSeparatorBypass` (real, non-catch-all `/api/premium` handler, `RAW_URI` simulated via `environ_overrides` since Werkzeug's own test client never populates it)

Ran `uv run pytest` from `python/x402/` — 2239 passed, 69 skipped (unrelated pre-existing environment issues excluded: `abnf` grammar version drift, MCP SDK `structuredContent` rename).

## Checklist


python/x402/http/middleware/fastapi.py
```diff
@@ -249,12 +249,13 @@ async def middleware(
...
         adapter = FastAPIAdapter(request)
-        # Routers dispatch on the escaped path, so route matching must use the
-        # raw request path rather than the decoded URL path.
+        # Starlette dispatches literal routes on the decoded path but
+        # wildcard/param routes on the escaped one, so match both.
         raw_path = request.scope["raw_path"].decode("ascii").split("?")[0]
         context = HTTPRequestContext(
             adapter=adapter,
             path=raw_path,
+            decoded_path=request.url.path,
```
phdargen, line 258: request.url.path is scope["path"], and uvicorn prefixes that with --root-path. Starlette matches get_route_path(scope) (prefix stripped), so GET /api/premium still fail-opens for both /api/premium and /api%2Fpremium when root_path=/svc. Use the stripped route path:
```suggestion
raw_path = request.scope["raw_path"].decode("ascii").split("?")[0]
path = str(request.scope.get("path") or "/")
root_path = str(request.scope.get("root_path") or "")
if root_path and path.startswith(root_path):
    if path == root_path:
        path = ""
    elif path[len(root_path)] == "/":
        path = path[len(root_path) :]
context = HTTPRequestContext(
    adapter=adapter,
    path=raw_path,
    decoded_path=path,
```

phdargen (approved):

PR #3521, opened 2026-09-18
fix(evm): reject malformed CAIP-2 network identifiers in getEvmChainId

HereForTheTechNFT:
## Description

`getEvmChainId` parses the CAIP-2 reference with `parseInt(idStr, 10)`, which stops at the first non-digit instead of rejecting the input, so a malformed network resolves to a chain ID the caller never supplied:

| `network` | TS (before) | Go `GetEvmChainId` | Python `get_evm_chain_id` |
| --- | --- | --- | --- |
| `eip155:0x2105` | `0` | error | `ValueError` |
| `eip155:8453abc` | `8453` | error | `ValueError` |
| `eip155:8453.9` | `8453` | error | `ValueError` |
| `eip155:1e3` | `1` | error | `ValueError` |
| `eip155:9007199254740993` | `9007199254740992` | exact (`big.Int`) | exact |

`NetworkSchemaV2` only requires a colon, so these reach the mechanism. The value is then signed into the EIP-712 domain — `getBatchSettlementEip712Domain(getEvmChainId(network))` in `batch-settlement/server/verify.ts` and `batch-settlement/utils.ts`, and the Permit2 / EIP-3009 domains via `shared/permit2.ts` — and selects the RPC in `shared/rpc.ts`.

The fix requires a bare decimal reference. `/^eip155:(\d+)$/` is the same pattern the SIWX extension already uses (`packages/extensions/src/sign-in-with-x/evm.ts`). Chain IDs past `Number.MAX_SAFE_INTEGER` now throw instead of rounding, since `number` cannot represent what Go's `big.Int` and Python's `int` can.

Well-formed identifiers are unaffected and both existing error messages are unchanged.

## Tests

- `pnpm --filter @x402/evm test`: 40 files, **1129 passed** (1121 on `main`, +8 new).
- The 8 new cases fail on `main` (`AssertionError: expected [Function] to throw an error`) and pass with the fix.
- `eslint` and `prettier --check` clean for the package.

Not run locally: the full `pnpm test` matrix and the Go/Python suites (unchanged here) — CI covers those. I also grepped every `"eip155:…"` literal under `packages/`; the only ones the stricter pattern rejects are negative-test fixtures.

## Checklist

---

AI-assisted, per [AI-Assisted Contributions](https://github.com/x402-foundation/x402/blob/main/CONTRIBUTING.md#ai-assisted-contributions): the patch was produced with Claude. The diff, the before/after behaviour table and the local test runs were reviewed before marking this ready.


phdargen (approved):
Thanks for the contribution @HereForTheTechNFT

